PR #17, opened 2018-02-15
Fix CircleCI tests

ilikebits:
This PR resolves test breakage resulting from #15. It works by doing integration testing using the in-repository test fixtures and spinning up a Docker image with all the required test tools.

Future work:

- Figuring out how to do unit tests (get rid of effects :tada: or maybe dependency injection?)
- Getting JUnit test reports in CircleCI (looks like Ginkgo is a good fit for use as a testing library)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,27 +2,16 @@ version: 2
 jobs:
   build:
     docker:
-      - image: circleci/golang:1.9
-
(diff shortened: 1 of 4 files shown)


ilikebits:
Overriding license check: dependencies are being overreported, and this will be fixed by dogfooding the CLI in a coming PR.

PR #22, opened 2018-02-17
Refactor/common build utils

### LINKED ISSUE #20: Refactor common build utilities
There are a few common build utilities that should be refactored:

- Checking which command to run (given a default, some fallbacks, and an environment variable)
- Searching for project manifests
- Running commands in a specified directory, and getting `stdout` and `stderr` as strings.

ilikebits:
Refactors builders to use standard common utilities for tasks like parsing manifests, running commands, and logging.

Fixes #20.

### DIFF
--- a/.fossa.yaml
+++ b/.fossa.yaml
@@ -16,7 +16,7 @@ analyze:
   #     type: folder
   modules:
     - name: fossa-cli
-      path: github.com/fossas/fossa-cli/cmd/fossa
+      path: ./cmd/fossa
       type: gopackage
     # - name: vendored-jquery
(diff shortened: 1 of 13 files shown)


.fossa.yaml
```diff
@@ -16,7 +16,7 @@ analyze:
   #     type: folder
   modules:
     - name: fossa-cli
-      path: github.com/fossas/fossa-cli/cmd/fossa
+      path: ./cmd/fossa
```
xizhao: awesome. is this backwards compatible?  we should be OK w/ breaking compatibility until 1.0

Issue #24, opened 2018-02-17
Separate REST operations into common utils


xizhao:
As we're going to have more commands operate as a REST client, there should be a common set of utils to make requests and handle common responses (i.e. invalid API keys, etc...)

Issue #12, opened 2018-02-07
Add `fossa test` command


xizhao:
Eclipse functions of `license-cli`

ilikebits:
Version 1 implemented by #25.

PR #25, opened 2018-02-17
feat(test): add test command

xizhao:
PR to address #12 #24

### DIFF
--- a/build/archive.go
+++ b/build/archive.go
@@ -4,7 +4,7 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"encoding/json"
-	"fmt"
+	"errors"
 	"io"
 	"os"
 	"path/filepath"
@@ -38,7 +38,7 @@ func findNodePkgs(path string) []string {
 
 // Build does nothing for VendoredArchives
 func (builder *VendoredArchiveBuilder) Build(m module.Module, force bool) error {
-	return fmt.Errorf("VendoredArchive packages are expected to be pre-built")
+	return errors.New("VendoredArchive packages are expected to be pre-built")
 }
 
 // Analyze scans for archives and inspects them
@@ -119,10 +119,12 @@ func (builder *VendoredArchiveBuilder) IsBuilt(m module.Module, allowUnresolved 
 	return true, nil
 }
 
+// IsModule is not implemented
 func (builder *VendoredArchiveBuilder) IsModule(target string) (bool, error) {
-	return false, fmt.Errorf("IsModule is not implemented for VendoredArchiveBuilder")
+	return false, errors.New("IsModule is not implemented for VendoredArchiveBuilder")
 }
 
+// InferModule is not implemented
 func (builder *VendoredArchiveBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, fmt.Errorf("InferModule is not implemented for VendoredArchiveBuilder")
+	return module.Module{}, errors.New("InferModule is not implemented for VendoredArchiveBuilder")
 }
--- a/build/archive.go
+++ b/build/archive.go
@@ -4,7 +4,7 @@ import (
 	"archive/tar"
 	"compress/gzip"
 	"encoding/json"
-	"errors"
+	"fmt"
 	"io"
 	"os"
 	"path/filepath"
@@ -38,7 +38,7 @@ func findNodePkgs(path string) []string {
 
 // Build does nothing for VendoredArchives
 func (builder *VendoredArchiveBuilder) Build(m module.Module, force bool) error {
-	return errors.New("VendoredArchive packages are expected to be pre-built")
+	return fmt.Errorf("VendoredArchive packages are expected to be pre-built")
 }
 
 // Analyze scans for archives and inspects them
@@ -120,9 +120,9 @@ func (builder *VendoredArchiveBuilder) IsBuilt(m module.Module, allowUnresolved 
 }
 
 func (builder *VendoredArchiveBuilder) IsModule(target string) (bool, error) {
-	return false, errors.New("IsModule is not implemented for VendoredArchiveBuilder")
+	return false, fmt.Errorf("IsModule is not implemented for VendoredArchiveBuilder")
 }
 
 func (builder *VendoredArchiveBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for VendoredArchiveBuilder")
+	return module.Module{}, fmt.Errorf("InferModule is not implemented for VendoredArchiveBuilder")
 }
--- a/cmd/fossa/analyze.go
+++ b/cmd/fossa/analyze.go
@@ -70,12 +70,12 @@ func doAnalyze(modules []moduleConfig, allowUnresolved bool) (analysis, error) {
 	for _, moduleConfig := range modules {
 		builder, m, err := resolveModuleConfig(moduleConfig)
 		if err != nil {
-			return nil, fmt.Errorf("failed to resolve modules: " + err.Error())
+			return nil, fmt.Errorf("failed to resolve modules: %s", err.Error())
 		}
 
 		err = builder.Initialize()
 		if err != nil {
-			return nil, fmt.Errorf("failed to initialize build: " + err.Error())
+			return nil, fmt.Errorf("failed to initialize build: %s", err.Error())
 		}
 
 		isBuilt, err := builder.IsBuilt(m, allowUnresolved)
@@ -83,12 +83,12 @@ func doAnalyze(modules []moduleConfig, allowUnresolved bool) (analysis, error) {
 			return nil, fmt.Errorf("could not determine whether module %#v is built: %#v", m.Name, err.Error())
 		}
 		if !isBuilt {
-			return nil, fmt.Errorf("module " + m.Name + " does not appear to be built (try first running your build or `fossa build`, and then running `fossa`)")
+			return nil, fmt.Errorf("module %s does not appear to be built (try first running your build or `fossa build`, and then running `fossa`)", m.Name)
 		}
 
 		deps, err := builder.Analyze(m, allowUnresolved)
 		if err != nil {
-			return nil, fmt.Errorf("analysis failed on module " + m.Name + ": " + err.Error())
+			return nil, fmt.Errorf("analysis failed on module %s: %s", m.Name, err.Error())
 		}
 		dependencies[analysisKey{
 			builder: builder,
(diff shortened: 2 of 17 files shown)


cmd/fossa/test.go
```diff
@@ -0,0 +1,127 @@
...
+		if timing >= config.timeout {
+			return errors.New("request series timed out")
+		}
+		time.Sleep(time.Duration(PollRequestDelay) * time.Millisecond)
+
+		return confirmScan(config, timing+PollRequestDelay)
+	}
+
+	issueCount := 0
+	gjson.Get(revisionData, "issues.#.resolved").ForEach(func(key, value gjson.Result) bool {
```
ilikebits: This is not JavaScript. Unmarshal this into a struct, and write a for loop over the list of issues.

cmd/fossa/test.go
```diff
@@ -0,0 +1,127 @@
...
+		if value.Bool() == false {
+			issueCount++
+		}
+		return true // keep iterating
+	})
+	if issueCount > 0 {
+		return errors.New(strconv.Itoa(issueCount) + " issues found")
```
ilikebits: `fmt.Errorf`. Format:clap:strings:clap:are:clap:your:clap:friend:clap:.

cmd/fossa/test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	if err := confirmBuild(config, 0); err != nil {
+		s.Stop()
+		testLogger.Fatalf("Error executing test: %#v", err)
+	}
+
+	s.Suffix = " Waiting for FOSSA scan results..."
+	s.Restart()
+
+	if err := confirmScan(config, 0); err != nil {
```
ilikebits: Errors are a terrible way to express a return result. This conflates an actual error occurrence with an expected output. Instead `confirmScan` should return `(int, error)`.

Also, `confirmScan` should _really_ just return the whole issues struct so the `testCmd` can do whatever processing it needs. For example, `testCmd` should have a mode where it outputs the reports as machine-readable JSON so that other processes can use its output.

cmd/fossa/common.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusForbidden {
+		return nil, errors.New("invalid API key, check the $FOSSA_API_KEY environment variable")
+	} else if resp.StatusCode != http.StatusOK {
+		return nil, errors.New("bad server response (" + string(resp.StatusCode) + ")")
```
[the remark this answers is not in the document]
xizhao: In common functions, the error should be the way for internal APIs to communicate failure with a different layer (i.e. front-facing command functions) handling how failures are expressed to users (i.e. terminal output).

cmd/fossa/test.go
xizhao: I think we should actually pull out the request functions into a separate file, as we may want to use the output for other reporting functions, and we can leave `test` as just an interface for the user.

cmd/fossa/common.go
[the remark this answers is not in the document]
ilikebits: I understand that, but using format strings means you don't have to do all those `string()` conversions.

cmd/fossa/test.go
```diff
@@ -0,0 +1,127 @@
...
+	reqURL := fossaBaseURL.ResolveReference(reqRef).String()
+
+	testLogger.Debugf("Querying <%#v>", reqURL)
+	resp, err := makeAPIRequest("GET", reqURL, nil, config.apiKey)
+	if err != nil {
+		return err
+	}
+
+	revisionData := string(resp)
+	if gjson.Get(revisionData, "meta.0.last_scan").String() == "" {
```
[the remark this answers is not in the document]
xizhao: Parsing this api will suck and I don't think this API structure will live very long on our side.  I think a better ideal is creating a `/revisions/:locator/issues` endpoint that's actually sane.

cmd/fossa/test.go
xizhao: Why? This is effectively what gjson will do, but I believe gjson will provide some level of safety in case entries don't exist.

cmd/fossa/test.go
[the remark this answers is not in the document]
ilikebits: You don't have to write out the entire struct. Just write out the fields and shape you need, and `Unmarshal` will ignore the rest.

cmd/fossa/test.go
```diff
@@ -0,0 +1,127 @@
...
+	default:
+	}
+
+	if timing >= config.timeout {
+		return errors.New("request series timed out")
+	}
+
+	time.Sleep(time.Duration(PollRequestDelay) * time.Millisecond)
+
+	return confirmBuild(config, timing+PollRequestDelay)
```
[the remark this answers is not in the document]
xizhao: Agreed, it's not an accurate timeout.

cmd/fossa/test.go
ilikebits: What? The documentation for `gjson` says that behaviour is undefined when the input is malformed. No panic or error. You have _no idea_ what `gjson` will do.

cmd/fossa/common.go
```diff
@@ -0,0 +1,30 @@
...
+
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusForbidden {
+		return nil, errors.New("invalid API key, check the $FOSSA_API_KEY environment variable")
```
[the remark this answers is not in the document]
xizhao: consistent with what, other error formats?

cmd/fossa/test.go
ilikebits: `gjson` is the opposite of safe. Here is the exact literal comment on `gjson.Get`:

```
// This function expects that the json is well-formed, and does not validate.
// Invalid json will not panic, but it may return back unexpected results.
```

cmd/fossa/common.go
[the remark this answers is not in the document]
ilikebits: With the other errors that FOSSA emits. See e.g. all the errors under `build/`.

cmd/fossa/main.go
```diff
@@ -96,12 +109,22 @@ type cliConfig struct {
...
+	timeout  int
 
 	defaultConfig defaultConfig
 	analyzeConfig analyzeConfig
 	buildConfig   buildConfig
 }
 
+func (conf *cliConfig) getVcsLocator() string {
+	normalizedPackageSpec := strings.TrimSuffix(conf.project, ".git")
+	normalizedPackageSpec = strings.TrimPrefix(conf.project, "git+")
```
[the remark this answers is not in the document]
xizhao: Whoops

cmd/fossa/test.go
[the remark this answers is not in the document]
ilikebits: This should either be implemented correctly as a timeout or documented correctly as a retry count. I think the former is the better approach -- timeouts in Go are not hard, and a retry count is significantly less useful.

cmd/fossa/common.go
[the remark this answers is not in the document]
xizhao: Oh. `fmt.Errorf` Returns an error.

build/archive.go
```diff
@@ -38,7 +38,7 @@ func findNodePkgs(path string) []string {
 
 // Build does nothing for VendoredArchives
 func (builder *VendoredArchiveBuilder) Build(m module.Module, force bool) error {
-	return errors.New("VendoredArchive packages are expected to be pre-built")
+	return fmt.Errorf("VendoredArchive packages are expected to be pre-built")
```
ilikebits: You don't need to use `fmt.Errorf` when you're not using a format string. Also, you should try to use the least functionality you require for a task (this helps communicate intent). This would be an appropriate place ot use `errors.New`.

The same goes for the other files in `build/`.

cmd/fossa/analyze.go
```diff
@@ -71,25 +70,25 @@ func doAnalyze(modules []moduleConfig, allowUnresolved bool) (analysis, error) {
 	for _, moduleConfig := range modules {
 		builder, m, err := resolveModuleConfig(moduleConfig)
 		if err != nil {
-			return nil, errors.New("failed to resolve modules: " + err.Error())
+			return nil, fmt.Errorf("failed to resolve modules: " + err.Error())
```
ilikebits: The point of using `fmt.Errorf` is to use the format string. Here, this would be `fmt.Errorf("failed to resolve modules: %s", err.Error())`. Likewise below.

cmd/fossa/common.go
```diff
@@ -2,14 +2,17 @@ package main
...
 	"io/ioutil"
 	"net/http"
 )
 
 // Common utilities among commands
 func makeAPIRequest(requestType string, url string, payload []byte, apiKey string) ([]byte, error) {
-	req, _ := http.NewRequest(requestType, url, bytes.NewReader(payload))
+	req, err := http.NewRequest(requestType, url, bytes.NewReader(payload))
+	if err != nil {
+		return nil, err
```
ilikebits: This error should probably be annotated with `fmt.Errorf` for debuggability.

cmd/fossa/common.go
```diff
@@ -21,10 +24,9 @@ func makeAPIRequest(requestType string, url string, payload []byte, apiKey strin
 	defer resp.Body.Close()
 
 	if resp.StatusCode == http.StatusForbidden {
-		return nil, errors.New("invalid API key, check the $FOSSA_API_KEY environment variable")
+		return nil, fmt.Errorf("invalid API key (check the $FOSSA_API_KEY environment variable)")
 	} else if resp.StatusCode != http.StatusOK {
-		return nil, errors.New("bad server response (" + string(resp.StatusCode) + ")")
+		return nil, fmt.Errorf("bad server response (" + string(resp.StatusCode) + ")")
```
ilikebits: Again, you should use the format string.

cmd/fossa/common.go
```diff
@@ -21,10 +24,9 @@ func makeAPIRequest(requestType string, url string, payload []byte, apiKey strin
...
 	}
-	responseBytes, _ := ioutil.ReadAll(resp.Body)
-	return responseBytes, nil
+	return ioutil.ReadAll(resp.Body)
```
ilikebits: This function would probably be more convenient if it returned a `string`.

xizhao:
Oh word i wasn't don't yet but thanks for the comments.  Feel free to bring this branch home.

Issue #23, opened 2018-02-17
Implement update checker and `update` cmd


xizhao:
https://github.com/rhysd/go-github-selfupdate

PR #36, opened 2018-02-20
feat(update): #23 add fossa update command

xizhao:
Uses `github.com/rhysd/go-github-selfupdate/selfupdate` to update binary.

Not tested yet, do not merge.

### DIFF
--- a/build/sbt.go
+++ b/build/sbt.go
@@ -1,7 +1,6 @@
 package build
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -144,10 +143,10 @@ func (builder *SBTBuilder) IsBuilt(m module.Module, allowUnresolved bool) (bool,
 
 // IsModule is not implemented
 func (builder *SBTBuilder) IsModule(target string) (bool, error) {
-	return false, errors.New("IsModule is not implemented for SBTBuilder")
+	return false, fmt.Errorf("IsModule is not implemented for SBTBuilder")
 }
 
 // InferModule is not implemented
 func (builder *SBTBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for SBTBuilder")
+	return module.Module{}, fmt.Errorf("InferModule is not implemented for SBTBuilder")
 }
--- a/build/sbt.go
+++ b/build/sbt.go
@@ -1,6 +1,7 @@
 package build
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"strings"
@@ -143,10 +144,10 @@ func (builder *SBTBuilder) IsBuilt(m module.Module, allowUnresolved bool) (bool,
 
 // IsModule is not implemented
 func (builder *SBTBuilder) IsModule(target string) (bool, error) {
-	return false, fmt.Errorf("IsModule is not implemented for SBTBuilder")
+	return false, errors.New("IsModule is not implemented for SBTBuilder")
 }
 
 // InferModule is not implemented
 func (builder *SBTBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, fmt.Errorf("InferModule is not implemented for SBTBuilder")
+	return module.Module{}, errors.New("InferModule is not implemented for SBTBuilder")
 }
--- a/cmd/fossa/main.go
+++ b/cmd/fossa/main.go
@@ -252,7 +252,7 @@ func defaultCmd(c *cli.Context) {
 		mainLogger.Fatalf("Could not load configuration: %s", err.Error())
 	}
 
-	if ok, err := checkUpdate(); ok && err != nil {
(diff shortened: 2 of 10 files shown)


build/sbt.go
```diff
@@ -144,10 +143,10 @@ func (builder *SBTBuilder) IsBuilt(m module.Module, allowUnresolved bool) (bool,
 
 // IsModule is not implemented
 func (builder *SBTBuilder) IsModule(target string) (bool, error) {
-	return false, errors.New("IsModule is not implemented for SBTBuilder")
+	return false, fmt.Errorf("IsModule is not implemented for SBTBuilder")
```
ilikebits: You don't need to use `fmt.Errorf` if you don't need the format string.

cmd/fossa/update.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/blang/semver"
+	logging "github.com/op/go-logging"
+	"github.com/rhysd/go-github-selfupdate/selfupdate"
+	"github.com/urfave/cli"
+)
+
+var updateLogger = logging.MustGetLogger("update")
+var updateEndpoint := "fossas/fossa-cli"
+
+func checkUpdate() error {
```
ilikebits: This should return `(bool, error)`. Don't use `error` to represent things that are not errors.

cmd/fossa/update.go
```diff
@@ -0,0 +1,49 @@
...
+	if latest.Version.Equals(v) {
+		// latest version is the same as current version. It means current binary is up to date.
+		return fmt.Errorf("no update required; currently on latest")
+	}
+	return nil
+}
+
+func updateCmd(c *cli.Context) {
+	if err := checkUpdate(); err != nil {
+		updateLogger.Fatalf("unable to update (%s)", err)
```
ilikebits: _Logged_ messages (distinct from _error messages_) should generally be `Sentence case`. The reason why error messages should _not_ be `Sentence case` is because they're usually embedded within a sentence when they're finally logged to display to the user (e.g. `logger.Fatalf("Problem happened: %s", err.Error())`).

ilikebits: `err` is of type `error`. You probably mean `err.Error()`, which is of type `string`.

ilikebits: Why the wrapping parentheses? Error messages should _always_ be readable as conventional strings. If you need to display the value of the `error` itself, you should use `%#v`.

build/sbt.go
ilikebits: This still needs to be fixed. Also below, in the same file.

cmd/fossa/main.go
```diff
@@ -244,6 +252,10 @@ func defaultCmd(c *cli.Context) {
 		mainLogger.Fatalf("Could not load configuration: %s", err.Error())
 	}
 
+	if ok, err := checkUpdate(); ok && err == nil {
```
ilikebits: [nit] Would prefer `err == nil && ok`. Generally, error checking goes first.

cmd/fossa/main.go
```diff
@@ -244,6 +252,10 @@ func defaultCmd(c *cli.Context) {
...
+		mainLogger.Infof("An update is available for this CLI; run `fossa update` to get the latest version.")
```
ilikebits: This won't actually log anything for most users. See logging levels from `go-logging`:
```go
const (
	CRITICAL Level = iota
	ERROR
	WARNING
	NOTICE
	INFO
	DEBUG
)
```
We currently only log `WARNING` and above by default (see `initialize` in `config.go`). I'm alright with changing the default logging level to `INFO` and above, although I'd actually consider this a `NOTICE`.

cmd/fossa/update.go
```diff
@@ -0,0 +1,73 @@
...
+	if version[0] != 'v' {
+		return ""
+	}
+	return version[1:]
+}
+
+func checkUpdate() (bool, error) {
+	latest, found, err := selfupdate.DetectLatest(updateEndpoint)
+	if err != nil {
+		return false, err // unable to update due to GH error
```
ilikebits: You should annotate this error with `fmt.Errorf`. For example, `return false, fmt.Errorf("could not detect latest comment: %s", err.Error())`.

cmd/fossa/update.go
```diff
@@ -0,0 +1,73 @@
...
+
+func doSelfUpdate() error {
+	parsedVersion := getSemver()
+	v, err := semver.Parse(parsedVersion)
+	if err != nil {
+		return errors.New("invalid version; you may be using a development binary")
+	}
+	latest, err := selfupdate.UpdateSelf(v, updateEndpoint)
+	if err != nil {
+		return err
```
ilikebits: Use `fmt.Errorf` to annotate this error. It helps us trace errors through the stack.

ilikebits:
I'm merging this. I know it works, but our build is breaking because of backend issues causing our license check to fail.

PR #48, opened 2018-02-23
Refactor/refactor builders

xizhao:
1. Refactor `build` --> `builders`
2. Use `ModuleType` as our switch keys when we have special handling logic for different module types
3. Properly break out initialization logic into constructors in `builders.New` and `module.New`
4. Utilize constructors to slim insane switch statements in `main.go`

### DIFF
--- a/builders/builder.go
+++ b/builders/builder.go
@@ -1,4 +1,4 @@
-package build
+package builders
 
 import (
 	"github.com/fossas/fossa-cli/config"
@@ -21,4 +21,27 @@ type Builder interface {
 	IsModule(configKey string) (bool, error)
 	// DiscoverModules finds what modules are available for analysis in a given directory.
 	DiscoverModules(dir string) ([]config.ModuleConfig, error)
+}
+
+// New instantiates a Builder given a ModuleType
+func New(moduleType config.ModuleType) Builder {
+	switch moduleType {
+	case config.Bower:
+		return &BowerBuilder{}
+	case config.Composer:
+		return &ComposerBuilder{}
+	case config.Golang:
+		return &GoBuilder{}
+	case config.Maven:
+		return &MavenBuilder{}
+	case config.Nodejs:
+		return &NodeJSBuilder{}
+	case config.Ruby:
+		return &RubyBuilder{}
+	case config.SBT:
+		return &SBTBuilder{}
+	case config.VendoredArchives:
+		return &VendoredArchiveBuilder{}
+	}
+	return nil
 }
--- a/cmd/fossa/analyze.go
+++ b/cmd/fossa/analyze.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/fossas/fossa-cli/build"
 	"github.com/fossas/fossa-cli/config"
 	"github.com/fossas/fossa-cli/module"
 	logging "github.com/op/go-logging"
@@ -54,7 +55,7 @@ func analyzeCmd(c *cli.Context) {
 }
 
 type analysisKey struct {
-	builder module.Builder
+	builder build.Builder
 	module  module.Module
 }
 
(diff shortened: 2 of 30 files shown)


xizhao:
My thinking is: 

 1. `config` should be responsible for managing environment state and normalizing it into hard structure for the rest of the program.  

 2. Any kind of instantiation we do on `Builders` or `Module` must be done through an API that accepts an enum or type like `ModuleType`, as we will be doing this a lot through the app.

 3. Messy "switch-based" instantiation logic should be as minimal and in as few places as possible.  That way, as we implement, we only have thing to take care of.  I'm OK with merging builders/modules as long as there is only one interface for creating them together.

 4.  I could imagine that a module could have multiple builders one day, if we say that a `module` can have multiple build paths, or be an aggregate of many build paths.  I.e. one module may both use npm but also tarballs/vendored deps -- builders are for dependency discovery while modules group a collection of artifacts together even if they span multiple builds.

 5.  It makes no sense for `Builder` not to live with all the instances of `Builders`.  If you believe we should unify the constructors for now, I suggest we move the `Builder` constructor the `Module`, though I dislike the idea that you call `module.New()` to get the result of `builder.New()`

 6.  This refactor I think is at least a cleaner and slimmer implementation of the giant switch in `main.go`, we could run with this and get a v2 unless you disagree.

cmd/fossa/main.go
```diff
@@ -115,136 +114,44 @@ func main() {
...
+	if moduleType == "" {
+		mainLogger.Debug("Got unknown module.")
+		return builder, m, fmt.Errorf("unknown module type: %s", moduleConfig.Type)
 	}
 
-	moduleName := config.Name
-	if moduleName == "" {
-		moduleName = config.Path
-	}
+	mainLogger.Debugf("Got <%s> module.", moduleType)
```
ilikebits: You don't need to surround `%s` with brackets. If you feel like you need to, you should probably be using `%#v` instead.

cmd/fossa/main.go
```diff
@@ -115,136 +114,44 @@ func main() {
...
-		}
-	case "gopackage":
-		fallthrough
-	case "golang":
-		fallthrough
-	case "go":
-		mainLogger.Debug("Got Go module.")
-		builder = &build.GoBuilder{}
-		m, err = setupModule(moduleConfig, "", module.Golang)
-		// Target should be relative to $GOPATH
```
ilikebits: This comment should be preserved in `module.New`.

builders/builder.go
```diff
@@ -0,0 +1,47 @@
...
+	Initialize() error
+	// Build runs a best-effort attempt at building the module.
+	Build(m module.Module, force bool) error
+	// Analyze returns the dependencies of a module.
+	Analyze(m module.Module, allowUnresolved bool) ([]module.Dependency, error)
+
+	// IsBuilt checks whether a module has been built.
+	IsBuilt(m module.Module, allowUnresolved bool) (bool, error)
+
+	// IsModule takes a configKey and returns whether it is valid to elect this Builder.
```
ilikebits: What is a `configKey`? What is electing, and why is that different from choosing?

If you're going to use weird vocabulary, you should define it because others will assume there is a specific definition.

cmd/fossa/analyze.go
```diff
@@ -38,39 +34,39 @@ func analyzeCmd(c *cli.Context) {
...
 type analysis map[analysisKey][]module.Dependency
 
-func doAnalyze(modules []moduleConfig, allowUnresolved bool) (analysis, error) {
+func doAnalyze(modules []config.ModuleConfig, allowUnresolved bool) (analysis, error) {
 	analysisLogger.Debugf("Running analysis on modules: %#v", modules)
 	dependencies := make(analysis)
 
-	for _, moduleConfig := range modules {
-		builder, m, err := resolveModuleConfig(moduleConfig)
+	for _, ModuleConfig := range modules {
```
ilikebits: Don't capitalise local variables.

xizhao:
Ty for comments, merging into #49

Issue #46, opened 2018-02-23
Add `fossa init`


xizhao:
Add a command that re-writes `analyze.modules` in `.fossa.yml` based off any discovered modules.

- [x] Ensure interface to easily serialize and de-serialize config from/to yml
- [x] Add fossa-cli comment header on config write
- [x] Rename `builder.InferModule(target string)` to `builder.DiscoverModules(dir string)`
- [x] Implement `builder.DiscoverModule` for node
- [x] Add behavior to default cmd if we are missing critical config
- [x] Add flag to default cmd to force updating config
- [x] Implement "fail-open" flag if a module has an error in default cmd
- [x] fully implement `DiscoverModules` across all currently-supported build systems

xizhao:
Only golang left for DiscoverModules, @ilikebits want to take golang over?

xizhao:
RFC: #49

PR #49, opened 2018-02-23
Feat/fossa init cmd 

xizhao:
Implements `fossa init` with `nodejs builder`

#46 

Not ready for merge.


builders/nodejs.go
```diff
@@ -161,7 +161,43 @@ func (builder *NodeJSBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for NodeJSBuilder")
 }
 
-// DiscoverModules is not implemented
+// DiscoverModules builds ModuleConfigs for any package.jsons that are not contained in a node_modules dir
 func (builder *NodeJSBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
-	return []config.ModuleConfig{}, errors.New("DiscoverModules is not implemented for NodeJSBuilder")
+	moduleConfigs := []config.ModuleConfig{}
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
ilikebits: Why not use glob?

cmd/fossa/init.go
```diff
@@ -9,14 +10,34 @@ import (
...
+	}
+
+	initLogger.Warningf("%v", conf)
 }
 
-// doInit calls DiscoverModules() on all available integrations and writes to
-func doInit(conf config.CliConfig) {
-	return
+// findModules calls DiscoverModules() on all available integrations and returns a new config state
+// to implement fail-open behavior, do not handle the err returned by this function
```
ilikebits: I'm not sure what you mean by "fail-open" here.

cmd/fossa/init.go
```diff
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"github.com/fossas/fossa-cli/builders"
```
ilikebits: Internal imports belong in their own import section.

builders/nodejs.go
xizhao: Glob is really bad at excluding stuff, need to use walk.

cmd/fossa/init.go
xizhao: some integrations throw an error and we're still OK

builders/nodejs.go
ilikebits: Glob and then filter the output. It'll be faster. Standard library Glob uses a syscall instead of walking to every file.

ilikebits: Actually, it looks like you'll need a doublestar glob and I'm not sure what that performance profile looks like. We can roll with this for now and change it if it's noticeably bad.

builders/bower.go
```diff
@@ -141,7 +142,48 @@ func (builder *BowerBuilder) IsModule(target string) (bool, error) {
...
-	return module.Module{}, errors.New("InferModule is not implemented for BowerBuilder")
+// DiscoverModules is not implemented
+func (builder *BowerBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	moduleConfigs := []config.ModuleConfig{}
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			bowerLogger.Debugf("failed to access path %s: %s\n", path, err.Error())
+			return err
+		}
+		// skip **/node_modules
```
ilikebits: Generally, comments should be `Sentence case`.

builders/bower.go
```diff
@@ -141,7 +142,48 @@ func (builder *BowerBuilder) IsModule(target string) (bool, error) {
...
+		// skip **/bower_components
+		if info.IsDir() && info.Name() == "bower_components" {
+			bowerLogger.Debugf("skipping bower_components directory: %s", info.Name())
+			return filepath.SkipDir
+		}
+
+		if !info.IsDir() && info.Name() == "bower.json" {
+			moduleName := filepath.Base(filepath.Dir(path))
+
+			// parse from package.json and set moduleName if successful
```
ilikebits: This comment is wrong.

builders/bower.go
```diff
@@ -141,7 +142,48 @@ func (builder *BowerBuilder) IsModule(target string) (bool, error) {
...
+			return err
+		}
+		// skip **/node_modules
+		if info.IsDir() && info.Name() == "node_modules" {
+			bowerLogger.Debugf("skipping node_modules directory: %s", info.Name())
+			return filepath.SkipDir
+		}
+
+		// skip **/bower_components
+		if info.IsDir() && info.Name() == "bower_components" {
```
ilikebits: Why not combine this with the `node_modules` check?

builders/composer.go
```diff
@@ -137,7 +139,43 @@ func (builder *ComposerBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for ComposerBuilder")
 }
 
-// InferModule is not implemented
-func (builder *ComposerBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for ComposerBuilder")
+// DiscoverModules is not implemented
```
ilikebits: This comment is outdated.

builders/composer.go
```diff
@@ -137,7 +139,43 @@ func (builder *ComposerBuilder) IsModule(target string) (bool, error) {
...
+func (builder *ComposerBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	moduleConfigs := []config.ModuleConfig{}
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			composerLogger.Debugf("failed to access path %s: %s\n", path, err.Error())
+			return err
+		}
+
+		// skip the /vendor/ folder
```
ilikebits: `Sentence case`.

builders/golang.go
```diff
@@ -589,7 +590,7 @@ func (builder *GoBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for GoBuilder")
 }
 
-// InferModule is not implemented
-func (builder *GoBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for GoBuilder")
+// DiscoverModules is not implemented
+func (builder *GoBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	return []config.ModuleConfig{}, errors.New("DiscoverModules is not implemented for GoBuilder")
```
ilikebits: You can return `nil` for slices.

builders/maven.go
```diff
@@ -35,6 +39,17 @@ func (m MavenArtifact) Revision() string {
 	return m.Version
 }
 
+// PomFile represents the schema of a common pom.xml file
+type PomFile struct {
```
ilikebits: Is `builders.PomFile` meant to be public?

builders/maven.go
```diff
@@ -138,7 +153,52 @@ func (builder *MavenBuilder) IsModule(target string) (bool, error) {
...
 }
 
-// InferModule is not implemented
-func (builder *MavenBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for MavenBuilder")
+// DiscoverModules finds either a root pom.xml file or all pom.xmls in the specified dir
+func (builder *MavenBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	_, err := os.Stat(filepath.Join(dir, "pom.xml"))
+	if err == nil {
+		// root pom found; parse and return
```
ilikebits: `Sentence case`.

builders/ruby.go
```diff
@@ -148,7 +152,34 @@ func (builder *RubyBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for RubyBuilder")
 }
 
-// InferModule is not implemented
-func (builder *RubyBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for RubyBuilder")
+// DiscoverModules returns ModuleConfigs that match Gemfiles in the directory
+func (builder *RubyBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	gemFilePaths, err := doublestar.Glob(filepath.Join(dir, "**", "Gemfile"))
```
ilikebits: Should we be matching all Gemfiles?

ilikebits: cc @lackstein, who has the most Ruby experience. I'm not sure this method of discovering Ruby modules is idiomatic.

builders/sbt.go
```diff
@@ -147,7 +150,38 @@ func (builder *SBTBuilder) IsModule(target string) (bool, error) {
...
+	}
+
+	// no pom in root directory; find and parse all of them
+	sbtFilePaths, err := doublestar.Glob(filepath.Join(dir, "**", "build.sbt"))
+	if err != nil {
+		return nil, err
+	}
+	moduleConfigs := make([]config.ModuleConfig, len(sbtFilePaths))
+	for i, path := range sbtFilePaths {
+		artifactName := filepath.Dir(path) // use the dirname as it's too hard to parse from build.sbt
```
ilikebits: Isn't `build.sbt` a Scala file? That makes it literally impossible to parse without executing.

cmd/fossa/build.go
```diff
@@ -4,26 +4,24 @@ import (
...
 
-	logging "github.com/op/go-logging"
 	"github.com/urfave/cli"
-)
 
-type buildConfig struct {
-	force bool
-}
+	config "github.com/fossas/fossa-cli/config"
+	logging "github.com/op/go-logging"
```
ilikebits: Separate internal and third-party imports.

cmd/fossa/build.go
```diff
@@ -4,26 +4,24 @@ import (
...
 	"os"
 
-	logging "github.com/op/go-logging"
 	"github.com/urfave/cli"
-)
 
-type buildConfig struct {
-	force bool
-}
+	config "github.com/fossas/fossa-cli/config"
```
ilikebits: `config` doesn't need to be a named import. This seems to happen a lot, you might want to check if you have a misconfigured IDE or `goimports`.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+func doInit(conf *config.CliConfig, force bool, includeAll bool) error {
+	var err error
+	if len(conf.Modules) == 0 || force {
+		conf.Modules, err = findModules(".")
+		if err != nil {
+			initLogger.Warningf("A non-fatal error occured during module discovery: %s", err.Error())
+		}
+
+		if !includeAll {
+			// fitler suspicious modules
```
ilikebits: Typo in comment. Also, `Sentence case`.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		initLogger.Warningf("Config for %d modules written to `%s`.", len(conf.Modules), conf.ConfigFilePath)
+	}
+
+	fmt.Println("`fossa` is initialized")
+}
+
+func doInit(conf *config.CliConfig, force bool, includeAll bool) error {
+	var err error
```
ilikebits: Why is this declared up here?

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+	if len(conf.Modules) == 0 || force {
```
ilikebits: I'm not sure I understand -- why would you ever want to `--force` an init to use autoconfiguration rather than the user's configuration?

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+		conf.Modules, err = findModules(".")
```
ilikebits: You should pass an absolute path instead, for robustness. (Personally, I'd use a type alias to ensure you use an absolute path.)

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+			conf.Modules = filteredModuleConfigs
+		}
+	} else {
+		initLogger.Warningf("%d module(s) found in config file (`%s`); skipping initialization.", len(conf.Modules), conf.ConfigFilePath)
+	}
+	return nil
+}
+
+// findModules calls DiscoverModules() on all available integrations and returns a new config state
```
ilikebits: *\`findModules\`.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+// to implement fail-open behavior, do not handle the err returned by this function
```
ilikebits: That's not what fail-open means, get rid of this part of the comment.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+
+	fmt.Println("`fossa` is initialized")
+}
+
+func doInit(conf *config.CliConfig, force bool, includeAll bool) error {
+	var err error
+	if len(conf.Modules) == 0 || force {
+		conf.Modules, err = findModules(".")
+		if err != nil {
+			initLogger.Warningf("A non-fatal error occured during module discovery: %s", err.Error())
```
ilikebits: Typo. Also, this warning is extremely unhelpful. Much better:
```go
initLogger.Warningf("Error during autoconfiguration: %s", err.Error())
```

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		if !includeAll {
+			// fitler suspicious modules
+			filteredModuleConfigs := []config.ModuleConfig{}
```
ilikebits: You should use the `var slice []T` form unless the length is known, in which case use `make`.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+			for _, c := range conf.Modules {
+				if matched, err := regexp.MatchString("(docs?/|test|example|vendor/)", c.Path); err != nil || matched != true {
+					filteredModuleConfigs = append(filteredModuleConfigs, c)
+				} else {
+					initLogger.Warningf("Filtering out suspcious module: %s (%s)", c.Name, c.Path)
```
ilikebits: Typo.

cmd/fossa/init.go
```diff
@@ -0,0 +1,77 @@
...
+		initLogger.Warningf("%d module(s) found in config file (`%s`); skipping initialization.", len(conf.Modules), conf.ConfigFilePath)
+	}
+	return nil
+}
+
+// findModules calls DiscoverModules() on all available integrations and returns a new config state
+// to implement fail-open behavior, do not handle the err returned by this function
+func findModules(dir string) ([]config.ModuleConfig, error) {
+	var lastError error
+	moduleConfigs := []config.ModuleConfig{}
```
ilikebits: You should use the `var slice []T` form unless the length is known, in which case use `make`.

cmd/fossa/main.go
```diff
@@ -49,13 +47,23 @@ func main() {
...
 		cli.BoolFlag{Name: "debug", Usage: debugUsage},
 	}
 
 	app.Commands = []cli.Command{
+		{
+			Name:   "init",
+			Usage:  "Scans your environment for code module entry points and writes to config",
+			Action: initCmd,
+			Flags: []cli.Flag{
+				cli.BoolFlag{Name: "f, force", Usage: "rescan and overwrite modules in config even if they exist"},
```
ilikebits: This doesn't seem like a real use case to me. I would remove this until someone requests it.

cmd/fossa/main.go
```diff
@@ -49,13 +47,23 @@ func main() {
...
+				cli.BoolFlag{Name: "include-all", Usage: "include suspicious modules (`docs`, `test` or `example` in name)"},
+				cli.BoolFlag{Name: "skip-write", Usage: "do not write config if successful"},
```
ilikebits: Why would I ever want this? I think a more sensible flag would be `--output`, which prints the configuration to `stdout` and then exits.

builders/bower.go
```diff
@@ -141,7 +142,48 @@ func (builder *BowerBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for BowerBuilder")
 }
 
-// InferModule is not implemented
-func (builder *BowerBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for BowerBuilder")
+// DiscoverModules is not implemented
+func (builder *BowerBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	moduleConfigs := []config.ModuleConfig{}
```
ilikebits: You should use the `var slice []T` form unless the length is known, in which case use `make`.

builders/composer.go
```diff
@@ -137,7 +139,43 @@ func (builder *ComposerBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for ComposerBuilder")
 }
 
-// InferModule is not implemented
-func (builder *ComposerBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for ComposerBuilder")
+// DiscoverModules is not implemented
+func (builder *ComposerBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	moduleConfigs := []config.ModuleConfig{}
```
ilikebits: You should use the `var slice []T` form unless the length is known, in which case use `make`.

builders/nodejs.go
```diff
@@ -160,7 +161,48 @@ func (builder *NodeJSBuilder) IsModule(target string) (bool, error) {
 	return false, errors.New("IsModule is not implemented for NodeJSBuilder")
 }
 
-// InferModule is not implemented
-func (builder *NodeJSBuilder) InferModule(target string) (module.Module, error) {
-	return module.Module{}, errors.New("InferModule is not implemented for NodeJSBuilder")
+// DiscoverModules builds ModuleConfigs for any package.jsons that are not contained in a node_modules dir
+func (builder *NodeJSBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+	moduleConfigs := []config.ModuleConfig{}
```
ilikebits: You should use the `var slice []T` form unless the length is known, in which case use `make`.

cmd/fossa/main.go
```diff
@@ -123,131 +130,39 @@ func main() {
 	app.Run(os.Args)
 }
 
-func setupModule(config moduleConfig, manifestName string, moduleType module.Type) (module.Module, error) {
+func resolveModuleConfig(moduleConfig config.ModuleConfig) (builders.Builder, module.Module, error) {
+	mainLogger.Debugf("Resolving ModuleConfig: %#v", moduleConfig)
```
ilikebits: *`moduleConfig`, referring to the variable (as opposed to the type).

cmd/fossa/main.go
```diff
@@ -123,131 +130,39 @@ func main() {
...
-		mainLogger.Debug("Got unknown module.")
-		return builder, m, fmt.Errorf("unknown module type: %s", moduleConfig.Type)
+	m, err = module.New(moduleType, moduleConfig)
+	if err != nil {
+		mainLogger.Debug("Unable to resolve module config")
+		return builder, m, fmt.Errorf("unable to setup module type: %s", moduleConfig.Type)
 	}
 
-	mainLogger.Debugf("Resolved moduleConfig to: %#v, %#v", builder, m)
+	mainLogger.Debugf("Resolved ModuleConfig to: %#v, %#v", builder, m)
```
ilikebits: *`moduleConfig`, referring to the variable (as opposed to the type).

cmd/fossa/main.go
```diff
@@ -328,21 +246,27 @@ func defaultCmd(c *cli.Context) {
 			mainLogger.Fatalf("Could marshal analysis results: %s", err.Error())
 		}
 		fmt.Println(string(buildData))
+		return
 	}
 
-	if config.analyzeConfig.noUpload {
-		return
+	s.Stop()
+	s.Suffix = fmt.Sprintf(" Writing configuration...")
```
ilikebits: Why does this happen at the end?

cmd/fossa/main.go
```diff
@@ -328,21 +246,27 @@ func defaultCmd(c *cli.Context) {
...
+	s.Restart()
+
+	if err := config.WriteConfigFile(&conf); err != nil {
```
ilikebits: Does this work if you run the default command but already have a configuration file?

config/config.go
```diff
@@ -0,0 +1,333 @@
+package config
```
ilikebits: Use `git mv` to preserve file history.

config/config.go
```diff
@@ -0,0 +1,333 @@
...
+	/*configHeader := `
+	# Generated by FOSSA CLI (https://github.com/fossas/fossa-cli)
+	# Visit https://fossa.io to learn more
+	`*/
+
+	return ioutil.WriteFile(conf.ConfigFilePath, yamlConfig, 0777)
+}
+
+// ParseConfigFile parses a file and returns a config state
+func ParseConfigFile(filename string) (configFileV1, error) {
```
ilikebits: Why is this public?

config/config.go
```diff
@@ -0,0 +1,333 @@
...
+			Modules: conf.Modules,
+		},
+	}
+
+	yamlConfig, err := yaml.Marshal(writeConfig)
+	if err != nil {
+		return err
+	}
+
+	/*configHeader := `
```
ilikebits: Either remove this or add a `TODO`.

docs/user-guide.md
```diff
@@ -214,11 +214,27 @@ Print the version, then exit.
 ##### `-h, --help`
 Print a help message, then exit.
 
-<!-- ### `fossa init`
+### `fossa init`
 
-Makes a best-effort attempt at inferring the correct configuration, then outputs the configuration to `stdout`.
+Makes a best-effort attempt at inferring the correct configuration from current system state. If successful, it will write the configuration to a new or existing config file (defaults to `.fossa.yml`).
+
+If there are no modules defined in the configuration, it will scan the working directory for code modules to analyze.  You can pass the `--force` to
```
ilikebits: This sentence is not finished.

docs/user-guide.md
```diff
@@ -304,7 +317,7 @@ Uploads user-provided build data to FOSSA. This allows users to manually provide
 ```bash
 # You can manually override the project and revision name.
 # This is useful when `fossa` can't manually infer them from `git`.
-FOSSA_API_KEY=YOUR_API_KEY_HERE fossa upload --project=PROJECT_NAME --revision=SOME_HASH --data=$(fossa analyze --no-upload --output)
+FOSSA_API_KEY=YOUR_API_KEY_HERE fossa upload --project=PROJECT_NAME --revision=SOME_HASH --data=$(fossa analyze --output)
```
ilikebits: Why has this been changed? The code hasn't changed. This snippet will just upload twice.

module/module.go
```diff
@@ -1,37 +1,20 @@
...
-	// gopath vendoring
-	Golang = Type("golang")
-
-	VendoredArchives = Type("vendoredarchives")
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/fossas/fossa-cli/config"
```
ilikebits: Separate internal imports.

module/module.go
```diff
@@ -43,19 +26,64 @@ type Module struct {
...
-	// IsModule checks whether a build target is a valid module. This is used for
-	// inferring default configurations.
-	IsModule(target string) (bool, error)
-	InferModule(target string) (Module, error)
+// New instantiates and sets up a Module for a given ModuleType
+func New(moduleType config.ModuleType, conf config.ModuleConfig) (Module, error) {
+	var manifestName string
+	var moduleTarget string
+
+	// derive root dir of module
```
ilikebits: "Derive"? Also, `Sentence case`.

builders/ruby.go
lackstein: Pretty much every Ruby project will end up having its own `Gemfile`, so this is probably an okay way of finding them throughout the directory tree. But if they've got vendored dependencies, then those likely also have Gemfiles, and you'd be picking them up as well.

builders/maven.go
xizhao: Yes, we may want to reuse this among other builders, and potentially move it into `common.go`

builders/ruby.go
xizhao: That's fine, we will handle vendored deps separately in #52 via config, and this is desired behavior.

builders/sbt.go
xizhao: No, you can grep with reasonable success rate (maybe 75%?) like we do with the ruby implementation; from what i've seen in scala code so far.

cmd/fossa/init.go
xizhao: Tou may want to trust fossa's configuration to discover new modules (i.e. vendored deps) as a codebase changes and grows.

cmd/fossa/init.go
xizhao: What do you mean?

cmd/fossa/main.go
xizhao: We only want to write configuration if the defaultCmd was successful, esp useful in combination with `--force`, where we only want to re-write the configuration if successful.

cmd/fossa/main.go
xizhao: Also if the user doesn't want to reformat stuff

cmd/fossa/main.go
xizhao: Yes, it should normalize the config file

builders/sbt.go
ilikebits: I meant impossible in general. It's a bit misleading for this comment to suggest it's just "too hard" (unless you would consider the halting problem to be the same level of "too hard").

cmd/fossa/init.go
ilikebits: I mean, in reference to "to implement fail-open behavior, do not handle the err returned by this function":

1. "Fail-open" is only meaningful in the context of security, and doesn't mean anything here. It would be more accurate to say "fail-safe" but that's not quite accurate either.
2. Everybody knows they can ignore errors, this part of the comment adds no new information and is potentially confusing.

docs/user-guide.md
xizhao: I changed the behavior to remove `--no-upload` and just have `--output`, which will output and exist.

cmd/fossa/main.go
ilikebits: I don't understand. This is a flag on `fossa init`, a command which does nothing but generate an automatic configuration. This flag then throws away that configuration, causing `fossa init` to do nothing. Is that intended?

cmd/fossa/init.go
ilikebits: I don't think we should support this feature. Allowing the user to do this places a much higher burden on the expected quality of `fossa init`, because getting things wrong may silently break a codebase's build with no apparent user action.

builders/ruby.go
ilikebits: Bundler puts vendorised gems into `vendor` by default. We should filter those out, unless `includeAll` is specified.

docs/user-guide.md
ilikebits: I don't see those changes, it doesn't look like `--output` exits.

PR #66, opened 2018-02-28
Add `fossa report`

ilikebits:
This PR adds basic NOTICE generation. Reference links:

- https://github.com/apache/spark/blob/master/NOTICE
- https://github.com/nodejs/node/blob/master/LICENSE

### DIFF
--- a/cmd/fossa/main.go
+++ b/cmd/fossa/main.go
@@ -141,6 +141,7 @@ func main() {
 				cli.StringFlag{Name: "r, revision", Usage: revisionUsage},
 				cli.StringFlag{Name: "e, endpoint", Usage: endpointUsage},
 				cli.BoolFlag{Name: "debug", Usage: debugUsage},
+				cli.StringFlag{Name: "t, type", Usage: "the type of report to generate (either \"dependencies\" or \"licenses\""},
 			},
 		},
 	}
--- a/cmd/fossa/main.go
+++ b/cmd/fossa/main.go
@@ -131,6 +131,18 @@ func main() {
 				cli.BoolFlag{Name: "debug", Usage: debugUsage},
 			},
 		},
+		{
+			Name:   "report",
+			Usage:  "Generates a license report",
+			Action: reportCmd,
+			Flags: []cli.Flag{
+				cli.StringFlag{Name: "c, config", Usage: configUsage},
+				cli.StringFlag{Name: "p, project", Usage: projectUsage},
+				cli.StringFlag{Name: "r, revision", Usage: revisionUsage},
+				cli.StringFlag{Name: "e, endpoint", Usage: endpointUsage},
+				cli.BoolFlag{Name: "debug", Usage: debugUsage},
+			},
+		},
 	}
 
 	app.Run(os.Args)
--- a/cmd/fossa/main.go
+++ b/cmd/fossa/main.go
@@ -137,6 +137,7 @@ func main() {
 			Action: reportCmd,
 			Flags: []cli.Flag{
 				cli.StringFlag{Name: "c, config", Usage: configUsage},
+				cli.StringFlag{Name: "fetcher", Usage: fetcherUsage},
 				cli.StringFlag{Name: "p, project", Usage: projectUsage},
 				cli.StringFlag{Name: "r, revision", Usage: revisionUsage},
 				cli.StringFlag{Name: "e, endpoint", Usage: endpointUsage},
--- a/cmd/fossa/report.go
+++ b/cmd/fossa/report.go
@@ -28,7 +28,7 @@ type dependencyResponse struct {
 }
 
 type licenseResponse struct {
-	ID       string
+	ID       string `json:"spdx_id"`
 	Title    string
 	FullText string
 }
--- a/cmd/fossa/report.go
+++ b/cmd/fossa/report.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"net/url"
+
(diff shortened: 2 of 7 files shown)


cmd/fossa/main.go
```diff
@@ -125,6 +125,18 @@ func main() {
 				cli.BoolFlag{Name: "debug", Usage: debugUsage},
 			},
 		},
+		{
+			Name:   "report",
```
xizhao: I think we should allow for multiple report formats.

`fossa report dependencies` (default), `fossa report licenses`, `fossa report vulns`

cmd/fossa/report.go
```diff
@@ -0,0 +1,87 @@
...
+	Licenses []licenseResponse
+	Project  struct {
+		Title   string
+		URL     string
+		Authors []string
+	}
+}
+
+type licenseResponse struct {
+	ID       string
```
xizhao: should include SPDX id; we should prefer using SPDX in general

cmd/fossa/report.go
```diff
@@ -0,0 +1,87 @@
...
+	conf, err := config.New(c)
+	if err != nil {
+		reportLogger.Fatalf("Could not load configuration: %s", err.Error())
+	}
+
+	server, err := url.Parse(conf.Endpoint)
+	if err != nil {
+		reportLogger.Fatalf("Invalid FOSSA endpoint: %s", err.Error())
+	}
+	api, err := server.Parse(fmt.Sprintf("/api/revisions/%s/dependencies", url.PathEscape(config.MakeLocator(conf.Project, conf.Revision))))
```
lackstein: Why is `MakeLocator` in the `config` package?

docs/user-guide.md
```diff
@@ -342,19 +342,21 @@ Print debugging information to `stderr`.
...
 Print the project's license report.
 
 #### Flags
-##### `-t, --type report_type`
-Print a specific type of license report for automatically creating attribution files. Possible report types include:
-- `NOTICE`: generate a `NOTICE` file for your dependencies
-- `ATTRIBUTION`: generate an `ATTRIBUTION` file for your dependencies
-
-##### `-j, --json`
+<!-- ##### `-j, --json`
```
xizhao: Why remove report type?

cmd/fossa/report.go
```diff
@@ -0,0 +1,87 @@
...
+
+The following software have components provided under the terms of this license:
+
+`, license.Title)
+		for _, dep := range deps {
+			fmt.Printf("- %s (from %s)\n", dep.Project.Title, dep.Project.URL)
+		}
+	}
+	fmt.Println()
+}
```
xizhao: A thought -- in the future maybe we could fallback to querying package registries if we're missing license data.

cmd/fossa/report.go
xizhao: This depends on `fossa upload` succeeding as well as all deep deps to be resolved.  I would imagine the implementation would query our `/api/revisions` and `/api/licenses` api.  

I prefer for the `report` command to be independent from if the user was integrated with FOSSA; and we can offer "better" data if you used a FOSSA account.

In cases where we hit stuff that hasn't been scanned, we could rely on the `fetcher.getMetadata()` responses which we'd need to build support for in FOSSA.

If we're running with this path as our v1 implementation, I would:

 1. File an issue proposing the fossa-independent report design
 2. Give user feedback that you need `fossa upload` to succeed for this command to succeed

ilikebits: Legacy. I'll refactor it out later, I think.

xizhao: It kinda makes sense as a config util, to take config params and make it into a certain form.

cmd/fossa/report.go
ilikebits: Fixed.

cmd/fossa/main.go
ilikebits: This is a really good idea.

docs/user-guide.md
ilikebits: I was originally thinking of report types in the context of licenses i.e. generating attribution vs. notice files. I couldn't figure out a good way to automatically generate different types of license reports, so I removed this.

This is going to be added back in to differentiate between e.g. dependency vs. license reports.

cmd/fossa/main.go
ilikebits: This is a good idea. I was thinking of report types as different kinds of notices to generate (e.g. NOTICE, ATTRIBUTION, LICENSE) but this makes a lot more sense. I'll implement this.

cmd/fossa/report.go
ilikebits: Will fix.

docs/user-guide.md
ilikebits: See above.

cmd/fossa/report.go
ilikebits: Will implement.

ilikebits:
PTAL.

cmd/fossa/main.go
```diff
@@ -131,6 +131,19 @@ func main() {
...
+			Usage:  "Generates a license report",
+			Action: reportCmd,
+			Flags: []cli.Flag{
+				cli.StringFlag{Name: "c, config", Usage: configUsage},
```
anuccio1: fetcher?

ilikebits: Good catch. We're going to switch these to use global flags.

ilikebits:
As per offline discussion, I'm going to merge this in and override merge checks.

PR #74, opened 2018-03-02
Gradle support

ilikebits:
Follow-up work:

- Install Android SDK prerequisites for Gradle fixture inside Dockerfile (or document how to, since we probably don't want to accept licenses on behalf of users)
- Add syntax for providing module names on the command line (for Gradle builder)

### DIFF
--- a/builders/gradle.go
+++ b/builders/gradle.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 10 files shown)


builders/gradle.go
```diff
@@ -0,0 +1,149 @@
...
+// Initialize collects metadata on Java and Maven binaries
+func (builder *GradleBuilder) Initialize() error {
+	gradleLogger.Debug("Initializing Gradle builder...")
+
+	// Set Gradle context variables
+	gradleCmd, gradleVersionOut, err := which("--version --offline", os.Getenv("GRADLE_BINARY"), "./gradlew", "gradle")
+	if err == nil {
+		builder.GradleCmd = gradleCmd
+
+		gradleVersionMatchRe := regexp.MustCompile(`Gradle ([0-9]+\.[0-9]+.\w+)`)
```
ilikebits: It's probably more useful (for debugging purposes) to just capture the entire `--version` output rather than checking for a version.

builders/gradle.go
```diff
@@ -0,0 +1,149 @@
...
+	return nil
+}
+
+// Analyze parses the output of `gradle -q app:dependencies`
+func (builder *GradleBuilder) Analyze(m module.Module, allowUnresolved bool) ([]module.Dependency, error) {
+	gradleLogger.Debugf("Running Gradle analysis: %#v %#v in %s", m, allowUnresolved, m.Dir)
+
+	// TODO: We need to let the user configure the right configurations
+	dependenciesOutput, err := exec.Command(builder.GradleCmd, m.Name+":dependencies", "-q", "--configuration=compile", "--offline", "-a").Output()
+	// Do not handle error, as gradle will exit with status 1 regardless
```
ilikebits: I'm not sure what this comment means -- aren't you handling `err` just below?

builders/gradle.go
```diff
@@ -0,0 +1,149 @@
...
+	if len(dependenciesOutput) == 0 || err != nil {
+		return nil, fmt.Errorf("could not run Gradle task %s:dependencies", m.Name)
+	}
+
+	var deps []module.Dependency
+	dependenciesRe := regexp.MustCompile(`- ([\w\.-]+):([\w\.-]+):([\w\.-]+)( -> ([\w\.-]+))?`)
```
ilikebits: This regex is unmaintainable without a very detailed comment. I think it would be much cleaner to loop through each line and check for `---` instead (which prefixes all lines with dependency names).

builders/gradle.go
```diff
@@ -0,0 +1,149 @@
...
+	return false, errors.New("IsModule is not implemented for GradleBuilder")
+}
+
+// DiscoverModules finds either a root build.gradle file in the specified dir
+func (builder *GradleBuilder) DiscoverModules(dir string) ([]config.ModuleConfig, error) {
+
+	if err := builder.Initialize(); err != nil {
+		return nil, err
+	}
+	// Look for the root Gradle build
```
ilikebits: These comments are out of date.

builders/gradle.go
```diff
@@ -0,0 +1,149 @@
...
+		if len(taskListOutput) > 0 && gradleTaskErr == nil {
+			// Search for subprojects using Gradle task list instead of grepping for build.gradle
+			var moduleConfigurations []config.ModuleConfig
+			// NOTE: this leaves out the root ("") dependencies task. To include, replace with `(\w+:)?dependencies -`
+			taskListRe := regexp.MustCompile(`\w+:dependencies -`)
+			for _, line := range strings.Split(string(taskListOutput), "\n") {
+				trimmed := strings.TrimSpace(line)
+				if len(trimmed) > 0 {
+					depMatchIndicies := taskListRe.FindStringIndex(trimmed)
+					if len(depMatchIndicies) > 0 && depMatchIndicies[0] == 0 {
```
ilikebits: *`Indices`

builders/gradle.go
xizhao: That's not true, since it renders as a tree the parent nodes have a different prefix.  There is no way to guarantee a specific format for each level as well.

The only common prefix i found was `- `

builders/gradle.go
xizhao: How so?

builders/gradle.go
ilikebits: Really? I see all nodes being prefixed with either `\--- ` or `+--- `. Is this because we're using different versions? Do you have an example project where this isn't true?

builders/gradle.go
ilikebits: We're parsing `gradle tasks` output instead of looking for `*.gradle` files now. I might have been looking at an older commit.

ilikebits:
LGTM, modulo nits. @xizhao, can you LGTM this PR when ready? I opened it so it won't let me LGTM.

xizhao:
Gradle still needs the build cmd, it's missing APIs

PR #75, opened 2018-03-02
Fix/fix builder paths

xizhao:
Some fixes and ignores, please merge if approved.


builders/nodejs.go
```diff
@@ -185,6 +185,7 @@ func (builder *NodeJSBuilder) DiscoverModules(dir string) ([]config.ModuleConfig
 			}
 
 			nodejsLogger.Debugf("Found NodeJS package: %s (%s)", path, moduleName)
+			path, _ = filepath.Rel(dir, path)
```
ilikebits: I'd prefer if you didn't mutate `path`, and instead gave this a different name (e.g. `manifestPath`).

PR #89, opened 2018-03-07
added ability to specify the logging level you want output from our l…

anuccio1:


config/config.go
```diff
@@ -110,13 +111,42 @@ func parseModulesFlag(moduleFlag string) ([]ModuleConfig, error) {
...
+		return logging.WARNING, nil
+	case "notice":
+		return logging.NOTICE, nil
+	case "info":
+		return logging.INFO, nil
+	case "debug":
+		return logging.DEBUG, nil
+	}
+
+	return -1, fmt.Errorf("unknown logging level: %s", logFlag)
```
anuccio1: not 100% what best practice is here. 0 technically is logging.DEBUG

ilikebits:
I'm not convinced that logging level should be configurable beyond debug/not-debug. My reasoning:

- We should keep configuration as simple as possible.
- I'm not aware of a concrete use case where a user would want granular control over the messages they get from `fossa`. All user-interactive messages are printed to `STDERR` and is separate from actual output in `STDOUT`, and a user that _really_ cares can run `--debug` which emits interactive messages in a machine-filterable format.

anuccio1:
If they're parsing stdout JSON and don't want warning messages

ilikebits:
Those messages should be written to `STDERR`. If they're not, the better fix would be to make sure they're sent to `STDERR`.

ilikebits:
Closing. Please reopen if you feel this is important.

PR #96, opened 2018-03-09
Do Go import tracing with `go list` instead of `depth`

### LINKED ISSUE #63: Go analysis fails from within Docker containers
Steps to reproduce:

1. Download alvin-huang/go-hello-world.
2. Download `fossa` v0.3.1.
3. Mount `go-hello-world` in its $GOPATH location within the `golang:1.10-stretch` container
4. Run `fossa analyze --modules go:.` from within the `go-hello-world` directory in the container.

Interestingly, `fossa` works if you build it from source from within the Docker container. I suspect this has to do with our usage of KyleBanks/depth for import analysis -- something about a difference between the environment's Go and the Go used to compile `fossa` is breaking things.

ilikebits:
Fixes #63.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -8,12 +8,6 @@
   version = "v0.3.0"
 
 [[projects]]
-  name = "github.com/KyleBanks/depth"
-  packages = ["."]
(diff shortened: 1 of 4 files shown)


ilikebits:
@xizhao is OOO, but please review retroactively. Merging.

PR #97, opened 2018-03-09
Implement Go module discovery

ilikebits:
I'm a bit worried here about using `go/parser` and `go/token` because of our previous issues with #63, but parsing should work consistently across Go versions because of the Go 1 compatibility promise.

This PR also uses `fossa init` to regenerate our configuration, as a sanity check.


ilikebits:
@xizhao Please review retroactively.

PR #98, opened 2018-03-09
Improve Makefile, add multiple targets

seemethere:
Was trying to do a `make build` on the project and found that it didn't work. 

Adds targets to do things like:
* build
* install
* vendor

There'll be a follow up PR to do the vendoring as well so that you can actually build from the source directory instead of having to install all the dependencies into your `$GOPATH/src/`

ping @ilikebits 

Signed-off-by: Eli Uriegas <eli.uriegas@docker.com>


ilikebits:
Gotcha. Saw `vendor` PR before this, and this makes a lot of sense now. LGTM, will merge in once we get a CLA set up.

ilikebits:
We've got a CLA set up using SAP's CLA assistant (see "details" link in the PR merge check or direct link [here](https://cla-assistant.io/fossas/fossa-cli?pullRequest=98)). Our CLA is modelled after the ASF's, so everything in it should be pretty standard. The intent here is to allow us to potentially relicense down the road without needing to chase down all contributors. Happy to merge this and #99 in as soon as you sign!

PR #99, opened 2018-03-09
 Introduce vendor directory into the repo

seemethere:
Was trying to build the CLI from the source directory and I could not,
this re-adds the vendor directory so we can build the CLI from the
source repository instead of having to `dep ensure` everytime

ping @ilikebits


seemethere:
Not really seeing why the tests are failing?

ilikebits:
The tests are failing because our tests have a dependency on the FOSSA API, which requires us to pass a secret API key to the tests. This is not passed to tests running on forked PR branches for security reasons. We should probably turn off the API tests for non-master commits.

I'm not sure that committing `vendor` is a best practice, and installing with `go get -u github.com/fossas/fossa-cli/cmd/fossa` works for me. What are the steps for reproducing your installation failure, and what are the strong arguments for checking in `vendor`? (The main argument I can think of against committing `vendor` is repository size.)

seemethere:
@ilikebits Committing a `vendor` is standard practice for almost every `golang` project on Github.

See:

* [kubernetes/kuberenetes](https://github.com/kubernetes/kubernetes/tree/master/vendor)
* [moby/moby](https://github.com/moby/moby/tree/master/vendor)
* [ethereum/go-ethereum](https://github.com/ethereum/go-ethereum/tree/master/vendor)
* [containous/traefik](https://github.com/containous/traefik/tree/master/vendor)
* [gin-gonic/gin](https://github.com/gin-gonic/gin/tree/master/vendor)

What it allows you to do is not pollute your `$GOPATH/src/` with dependencies from a `go get -u`

Issue #114, opened 2018-03-20
Support gradle root dependencies task


xizhao:
Currently we intentionally filter out the root dependencies task for gradle.

If we discover no other dependencies task other than the root dependencies, we should configure to include the root dep task.

Issue #10, opened 2018-02-06
Better README and docs


xizhao:
Improve README and docs, some inspiration:
 - https://github.com/matiassingers/awesome-readme
 
Add a "who's using" section with customer logo list.

ilikebits:
Closed for now by #34.

PR #116, opened 2018-03-20
Feat/support gradle root deps

xizhao:
This PR adds support for Gradle projects that manage their dependencies via a root `dependencies` task rather than within a subtask.

Refs: #74 #114 #10 

# Changes
- When `fossa init` does not find dependency subtasks, have it elect the root dependencies task for the `name` field in `fossa.yml` configuration (represented as an empty taskname + config `:compile`)
- Support falling back to root dependency tasks when task name is unspecified in config
- Strip ANSI from dependency output for more reliable parsing
- Add extensive documentation + troubleshooting for our gradle integration
- Add some configuration handling improvements

# How to configure
For users that want to support root dep tasks, this should be a non-breaking change to existing config as well as provide OOTB support for new `fossa init`.

Options: 
 1. Overwrite existing config with `fossa init -O` and confirm the right task/configuration in `fossa.yml`

OR

 2. Change your `.fossa.yml` file to have an empty task name; optionally specify your taskConfiguration:

```yaml
analyze:
  modules:
    - name: :compile # this is valid, and will run `gradle dependencies --configuration=compile`
      path: build.gradle
      type: gradle
    - name: ''  # this is also valid, and will run `gradle dependencies --configuration=compile`
      path: build.gradle
      type: gradle
```


docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+### Reporting issues
+
+If you're still having problems, open up an issue on this repo with the full logs/output of `fossa -o --debug` and your `fossa.yml` file.
+
+Optionally, provide your `build.gradle` for more context.
+
+## Roadmap
 
-### Analysis
+If you'd like to help us improve our Gradle integration, we have
```
ilikebits: ?

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+
+Optionally, provide your `build.gradle` for more context.
+
+## Roadmap
 
-### Analysis
+If you'd like to help us improve our Gradle integration, we have 
 
-Analysis will support any tasks you can run `:dependendencies` on with the "compile" configuration.  This is equivalent to `gradle app:dependencies --configuration=compile`.
+ 1. Support multiple independent "root" Gradle builds that may be segregated in a codebase
```
ilikebits: These should probably also be tracked as issues, tagged with `help wanted`.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
 
 ### Automatic Configuration
 
-When running `fossa init`, FOSSA will look for a root Gradle build (`build.gradle`) and interrogate it for build-able tasks.  From there, it will write module configurations into your `.fossa.yml` for every task that supports a `:dependency` sub-task.
+An average Gradle build, there are often hundreds of configuration paths that can influence what dependencies are included.  In order to determine what tasks and configurations you care about, `fossa` will attempt a "best-effort" strategy during `fossa init` to discover the most likely configuration for a production-build scenario.
```
ilikebits: Grammar.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+
+When running `fossa init`, FOSSA executes the following steps:
+
+ 1. Look for a root Gradle build (`build.gradle`)
+ 2. Run and parse the output of `gradle tasks --all -q -a --offline` to seek all available sub-tasks that support a `:dependencies` command
```
ilikebits: s/seek/find

docs/integrations/gradle.md
ilikebits: This is all kind of just marketing noise. I'd remove it.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+ 3. If task list succeeds but no sub-tasks are available, fallback to the root `dependencies` task in `build.gradle`
+ 4. Filter out any suspicious-looking tasks (i.e. labeled `test` or `testCompile`)
```
ilikebits: You should mention that this can be disabled by flag.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+analyze:
+  modules:
+    - name: {task}:{configuration}
+      path: {path-to-build.gradle}
+      type: gradle
+```
+
+For Gradle modules, `fossa` requires a few peices of information to run dependency analysis:
+
+- `task` - the name of the task/subtask used to generate your build (most often this is `app` or empty `` for root)
```
ilikebits: Empty is `''`.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,103 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
...
+- `configuration` - the configuration your task runs during a production build (if undefined, defaults to `compile`)
+- `path` - path to your `*.gradle` build file (usually this is just `build.gradle`)
```
ilikebits: Relative or absolute? Either? If relative, relative to what?

builders/gradle.go
```diff
@@ -58,14 +58,20 @@ func (builder *GradleBuilder) Analyze(m module.Module, allowUnresolved bool) ([]
...
+	if taskName != "" {
+		// this is not the root gradle project
+		taskName = taskName + ":"
+	}
+	if len(moduleConfigurationKey) == 2 && moduleConfigurationKey[1] != "" {
 		taskConfiguration = moduleConfigurationKey[1]
 	}
 
-	// TODO: We need to let the user configure the right configurations
 	// NOTE: we are intentionally using exec.Command over runLogged here, due to path issues with defining cmd.Dir
```
ilikebits: What is the issue described in this comment?

builders/gradle.go
```diff
@@ -58,14 +58,20 @@ func (builder *GradleBuilder) Analyze(m module.Module, allowUnresolved bool) ([]
 	moduleConfigurationKey := strings.Split(m.Name, ":")
 	taskConfiguration := "compile"
 	taskName := moduleConfigurationKey[0]
-	if len(moduleConfigurationKey) == 2 {
+	if taskName != "" {
+		// this is not the root gradle project
+		taskName = taskName + ":"
```
ilikebits: 1. Avoid mutation, it makes your intent much less clear.
2. I'd locate this closer to where the variable is used, so it's clear that this is formatting for usage as an argument to `gradle`.

builders/gradle.go
```diff
@@ -58,14 +58,20 @@ func (builder *GradleBuilder) Analyze(m module.Module, allowUnresolved bool) ([]
...
+	}
+	if len(moduleConfigurationKey) == 2 && moduleConfigurationKey[1] != "" {
```
ilikebits: Move this up to be next to the definition of `taskConfiguration`, so it's clear that your intent is to use a default value.

builders/gradle.go
```diff
@@ -139,10 +145,21 @@ func (builder *GradleBuilder) DiscoverModules(dir string) ([]module.Config, erro
...
+				{
+					Name: ":compile",
+					Path: "build.gradle",
+					Type: "gradle",
+				},
+			}, nil
 		}
 
-		// Fall back to "app" as default task, even though technically it would be "" (root)
+		// If task list fails, fall back to "app" as default task, even though technically it would be "" (root)
```
ilikebits: Is this behaviour correct? Should we expect `app:compile` to work even when `gradle tasks` fails, or should this be a fatal error?

xizhao: I'm 50/50 on having this fallback vs erroring if `gradle tasks` fails, but:

 1. I believe `app:compile` is a sane default if a `build.gradle` is found
 2. Checking for a `build.gradle` is a valid way of identifying a module, and much simpler/more reliable then parsing output of `gradle tasks`

builders/gradle.go
```diff
@@ -56,18 +56,21 @@ func (builder *GradleBuilder) Analyze(m module.Module, allowUnresolved bool) ([]
...
 
 	moduleConfigurationKey := strings.Split(m.Name, ":")
-	taskConfiguration := "compile"
-	taskName := moduleConfigurationKey[0]
-	if taskName != "" {
-		// this is not the root gradle project
-		taskName = taskName + ":"
+
+	taskName := "dependencies" // defaults to root gradle dependencies task
+
```
ilikebits: Remove newline (makes default-value intention clearer).

builders/gradle.go
ilikebits: I think the question is whether `app:compile` is likely to work. If it works, then this is great. If it doesn't work, then we've kicked a configuration error down the line into a much more obscure runtime error.

I don't have much experience with the Gradle ecosystem, so I trust your intention on this.

docs/integrations/gradle.md
```diff
@@ -11,24 +11,101 @@ Gradle support in FOSSA CLI depends on the following tools existing in your envi
 
 ### Automatic Configuration
 
-When running `fossa init`, FOSSA will look for a root Gradle build (`build.gradle`) and interrogate it for build-able tasks.  From there, it will write module configurations into your `.fossa.yml` for every task that supports a `:dependency` sub-task.
+`fossa init` will attempt a "best-effort" strategy to look through all available Gradle tasks/configurations and elect the most likely ones used for a production build.
+
+ 1. Look for a root Gradle build (`build.gradle`)
+ 2. Run and parse the output of `gradle tasks --all -q -a --offline` to find all available sub-tasks that support a `:dependencies` command
+ 3. If task list succeeds but no sub-tasks are available, fallback to the root `dependencies` task in `build.gradle`
+ 4. Filter out any suspicious-looking tasks (i.e. labeled `test` or `testCompile`)
```
ilikebits: You should really mention that there's a flag to disable this.

PR #134, opened 2018-05-09
Feat/ant support

xizhao:
This provides support for Any / Ivy builds that vendorize Jar dependencies in library directories


builders/ant.go
```diff
@@ -0,0 +1,222 @@
...
+		var pomFilePath string
+		for _, file := range info.Files {
+			if strings.HasPrefix(file, "META-INF") && strings.HasSuffix(file, "pom.xml") && (pomFilePath == "" || len(pomFilePath) > len(file)) {
+				pomFilePath = file
+			}
+		}
+
+		if pomFilePath != "" {
+			// open the jar
+			if jarFile, err := os.Open(path); err == nil {
```
ilikebits: You should return on `err != nil` instead to avoid the pyramid of doom.

builders/ant.go
```diff
@@ -0,0 +1,222 @@
...
+	return errors.New("Ant builder is not implemented")
+}
+
+// Analyze resolves a lib directory and parses the jars inside
+func (builder *AntBuilder) Analyze(m module.Module, allowUnresolved bool) ([]module.Dependency, error) {
+	antLogger.Debugf("Running Ant analysis: %#v %#v in %s", m, allowUnresolved, m.Dir)
+
+	options := m.Context.(map[string]interface{})
+
+	libdir := options["libdir"].(string)
```
ilikebits: Have you tested this? I'm not sure this is actually plumbed through.

builders/ant.go
```diff
@@ -0,0 +1,222 @@
...
+			})
+		} else {
+			antLogger.Warningf("unable to resolve Jar: %s", jarFilePath)
+		}
+	}
+
+	return dependencies, nil
+}
+
+// locatorFromJar resolves a locator from
```
ilikebits: This is not a complete sentence.

builders/ant.go
xizhao: See: https://github.com/fossas/fossa-cli/pull/134/files#diff-88e008ed0bc6f169079d9bde601e0edcR107

builders/ant.go
xizhao: I'd like to fall back to decoding META-INF and then jarname, and it seems easier to do this with nesting `err == nil` then returning an error case every time

ilikebits: You should refactor this pyramid of doom into a separate function that returns on `err != nil` and then call that function. This nesting seriously affects readability.

PR #143, opened 2018-06-01
Feat/compute dependency hashes

xizhao:
WIP, proposal to support using dependency hashes as an alternative resolution method for builds that heavily depend on vendorized dependencies.


builders/builderutil/hashes.go
```diff
@@ -0,0 +1,42 @@
+package builderutil
+
+import (
+	"crypto/md5"
```
golangci: G501: Blacklisted import crypto/md5: weak cryptographic primitive

builders/builderutil/hashes.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	defer f.Close()
+
+	sha1Hash := sha1.New()
+	if _, err := io.Copy(sha1Hash, f); err != nil {
+		return hashes, err
+	}
+	hashes.SHA1 = string(sha1Hash.Sum(nil))
+
+	md5Hash := md5.New()
```
golangci: G401: Use of weak cryptographic primitive

builders/builderutil/hashes.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	defer f.Close()
+
+	sha1Hash := sha1.New()
+	if _, err := io.Copy(sha1Hash, f); err != nil {
+		return hashes, err
+	}
+	hashes.SHA1 = hex.EncodeToString(sha1Hash.Sum(nil))
+
+	md5Hash := md5.New()
```
golangci: G401: Use of weak cryptographic primitive

builders/builderutil/hashes.go
```diff
@@ -0,0 +1,43 @@
+package builderutil
+
+import (
+	"crypto/md5" // #nosec
```
golangci: G501: Blacklisted import crypto/md5: weak cryptographic primitive

builders/builderutil/hashes.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+	defer f.Close()
+
+	sha1Hash := sha1.New()
+	if _, err := io.Copy(sha1Hash, f); err != nil {
+		return hashes, err
+	}
+	hashes.SHA1 = hex.EncodeToString(sha1Hash.Sum(nil))
+
+	md5Hash := md5.New() // #nosec
```
golangci: G401: Use of weak cryptographic primitive

ilikebits:
I'm not convinced that dependency hashing is the right way forward for detecting vendored dependencies. I'm _especially_ not convinced that using standard MD5/etc. hashing is the correct way to implement dependency hashing. I suspect this implementation might be JAR-specific.

Here are two other examples where hashing might help us:
1. Vendored Go dependencies. They're generally unmodified, but the `vendor/` folder doesn't contain information about which _revision_ of a package was used. Go has its own weird quirks about parts within a repository are and are not considered part of a package, and each tool has its own idiosyncrasies on top of that. In this case, I think that we would want to do a tree hash of only `*.go` files.
2. Vendored JavaScript dependencies. If somebody drops `jquery.min.js` into a folder, we want to detect that. The issue here is that `jquery.js` comes in many forms: source, minified, uglified, compressed, comments stripped, etc. We might be able to hash and then match many hashes to a single revision. Alternative strategies include looking exclusively for file name (in one extreme) to doing AST parsing (at the other extreme).

I don't think it's clear that this is a good design for hashing dependencies in general, but I'm happy to LGTM for JARs specifically if that is the intention and you move this into the Ant builder package.

xizhao:
We need a narrative to address a package that we truly don't have access to, but is considered 3rd-party; like an old vendorized JAR from who knows where.

module/hashes.go
```diff
@@ -0,0 +1,8 @@
+package module
+
+// Hashes contains hexadecimal checksums of code libraries brought in by running a Build
+type Hashes struct {
```
ilikebits: Why are we computing all of these? Isn't there just one hash that's relevant per dependency?

PR #144, opened 2018-06-01
fix(builders): fix go-bindata error

xizhao:


ilikebits:
This build error occurs on an out-of-date commit.

PR #149, opened 2018-06-05
feat(install): modify windows install script

xizhao:
Repurposed @joshuapetryk's execution script as a windows installer and updated our README.

@joshuapetryk I'm not on a Windows machine -- would appreciate your review on this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,12 +34,17 @@
 
 ## Installation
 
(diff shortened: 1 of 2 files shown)


joshuapetryk:
@xizhao sorry, wasn't able to look at it this week, but I am looking at it now. There is a small syntax issue with the Join-Path command in install.ps1, I will submit a PR.

PR #151, opened 2018-06-09
Fixes for Powershell install script

joshuapetryk:
@xizhao I made some fixes to the Powershell install script. Also added support for Powershell streams, so someone could run `$fossa = install.ps1` and the variable will have the .exe path.


xizhao:
Amazing, thank you @joshuapetryk

PR #159, opened 2018-06-13
Properly split arguments

renaud-yahoo:
Arguments not properly whitespace-split may result in "two" arguments being considered as one unique, long string, and interpreted by the command to execute as one argument.


ilikebits:
Thanks for the PR! We fixed this with 62ae5104247345014620f2703c5b16b746d08d92, which changes the application call instead of the library.

PR #160, opened 2018-06-15
fix(builders): fix hash calculation

xizhao:
This PR fixes a bug that we have with our hash calculation.

Since we were using the same file stream to calculate the hash in serial, the reader was reaching the end of the file for subsequent hashes after sha1, resulting all other hashes to return the values of hashing empty data.

This solves this problem by paging the file into a multiplexer, which allows us to calculate all hashes simultaneously with one read.

### DIFF
--- a/builders/builderutil/hashes.go
+++ b/builders/builderutil/hashes.go
@@ -1,6 +1,7 @@
 package builderutil
 
 import (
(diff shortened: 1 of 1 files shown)


ilikebits:
LGTM.

PR #161, opened 2018-06-18
Add ability to specify an output file and a template file

djgilcrease:
* Changes the -o / --output option to be a file `fossa -o deps.json` for the original behavior you would do `fossa -o -` or `fossa report --type licenses -o NOTICE.txt`
* Added the -t / --template option to process the output via a template file first.

Default template for licenses report is
```
# 3rd-Party Software License Notice
   
Generated by fossa-cli (https://github.com/fossas/fossa-cli).

This software includes the following software and licenses:

{{range $license, $deps := .}}
{{$license.Title}}

The following software have components provided under the terms of this license:
{{range $i, $dep := $deps}}
- {{$dep.Project.Title}} (from {{$dep.Project.URL}})
{{- end}}
{{end}}
```
which will produce the same output as before.


cmd/fossa/analyze.go
```diff
@@ -31,22 +32,56 @@ func analyzeCmd(c *cli.Context) {
...
+		}
+	} else if len(msg) == 0 {
+		msg, err = json.Marshal(normalModules)
+		if err != nil {
+			log.Logger.Fatalf("Could not marshal analysis results: %s", err.Error())
+		}
+	}
+
+	defer out.Close()
+	out.Write(msg)
```
golangci: Error return value of `out.Write` is not checked

cmd/fossa/report.go
```diff
@@ -156,18 +189,27 @@ func reportCmd(c *cli.Context) {
...
+				log.Logger.Fatalf("Could not marshal analysis: %s", err.Error())
+			}
 		}
-		fmt.Println(string(out))
 	default:
 		log.Logger.Fatalf("Report type is not recognized (supported types are \"dependencies\" or \"licenses\": %s", conf.ReportCmd.Type)
 	}
+
+	defer out.Close()
+	out.Write(msg)
```
golangci: Error return value of `out.Write` is not checked

djgilcrease:
Closing to fix CLA signing requirement

PR #171, opened 2018-06-21
Implement the report command.

djgilcrease:
* Changes the -o / --output option to be a file `fossa analyze -o deps.json` for the original behavior you would do `fossa analyze -o -`
* Split the report types into sub commands `fossa report licenses -o NOTICE.txt` || `fossa report dependencies --output deps.json`
* Added the -t / --template option to process the output via a template file first. `fossa report dependencies --output deps.txt --template deps.tmpl`

Default template for licenses report is
```
# 3rd-Party Software License Notice
Generated by fossa-cli (https://github.com/fossas/fossa-cli).
This software includes the following software and licenses:
{{range $license, $deps := .}}
{{$license}}
The following software have components provided under the terms of this license:
{{range $i, $dep := $deps}}
- {{$dep.Project.Title}} (from {{$dep.Project.URL}})
{{- end}}
{{end}}
```
which will produce the same output as before.

Since `fossa analyze -o deps.json` && `fossa report dependencies --output deps.json` should output the same info should analyze still have the `-o` or should people just use `fossa report dependencies`?


api/fossa/revisions.go
```diff
@@ -0,0 +1,109 @@
+package fossa
+
+import (
+	"net/url"
+
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+type License struct {
+	Id int64 `json:"id"`
```
golangci: File is not goimports-ed

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -1,8 +1,11 @@
 package analyze
 
 import (
+	"os"
```
golangci: File is not goimports-ed

cmd/fossa/cmd/report/dependencies.go
```diff
@@ -0,0 +1,41 @@
...
+	"text/template"
+
+	"github.com/fossas/fossa-cli/cmd/fossa/flags"
+	"github.com/fossas/fossa-cli/log"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/urfave/cli"
+)
+
+var dependenciesCmd = cli.Command{
+	Name:      "dependencies",
```
golangci: File is not goimports-ed

api/fossa/revisions.go
```diff
@@ -0,0 +1,109 @@
...
+	LicenseId string `json:"licenseId"`
```
golangci: struct field LicenseId should be LicenseID

api/fossa/revisions.go
```diff
@@ -0,0 +1,109 @@
...
+	RevisionID string `json:"revisionId"`
+	LicenseGroupId int64 `json:"licenseGroupId"`
```
golangci: struct field LicenseGroupId should be LicenseGroupID

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -45,7 +50,26 @@ func Run(ctx *cli.Context) error {
...
+	}
+
+	if ctx.String(Output) != "" || ctx.String(Template) != "" {
+		return outputAnalysis(ctx.String(Output), ctx.String(Template), normalized)
+	}
+
+	return uploadAnalysis(normalized)
+}
+
+func AnalyzeModules(modules []module.Module) (analyzed []module.Module, err error) {
```
golangci: func name will be used as analyze.AnalyzeModules by other packages, and that stutters; consider calling this Modules

api/fossa/revisions.go
```diff
@@ -0,0 +1,109 @@
+package fossa
+
+import (
+	"net/url"
+
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+type License struct {
+	Id             int64  `json:"id"`
```
golangci: struct field Id should be ID

api/fossa/locator.go
```diff
@@ -69,8 +80,8 @@ func ReadImportPath(s ImportPathString) ImportPath {
 	return out
 }
 
-func LocatorOf(id pkg.ID) Locator {
-	return Locator{
+func LocatorOf(id pkg.ID) *Locator {
```
ilikebits: Why return a pointer here?

cmd/fossa/main.go
```diff
@@ -3,6 +3,7 @@ package main
 import (
 	"os"
 
+	"github.com/fossas/fossa-cli/cmd/fossa/cmd/report"
```
ilikebits: [nit] This import should go in the block with all the other `cmd`s.

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -25,7 +29,8 @@ var Cmd = cli.Command{
 	Action:    Run,
 	ArgsUsage: "MODULE",
 	Flags: flags.WithGlobalFlags(flags.WithAPIFlags(flags.WithModulesFlags([]cli.Flag{
-		cli.BoolFlag{Name: flags.Short(Output), Usage: "show analysis output instead of uploading to FOSSA"},
+		cli.StringFlag{Name: flags.Short(Output), Usage: "send analysis to output file instead of uploading to FOSSA (default: -)"},
```
ilikebits: I'd rather default to output on `STDOUT`. It seems like it should already be easy for a user to `fossa analyze --output > file`.

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -25,7 +29,8 @@ var Cmd = cli.Command{
...
+		cli.StringFlag{Name: flags.Short(Template), Usage: "process analysis via template prior to sending it to output"},
```
ilikebits: I'd like this usage string to mention that this should be a file name (as opposed to a raw string template).

cmd/fossa/cmd/report/dependencies.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/fossas/fossa-cli/log"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/urfave/cli"
+)
+
+var dependenciesCmd = cli.Command{
+	Name:  "dependencies",
+	Usage: "Generate dependencies report",
+	Flags: flags.WithGlobalFlags(flags.WithAPIFlags(flags.WithModulesFlags([]cli.Flag{
+		cli.StringFlag{Name: flags.Short(Output), Destination: &outputFlag, Value: "-", Usage: "Output file for report"},
```
ilikebits: It's helpful to add flags used in multiple places to `cmd/fossa/flags`. It conveys the intent that these flags are identical, and prevents their usage strings from drifting out of sync.

cmd/fossa/cmd/report/licenses.go
```diff
@@ -0,0 +1,90 @@
...
+func generateLicenses(ctx *cli.Context) (err error) {
+	defer log.StopSpinner()
+	revs := make(fossa.Revisions, 0)
+	for _, module := range analyzed {
+		if ctx.Bool(Unknown) {
+			totalDeps := len(module.Deps)
+			i := 0
+			for _, dep := range module.Deps {
+				i++
+				log.ShowSpinner(fmt.Sprintf("Fetching Licence Info (%d/%d): %s", i+1, totalDeps, dep.ID.Name))
```
ilikebits: [nit] In general, prefer American English common usage ("license") for the sake of uniformity.

cmd/fossa/cmd/analyze/analyze.go
djgilcrease: using `-` to mean `STDOUT` seems to be a common practice with linux cli's per the POSIX guidlines http://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html#tag_12_02

```
For utilities that use operands to represent files to be opened for either reading or writing, the '-' operand should be used only to mean standard input (or standard output when it is clear from context that an output file is being specified).
```

api/fossa/locator.go
djgilcrease: Because every usage of it I could find immediately called the `.String` method, so why create a new object and consume more memory when the usage just wants the string output anyways.

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -45,7 +39,37 @@ func Run(ctx *cli.Context) error {
...
+	if ctx.String(flags.ShowOutput) != "" || ctx.String(flags.Template) != "" {
+		var tmpl *template.Template
+		if ctx.String(flags.Template) != "" {
+			tmpl, err = template.ParseFiles(ctx.String(flags.Template))
+			if err != nil {
+				log.Logger.Fatalf("Could not parse template data: %s", err.Error())
+			}
+
+			if ctx.String(flags.ShowOutput) == "" {
+				ctx.Set(flags.ShowOutput, "-")
```
golangci: Error return value of `ctx.Set` is not checked

cmd/fossa/cmd/report/report.go
```diff
@@ -1,173 +1,79 @@
...
+}
 
-	s.Suffix = " Analyzing modules..."
-	s.Start()
-	analyses, err := doAnalyze(conf.Modules, conf.AnalyzeCmd.AllowUnresolved)
-	s.Stop()
-	if err != nil {
-		log.Logger.Fatalf("Could not complete analysis (is the project built?): %s", err.Error())
-	}
+func outputReport(outputFile string, tmpl *template.Template, data interface{}) (err error) {
```
golangci: `outputReport` is unused

buildtools/bower/bower.go
```diff
@@ -10,24 +16,63 @@ type PackageMetadata struct {
...
+	}
+	bower := Bower{
+		Cmd:    cmd,
+		Config: config,
+	}
+	return &bower, nil
+}
+
+func (b *Bower) List() (Package, error) {
+	stdout, _, err := exec.Run(exec.Cmd{
```
golangci: ineffectual assignment to `err`

djgilcrease:
Addressed the issues you commented on. And I have allow edit for maintainers turned on.

Issue #115, opened 2018-03-20
Tests for PRs should not need `$FOSSA_API_KEY`


liftM:
The requirement for `$FOSSA_API_KEY` to upload an analysis causes both Circle CI and FOSSA to fail on PRs.

- Circle CI fails because tests run on PRs don't have access to `$FOSSA_API_KEY`, so any upload fails
- FOSSA fails because the failed upload prevents up from using provided builds

ilikebits:
We need push-only API keys for this.

ilikebits:
We now have support for "push-only" API keys, which only allow access to the API for uploading custom builds. Unfortunately, this currently is insufficient. We also use a variety of other APIs:

- The users API for determining organisation (for building custom locators).
- The licenses API for reporting.
- The issues API for testing.

liftM:
Fixed.

PR #197, opened 2018-07-25
fix(install): avoid hitting rate limit on install script

### LINKED ISSUE #180: Avoid hitting rate limit on installation script
Our installation script currently queries the GitHub Releases API to discover the latest release for download. When it is run many times in rapid succession (e.g. in parallel across many CI machines), this hits the rate limit and causes the installation script to fail.

Several options:

1. Generate a new installation script for every release, hardcoding the release download link. Downloading releases or from `raw.githubusercontent.com` is not rate limited.
2. Add a user option to specify a `GITHUB_TOKEN` to the request. Authenticated requests are not rate limited.
3. Create our own `GITHUB_TOKEN` with no permissions, and use this to authenticate the request.

Personally, option 1 seems like the best choice. It adds the most complexity, but provides the best user experience and seems easily maintainable.

See also:

- [Installer code that queries the Releases API](https://github.com/fossas/fossa-cli/blob/master/install.sh#L108)
- [Releases API documentation](https://developer.github.com/v3/repos/releases/)
- [Rate limit documentation](https://developer.github.com/v3/#rate-limiting)

emilyzhang:
Fixes #180 by generating a new install.sh (from install_tpl.sh) for each new release. Does not (yet) fix the issue for Windows users, however.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,3 +80,6 @@ clean:
 .PHONY: release
 release:
 	GOVERSION=$$(go version) goreleaser --rm-dist
+	RELEASE=$(git tag -l --points-at HEAD)
+	VERSION=${RELEASE#v}
+	sed "s/# RELEASE=/RELEASE=\'$RELEASE\'/; s/# VERSION=/VERSION=\'$VERSION\'/" install_tpl.sh > install.sh
--- a/install_tpl.sh
+++ b/install_tpl.sh
@@ -105,10 +105,10 @@ function install {
   cd $TMP_DIR || fail "changing directory to $TMP_DIR failed"
 
   # Download and validate release
-  bash -c "$GET $GH_API/repos/$USER/$REPO/releases/latest" > latest || fail "downloading latest release metadata failed"
-  RELEASE=$(grep tag_name latest | cut -d'"' -f4)
-  VERSION=${RELEASE#v} # remove prefix "v"
-
+  # DO NOT REMOVE THE FOLLOWING TWO LINES. They are used to generate
+  # installation scripts for new releases.
+  # RELEASE=
+  # VERSION=
   echo "Installing $USER/$REPO $RELEASE..."
   RELEASE_URL="$GH/$USER/$REPO/releases/download/$RELEASE"
   bash -c "$GET $RELEASE_URL/${REPO}_${VERSION}_${OS}_${ARCH}.tar.gz" > release.tar.gz || fail "downloading release failed"


Makefile
```diff
@@ -79,4 +79,7 @@ clean:
 
 .PHONY: release
 release:
+	SHELL:=/bin/bash
```
ilikebits: What does this line do?

install_tpl.sh
```diff
@@ -10,6 +10,7 @@
 
 TMP_DIR="/tmp/install-fossa-cli"
 
+
```
ilikebits: [nit] Extra whitespace.

emilyzhang: It's intentional, because a line will be added at line 12 when each new file is generated.

Makefile
emilyzhang: Sets the shell to be bash for the release target, and only the release target. Might not be a good idea for portability reasons (?) but I wanted to make sure that I could use sed in the Makefile. It might be unneeded.

install_tpl.sh
ilikebits: You should add a comment indicating this. Another option might be e.g. a "REPLACE ME" comment that you use as a replacement target for `sed`.

Makefile
ilikebits: This isn't necessary. Our release process always runs in the the `quay.io/fossas/fossa-cli` Docker container, which uses `bash`.

Makefile
```diff
@@ -79,4 +79,7 @@ clean:
 
 .PHONY: release
 release:
-	GOVERSION=$$(go version) goreleaser --rm-dist
+	GOVERSION=$(go version) goreleaser --rm-dist
```
ilikebits: This should be `$$(go version)` instead of `$(go version)`. Note the extra `$`.

install_tpl.sh
```diff
@@ -105,10 +105,10 @@ function install {
   cd $TMP_DIR || fail "changing directory to $TMP_DIR failed"
 
   # Download and validate release
-  bash -c "$GET $GH_API/repos/$USER/$REPO/releases/latest" > latest || fail "downloading latest release metadata failed"
-  RELEASE=$(grep tag_name latest | cut -d'"' -f4)
-  VERSION=${RELEASE#v} # remove prefix "v"
-
+  # Do not remove the following two lines. They are used to generate
```
ilikebits: I would prefer these in ALLCAPS.

ilikebits:
PR builds are failing due to #115. This LGTM; merging.

PR #200, opened 2018-07-30
feat(ruby): Configurable `Gemfile.lock` path

### LINKED ISSUE #183: Configurable `Gemfile.lock` location
Right now, we assume that the location of `Gemfile.lock` is fixed when we use Ruby analysis strategies that parse the lockfile.

For some build processes, this is not the case (e.g. when creating multiple lockfiles for different dependency groups). In order to allow custom `Gemfile.lock` locations, this should be configurable in `ruby.Options`.

cc @alvin-huang

delikat:
Fixes #183.


analyzers/ruby/ruby_test.go
```diff
@@ -0,0 +1,41 @@
+package ruby_test
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/fossas/fossa-cli/analyzers/ruby"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
```
ilikebits: [nit] Separate import groups of third-party and first-party code. In general, import group ordering (top-to-bottom) is standard library, third party, first-party.

ilikebits:
@delikat Can you set up commit signing for future contributions?

PR #203, opened 2018-07-31
fix(builders): add maven dependency list fallback

xizhao:
haven't tested this yet, but it takes freaking forever for a large project -- maybe we want to have some sort of escape hatch or alternative method since dependency:list is very expensive

if dependency:list fails once, for instance, we can prevent it from executing for the rest of the Discover run, what do you think @ilikebits ?

### DIFF
--- a/analyzers/maven/maven.go
+++ b/analyzers/maven/maven.go
@@ -58,7 +58,6 @@ func (a *Analyzer) Discover(dir string) ([]module.Module, error) {
 	log.Logger.Debugf("%#v", dir)
 	var modules []module.Module
 
-	skipDependencyList := false
 	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
 			log.Logger.Debugf("Failed to access path %s: %s\n", path, err.Error())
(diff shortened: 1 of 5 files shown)


cmd/fossa/cmd/init/init.go
```diff
@@ -43,10 +43,19 @@ func Run(ctx *cli.Context) error {
...
 		log.Logger.Fatalf("Could not run init: %s", err.Error())
 	}
+
 	err = config.WriteFile(modules)
 	if err != nil {
 		log.Logger.Fatalf("Could not write config: %s", err.Error())
+		return nil
+	}
+
+	if len(modules) > 0 {
```
ilikebits: Can we add this in a separate PR?

analyzers/maven/maven.go
```diff
@@ -73,19 +75,46 @@ func (a *Analyzer) Discover(dir string) ([]module.Module, error) {
 
 			dir := filepath.Dir(path)
 			submodules, err := a.Maven.Modules(dir)
-			if err != nil {
-				log.Logger.Debugf("Modules err: %#v %#v", err.Error(), err)
-				return err
-			}
+			if len(submodules) == 0 {
```
ilikebits: We should also explicitly check whether `err != nil`. If `err != nil`, then `submodules` is not guaranteed to be 0, and using its values is UB.

analyzers/maven/maven.go
```diff
@@ -73,19 +75,46 @@ func (a *Analyzer) Discover(dir string) ([]module.Module, error) {
...
+				// Failed to parse modules via calling into `mvn` or this is a single-module build
+				// Fallback to parsing the pom.xml and skipping the underlying file tree assuming they are subprojects
+				if err != nil {
+					log.Logger.Debugf("Unable to parse modules through `mvn`: %#v %#v", err.Error(), err)
+				}
+				var parsedPom maven.Manifest
```
ilikebits: You can probably just name this `pom`.

analyzers/maven/maven.go
```diff
@@ -73,19 +75,46 @@ func (a *Analyzer) Discover(dir string) ([]module.Module, error) {
...
+				pomParseError := files.ReadXML(&parsedPom, path, "pom.xml")
```
ilikebits: Why not name this `err`?

analyzers/maven/maven.go
```diff
@@ -73,19 +75,46 @@ func (a *Analyzer) Discover(dir string) ([]module.Module, error) {
...
+				groupId := parsedPom.GroupID
+				if groupId == "" {
+					groupId = parsedPom.Parent.GroupID
+				}
+
+				if parsedPom.Name == "" || groupId == "" || parsedPom.ArtifactID == "" {
+					pomParseError = errors.Errorf("invalid manifest at %s/pom.xml", path)
+				}
+
+				if pomParseError != nil {
```
ilikebits: This error check must occur before `pom` is used. Using return values of function calls with non-nil errors is UB.

analyzers/maven/maven.go
```diff
@@ -1,6 +1,7 @@
 // Package maven implements Maven analysis.
 //
-// A `BuildTarget` for Maven is the Maven project name.
+// A `BuildTarget` for Maven is a reactor build target
+// specified by [groupId]:artifactId or by its relative path
```
ilikebits: When is relative path used as a `BuildTarget`?

xizhao: This is taken from the `mvn dependency:tree` docs

xizhao:
Idk if #240 addresses this, but bump.  

@liftM can you take over this PR?

dchenk:
Sadly much of the code in this PR can't be used anymore because the repo code has diverged significantly since July 2018. I'll take over this and try to salvage what we can (may need to just close the PR). Actually the module finder here is dealing with Maven modules correctly—more correctly than the current [`Discover`](https://github.com/fossas/fossa-cli/blob/master/analyzers/maven/maven.go#L62) and [`Modules`](https://github.com/fossas/fossa-cli/blob/master/buildtools/maven/maven.go#L62) functions. We need to distinguish between module _identifers_ (e.g., `groupId:artifactId`) from the relative path to the module. Currently the CLI conflates that and is likely not discovering all the dependencies that it's supposed to.

analyzers/maven/maven.go
dchenk: BuildTarget is not going to be very helpful anymore, partly because of how it plays these two different roles. I'm working on a fix that relies on the `Module.Dir` property instead.

zlav:
Closing due to inactivity. Since this PR was opened we have introduced other methods of fallbacks that can be implemented, such as pom file parsing and command timeouts.

PR #205, opened 2018-07-31
feat(vcs): Correctly infer VCS settings when within a repo but not at the root

### LINKED ISSUE #190: Correctly infer VCS settings when within a repository but not at the root
In `package config`, we currently infer VCS settings only if the current working directory is the root of a VCS repository. Ideally, we should be able to infer settings as long as we're _within_ a VCS repository. (In practice, `fossa` is almost always run from the root of VCS, so this isn't a problem that occurs very much.)

There is starter code for this in `package vcs`, and related code in `package golang`.

delikat:
Fixes #190.


cli.go
```diff
@@ -0,0 +1,13 @@
+// Package cli encompasses domain types for the CLI app.
+package cli
+
+// VCSType represents a type of version control system.
```
ilikebits: [nit] This can just be `VCS`.

monad/monad.go
```diff
@@ -25,3 +27,28 @@ func (r *EitherStr) Bind(f EitherStrFunc) *EitherStr {
...
 	return r
 }
+
+// EitherVCSType is an Either monad specialized to VCSType.
+type EitherVCSType struct {
+	Result cli.VCSType
+	Err    error
+}
+
+// EitherVCSTypeFunc defines monadic EitherStr functions.
```
ilikebits: [nit] "defines monadic _EitherVCSType_ functions".

vcs/vcs_test.go
```diff
@@ -0,0 +1,25 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/errutil"
+	"github.com/fossas/fossa-cli/vcs"
+)
+
+func TestGetRepositoryBelowRoot(t *testing.T) {
```
ilikebits: Can you also add a test for being _at_ the root?

PR #207, opened 2018-08-02
Enable unit tests in CI

ilikebits:
Unit tests were previously disabled in CI for a variety of historical reasons (principally: they didn't work).

The VCS tests rely on the existence of mock repository folders. These weren't committed because `git` does not automatically commit empty folders, and we didn't realise the error since the unit tests didn't run in CI.

I've corrected this by adding a `.gitkeep` to the mock Subversion repository and switching the tests to use the actual `git` repository for testing (if we try to add a mock repository, we have to jump through a whole set of submodule hoops).

Remaining work on this PR:
- [x] Add JUnit reporting for CircleCI ([reference](https://circleci.com/docs/2.0/language-go/))

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,13 +34,6 @@ jobs:
           paths:
             - "/home/fossa/go/pkg"
             - "/home/fossa/go/src/github.com/fossas/fossa-cli/vendor"
-      # - run:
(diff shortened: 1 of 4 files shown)


ilikebits:
Force-merging. Tests have previously run successfully, and the rebase is orthogonal.

PR #213, opened 2018-08-03
feat(analyzers): allow npm errors by default

xizhao:
This PR turns strict mode off by default for the `npm` analyzer, allowing analysis to still complete even when there's a non-zero code returned by npm.

This is because `npm ls` can still give the user lots of useful data, but will error if the build state isn't perfect.  In this case we should just warn the user, but still proceed since, to be honest, build state is never perfect in npm land.

Also: all error messages should be more user friendly, "NPM had a non-zero exit code" requires knowledge of internals that we're talking to npm.


ilikebits:
This is already the NPM behaviour.

PR #214, opened 2018-08-03
fix(buildtools): add bower config constructor to apply defaults

xizhao:
This PR ensures configuration defaults to `bower_components` when unspecified by an existing but empty `.bowerrc` by implementing a constructor pattern.

Question, if the `.bowerrc` is empty, will `NewConfig(dir).CWD` be overwritten to be empty too?

### DIFF
--- a/buildtools/bower/config.go
+++ b/buildtools/bower/config.go
@@ -11,27 +11,29 @@ type Config struct {
 	Directory string `json:"directory"`
 }
 
+func NewConfig(cwd string) Config {
+	return Config{
+		CWD:       cwd,
+		Directory: filepath.Join(cwd, "bower_components"),
+	}
+}
+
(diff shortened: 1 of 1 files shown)


analyzers/bower/bower.go
```diff
@@ -133,7 +133,11 @@ func (a *Analyzer) IsBuilt(m module.Module) (bool, error) {
 	}
 	// TODO: Check if the installed modules are consistent with what's in the
 	// actual manifest.
-	isBuilt, err := files.ExistsFolder(config.Directory)
+	bowerComponentsDir := config.Directory
+	if !filepath.IsAbs(bowerComponentsDir) {
+		bowerComponentsDir = filepath.join(m.Dir, config.Directory)
```
golangci: join not exported by package filepath

PR #219, opened 2018-08-07
Remove polymorphic monads

### LINKED ISSUE #210: Specialise all `package monad` instances
The current interface for `package monad` is unwieldy and is only ever used for finding files or folders recursively in ancestors. We should specialise this functionality in `package files` and remove `package monad`.

This also has the happy side effect of allowing us to fold VCS types back in to `package vcs`.

ilikebits:
Fixes #210.


vcs/vcs.go
```diff
@@ -18,63 +16,61 @@ var (
...
+	return 0, ErrNoVCSInDir
 }
 
-// NearestVCS returns the type and directory of the nearest VCS repository
+// Nearest returns the type and directory of the nearest VCS repository
 // containing the directory, or ErrNoNearestVCS if none is found.
-func NearestVCS(dirname string) (vcsType cli.VCS, vcsDir string, err error) {
-	vcsDir, err = files.WalkUp(dirname, func(d string) error {
-		tool, err := vcsIn(d)
+func Nearest(dirname string) (VCS, string, error) {
```
delikat: 👍 to this and other renames

files/files_test.go
```diff
@@ -0,0 +1,16 @@
+package files_test
```
delikat: hooray tests

vcs/vcs.go
```diff
@@ -18,63 +16,61 @@ var (
...
+		}
 	}
-	return result.Result, nil
+	return 0, ErrNoVCSInDir
 }
 
-// NearestVCS returns the type and directory of the nearest VCS repository
+// Nearest returns the type and directory of the nearest VCS repository
 // containing the directory, or ErrNoNearestVCS if none is found.
-func NearestVCS(dirname string) (vcsType cli.VCS, vcsDir string, err error) {
```
delikat: Why the move away from named returns?

ilikebits: I just avoid them where possible, because assignment in Go follows unintuitive shadowing rules. I usually only use named returns when returning multiple values of the same type.

Example of weird shadowing:
```go
func Foo() (bar string) {
	bar, err := Baz() // This is a _different_ bar, shadowing the original
	return            // Implicitly returns `bar = ""`, the string type's zero value
}
```

- [Documentation on named returns](https://tour.golang.org/basics/7)
- More shadowing examples:
	- [Go 2 proposal removing shadowing](https://github.com/golang/go/issues/21114)
	- [Blog post](https://blog.charmes.net/post/scope-shadowing-go/)
	- [Mailing list thread](https://groups.google.com/forum/#!topic/golang-nuts/VeJTB9dhhK4)

PR #227, opened 2018-08-21
Third-party tarballs

liftM:
TBR. Merging due to urgency. There will be a follow-up PR correcting the `TODO`s in this.


api/fossa/tar.go
```diff
@@ -0,0 +1,201 @@
...
+		return Locator{}, err
+	}
+
+	// Run second pass: multi-part uploading.
+	r, w := io.Pipe()
+	// In parallel, stream temporary file to PUT.
+	go func() {
+		defer w.Close()
+		defer tarball.Close()
+		tarball.Seek(0, 0)
```
golangci: Error return value of `tarball.Seek` is not checked

api/fossa/tar.go
```diff
@@ -0,0 +1,201 @@
...
+
+// CreateTarball archives and compresses a directory's contents to a temporary
+// file while simultaneously computing its MD5 hash. The caller is responsible
+// for closing the file handle.
+func CreateTarball(dir string) (*os.File, []byte, error) {
+	tmp, err := ioutil.TempFile("", "fossa-tar-"+filepath.Base(dir)+"-")
+	if err != nil {
+		return nil, nil, err
+	}
+	defer tmp.Sync()
```
golangci: Error return value of `tmp.Sync` is not checked

PR #237, opened 2018-08-28
Added Jira project key and Project URL flags

anuccio1:
Added Jira Project Key and Project URL flags to fossa analyze


api/fossa/upload.go
```diff
@@ -55,6 +55,12 @@ func Upload(fetcher, project, revision, title, branch string, data []SourceUnit)
 	if branch != "" {
 		q.Add("branch", branch)
 	}
+	if projectURL != "" {
```
liftM: I think we should refactor optional properties (title, branch, projectURL, jiraProjectKey) into a separate `type UploadOptions struct`. Thoughts?

Issue #232, opened 2018-08-27
Allow installing without root


grosser:
`install.sh | bash` requires root or it fails, please add a `LOCAL=true` or so flag so this is not required anymore and I can still have a successful install

PR #238, opened 2018-08-28
allow local install without sudo using LOCAL=true

grosser:
fixes https://github.com/fossas/fossa-cli/issues/232

@liftM 
/cc @plonergan


grosser:
errors are not from this PR ... so ready to merge ?

PR #240, opened 2018-08-29
[WIP]: add runtime scope to mvn dependency:tree

### LINKED ISSUE #202: Maven integration uses `test` scope by default
`mvn dependency:list` seems to automatically rely on the `test` scope, which is the wrong default scope we want.

We should be including only `compile, runtime` as our default scopes in order to replicate a production build.  This should be configurable to the user via module options.

See:

https://github.com/fossas/fossa-cli/blob/master/buildtools/maven/maven.go#L80

iamoperand:
fixes #202
@liftM


liftM:
Can you add a test verifying that this works? I believe that this fix doesn't work due to a long-standing bug in Maven itself. See https://issues.apache.org/jira/projects/MSHARED/issues/MSHARED-4?filter=allopenissues for more details.

iamoperand:
Taken from mvn dependency:tree docs...

It provides a flag called **scope** which lets us use any scope that we want i.e. we can just use
```
mvn dependency:tree --batch-mode -Dscope=runtime --projects <project_name>
```

and it will give both **compile** and **runtime** dependencies.

iamoperand:
@liftM  I have put together a [pdf](https://github.com/fossas/fossa-cli/files/2331851/Fossa_.Issue.202.1.pdf) too. Basically, testing things on my local machine for a new maven project.

It was working as expected, but yes, let me add a test and check it.

iamoperand:
@liftM I have added the [test](https://github.com/fossas/fossa-cli/blob/3a922c5c8756038a47e1a1610ae87fe8a97da48a/analyzers/maven/maven_test.go).

Would love to get your feedback about it.

iamoperand:
Also, @liftM can you tell me a bit more about making it **user configurable** through **module options**?

**What does that mean in particular?**

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,39 @@
...
+	assert.NoError(t, err)
+
+	analyzed, err := analyzer.Analyze()
+	assert.NoError(t, err)
+
+	// convert it into a string
+	stringifiedAnalyzed := fmt.Sprintf("%#v", analyzed)
+
+	assert.Contains(t, stringifiedAnalyzed, "com.amazonaws")
+	assert.NotContains(t, stringifiedAnalyzed, "junit")
```
liftM: This test is great! You might want to inspect the values within `analyzed` rather than printing them. In particular, you can loop through the dependencies and inspect the dependency names rather than relying on the stringified format.

Inspecting the values will produce a less brittle test than checking the string.

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/analyzers/maven"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+func TestCustomMavenPOMFile(t *testing.T) {
+	buildTarget := "github.com/fossas/fossa-cli/cmd/fossa"
```
liftM: I don't think this build target correctly corresponds to a Maven subproject. In this case, I think there are no subprojects (so the correct build target would be an empty string).

I'm not sure we handle the empty subproject case correctly in `DependencyTree` -- you might need to add in additional logic to avoid passing the project flag if the parameter is not provided.

buildtools/maven/maven.go
```diff
@@ -97,7 +97,7 @@ func (m *Maven) DependencyTree(dir, project string) ([]Dependency, map[Dependenc
 	output, _, err := exec.Run(exec.Cmd{
 		Name: m.Cmd,
 		Dir:  dir,
-		Argv: []string{"dependency:tree", "--batch-mode", "--projects", project},
+		Argv: []string{"dependency:tree", "--batch-mode", "-Dscope=runtime"},
```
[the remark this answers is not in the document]
iamoperand: I didn't see any use of putting it in the docs.

Let me search for it again 👍

analyzers/maven/maven_test.go
iamoperand: Yeah, totally get it.

analyzers/maven/maven_test.go
iamoperand: Ahhh, that is why I got confused with it.

*On it!*

buildtools/maven/maven.go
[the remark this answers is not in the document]
xizhao: The behavior we want is:

 1. get the whole dep tree
 2. filter for the given project (or artifact) for dependency analysis
 3. filter for runtime scope only by default, but have this be configurable

Options 2/3 SHOULD be provided by the plugin, but [MSHARED-4](https://issues.apache.org/jira/projects/MSHARED/issues/MSHARED-4?filter=allopenissues) says that these don't function correctly due to the bug.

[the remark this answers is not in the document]
xizhao: Also the default might not be just runtime, we might also want compile as well.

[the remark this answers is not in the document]
iamoperand: @xizhao 
1. Got it.
2. *-Dscope=runtime* seems to work even after *MSHARED-4*. That's weird, but the test shows that it works :/
3. Got it.


*runtime scope* gives both the *compile-time* as well as *run-time* dependencies. 
So that's fixed. 
Right?

analyzers/maven/maven_test.go
iamoperand: This has been resolved!

analyzers/maven/maven_test.go
iamoperand: Did it 😄

iamoperand:
@liftM @xizhao 

I have made the required fixes. 
Waiting for your feedback 😄

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	analyzed, err := analyzer.Analyze()
+	assert.NoError(t, err)
+
+	directDependencies := analyzed.Direct
+
+	compileTimeDependencyfound := false
+	runtTimeDependencyFound := false
+	testDependencyFound := false
+	for dependencyIndex := 0; dependencyIndex < len(directDependencies); dependencyIndex++ {
```
liftM: You can use `for _, dep := range directDependencies {}` as a for-each loop in Go.

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		if dependencyName == "com.thoughtworks.xstream:xstream" {
+			runtTimeDependencyFound = true
+		}
+		if dependencyName == "junit:junit" {
+			testDependencyFound = true
+		}
+	}
+
+	assert.Equal(t, compileTimeDependencyfound, true, "Compile time dependency found")
```
liftM: You can use `assert.True`.

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,143 @@
...
+	assert.NoError(t, err)
+
+	analyzedRuntimeScope, err := analyzerWithRuntimeScope.Analyze()
+	assert.NoError(t, err)
+
+	// analyzedDefaultScope != analyzedTestScope
+	assert.NotEqual(t, analyzedDefaultScope, analyzedTestScope)
+
+	// analyzedDefaultScope == analyzedRuntimeScope
+	assert.Equal(t, analyzedDefaultScope, analyzedRuntimeScope)
```
liftM: Can you also add a test that the runtime scope contains no test dependencies?

analyzers/maven/maven_test.go
iamoperand: Did it 👍

analyzers/maven/maven_test.go
iamoperand: Done 👍

analyzers/maven/maven_test.go
iamoperand:
```golang
assert.False(t, testDepsExist, "Test dependency not found")
```

Are you talking about this?

iamoperand: https://github.com/fossas/fossa-cli/pull/240/files#diff-b9f26dd681f3f09b56b8f03a3f8b2af2R51

analyzers/maven/maven.go
```diff
@@ -134,7 +136,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	var deps map[maven.Dependency][]maven.Dependency
 	var err error
 	if a.Options.Command == "" {
-		imports, deps, err = a.Maven.DependencyTree(a.Module.Dir, a.Module.BuildTarget)
+		imports, deps, err = a.Maven.DependencyTree(a.Module.Dir, a.Module.BuildTarget, a.Options.Scope)
```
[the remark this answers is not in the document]
iamoperand: I'm making the runtime as the default scope [here](https://github.com/fossas/fossa-cli/pull/240/files#diff-470cd8df4052a403baedd4ed162c14faR102)

zlav:
@iamoperand I am closing this due to inactivity from fossa and code changes. I believe that due to the conflicts that are now in this PR the best path forward would be for a maintainer to take over and open a PR. The fixes you worked on are really appreciated and we will make sure that any future changes will be related to this PR.

PR #245, opened 2018-08-31
Fix shasum failed

mcsec:
Later lines specifically refer to $FILE, but the release was downloaded as "release.tar.gz." Checking the shasum failed because the file was under that name. I doubt "> $FILE" is necessary, as there's no need to rename, but at least these changes allowed me to install FOSSA.

This change appears to already exist in https://github.com/fossas/fossa-cli/blob/master/install_template.sh, but was simply overlooked in this, install.sh.


mcsec:
Looks like this file is generated automatically. I believe just running that generation process again should make it work. Removing.

PR #249, opened 2018-08-31
Correctly upload bad input with empty lines

liftM:


api/fossa/upload.go
```diff
@@ -32,15 +32,15 @@ type UploadOptions struct {
 
 // Upload uploads a project's analysis.
 func Upload(title string, locator Locator, options UploadOptions, data []SourceUnit) (Locator, error) {
-	// Check preconditions
+	// Check preconditions.
 	if locator.Fetcher != "custom" && locator.Revision == "" {
```
anuccio1: This makes a little more sense as `if locator.Fetcher == "git" && locator.Revision == "" {`.

cmd/fossa/cmd/upload/upload.go
```diff
@@ -52,18 +53,31 @@ var Cmd = cli.Command{
...
+	if locators == "" {
+		return fossa.SourceUnit{}, errors.New("upload did not receive any input")
+	}
+
+	// Handle bad input with empty lines.
 	var deps []fossa.SourceUnitDependency
 	lines := strings.Split(locators, "\n")
 	for _, line := range lines {
-		deps = append(deps, fossa.SourceUnitDependency{Locator: line})
+		if line == "" {
```
anuccio1: what about a regex comparison instead? `\s+\s$\s` ?

PR #254, opened 2018-09-06
feat(ruby analyzer) add fallback mechanism for ruby analyzers when bundler fails

microsoftly:


analyzers/ruby/ruby.go
```diff
@@ -127,16 +127,26 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
 		lockfilePath = a.Options.LockfilePath
 	}
 
-	switch a.Options.Strategy {
+	switch strategy {
 	case "list":
 		gems, err := a.Bundler.List()
 		if err != nil {
-			return graph.Deps{}, err
+			return a.analyzeHelper("lockfile")
```
liftM: The potential for accidental infinite recursion is a bit scary here. Given that this function is a gigantic switch anyway, I think it makes sense to break out these strategies into separate functions.

analyzers/ruby/ruby_test.go
```diff
@@ -43,3 +45,48 @@ func TestCustomGemfileLockPath(t *testing.T) {
 
 	assert.NotEqual(t, analyzed, analyzedCustom)
 }
+
+func TestFallbackOnMissingBundler(t *testing.T) {
+	buildTarget := "github.com/fossas/fossa-cli/cmd/fossa"
+	useLockfileOptions := map[string]interface{}{
+		"strategy":          "lockfile",
+		"gemfile-lock-path": filepath.Join("testdata", "Gemfile.lock"),
```
liftM: This doesn't need to be set by default.

analyzers/ruby/ruby_test.go
```diff
@@ -43,3 +45,48 @@ func TestCustomGemfileLockPath(t *testing.T) {
 
 	assert.NotEqual(t, analyzed, analyzedCustom)
 }
+
+func TestFallbackOnMissingBundler(t *testing.T) {
+	buildTarget := "github.com/fossas/fossa-cli/cmd/fossa"
```
liftM: This is incorrect. As documented in the package comment for the Ruby analyser, the build target of a Ruby module is the directory containing the project. In this case, it should be `testdata`. (Tests are always run with the CWD set to the test file's folder.)

analyzers/ruby/ruby_test.go
```diff
@@ -43,3 +45,48 @@ func TestCustomGemfileLockPath(t *testing.T) {
...
+		Name:        "test",
+		Type:        pkg.Ruby,
+		BuildTarget: buildTarget,
+		Options:     useBundlerWithLockfilePathOptions,
+	}
+
+	lockfileBasedAnalyzer, lockfileBasedAnalyzerErr := ruby.New(gemModuleUsingLockfile)
+	fallbackBasedAnalyzer, fallbackBasedAnalyzerErr := ruby.New(gemModuleUsingFallbackToLockfile)
+
+	assert.NoError(t, lockfileBasedAnalyzerErr)
```
liftM: You should `assert.NoError` directly after the call. The tests will fail faster.

Makefile
```diff
@@ -57,6 +57,12 @@ dev: docker-test-base
 		-v $$GOPATH/bin:/home/fossa/go/bin \
 		quay.io/fossa/fossa-cli-test-base /bin/bash
 
+.PHONY: dev-mac
+dev-mac: docker-test-base
```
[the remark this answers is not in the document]
microsoftly: `dev-osx`?

Makefile
```diff
@@ -57,6 +57,13 @@ dev: docker-test-base
 		-v $$GOPATH/bin:/home/fossa/go/bin \
 		quay.io/fossa/fossa-cli-test-base /bin/bash
 
+.PHONY: dev-osx
+dev-osx: docker-test-base
+	docker run --rm -it \
+		-p 2345:2345 \
```
liftM: Is this for the debugger? Can you add this to all the Docker invocations in a separate PR?

docker/test-base/Dockerfile
```diff
@@ -184,4 +184,6 @@ RUN git clone --depth=1 https://github.com/MacDownApp/macdown
 RUN git clone --depth=1 https://github.com/fossas/ant-example-project
 
 # Raw projects:
-RUN git clone --depth=1 https://github.com/jelaas/bifrost-framework
\ No newline at end of file
+RUN git clone --depth=1 https://github.com/jelaas/bifrost-framework
+
+RUN go get -u github.com/derekparker/delve/cmd/dlv
```
liftM: This should really be in the base Dockerfile, which will cause all the Docker images to need a rebuild (groan).

analyzers/ruby/ruby_test.go
```diff
@@ -43,3 +45,51 @@ func TestCustomGemfileLockPath(t *testing.T) {
...
+	}
+
+	lockfileBasedAnalyze, lockfileBasedAnalyzerErr := lockfileBasedAnalyzer.Analyze()
+	assert.NoError(t, lockfileBasedAnalyzerErr)
+
+	fallbackBasedAnalyze, fallbackBasedAnalyzerErr := fallbackBasedAnalyzer.Analyze()
+	assert.NoError(t, fallbackBasedAnalyzerErr)
+
+	// ensure that the arrays are actually populated and we aren't comparing equivalent empty results
+	assert.NotZero(t, len(fallbackBasedAnalyze.Direct))
```
liftM: `assert.NotEmpty`

analyzers/ruby/ruby_test.go
```diff
@@ -43,3 +45,51 @@ func TestCustomGemfileLockPath(t *testing.T) {
 
 	assert.NotEqual(t, analyzed, analyzedCustom)
 }
+
+func TestFallbackOnMissingBundler(t *testing.T) {
+	buildTarget := "testdata"
+	useLockfileOptions := map[string]interface{}{
+		"strategy":          "lockfile",
+		"gemfile-lock-path": filepath.Join("testdata", "Gemfile.lock"),
```
liftM: Is this still needed?

analyzers/ruby/ruby_test.go
```diff
@@ -4,6 +4,8 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/fossas/fossa-cli/buildtools/bundler"
```
liftM: This should be grouped with the other first-party imports.

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
+
+	imports, deps := FilteredLockfile(gems, lockfile)
+
+	return graph.Deps{
+		Direct:     imports,
+		Transitive: deps,
+	}, nil
+}
+
+func (a *Analyzer) getLockfilePath() string {
```
liftM: You can simplify this function a lot. Ex:

```go
func (a *Analyzer) lockfilePath() string {
	if a.Options.LockfilePath != "" {
		return a.Options.LockfilePath
	}
	return filepath.Join(a.Module.Dir, "Gemfile.lock")
}
```

Also, the `getX` prefix is unidiomatic in Go.

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
+		return a.defaultAnalyzerStrategy(lockfilePath)
+	}
+}
+
+func (a *Analyzer) bundlerListAnalyzerStrategy() (graph.Deps, error) {
+	gems, err := a.Bundler.List()
+	if err != nil {
+		lockfilePath := a.getLockfilePath()
+
+		return a.lockfileAnalyzerStrategy(lockfilePath)
```
liftM: I'd rather have the fallback occuring in the caller of each Analyzer strategy. This seems easier to reason about.

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
+	return graph.Deps{
+		Direct:     imports,
+		Transitive: deps,
+	}, nil
+}
+
+func (a *Analyzer) lockfileAnalyzerStrategy(lockfilePath string) (graph.Deps, error) {
+	lockfile, err := bundler.FromLockfile(lockfilePath)
+	if err != nil {
+		return a.defaultAnalyzerStrategy(lockfilePath)
```
liftM: This doesn't seem like the correct fallback order. It should be: list-lockfile (default) -> lockfile -> list.

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
-	if a.Options.LockfilePath != "" {
-		lockfilePath = a.Options.LockfilePath
+	strategy := "list"
+
+	if a.Options.Strategy != "" {
+		strategy = a.Options.Strategy
 	}
 
-	switch a.Options.Strategy {
+	return a.analyzeHelper(strategy)
```
liftM: Is this helper still needed?

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
-		gems, err := a.Bundler.List()
-		if err != nil {
-			return graph.Deps{}, err
-		}
-		imports, deps := FilteredLockfile(gems, lockfile)
-		return graph.Deps{
-			Direct:     imports,
-			Transitive: deps,
-		}, nil
+		return a.defaultAnalyzerStrategy(lockfilePath)
```
liftM: I'd rather have the fallback logic occurring here.

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,85 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
-		gems, err := a.Bundler.List()
-		if err != nil {
-			return graph.Deps{}, err
-		}
-		imports, deps := FromGems(gems)
-		return graph.Deps{
-			Direct:     imports,
-			Transitive: deps,
-		}, nil
+		return a.bundlerListAnalyzerStrategy()
```
liftM: We should not implicitly fall back if a strategy is set explicitly and manually by the user.

docker/test-base/Dockerfile
microsoftly: we can move this to the debug pr.

analyzers/ruby/ruby_test.go
microsoftly: without it, the tests timeout

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,94 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
+
+	imports, deps := FromLockfile(lockfile)
+
+	return graph.Deps{
+		Direct:     imports,
+		Transitive: deps,
+	}, nil
+}
+
+func (a *Analyzer) listLockfileAnalyzerStrategy(lockfilePath string) (graph.Deps, error) {
```
golangcibot: func `(*Analyzer).listLockfileAnalyzerStrategy` is unused

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,86 @@ func (a *Analyzer) IsBuilt() (bool, error) {
 }
 
 func (a *Analyzer) Analyze() (graph.Deps, error) {
-	lockfilePath := filepath.Join(a.Module.Dir, "Gemfile.lock")
-	if a.Options.LockfilePath != "" {
-		lockfilePath = a.Options.LockfilePath
+	shouldFallback := true
```
liftM: It would be easier to remove the fallback logic from the strategy helpers, and to instead only do fallback on the `default` case of the switch in this function.

analyzers/ruby/ruby_test.go
```diff
@@ -7,22 +7,27 @@ import (
...
 	m := module.Module{
 		Name:        "test",
 		Type:        pkg.Ruby,
 		BuildTarget: buildTarget,
+		Dir:         buildTarget,
 	}
 
+	p := filepath.Join("testdata", "Gemfile.lock")
+
+	print(p)
```
liftM: What is this?

analyzers/ruby/ruby.go
```diff
@@ -127,47 +127,92 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
```
liftM: Is `shouldFallback` still needed?

microsoftly: yes. We don't know if `list-lockfile` is explicitly provided or implicitly chosen.

anuccio1:
test

Issue #235, opened 2018-08-28
Filter development dependencies in Node.JS


liftM:
Currently, we capture the Node.JS dependency graph by running `npm list --json`, which enumerates all the installed dependencies.

For users who don't want to include development dependencies, the _correct_ solution is to run `npm install --production`, and to not install development dependencies. Unfortunately, we have come across a number of cases where this is not feasible. Most of the time, this is because some development dependency (e.g. webpack) is needed to run a post-installation hook.

We need to accommodate these use cases. The most immediate design for a solution is to do the filtering by hand:

1. Parse the full dependency graph from `npm list --json`, as before.
2. Construct a set of production dependencies starting from the module's `package.json` and recursively reading `dependencies` in each dependency's manifest (found in `node_modules`).
3. Filter the full dependency graph, removing all development dependencies.

bsima:
Update on this: I've got the code building properly (took a bit of time because I'm not familiar with go tooling, but its working now).

Any advice on how to handle tests? Right now there are none for the analyzers/nodejs module.

liftM:
Our story for testing is not great.

For things like parsers, we generally unit test them using the `*_test.go` package idiom (for example, you can check out the way we unit test our Ruby lockfile parser in `buildtools/bundler`). See `make unit-test` for the implementation.

For integration testing, we load fixtures into a Docker image and then run `test.sh` in `make integration-test`. We use the Docker image because we often clone entire projects, and we don't want to put these directory in our repository. These integration tests are often sanity checks ("does this exit 0?") as opposed to correctness checks.

This is far from ideal. I'm all ears if you've got a better plan for this. Something I've been considering is adding `integration_test.go` files into each package that we only run if `!testing.Short()`, and doing any fixture setup, analysis, and correctness check in there. I haven't gotten around to doing this, but it might be an interesting approach to explore.

bsima:
Okay, I will try and come up with a way to test this, I have an idea or two.

I just finished moving over the weekend and am without internet at home this
week, so bear with me. I will try to get back to working on this tomorrow.

liftM:
No worries, thanks for the update.

PR #255, opened 2018-09-07
235: Filter NodeJS dev dependencies

bsima:
I'm really not sure how to test this - I looked at "testing/quick" but couldn't figure out how to generate a `pkgMap` to run against a unit test. Any advice is appreciated.

Closes https://github.com/fossas/fossa-cli/issues/235


pkg/package.go
```diff
@@ -57,7 +57,7 @@ type Package struct {
 
 	Imports []Import
 	// SourceFiles []File
-	// Qualifier   Qualifier
+	Qualifier   Qualifier
```
bsima: Is this being actively detected, or do I have to write a function to detect Production vs Development?

liftM: I don't think these are implemented yet.

Makefile
```diff
@@ -38,21 +38,21 @@ $(PREFIX)/fossa: $(BIN)/fossa
 
 # Building various Docker images.
 docker-base: ./docker/base/Dockerfile
-	sudo docker build -t quay.io/fossa/fossa-cli-base -f ./docker/base/Dockerfile .
+	docker build -t quay.io/fossa/fossa-cli-base -f ./docker/base/Dockerfile .
```
liftM: These are `sudo` for Linux development. You might want to add MacOS-specific targets without `sudo`.

analyzers/nodejs/nodejs.go
```diff
@@ -276,6 +276,15 @@ func recurseDeps(pkgMap map[pkg.ID]pkg.Package, p npm.Output) {
 		if ok {
 			continue
 		}
+
+        // Remove dev dependencies
+        if removeDev {
+            o := pkgMap[id]
+            if o.Qualifier == 2 {
```
liftM: What is `o.Qualifier`?

Makefile
bsima: I do use Linux - sudo is only necessary if your user isn't part of the docker group, this is true of Linux and MacOS

    sudo groupadd docker
    sudo gpasswd -a $USER docker

c.f. https://askubuntu.com/questions/477551/how-can-i-use-docker-without-sudo/477554#477554

analyzers/nodejs/nodejs.go
bsima: This refers to the `Qualifier` field in the `Package` struct: https://github.com/fossas/fossa-cli/blob/49419e96778b772845535654cb0f813bb922b6d9/pkg/package.go#L60

Makefile
liftM: That's _true_, but I'm not sure that "being in the `docker` group" a good assumption.

1) Being in the `docker` group is effectively equivalent to being root, and I want to discourage this behaviour as much as possible.
2) You're effectively root anyway, so `sudo docker ...` is only more typing, not actually less secure.

bsima:
Okay I’ll rebase that commit out in the morning

bsima:
Crap, i rebased improperly and lost my commits... will fix this and re-open.

bsima:
Superseded by https://github.com/fossas/fossa-cli/pull/270

PR #259, opened 2018-09-12
 feat(npm buildtool) update FromManifest to return a package and rename it accordingly

microsoftly:


buildtools/npm/manifest.go
```diff
@@ -35,3 +36,34 @@ type Lockfile struct {
...
+		imports = append(imports, createImport(depName, version))
+	}
+
+	return pkg.Package{
+		ID:      id,
+		Imports: imports,
+	}
+}
+
+func createImport(packageName string, version string) pkg.Import {
```
liftM: I would inline this unless there's a good reason to separate it e.g.:

- This is called in multiple places.
- This is an independently tested function.

Descriptive function names are valuable, but I think comments work better in cases where there's no control flow reason to extract a function.

microsoftly:
closing in favor of PR with implementation due to # of changes

PR #261, opened 2018-09-14
feat(fromNodeModules) define and test fromNodeModules func

microsoftly:
* Tests use node `v0.10.X` and `v8.x.x` to introduce two different `node_module` schemes
  * these fixtures are added with all `.js`, `.ts`, and `.md` files removed
  * ensures that output state is equivalent regardless of install strategy
* Included deps are relatively simple but have a decent amount of case coverage
  * deduped deps 
  * depth of up to 3  for deps
  * multiple versions of same module for different dep levels


buildtools/npm/manifest_test.go
```diff
@@ -1,65 +1,95 @@
...
+		Location: "",
+		Name:     "type-detect",
+		Revision: "4.0.8",
+		Type:     pkg.NodeJS,
+	}
+
+	// ensure project that is both a transitive and direct dep have correct versions based on where they live
+	assert.Contains(t, depGraph.Transitive, typeDetectTransitiveDepKey)
+
+	chaiProject := depGraph.Transitive[chaiDirectDep.Resolved]
```
liftM: Can you also check the imports of `type-detect`? Checking the absence of bad data is just as important as the presence of good data.

At a higher level, it would be great if we had some facility for conveniently describing mock package graphs.

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,101 @@ type Lockfile struct {
...
+			return nil, err
+		}
+
+		addNextDepthSubModulesToSubModuleAccumulator(submoduleProjects, nextLevelSubModules)
+	}
+
+	return submoduleProjects, nil
+}
+
+func addNextDepthSubModulesToSubModuleAccumulator(accumulator map[pkg.ID]pkg.Package, nextDepthModules map[pkg.ID]pkg.Package) {
```
liftM: What is the purpose of this function?

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,101 @@ type Lockfile struct {
...
+		Revision: version,
+	}
+
+	return pkg.Import{
+		Target:   packageName,
+		Resolved: id,
+	}
+}
+
+func fromSubNodeModules(currentDir string, rootNodeModuleDir string, previousPackage pkg.Package) (map[pkg.ID]pkg.Package, error) {
```
liftM: This logic is difficult to follow, and I think it's because you're doing raw recursion. I would recommend using `filepath.Walk` instead.

buildtools/npm/manifest_test.go
```diff
@@ -4,17 +4,45 @@ import (
...
 
-	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
 
 	"github.com/fossas/fossa-cli/buildtools/npm"
-	"github.com/stretchr/testify/assert"
+	"github.com/fossas/fossa-cli/pkg"
 )
 
+func getDirectDepImports() (chaiDirectDep pkg.Import, typeDetectDirectDep pkg.Import) {
```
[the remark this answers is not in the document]
microsoftly: it's used in multiple tests. Just slightly DRYer considering it's using the same fixture

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,98 @@ type Lockfile struct {
...
+		ID:      id,
+		Imports: imports,
+	}
+}
+
+func fromSubNodeModules(currentDir string, previousPackage pkg.Package) (map[pkg.ID]pkg.Package, error) {
+	submoduleProjects := make(map[pkg.ID]pkg.Package)
+
+	for i, imported := range previousPackage.Imports {
+		pathToSubModule, subProject, err := subModule(imported.Target, currentDir)
```
golangcibot: ineffectual assignment to `err`

microsoftly: addressed, should be hidden

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,102 @@ type Lockfile struct {
 func FromLockfile(filename string) (Lockfile, error) {
 	return Lockfile{}, errors.ErrNotImplemented
 }
+
+// TODO: add support for NODE_PATH and GLOBAL_FOLDERS.
+func subModulePath(moduleName string, currentDir string) (string, error) {
+	var modulePath string
+	parentPath := filepath.Dir(currentDir)
+	if filepath.Base(currentDir) == "node_modules" {
```
liftM: Why is this needed?

buildtools/npm/manifest_test.go
```diff
@@ -1,65 +1,111 @@
...
+			Type:     pkg.NodeJS,
+		},
+	}
+
+	return
+}
+
 func TestFromManifest(t *testing.T) {
-	manifest, err := npm.FromManifest("testdata/package.json")
+	pkg, err := npm.PackageFromManifest(filepath.Join("testdata", "nested_node_modules/"), "package.json")
```
liftM: `nested_node_modules/` should not have the trailing slash. This defeats the purpose of using `filepath.Join`.

buildtools/npm/manifest_test.go
```diff
@@ -1,65 +1,111 @@
...
+	"path/filepath"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 
 	"github.com/fossas/fossa-cli/buildtools/npm"
+	"github.com/fossas/fossa-cli/pkg"
 )
 
+func getDirectDepImports() (chaiDirectDep pkg.Import, typeDetectDirectDep pkg.Import) {
```
liftM: Again, why not make these package variables? There's no advantage to making this a function. I would prefer using the language construct that communicates intent most clearly rather than the technically safest one.

buildtools/npm/manifest_test.go
```diff
@@ -1,65 +1,111 @@
...
+	typeDetectProject := depGraph.Transitive[typeDetectDirectDep.Resolved]
+	assert.Equal(t, typeDetectDirectDep.Resolved, typeDetectProject.ID)
+	assert.Empty(t, typeDetectProject.Imports)
 }
 
-func containsDep(manifests []npm.Manifest, name string, version string) bool {
-	_, err := selectDep(manifests, name, version)
-	return err == nil
+func AssertImport(t *testing.T, imports pkg.Imports, name, revision string) {
+	// Inline the `ContainsImport` logic.
```
liftM: [nit] Extra comment.

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,102 @@ type Lockfile struct {
...
+		for pkgID, nextLevelSubModule := range nextLevelSubModules {
+			submoduleProjects[nextLevelSubModule.ID] = nextLevelSubModule
+			submoduleProjects[pkgID] = nextLevelSubModule
+		}
+	}
+
+	return submoduleProjects, nil
+}
+
+func subModule(moduleName string, currentDir string) (pathToSubModule string, modulePackage pkg.Package, err error) {
```
liftM: [nit] `moduleName, currentDir string`.

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,58 @@
...
 	}
-	return manifest, nil
+
+	return convertManifestToPkg(manifest), nil
 }
 
-func FromNodeModules(dir string) ([]Manifest, error) {
-	return nil, errors.ErrNotImplemented
+// FromNodeModules Generates the dep graph based on the manifest provided in the supplied directory
+func FromNodeModules(dir string, manifestFileName string) (graph.Deps, error) {
```
liftM: Why is `dir` still needed?

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,58 @@
...
+	}
+
+	transitiveDeps, err := fromSubNodeModules(dir, rootPackage)
+
+	if err != nil {
+		return graph.Deps{}, err
+	}
+
+	return graph.Deps{
+		Direct:     rootPackage.Imports,
```
liftM: Are these versions resolved? If so, where are they resolved?

buildtools/npm/manifest_test.go
```diff
@@ -1,65 +1,111 @@
...
+	"github.com/fossas/fossa-cli/pkg"
 )
 
+func getDirectDepImports() (chaiDirectDep pkg.Import, typeDetectDirectDep pkg.Import) {
+	chaiDirectDep = pkg.Import{
+		Target: "chai",
+		Resolved: pkg.ID{
+			Location: "",
+			Name:     "chai",
+			Revision: "4.1.2",
```
liftM: Your fixtures don't have unresolved direct dependencies.

buildtools/npm/manifest.go
liftM: Why use named returns?

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,102 @@ type Lockfile struct {
...
+	for i, imported := range previousPackage.Imports {
+		pathToSubModule, subProject, err := subModule(imported.Target, currentDir)
+		if err != nil {
+			return nil, err
+		}
+
+		submoduleProjects[subProject.ID] = subProject
+
+		// update previous project's revision resolved reference to stamp out non-deterministic behavior (e.g. semver defined versions in package.json)
+		previousPackage.Imports[i].Resolved.Revision = subProject.ID.Revision
```
liftM: I'm not certain that this actually _changes_ the imports of the previous package, because the package is passed by value.

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,102 @@ type Lockfile struct {
...
+
+		nextLevelSubModules, err := fromSubNodeModules(pathToSubModule, subProject)
+		if err != nil {
+			return nil, err
+		}
+
+		for pkgID, nextLevelSubModule := range nextLevelSubModules {
+			submoduleProjects[nextLevelSubModule.ID] = nextLevelSubModule
+			submoduleProjects[pkgID] = nextLevelSubModule
```
liftM: Isn't this doing the same thing twice?

buildtools/npm/manifest.go
```diff
@@ -35,3 +65,102 @@ type Lockfile struct {
...
+		imports = append(imports, depImport)
+	}
+
+	return pkg.Package{
+		ID:      id,
+		Imports: imports,
+	}
+}
+
+func fromSubNodeModules(currentDir string, previousPackage pkg.Package) (map[pkg.ID]pkg.Package, error) {
```
liftM: The naming of "submodule" makes this logic very confusing to follow, because Node has no notion of submodules. Can we sync over call about this?

buildtools/npm/manifest.go
microsoftly: It does, it's value is referencing an array on the heap, which does not get copied.

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,58 @@
...
+	filePath := filepath.Join(pathElems...)
+
+	err := files.ReadJSON(&manifest, filePath)
 	if err != nil {
-		return Manifest{}, err
+		return pkg.Package{}, err
 	}
-	return manifest, nil
+
+	return convertManifestToPkg(manifest), nil
```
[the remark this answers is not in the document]
microsoftly: I like this as a separate function. Considering it's ~20 lines, it looks cleaner/ is easier to read.

buildtools/npm/manifest.go
```diff
@@ -35,3 +69,152 @@ type Lockfile struct {
...
+		imports = append(imports, depImport)
+	}
+
+	return pkg.Package{
+		ID:      id,
+		Imports: imports,
+	}
+}
+
+func fromModulesHelper(currentDir string, previousPackage pkg.Package) (map[pkg.ID]pkg.Package, error) {
```
golangcibot: func `fromModulesHelper` is unused

buildtools/npm/manifest.go
```diff
@@ -35,3 +69,152 @@ type Lockfile struct {
...
+
+		for _, nextLevelSubModule := range nextLevelSubModules {
+			submoduleProjects[nextLevelSubModule.ID] = nextLevelSubModule
+		}
+	}
+
+	return submoduleProjects, nil
+}
+
+func subModule(moduleName string, currentDir string) (string, pkg.Package, error) {
```
golangcibot: func `subModule` is unused

microsoftly:
I updated the fixtures to include `"type-detect": "^3.0.0"` to ensure the solution fixes semver ambiguity on the direct deps

buildtools/npm/manifest.go
```diff
@@ -35,3 +71,96 @@ type Lockfile struct {
...
+		modulePackage, err := PackageFromManifest(pathToDepModule) //, "package.json")
+		if err != nil {
+			return nil, err
+		}
+
+		moduleProjects[modulePackage.ID] = modulePackage
+
+		// update previous project's revision resolved reference to stamp out non-deterministic behavior (e.g. semver defined versions in package.json)
+		currentModule.Imports[i].Resolved = modulePackage.ID
+		println("UPDATING " + currentModule.Imports[i].Target + "@" + currentModule.Imports[i].Resolved.Revision + " to " + modulePackage.ID.Revision)
```
liftM: Don't forget to remove debugging statements.

buildtools/npm/manifest.go
```diff
@@ -35,3 +71,96 @@ type Lockfile struct {
...
+
+		moduleProjects[currentModule.ID] = currentModule
+
+		nextLevelSubModules, err := fromModulesHelper(pathToDepModule)
+		if err != nil {
+			return nil, err
+		}
+
+		for _, nextLevelSubModule := range nextLevelSubModules {
```
liftM: [nit] This copying will get `O(scary)`. To avoid it, you can pass a pointer to a map used at all levels of recursion.

Unless you're seeing performance issues right now, I wouldn't worry about it.

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,64 @@
...
+	Name         string
+	Version      string
 	Dependencies map[string]string
 }
 
-func FromManifest(filename string) (Manifest, error) {
-	var manifest Manifest
-	err := files.ReadJSON(&manifest, filename)
+// PackageFromManifest generates a package definition for the provided manifest in the supplied directory
+func PackageFromManifest(pathElems ...string) (pkg.Package, error) {
```
liftM: I think it makes more sense for `PackageFromManifest` to contain the direct import resolution logic. This has two advantages:

1. Consumers don't need to remember the unstated contract that `PackageFromManifest` only provides unresolved direct imports, and that they need to do resolution separately. A simpler interface is less error-prone.
2. This will make the logic in `FromNodeModules` more readable. The existing logic looks something like:

	```
	func nodeModulesHelper(manifest):
		pkg := PackageFromManifest(manifest)
	
		for import in unresolvedPkg.Imports:
			1. resolve import
			2. set resolved import field in pkg
			3. recurse nodeModulesHelper(import filepath)
	
		return deps
	```
	
	The new logic can look something like:
	```
	func nodeModulesHelper(manifest):
		// the imports in here will already be resolved
		pkg := PackageFromManifest(manifest)
	
		for import in pkg.Imports:
			recurse nodeModulesHelper(import filepath)
		
		return deps
	```

	You can store the import filepath in the `Location` field of the `pkg.ID` of the `pkg.Import.Resolved` during `PackageFromManifest` to avoid doing that WalkUp twice, although I think doing it twice (i.e. invoking `modulePath` once in `PackageFromManifest` and again in `nodeModulesHelper`) is acceptable performance overhead for the increased clarity.

If you disagree, I'd still like to add comments to the individual parts of logic in `nodeModulesHelper` (i.e. resolving, setting the field, recursing). Go lint doesn't _require_ comments on private functions, but we still _should_ comment them in cases like this where it clarifies the function's intention. Function names are not always sufficient for clarity, and sometimes it's better to have a shorter name and a clear comment for ergonomic reasons.

microsoftly: I disagree here. `PackageFromManifest` is used in places where that resolution might not be possible (i.e. no node_modules folder; e.g. https://github.com/fossas/fossa-cli/pull/261/files#diff-612d68e420391fb3cc54bf19ef4b44f4R199).

I'll move forward with comments

liftM: 1. I don't see how resolution is not possible here. It's just not needed.
2. Maybe we should have different functions and data structures for "read a manifest" vs. "construct a package". I don't like how the constructed package implicitly has unresolved dependencies. Are there any cases where resolution is not needed _and_ we need a `pkg.Package`?

microsoftly: The resolution not being possible would be in the case of not having `node_modules` folder. If it requires a . `node_modules` folder, then the function would error out inside the `IsBuilt` function.  

I think a fair middleground would be attempting to resolve with node_modules if possible

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,94 @@
...
+	nodeModulesExists, err := files.ExistsFolder(nodeModuleDirectory)
+	if err != nil {
+		return pkg.Package{}, err
+	}
+
+	if !nodeModulesExists {
+		return manifestAsPackage, nil
+	}
+
+	resolveDirectDependencyVersions(&manifestAsPackage, nodeModuleDirectory)
```
golangcibot: Error return value of `resolveDirectDependencyVersions` is not checked

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,97 @@
...
+	nodeModulesExists, err := files.ExistsFolder(nodeModuleDirectory)
+	if err != nil {
+		return pkg.Package{}, err
+	}
+
+	if !nodeModulesExists {
+		return manifestAsPackage, nil
+	}
+
+	err = resolveDirectDependencyVersions(&manifestAsPackage, nodeModuleDirectory)
```
golangcibot: ineffectual assignment to `err`

buildtools/npm/manifest.go
liftM: You never _know_ if you have a `node_modules` folder. Even during the analysis, an unbuilt project is a failure mode we need to gracefully handle.

buildtools/npm/manifest.go
```diff
@@ -35,3 +107,113 @@ type Lockfile struct {
...
+
+/*
+	1. get package at pathToModule
+	2. add  currentPackage to the accumulator
+	3. for each of the dependencies found in currentPackage
+		3a. determine the correct path to that module
+		3b. create a package for that dependency and add it to the accumulator
+		3c. recurse to 1 using the path to the dependency
+*/
+func fromModulesHelper(pathToModule string, moduleProjects *map[pkg.ID]pkg.Package) error {
```
liftM: Maps are reference types. You don't need to pass a pointer to them.

buildtools/npm/manifest.go
```diff
@@ -1,28 +1,100 @@
...
+	filePath := filepath.Join(pathElems...)
+	manifest, err := FromManifest(filePath)
+	if err != nil {
+		return pkg.Package{}, err
+	}
+
+	manifestAsPackage := convertManifestToPkg(manifest)
+
+	// attempt to resolve revisions if node_modules folder exists and package any imports
+	if len(manifestAsPackage.Imports) == 0 {
```
liftM: Is this check necessary?

microsoftly: small optimization. If we expect 0 imports, we don't need to check for the existence of `node_modules`, saving slightly on i/o

PR #269, opened 2018-09-18
feat(npm fallback) use node_modules to build dep graph if npm fails

microsoftly:


analyzers/nodejs/mocks_test.go
```diff
@@ -25,3 +27,17 @@ func (n MockNPM) Clean(dir string) error {
 func (n MockNPM) Install(dir string) error {
 	return nil
 }
+
+type MockNPMFailure struct{}
+
+func (n MockNPMFailure) List(_ string) (npm.Output, error) {
+	return npm.Output{}, errors.New("expected failure")
```
liftM: I'd prefer if these error messages had different strings so it would be easier to distinguish them.

analyzers/nodejs/nodejs.go
```diff
@@ -221,6 +221,9 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	pkgs, err := a.Tool.List(filepath.Dir(a.Module.BuildTarget))
 	if err != nil {
 		log.Warnf("NPM had non-zero exit code: %s", err.Error())
+		log.Info("Using fallback of node_modules")
```
liftM: This should be a debug message. It means nothing to the user.

analyzers/nodejs/nodejs_test.go
```diff
@@ -109,25 +109,27 @@ func TestAnalyzeWithNpmLs(t *testing.T) {
...
+		Options:     map[string]interface{}{},
+	}
 
-	// sysTool := nodejs.SystemNpmTool(nodeModule.Options)
+	analyzer, err := nodejs.New(nodeModule)
+	assert.NoError(t, err)
 
-	// analyzer, err := nodejs.New(nodeModule)
-	// assert.NoError(t, err)
+	analyzer.Tool = MockNPMFailure{}
```
liftM: This is really clever. I like it.

analyzers/nodejs/nodejs_test.go
```diff
@@ -109,25 +109,27 @@ func TestAnalyzeWithNpmLs(t *testing.T) {
...
 
-	// analysisResults, err := analyzer.Analyze()
-	// assert.NoError(t, err)
+	for dep := range analysisResults.Transitive {
+		println(dep.Name + "@" + dep.Revision)
+	}
 
-	// assert.NotEmpty(t, analysisResults.Direct)
+	assert.Len(t, analysisResults.Direct, 1)
+	assert.Len(t, analysisResults.Transitive, 7)
```
liftM: Can you check contents as well?

PR #271, opened 2018-09-19
 feat(AnalyzerTest) canonical reference for slimming docker test image and native analyzer testing

microsoftly:
This should serve as an example for how to move away from having a bloated docker image and have native integration tests


analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,74 @@
...
+		Options: map[string]interface{}{"allow-npm-err": true},
+	}
+
+	analyzer, err := analyzers.New(module)
+	assert.NoError(t, err)
+
+	deps, err := analyzer.Analyze()
+	assert.NoError(t, err)
+	assert.NotEmpty(t, deps.Direct)
+	assert.NotEmpty(t, deps.Transitive)
```
microsoftly: While these are not in depth, they yield the same level of confidence that `test.sh` would give

.circleci/config.yml
```diff
@@ -88,7 +88,10 @@ jobs:
           command: |
             # Load shell helpers (e.g. sdkman)
             source /home/fossa/.bashrc
-            # Run tests
+            # Run integration tagged native tests
+            go test ./... -tags=integration
```
microsoftly: I've been running into some issues getting the tags to work as expected on the docker image, I'll take a longer look tomorrow

Makefile
```diff
@@ -91,11 +91,11 @@ test:
 
 .PHONY: unit-test
 unit-test:
-	go test ./...
+	go test ./... -short
 
 .PHONY: junit-test
 junit-test: $(GO_JUNIT_REPORT) $(GOVERALLS)
-	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) | go-junit-report
+	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
```
microsoftly: prevent the long running integration test from running

test/integration/analyzers/nodejs_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/test/project_initializer/node"
+	"github.com/fossas/fossa-cli/test/testtools"
+	"github.com/stretchr/testify/assert"
+)
+
+var nodeInitializer = node.New()
+
+func TestMain(m *testing.M) {
```
microsoftly: `TestMain` runs once and controls running of the test suite. after calling `flag.Parse`, we can short circuit all tests here if not running integration tests

test/testtools/testtools.go
```diff
@@ -0,0 +1,69 @@
...
+	FixtureDirectory() string
+	// BuildAll fetches source code and runs all install/build scrips necessary for analysis
+	BuildAll() error
+	// RemoveAll removes all artifacts from BuildAll
+	RemoveAll() error
+}
+
+// SetupTests executes the initializer's build all and returns a function to be deferred for teardown. SetupTests and it's returned function are meant to be called once per analyzer type
+func SetupTests(initializer Initializer) (func(), error) {
+	err := initializer.BuildAll()
```
liftM: It would make more sense for the consumer to call `initializer.BuildAll()` and then `defer initializer.RemoveAll()`.

analyzers/integration_tests/project_initializer/node/node_initializer.go
```diff
@@ -0,0 +1,153 @@
...
+	return nil
+}
+
+func (n NodeInitializer) RemoveAll() error {
+	testDir := n.FixtureDirectory()
+	return os.RemoveAll(testDir)
+}
+
+func (n NodeInitializer) FixtureDirectory() string {
+	return filepath.Join(os.TempDir(), "nodejs")
```
liftM: I would prefer a FOSSA CLI-specific path e.g. `filepath.Join(os.TempDir(), "fossa-cli-fixtures", "analyzers", "nodejs")`.

analyzers/integration_tests/analyzers/nodejs_test.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/fossas/fossa-cli/analyzers/integration_tests/testtools"
+	"github.com/fossas/fossa-cli/files"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
+)
+
+var nodeInitializer = node.New()
+
+func TestMain(m *testing.M) {
```
liftM: This should be part of the Node.JS analyzer tests. In general, tests should be colocated with the package being tested.

analyzers/integration_tests/testtools/testtools.go
```diff
@@ -0,0 +1,69 @@
+package testtools
+
+import (
+	"github.com/fossas/fossa-cli/exec"
+
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+type Initializer interface {
```
liftM: I don't think using an interface is the right level of abstraction. This seems too general to me.

The only consumer of this interface is `SetupTests`, which just calls `BuildAll` and `RemoveAll`.

analyzers/integration_tests/project_initializer/node/node_initializer.go
```diff
@@ -0,0 +1,153 @@
...
+
+	err = os.Mkdir(testDir, 0700)
+	if err != nil {
+		return err
+	}
+
+	var waitGroup sync.WaitGroup
+	waitGroup.Add(len(n.projects))
+
+	for _, project := range n.projects {
```
liftM: I think _this_ is the actual logic that will be common between packages. It would be more useful to provide `func CloneProjects(dir string, projects []Project) error` as a testing tool.

.circleci/config.yml
```diff
@@ -72,6 +72,7 @@ jobs:
     docker:
       - image: quay.io/fossa/fossa-cli-test:circleci
     steps:
+      - checkout
```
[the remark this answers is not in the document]
microsoftly: This was me trying to troubleshoot the issue I mentioned earlier

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,180 @@
...
+var nodeAnalyzerFixtureDir = filepath.Join(testUtils.FixtureDirectory(), "nodejs", "analyzer")
+
+func TestMain(m *testing.M) {
+	// flags are not parsed at this point. In order to have testing.Short() read actually provided values, this must be executed
+	flag.Parse()
+	if testing.Short() {
+		return
+	}
+
+	testUtils.Clone(nodeAnalyzerFixtureDir, projects)
```
golangcibot: Error return value of `testUtils.Clone` is not checked

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	println("projects cloned, starting initialization")
+
+	initializeProjects(nodeAnalyzerFixtureDir)
```
golangcibot: Error return value of `initializeProjects` is not checked

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	println("Initialization complete, running tests")
+
+	exitCode := m.Run()
+	defer os.Exit(exitCode)
+	defer cleanUp(nodeAnalyzerFixtureDir)
```
golangcibot: Error return value of `cleanUp` is not checked

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,203 @@
...
+	},
+	testUtils.ProjectFixture{
+		Name:   "request",
+		URL:    "https://github.com/request/request",
+		Commit: "8162961dfdb73dc35a5a4bfeefb858c2ed2ccbb7",
+		Env:    map[string]string{},
+	},
+}
+
+func cleanUp(dir string) {
```
golangcibot: `cleanUp` is unused

microsoftly: I'm still on the fence as to whether or not we should cleanup the fixtures after the run. It's not necessary for CI and it's also not necessary for local dev; it will likely save time. If you need a new run locally, you can just delete the fixtures manually

.circleci/config.yml
```diff
@@ -54,41 +52,35 @@ jobs:
...
           name: Run integration tests
           command: |
             # Load shell helpers (e.g. sdkman)
             source /home/fossa/.bashrc
-            # Run tests
+
+            # Run native integration tests
+            # as long as -short is not being called, integration tests get run
+            go test ./... -v -timeout 900s # 15 min
+            #Run shell based tests
```
liftM: [nit] Whitespace.

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,210 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/analyzers"
+	"github.com/fossas/fossa-cli/exec"
+	"github.com/fossas/fossa-cli/files"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	testUtils "github.com/fossas/fossa-cli/testing"
```
liftM: This naming is kind of annoying, given the clash with stdlib's `testing`. What are your thoughts on using subpackages? My initial thinking:
- `testing/fixtures` (for fixture functions: consumers would call e.g. `fixtures.Clone(...)`)
- `testing/runfossa` (for invoking `fossa` commands end-to-end e.g. `runfossa.Init(...)`).

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,210 @@
...
+var projects = []testUtils.ProjectFixture{
+	testUtils.ProjectFixture{
+		Name:   "puppeteer",
+		URL:    "https://github.com/GoogleChrome/puppeteer",
+		Commit: "b97bddf8e5750d20c6ba82392eebe2a3fd2dd218",
+		Env: map[string]string{
+			"PUPPETEER_SKIP_CHROMIUM_DOWNLOAD": "1",
+		},
+	},
+	testUtils.ProjectFixture{
```
liftM: Is it preferable to pull these fixtures from NPM or GitHub? I can see arguments for both, but I'm wondering what your thoughts are.

analyzers/nodejs/integration_test.go
liftM: We should not clean up fixtures. Instead, they should be left on disk (or saved in cache in CI) and we should verify their integrity before running (for git repositories, this is probably as simple checking that `git status` is clean and the revision hash is correct).

analyzers/nodejs/integration_test.go
microsoftly: I think we should pull this from git. We can version them ourselves and it gives us control. Pulling from npm means that we'll always be reliant on another toolchain we don't control

Additionally, pulling from git allows for a consistent repo for all fixtures, regardless of ecosystem

analyzers/nodejs/integration_test.go
microsoftly: I think that seems reasonable

testing/runfossa/runfossa.go
```diff
@@ -0,0 +1,20 @@
...
+// Init executes fossa init the provided directory
+func Init(projectDir string) error {
+	cmd := exec.Cmd{
+		Argv:    []string{"init"},
+		Name:    "fossa",
+		Dir:     projectDir,
+		Command: "fossa",
+	}
+
+	_, errMsg, err := exec.Run(cmd)
```
liftM: [nit] `s/errMsg/stderr`, but that's just my personal preference.

testing/fixtures/fixutres.go
```diff
@@ -0,0 +1,74 @@
+package fixtures
```
liftM: [nit] Filename: `s/fixutres/fixtures`.

testing/fixtures/fixutres.go
```diff
@@ -0,0 +1,74 @@
...
+	"path/filepath"
+	"sync"
+
+	"github.com/fossas/fossa-cli/files"
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ProjectFixture describes a git locator for a test fixture and env vars for when it is to be built
+type ProjectFixture struct {
```
liftM: I'd just call this `Project` since it's already in `package fixtures` (users will consume it as `fixtures.Project` which sounds better than `fixtures.ProjectFixture`).

testing/fixtures/fixutres.go
```diff
@@ -0,0 +1,74 @@
...
+	"os"
+	"path/filepath"
+	"sync"
+
+	"github.com/fossas/fossa-cli/files"
+	git "gopkg.in/src-d/go-git.v4"
+	"gopkg.in/src-d/go-git.v4/plumbing"
+)
+
+// ProjectFixture describes a git locator for a test fixture and env vars for when it is to be built
```
liftM: Describes a git _repository_.

testing/fixtures/fixutres.go
```diff
@@ -0,0 +1,74 @@
...
+type ProjectFixture struct {
+	Name   string
+	URL    string
+	Commit string
+	Env    map[string]string
+}
+
+// FixtureDirectory returns the directory under which tests fixtures should be placed
+func FixtureDirectory() string {
```
liftM: I'd name this `Directory` so it doesn't stutter (import will be `fixtures.Directory()`).

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+func TestAnalysisOutput(t *testing.T) {
+	t.Parallel()
+	for _, proj := range projects {
+		t.Run(proj.Name, func(t *testing.T) {
+			module := module.Module{
+				Dir:         filepath.Join(nodeAnalyzerFixtureDir, proj.Name),
+				Type:        pkg.NodeJS,
+				Name:        proj.Name,
+				Options:     map[string]interface{}{"allow-npm-err": true},
```
liftM: We might want to use a table-driven test to set options only on modules that need it. I'd rather not set `allow-npm-err` for fixtures that don't need it.

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,211 @@
...
+			assert.NoError(t, err)
+
+			deps, err := analyzer.Analyze()
+			assert.NoError(t, err)
+			// faker has no deps
+			if proj.Name == "fakerjs" {
+				assert.Empty(t, deps.Direct)
+				assert.Empty(t, deps.Transitive)
+			} else {
+				assert.NotEmpty(t, deps.Direct)
```
liftM: Likewise, we can use a table-driven test to assert on _contents_ (not just existence) here.

analyzers/nodejs/integration_test.go
```diff
@@ -0,0 +1,211 @@
...
+			ymlAlreadyExists, err := files.Exists(filepath.Join(projectDir, ".fossa.yml"))
+			if err != nil {
+				panic(err)
+			}
+			if ymlAlreadyExists {
+				return
+			}
+
+			// any key will work to prevent the "NEED KEY" error message
+			err = runfossa.Init(projectDir)
```
liftM: The need to run `fossa init` may be general enough among integration tests to extract it outside of this package.

In the `fixtures` package, it may be helpful to have `fixtures.Add(dir string, fixtures []Fixture)` which does a clone and then `fossa init`.

microsoftly:
@liftM My goal was to get the tests to have equivalent value as the shell script tests then refine them once we settled on a standard approach

.circleci/config.yml
```diff
@@ -76,7 +76,10 @@ jobs:
           command: |
             # Load shell helpers (e.g. sdkman)
             source /home/fossa/.bashrc
-            # Run tests
+            # Run native integration tests
+            # as long as -short is not being called, integration tests get run
+            go test ./... -v -timeout 900s # 15 min
```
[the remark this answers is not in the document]
microsoftly: would there be a good argument for why we should/shouldn't have these tests run the full suite and coverage here?

(we can keep unit tests separate to fail quickly)

analyzers/nodejs/integration_test.go
microsoftly: See comment below

analyzers/nodejs/integration_test.go
microsoftly: While possible, I think we may want to keep this separate for now. Once we establish a pattern it may be useful to DRY it up. 

E.g. there may be tests where we want to test `fossa init` or different config files

.circleci/config.yml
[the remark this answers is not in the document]
microsoftly: also, that's timeout after 10 minutes fo no output, not 10 minutes of runtime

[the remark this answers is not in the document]
liftM: > would there be a good argument for why we should/shouldn't have these tests run the full suite and coverage here?
> 
> (we can keep unit tests separate to fail quickly)

We'll be testing fallbacks as unit tests, so we want to run them in an environment without build tools.

Issue #264, opened 2018-09-17
Automatically infer ignored Go packages from manifest


liftM:
We can currently ignore Go packages by setting module options.

Some package managers (e.g. `dep`) allow you to specify packages to ignore. We should automatically read these package paths from the manifests, and set that accordingly.

PR #272, opened 2018-09-19
Fix/go dep ignore

zlav:
Issue [#264] - Add functionality for dep package manager to ignore packages

### DIFF
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -38,7 +38,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 		if a.Options.LockfilePath == "" {
 			return graph.Deps{}, errors.New("manifest strategy specified without lockfile path")
 		}
-		r, err = dep.FromFile(a.Options.LockfilePath)
+		r, err = dep.New(a.Options.LockfilePath, a.Options.ManifestPath)
 		if err != nil {
 			return graph.Deps{}, err
 		}
--- a/analyzers/golang/resolver/lockfile.go
+++ b/analyzers/golang/resolver/lockfile.go
@@ -2,6 +2,7 @@ package resolver
 
 import (
 	"errors"
+	"path/filepath"
 
 	"github.com/fossas/fossa-cli/buildtools/dep"
 	"github.com/fossas/fossa-cli/buildtools/gdm"
@@ -30,7 +31,7 @@ const (
 func FromLockfile(tool Type, dir string) (Resolver, error) {
 	switch tool {
 	case Dep:
-		return dep.New(dir)
+		return dep.New(filepath.Join(dir, "Gopkg.lock"), filepath.Join(dir, "Gopkg.toml"))
 	case Gdm:
 		return gdm.New(dir)
 	case Glide:
--- a/buildtools/dep/dep.go
+++ b/buildtools/dep/dep.go
@@ -11,14 +11,14 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Resolver contains both the lockfile and Manifest information. Resolver implements golang.Resolver.
+// Resolver contains both the lockfile and manifest information. Resolver implements golang.Resolver.
 type Resolver struct {
-	Manifest Manifest
+	manifest manifest
 	lockfile lockfile
 }
 
-// Manifest contains the ignored packages in a dep toml file. Manifest implements golang.Resolver.
-type Manifest struct {
+// manifest contains the ignored packages in a dep toml file. manifest implements golang.Resolver.
+type manifest struct {
 	Ignored []string
 }
 
@@ -40,7 +40,7 @@ type Project struct {
 // lockfile and checks to see if it should be ignored. If the package cannot be
 // ignored and is not found, buildtools.ErrNoRevisionForPackage is returned.
 func (r Resolver) Resolve(importpath string) (pkg.Import, error) {
-	if r.Manifest.isIgnored(importpath) {
+	if r.manifest.isIgnored(importpath) {
 		return pkg.Import{}, buildtools.ErrPackageIsIgnored
 	}
 
@@ -53,7 +53,7 @@ func (r Resolver) Resolve(importpath string) (pkg.Import, error) {
 }
 
 // isIgnored checks if a Go package can be ignored according to a dep manifest.
-func (m Manifest) isIgnored(importpath string) bool {
+func (m manifest) isIgnored(importpath string) bool {
 	for _, ignoredPackage := range m.Ignored {
 		if strings.HasSuffix(ignoredPackage, "*") {
 			ignoredPrefix := ignoredPackage[:len(ignoredPackage)-1]
@@ -80,7 +80,7 @@ func New(lockfilePath string, manifestPath string) (Resolver, error) {
 		return Resolver{}, err
 	}
 
-	resolver.Manifest, err = readManifest(manifestPath)
+	resolver.manifest, err = readManifest(manifestPath)
 	if err != nil {
 		return Resolver{}, err
 	}
@@ -117,13 +117,13 @@ func readLockfile(filepath string) (lockfile, error) {
 	return lock, nil
 }
 
-// readManifest returns a Manifest using the provided filepath.
-func readManifest(filepath string) (Manifest, error) {
-	var manifest Manifest
-	err := files.ReadTOML(&manifest, filepath)
+// readManifest returns a manifest using the provided filepath.
+func readManifest(filepath string) (manifest, error) {
+	var man manifest
+	err := files.ReadTOML(&man, filepath)
 	if err != nil {
-		return Manifest{}, errors.Wrap(err, "No manifest Gopkg.toml found")
+		return manifest{}, errors.Wrap(err, "No manifest Gopkg.toml found")
 	}
 
-	return manifest, nil
+	return man, nil
 }
--- a/buildtools/dep/dep.go
+++ b/buildtools/dep/dep.go
@@ -2,19 +2,33 @@
 package dep
 
 import (
-	"errors"
 	"path"
-	"path/filepath"
+	"strings"
 
 	"github.com/fossas/fossa-cli/buildtools"
 	"github.com/fossas/fossa-cli/files"
 	"github.com/fossas/fossa-cli/pkg"
+	"github.com/pkg/errors"
 )
 
-// ErrNoLockfile is returned if a dep manifest is found without an accompanying lockfile.
-var ErrNoLockfile = errors.New("found Gopkg.toml but not Gopkg.lock")
+// Resolver contains both the lockfile and Manifest information. Resolver implements golang.Resolver.
+type Resolver struct {
+	Manifest Manifest
+	lockfile lockfile
+}
 
-// A Project is a single imported repository within a dep project.
+// Manifest contains the ignored packages in a dep toml file. Manifest implements golang.Resolver.
+type Manifest struct {
+	Ignored []string
+}
+
+// lockfile contains the Projects in a dep lockfile and a corresponding map for retrieving project data.
+type lockfile struct {
+	Projects   []Project
+	normalized map[string]pkg.Import
+}
+
+// Project is a single imported repository within a dep project.
 type Project struct {
 	Name     string
 	Packages []string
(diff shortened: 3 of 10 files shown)


buildtools/dep/dep.go
```diff
@@ -13,6 +15,25 @@ import (
...
+
+// A Resolver is a wrapper for Lockfile to allow for dep specific toml file behavior. Inherits resolvers
+type Resolver struct {
+	TOMLFile TOMLFile
+	Lockfile Lockfile
+}
+
+// TOMLFile contains the contents of a dep toml file. TOMLFiles are resolvers.
+// Ignored is a result of toml unmarshalling. IgnoreMap is for easy search of ignored packages
+type TOMLFile struct {
```
liftM: I would rather call this `Manifest`. `TOMLFile` is general and a bit misleading (this data structure does not represent _all_ TOML files).

In general, a "manifest" is a value describing the components in a project and a "lockfile" describes resolved revisions of components.

buildtools/dep/dep.go
```diff
@@ -22,11 +43,32 @@ type Project struct {
...
+	if r.isIgnored(importpath) {
+		return pkg.Import{}, buildtools.ErrNoRevisionForPackageOk
+	}
 
-	normalized map[string]pkg.Import // A normalized map of package import paths to revisions.
+	return r.Lockfile.Resolve(importpath)
+}
+
+//isIgnored finds packages to ignore while handling the possibility of a wildcard
+func (r Resolver) isIgnored(importpath string) bool {
```
liftM: 1. From an ontological perspective, I think this method makes more sense on the manifest.
2. I would be okay with this being public, especially for testing purposes.
3. [nit] Add a space between the `//` and the sentence.

buildtools/dep/dep.go
```diff
@@ -13,6 +15,25 @@ import (
 
 // ErrNoLockfile is returned if a dep manifest is found without an accompanying lockfile.
 var ErrNoLockfile = errors.New("found Gopkg.toml but not Gopkg.lock")
+var ErrIsIgnored = errors.New("this is ignored")
+
+// A Resolver is a wrapper for Lockfile to allow for dep specific toml file behavior. Inherits resolvers
```
liftM: [nit] This comment is a bit misleading. Both the lockfile and the manifest are dep-specific. Also, it _implements_ `Resolver`. Go has no notion of inheritance (from a technical perspective, it has no notion of subtyping).

buildtools/dep/dep.go
```diff
@@ -22,11 +43,32 @@ type Project struct {
...
+
+//isIgnored finds packages to ignore while handling the possibility of a wildcard
+func (r Resolver) isIgnored(importpath string) bool {
+
+	for _, matcher := range r.TOMLFile.Ignored {
+		if matcher == importpath {
+			return true
+		}
+
+		if matcher[len(matcher)-1] == '*' {
```
liftM: Use `strings.HasSuffix`.

buildtools/dep/dep.go
```diff
@@ -22,11 +43,32 @@ type Project struct {
...
-	normalized map[string]pkg.Import // A normalized map of package import paths to revisions.
+	return r.Lockfile.Resolve(importpath)
+}
+
+//isIgnored finds packages to ignore while handling the possibility of a wildcard
+func (r Resolver) isIgnored(importpath string) bool {
+
+	for _, matcher := range r.TOMLFile.Ignored {
+		if matcher == importpath {
+			return true
```
liftM: [nit] Technically, you should do this _after_ the wildcard check. This might break on packages with stars in the name. Are those package names valid?

buildtools/dep/dep.go
```diff
@@ -98,12 +140,23 @@ func UsedIn(dirname string) (bool, error) {
 	return false, nil
 }
 
-// FromFile reads a raw Lockfile from a Gopkg.lock file.
-func FromFile(filename string) (Lockfile, error) {
-	var lockfile Lockfile
-	err := files.ReadTOML(&lockfile, filename)
+// FromFile reads a raw toml Lockfile from a Gopkg.lock file.
+func FromFile(filename string) (Resolver, error) {
```
liftM: This function name is a bit misleading. Feel free to rearrange and rename functions to make things work better.

Here's what I would do instead:

1. `func New(dir string) (Resolver, error)` that constructs a `Resolver` given a directory containing a `Gopkg.toml` and `Gopkg.lock`.
2. `func NewManifest(filename string) (Manifest, error)` that reads, parses, and constructs a `Manifest` given the path to the `Gopkg.toml`.
3. `func NewLockfile(filename string) (Lockfile, error)` that reads, parses, and constructs a `Lockfile` given the path to the `Gopkg.lock`.

buildtools/dep/dep.go
```diff
@@ -13,6 +15,25 @@ import (
 
 // ErrNoLockfile is returned if a dep manifest is found without an accompanying lockfile.
 var ErrNoLockfile = errors.New("found Gopkg.toml but not Gopkg.lock")
+var ErrIsIgnored = errors.New("this is ignored")
```
liftM: Is this ever used?

buildtools/errors.go
```diff
@@ -4,5 +4,6 @@ import "errors"
 
 // Analysis errors.
 var (
-	ErrNoRevisionForPackage = errors.New("no revision found for package name")
+	ErrNoRevisionForPackage   = errors.New("no revision found for package name")
+	ErrNoRevisionForPackageOk = errors.New("no revision found for package name")
```
liftM: 1. Avoid having the same error message for multiple errors; this makes it very difficult to debug a failure after the fact because it's hard to tell what error caused the failure.
2. This name is a bit misleading because it's too general. In general, you want to name for ergonomics, which sometimes means avoiding generality (I know this is a hard habit to get out of). In this case, I'd probably call it something like `ErrPackageIsIgnored`.

buildtools/dep/dep.go
zlav: They are invalid on Github. Theoretically you could create a local package with a wildcard in the name but that should break dep before it got to us. Also, the * is only allowed at the end of the package by dep, so if there was a wildcard at any point in the name it would be ignored by dep which allows us to ignore the behavior as well.

buildtools/dep/dep.go
zlav: No, that was actually left over from our remote session. Good catch I never changed that line but should have.

buildtools/dep/dep.go
```diff
@@ -69,41 +108,28 @@ func New(dirname string) (Lockfile, error) {
...
-		}
-		// If both exist, then dep is being used correctly.
-		if ok {
-			return true, nil
-		}
-		// If only a manifest exists, then return ErrNoLockfile.
-		return true, ErrNoLockfile
+	err := files.ReadTOML(&lockfile, filepath)
+	if err != nil {
+		return Lockfile{}, errors.New(fmt.Sprintf("No lockfile Gopkg.lock found: %+v", err))
```
golangcibot: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...))

buildtools/dep/dep.go
```diff
@@ -69,41 +108,28 @@ func New(dirname string) (Lockfile, error) {
...
-func FromFile(filename string) (Lockfile, error) {
-	var lockfile Lockfile
-	err := files.ReadTOML(&lockfile, filename)
+// ReadManifest accepts the filepath of a manifest which it parses into a manifest object
+func ReadManifest(filepath string) (Manifest, error) {
+	var manifest Manifest
+	err := files.ReadTOML(&manifest, filepath)
 	if err != nil {
-		return Lockfile{}, err
+		return Manifest{}, errors.New(fmt.Sprintf("No manifest Gopkg.toml found: %+v", err))
```
golangcibot: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...))

zlav: Fixed

analyzers/golang/analyze.go
```diff
@@ -38,10 +38,17 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 		if a.Options.LockfilePath == "" {
```
liftM: I think it actually makes sense to specify default manifest and lockfile paths for each strategy (using the package manager's default conventions). I'm okay with you adding that into this PR or a follow-up.

analyzers/golang/resolver/lockfile.go
```diff
@@ -30,7 +30,7 @@ const (
 func FromLockfile(tool Type, dir string) (Resolver, error) {
 	switch tool {
 	case Dep:
-		return dep.New(dir)
+		return dep.New(dir+"/Gopkg.lock", dir+"/Gopkg.toml")
```
liftM: Use `filepath.Join`.

buildtools/dep/dep.go
```diff
@@ -22,39 +37,62 @@ type Project struct {
...
+		}
+	}
+
+	return false
 }
 
 // Resolve returns the revision of an imported Go package contained within the
 // lockfile. If the package is not found, buildtools.ErrNoRevisionForPackage is
 // returned.
 func (l Lockfile) Resolve(importpath string) (pkg.Import, error) {
```
liftM: Let's inline this into `Resolver.Resolve`.

buildtools/dep/dep.go
```diff
@@ -22,39 +37,62 @@ type Project struct {
...
+		return pkg.Import{}, buildtools.ErrPackageIsIgnored
+	}
 
-	normalized map[string]pkg.Import // A normalized map of package import paths to revisions.
+	return r.Lockfile.Resolve(importpath)
+}
+
+// IsIgnored finds packages to ignore while handling the possibility of a wildcard
+func (m Manifest) IsIgnored(importpath string) bool {
+
```
liftM: [nit] Whitespace.

buildtools/dep/dep.go
```diff
@@ -22,39 +37,62 @@ type Project struct {
...
+	for _, ignoredPackage := range m.Ignored {
+		if strings.HasSuffix(ignoredPackage, "*") {
+			ignoredWildcard := ignoredPackage[:len(ignoredPackage)-1]
```
liftM: [nit] I would call this `ignoredPrefix`.

buildtools/dep/dep.go
```diff
@@ -22,39 +37,62 @@ type Project struct {
...
+func (r Resolver) Resolve(importpath string) (pkg.Import, error) {
+	if r.Manifest.IsIgnored(importpath) {
+		return pkg.Import{}, buildtools.ErrPackageIsIgnored
+	}
 
-	normalized map[string]pkg.Import // A normalized map of package import paths to revisions.
+	return r.Lockfile.Resolve(importpath)
+}
+
+// IsIgnored finds packages to ignore while handling the possibility of a wildcard
```
liftM: [nit] Here and elsewhere, declaration comments should always be complete sentences (with punctuation).

buildtools/dep/dep.go
```diff
@@ -22,39 +37,62 @@ type Project struct {
 	Version  string
 }
 
-// A Lockfile contains the contents of a dep lockfile. Lockfiles are resolvers.
-type Lockfile struct {
-	Projects []Project
+// Resolve determines if the Gopkg.toml dep file has any rules that apply
```
liftM: I'm not sure what `has any rules that apply` means.

buildtools/dep/dep.go
```diff
@@ -2,17 +2,32 @@
...
-// ErrNoLockfile is returned if a dep manifest is found without an accompanying lockfile.
-var ErrNoLockfile = errors.New("found Gopkg.toml but not Gopkg.lock")
+// A Resolver contains both the Lockfile and Manifest information. Resolver implements Resolver.
+type Resolver struct {
+	Manifest Manifest
+	Lockfile Lockfile
+}
+
+// Manifest contains the contents of a dep toml file. Manifest implements Resolver.
+// Ignored is a result of toml unmarshalling. IgnoreMap is for easy search of ignored packages
```
liftM: What is `IgnoreMap`?

buildtools/dep/dep.go
```diff
@@ -2,17 +2,32 @@
...
+	"strings"
 
 	"github.com/fossas/fossa-cli/buildtools"
 	"github.com/fossas/fossa-cli/files"
 	"github.com/fossas/fossa-cli/pkg"
 )
 
-// ErrNoLockfile is returned if a dep manifest is found without an accompanying lockfile.
-var ErrNoLockfile = errors.New("found Gopkg.toml but not Gopkg.lock")
+// A Resolver contains both the Lockfile and Manifest information. Resolver implements Resolver.
```
liftM: [nit] For clarity, I would say `Resolver implements golang.Resolver`.

buildtools/dep/dep.go
```diff
@@ -69,41 +107,28 @@ func New(dirname string) (Lockfile, error) {
...
 
-// UsedIn checks whether dep is used correctly within a project folder.
-func UsedIn(dirname string) (bool, error) {
-	// Check whether there exists a manifest.
-	ok, err := files.Exists(dirname, "Gopkg.toml")
-	if err != nil {
-		return false, err
-	}
+// ReadLockFile accepts the filepath of a lockfile which it parses into a lockfile object
+func ReadLockfile(filepath string) (Lockfile, error) {
```
liftM: The normalization logic should go in here.

buildtools/dep/dep_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"testing"
+
+	"github.com/fossas/fossa-cli/buildtools"
+	"github.com/fossas/fossa-cli/buildtools/dep"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestIsIgnored(t *testing.T) {
+	manifest := dep.Manifest{Ignored: []string{"apple", "orange*"}}
```
liftM: I would prefer if you used a manifest fixture and `ReadManifest`.

buildtools/dep/dep_test.go
```diff
@@ -0,0 +1,61 @@
...
+	// Wildcard ignore other packages
+	valid = manifest.IsIgnored("orange/blood")
+	assert.Equal(t, valid, true)
+
+	// Don't ignore file that isn't in the ignored list
+	valid = manifest.IsIgnored("apple/fuji")
+	assert.Equal(t, valid, false)
+}
+
+func TestResolverResolver(t *testing.T) {
```
liftM: Just `TestResolve`.

buildtools/dep/dep_test.go
```diff
@@ -0,0 +1,61 @@
...
+	resolver := dep.Resolver{
```
liftM: I would prefer if you used manifest and lockfile file fixtures.

analyzers/golang/analyze.go
```diff
@@ -38,10 +38,17 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
 			return graph.Deps{}, errors.New("manifest strategy specified without lockfile path")
 		}
-		r, err = dep.FromFile(a.Options.LockfilePath)
+		depResolver := dep.Resolver{}
```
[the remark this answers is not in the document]
zlav: Great question. The reason I did this was to follow the structure of the other cases which use FromFile(lockfilePath), the code I modeled ReadLockfile and ReadManifest after. What do you think about changing all of the other FromFile()'s to use their New() function?

analyzers/golang/analyze.go
zlav: Agreed, judging by the line you selected are you also suggesting we remove the empty string check becuase we will no longer support custom Lockfile paths(in which case it cannot be empty)?

analyzers/golang/resolver/lockfile.go
zlav: Thank you for the tip, I haven't used that before.

buildtools/dep/dep.go
zlav: A leftover field that I chose to delete. Good catch, thank you.

buildtools/dep/dep_test.go
```diff
@@ -0,0 +1,80 @@
...
+	expectedLockfile := dep.Lockfile{
+		Projects: []dep.Project{
+			dep.Project{
+				Name:     "cat/fossa",
+				Packages: []string{"."},
+				Revision: "1",
+				Version:  "v0.3.0",
+			},
+		},
+		Normalized: map[string]pkg.Import{
```
liftM: Ah, so this is why `Normalized` is public.

I'm not a big fan of this kind of test. We're testing that the `Lockfile`'s _structure_ is what we expect, but the contract that a `Lockfile` presents to its consumers is its behaviour (i.e. that `Resolve` will work correctly).

In this case, this would be a good reason for `Lockfile` to have its own implementation of `Resolve`: so we can test that `Resolve` works as expected rather than asserting against the internal structure of the lockfile.

Alternatively, we can make `lockfile`s themselves private. In that case, they expose no public contract so it's perfectly fine to write an internal test (i.e. one in `package dep`) to check that its internals work.

In general, _external_ tests (i.e. those in `package foo_test`) should ensure that contracts presented to consumers are upheld while _internal_ tests (i.e. those in `package foo`) check that internal data structures and algorithms are working correctly and tend to change more frequently as the internals of the package change.

zlav: I like the idea of making `Lockfile` private as well as `ReadLockfile`, `ReadManifest`, and `IsIgnored`. I think this clarifies how the file should be tested. Externally: `Resolve` should simply return the revision you are looking for or alert that it is ignored or missing. Internally: `IsIgnored`, `ReadLockfile` and `ReadManifest` can then be tested. 

I have not worked with internal go tests and your comment does a really great job of explaining the different approaches. I think splitting the test files clarifies the intention of each test.

zlav: On the other hand, `Lockfile` is public in every other file, as well as most of its fields.

buildtools/dep/dep.go
```diff
@@ -12,21 +11,21 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Resolver contains both the Lockfile and Manifest information. Resolver implements golang.Resolver.
+// Resolver contains both the lockfile and Manifest information. Resolver implements golang.Resolver.
 type Resolver struct {
 	Manifest Manifest
```
liftM: I think if the lockfile is private, the manifest should be too. It feels weird to have a struct with no exported fields, but this is actually quite idiomatic in Go.

buildtools/dep/dep_test.go
liftM: Feel free to fix that in a follow-on PR. I believe that this is currently the case because `Lockfile` _is_ the `golang.Resolver` for the other Go tools.

liftM:
FYI, merge checks are failing for this PR because the commits are not signed by a key that GitHub recognises. Feel free to override this time, but please configure commit signing for your key. If you want to fix this properly: configure commit signing, then re-run the commits (or squash them, that's fine too).

Issue #274, opened 2018-09-19
FOSSA cli feedback is not clear when no API key is provided


zlav:
Currently, Fossa returns the default help output when no API key is provided. This can make it confusing for operators who do not include an API key and are unaware that is stopping fossa from running.

PR #275, opened 2018-09-19
default error value changed when an API key is not provided

zlav:


zlav:
https://github.com/fossas/fossa-cli/issues/274

cmd/fossa/main.go
```diff
@@ -64,7 +66,8 @@ func Run(ctx *cli.Context) error {
 	}
 
 	if config.APIKey() == "" && !ctx.Bool(analyze.ShowOutput) {
-		return cli.ShowAppHelp(ctx)
+		fmt.Printf("Incorrect Usage. FOSSA_API_KEY must be set as an environment variable or provided in .fossa.yml\n\n")
```
liftM: [nit] Let's actually `log.Fatal` for now since error handling at the top level is kind of weird.

PR #276, opened 2018-09-20
Filter out deps with name 'php'

anuccio1:
FOSSA should be filtering any dependency that is `comp+php`


anuccio1:
Still need to add tests

PR #277, opened 2018-09-20
fix(.circleci/config.yml) use test base docker image and check out branch cli code into the image

microsoftly:


.circleci/config.yml
```diff
@@ -70,8 +70,12 @@ jobs:
     <<: *defaults
     working_directory: /home/fossa/go/src/github.com/fossas/fossa-cli
     docker:
-      - image: quay.io/fossa/fossa-cli-test:circleci
+      - image: quay.io/fossa/fossa-cli-test-base:circleci
     steps:
+      - checkout
+      - run: 
+          name: Compile fossa binary
+          command: make
```
liftM: If we're compiling here, let's get rid of all the workspace steps that load the compiled binary from the previous step.

.circleci/config.yml
```diff
@@ -70,19 +70,19 @@ jobs:
...
       - run:
           name: Make folders
           command: |
             mkdir -p $TEST_RESULTS
             mkdir -p $ARTIFACTS
       - attach_workspace:
```
liftM: You'll also want to remove this step, and the steps in the `build` stage for saving to the workspace. It'll help to go through the whole configuration and maybe read the Circle CI documentation on workspaces and steps.

PR #279, opened 2018-09-21
WIP Fix/resolver-fix 

zlav:
Bug fix inside of cli analyze.go and buildtools logic. Currently, when a user provides a custom manifest the resolver is initialized but never creates the normalized project map. More refactoring needs to be done until it can be merged.

### DIFF
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -1,6 +1,8 @@
 package golang
 
 import (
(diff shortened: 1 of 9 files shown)


zlav:
Closing due to inactivity. This PR is still valid and there is an issue open related to the problem. Updates are coming to the golang analyzer as a whole that will make this largely irrelevant.

Issue #247, opened 2018-08-31
test cannot find what analyze uploaded


grosser:
I'd guess there is something wrong about how it builds the query since the upload does not match the locator it uses:

```
fossa analyze --config config/fossa.yml
WARNING Unrecognized locator
WARNING Unrecognized locator
============================================================
    View FOSSA Report:
    https://app.fossa.io/projects/custom+4071%2Fgit@github.com:zendesk%2Fsamson.git/refs/branch/HEAD/c84308c4092ae8cc97abf145ce8b576e63070634
============================================================

fossa test --debug --config config/fossa.yml
2018-08-31T22:10:24.267Z DEBUG config/config.go:68/readFile Trying to find configuration file at "config/fossa.yml"
2018-08-31T22:10:24.269Z DEBUG files/files.go:46/Read Reading file `config/fossa.yml`
2018-08-31T22:10:24.27Z DEBUG file%2ev1/file.go:72/New config: v1.ModuleProperties{Name:"Samson", Ignore:false, Type:"npm", BuildTarget:".", Path:".", Options:map[string]interface {}(nil)}
2018-08-31T22:10:24.27Z DEBUG file%2ev1/file.go:72/New config: v1.ModuleProperties{Name:"Gemfile", Ignore:false, Type:"gem", BuildTarget:".", Path:".", Options:map[string]interface {}(nil)}
[...]

PR #281, opened 2018-09-24
Issue #247, fix how custom fetchers are handled

zlav:
Expected format of locator URL is:
`/api/cli/custom+6214/git@github.com:zlav/pastport.git$SHAVALUE/latest_build`
Output with NormalizeGitURL:
`/api/cli/custom+6214/github.com/zlav/pastport$SHAVALUE/latest_build`

### DIFF
--- a/api/fossa/locator.go
+++ b/api/fossa/locator.go
@@ -18,18 +18,21 @@ type Locator struct {
 	Revision string `json:"revision"`
 }
 
+// String returns a locator converted to a string as a URL path for API access.
 func (l Locator) String() string {
-	if l.Fetcher != "git" {
-		if l.Fetcher == "archive" {
-			orgID, err := GetOrganizationID()
-			if err != nil {
-				log.Warnf("Could not get OrganizationID while constructing locator")
-			}
-			l.Project = orgID + "/" + l.Project
-		}
-		return l.Fetcher + "+" + l.Project + "$" + l.Revision
+	if l.Fetcher == "git" {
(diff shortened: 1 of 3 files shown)


liftM:
This problem looks like the other way around: we are correctly normalising on test but are _not_ correctly normalising on upload. The `github.com/user/project` form is canonical.

The reasoning here is that GitHub repositories are sometimes cloned via the HTTPS link and sometimes via the SSH link, and these links should point to the same FOSSA project. This mix-up is common enough that we handle GitHub as a special case (and we should probably also handle BitBucket and GitLab).

If we "fix" this correctly (by normalising on upload), we'll break existing projects in a minor way. New uploads to the projects that use the HTTPS locator will instead use the canonical locator. This might cause project-ignored issues to reappear.

A couple approaches to this issue:

1. Fix this correctly, and migrate HTTPS custom locator projects in the backend to canonical locators. Users running old versions of the CLI will still experience confusion (their CLI will still upload to the HTTPS locator).
2. Fix this correctly, and increment the configuration file version. On older configuration file versions, preserve the existing upload behaviour.
3. Do not normalise at all (which is what this PR does), which can cause inadvertent mix-ups.
4. Fix this correctly, then add logic for backwards compatibility with older projects. Specifically, uploading and testing should use the exact locator when that project already exists and the canonical locator otherwise.
5. Fix this correctly, and don't do any data migration.

(1), (4), and (5) are likely to cause confusion. I'm leaning towards either (2) or (3).

PR #287, opened 2018-09-25
added flags for specifying the team in fossa, and a link

anuccio1:
… the revision in fossa


cmd/fossa/flags/flags.go
```diff
@@ -62,6 +62,10 @@ var (
 	ProjectURLF     = cli.StringFlag{Name: ShortUpper(ProjectURL), Usage: "this repository's home page"}
 	JIRAProjectKey  = "jira-project-key"
 	JIRAProjectKeyF = cli.StringFlag{Name: Short(JIRAProjectKey), Usage: "this repository's JIRA project key"}
+	Link            = "link"
+	LinkF           = cli.StringFlag{Name: Short(Link), Usage: "A link to attach to the current build"}
```
liftM: [nit] Capitalisation: `s/A link/a link`

config/file.v1/file.go
```diff
@@ -131,6 +133,14 @@ func (file File) JIRAProjectKey() string {
 	return file.CLI.JIRAProjectKey
 }
 
+func (file File) Link() string {
+	return file.CLI.Link
```
liftM: Note that these properties will not automatically be written to the configuration file unless you explicitly update `config.WriteFile`.

(In either case, we shouldn't be writing the link to the file anyway, since we expect it to change for every build.)

PR #289, opened 2018-09-26
Fix/test upload

zlav:
This PR handles the normalization between the frontend and the backend. First commit that works, minor changes will be necessary.

### DIFF
--- a/api/fossa/locator.go
+++ b/api/fossa/locator.go
@@ -55,7 +55,7 @@ func (l Locator) URL() string {
 	if err != nil {
 		log.Fatalf("Invalid FOSSA URL: %s", err.Error())
 	}
-	return strings.Replace(server.ResolveReference(url).String(), "%", "%%", -1)
+	return server.ResolveReference(url).String()
 }
 
 // ReportURL provides a formatted URL.
--- a/api/fossa/locator.go
+++ b/api/fossa/locator.go
@@ -20,6 +20,20 @@ type Locator struct {
 
 // String returns a locator converted to a string as a URL path for API access.
 func (l Locator) String() string {
+	if l.Fetcher != "git" {
+		if l.Fetcher == "archive" {
+			orgID, err := GetOrganizationID()
+			if err != nil {
+				log.Warnf("Could not get OrganizationID while constructing locator")
+			}
+			l.Project = orgID + "/" + l.Project
+		}
+		return l.Fetcher + "+" + l.Project + "$" + l.Revision
+	}
+	return "git+" + NormalizeGitURL(l.Project) + "$" + l.Revision
+}
+
+func (l Locator) OrgString() string {
 	if l.Fetcher == "git" {
 		return "git+" + NormalizeGitURL(l.Project) + "$" + l.Revision
 	}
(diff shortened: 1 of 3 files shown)


api/fossa/locator.go
```diff
@@ -92,6 +106,27 @@ func NormalizeGitURL(project string) string {
...
+	// Remove fetcher prefix (in case project is derived from splitting a locator on '$').
+	noFetcherPrefix := strings.TrimPrefix(project, "git+")
+
+	// Normalize Git URL format only if not ssh
+	if strings.HasPrefix(noFetcherPrefix, "http") {
+		noFetcherPrefix = strings.TrimSuffix(noFetcherPrefix, ".git")
+	}
+
+	// Will not need this part ever
+	// handleGitHubSSH := strings.Replace(noGitExtension, "git@github.com:", "github.com/", 1)
```
zlav: I left this comment for reference on this commit. Will remove on the next. This normalizes ssh github urls however because we do not normalize on upload this will direct requests to the wrong locator.

api/fossa/locator_test.go
```diff
@@ -37,13 +37,37 @@ func TestLocatorFetcher(t *testing.T) {
...
 		Revision: "SHAVALUE",
 	}
 
-	stringified := custom.String()
+	stringified := git.String()
 	expected := "git+github.com/fossas/fossa-cli$SHAVALUE"
 	assert.Equal(t, stringified, expected)
 }
+
+func TestOrgStringGit(t *testing.T) {
```
liftM: Let's also add a `TestOrgStringCustom`.

api/fossa/locator.go
```diff
@@ -92,6 +106,27 @@ func NormalizeGitURL(project string) string {
 	return noHTTPSPrefix
 }
 
+// NormalizeGitURL normalizes all forms of git remote URLs to a single standard
+// form. This works around the backend only normalizing strings starting with http.
+func NormalizeGitURLTest(project string) string {
```
liftM: Let's add a gigantic `HACK: this is because of backend weirdness` comment.

api/fossa/locator.go
```diff
@@ -20,6 +20,20 @@ type Locator struct {
 
 // String returns a locator converted to a string as a URL path for API access.
 func (l Locator) String() string {
+	if l.Fetcher != "git" {
```
liftM: This is better written as `if l.Fetcher == "git" { return one-liner}`.

api/fossa/locator.go
```diff
@@ -20,19 +20,21 @@ type Locator struct {
...
+			log.Warnf("Could not get OrganizationID while constructing locator")
 		}
-		return l.Fetcher + "+" + l.Project + "$" + l.Revision
+		l.Project = orgID + "/" + l.Project
 	}
-	return "git+" + NormalizeGitURL(l.Project) + "$" + l.Revision
+	return l.Fetcher + "+" + l.Project + "$" + l.Revision
 }
 
+// OrgString returns a locator converted to a string as a URL path for API access
```
liftM: [nit] These comments should be sentences.

api/fossa/users.go
```diff
@@ -7,13 +7,20 @@ import (
 
 const OrganizationAPI = "/api/cli/organization"
 
-var CachedOrganizationID = -1
+var (
+	CachedOrganizationID = -1
```
liftM: Let's make this private.

PR #290, opened 2018-09-26
test(yarn lockfile) improve fixture case coverage

microsoftly:
* `type-detect@^3.0.0` added as a direct dep to cause duped deps with different versions
  * because `^3.0.0` is unresolved, which uncovered an edge case with direct deps

### DIFF
--- a/buildtools/yarn/parse_test.go
+++ b/buildtools/yarn/parse_test.go
@@ -24,11 +24,24 @@ func TestSimpleLockfile(t *testing.T) {
 	deps, err := yarn.FromProject(filepath.Join("testdata", "package.json"), filepath.Join("testdata", "yarn.lock"))
 	assert.NoError(t, err)
 
-	assert.Len(t, deps.Direct, 1)
+	assert.Len(t, deps.Direct, 2)
 	assert.Len(t, deps.Transitive, 7)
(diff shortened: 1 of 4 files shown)


buildtools/yarn/parse_test.go
```diff
@@ -24,18 +24,32 @@ func TestSimpleLockfile(t *testing.T) {
...
+		typeDetectImport = deps.Direct[1]
+	} else {
+		chaiImport = deps.Direct[1]
+		typeDetectImport = deps.Direct[0]
+	}
+
+	assert.Equal(t, "chai", chaiImport.Target)
+	assert.Equal(t, "4.1.2", chaiImport.Resolved.Revision)
+	assert.Equal(t, "type-detect", typeDetectImport.Target)
+	assert.Equal(t, "3.0.0", typeDetectImport.Resolved.Revision)
```
liftM: I think this would be a bit more robust if the spec did not resolve to a suffix of itself. 

For example, a dumb implementation of resolving that merely stripped the spec suffix (e.g. resolving any `^x.y.z` to `x.y.z`) would pass this test, but something like `lodash@^4.0.0` can resolve to `4.17.11`.

PR #293, opened 2018-09-26
feat(yarn lockfile fallback) add yarn lockfile reading fallback

microsoftly:
also implemented `YarnTool` in a similar manner to `NPMTool`


analyzers/nodejs/nodejs.go
```diff
@@ -162,17 +157,13 @@ func (a *Analyzer) Build() error {
...
 
 	// Prefer Yarn where possible.
-	if ok, err := files.Exists(a.Module.Dir, "yarn.lock"); err == nil && ok && a.YarnCmd != "" {
-		_, _, err := exec.Run(exec.Cmd{
-			Name: a.YarnCmd,
-			Argv: []string{"install", "--production", "--frozen-lockfile"},
-			Dir:  a.Module.Dir,
-		})
+	if ok, err := files.Exists(a.Module.Dir, "yarn.lock"); err == nil && ok && a.YarnTool.Exists() {
+		a.YarnTool.Install(a.Module.Dir)
```
golangcibot: Error return value of `a.YarnTool.Install` is not checked

analyzers/nodejs/nodejs.go
```diff
@@ -218,40 +207,48 @@ func (a *Analyzer) IsBuilt() (bool, error) {
 func (a *Analyzer) Analyze() (graph.Deps, error) {
 	log.Debugf("Running Nodejs analysis: %#v", a.Module)
 
-	pkgs, err := a.Tool.List(filepath.Dir(a.Module.BuildTarget))
-	if err != nil {
-		log.Warnf("NPM had non-zero exit code: %s", err.Error())
-		log.Debug("Using fallback of node_modules")
+	pkgs, err := a.NPMTool.List(filepath.Dir(a.Module.BuildTarget))
+	if err == nil {
```
microsoftly: restructuring checks for `err == nil` allows this func to avoid a level of nesting every time we use a fallback

analyzers/nodejs/nodejs.go
```diff
@@ -33,10 +34,8 @@ type Analyzer struct {
 	NodeCmd     string
 	NodeVersion string
 
-	Tool npm.NPM
-
-	YarnCmd     string
-	YarnVersion string
+	NPMTool  npm.NPM
```
liftM: We can probably just call these `NPM` and `Yarn`.

analyzers/nodejs/nodejs.go
```diff
@@ -33,10 +34,8 @@ type Analyzer struct {
 	NodeCmd     string
```
liftM: Do we still need this now that we have the tools?

buildtools/yarn/yarn.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+type SystemYarn struct {
+	Cmd string
+}
+
+func New() YarnTool {
+	yarnCmd, _, yarnErr := exec.Which("-v", os.Getenv("FOSSA_NPM_CMD"), "yarn")
+	if yarnErr != nil {
+		log.Warnf("Could not find Yarn %s", yarnErr.Error())
```
liftM: We should return this error instead of logging.

In general, errors should be propagated outwards as far as possible. Fewer side effects makes code easier to reason about.

analyzers/nodejs/nodejs.go
```diff
@@ -91,10 +86,8 @@ func New(m module.Module) (*Analyzer, error) {
 		NodeCmd:     nodeCmd,
 		NodeVersion: nodeVersion,
 
-		Tool: npmTool,
-
-		YarnCmd:     yarnCmd,
-		YarnVersion: yarnVersion,
+		NPMTool:  npmTool,
+		YarnTool: yarn.New(),
```
liftM: We might want to move the `os.Getenv` checking logic into here, unless anything else uses `yarn.New()` or `npm.New()`.

PR #296, opened 2018-09-27
Improve performance of .NET Core Analyze

johnduhart:
Calling file.Sync() for each log message has an extremely negative performance impact. Additionally, logging each Package/Project reference in a mid-sized solution will result in a 2GB log file.

There's likely some additional perf gains to be had in this analyze implementation, but this is the low-hanging fruit.


liftM:
PR LGTM. Thanks for the contribution! We'll merge once we get the tests passing (this looks like CI misconfiguration on our end, we'll take care of it).

@zlav can you take a look at the CI failures here? They're related to the coverage statistics we added.

It looks like we might want to do something similar to cloudflare/cfssl#203.

liftM:
I'm a bit concerned that panics and `log.Fatal`s might not flush the log if we get rid of per-entry sync'ing.

To improve the performance of this, we should probably have a `defer func() { logFile.Sync() }` in `main`. We'll need to migrate away from `log.Fatal` usage before that, since `log.Fatal` calls `os.Exit` which bypasses deferred functions.

johnduhart:
@liftM I put the sync back in, it provided no perf benefit when the log message in question was removed.

PR #297, opened 2018-09-27
test(nodejs integration) add full nodejs integration test

microsoftly:
* remove `Env` field from `fixtures.Project` struct to make it more generic
* created a `NodejsProject` and fixture array of that type to feed the table driven tests
* fixes a scoping bug where the value being used in parallel in the table driven test would get overriden when running in parallel mode
  * simply set the var within the scope, e.g. `proj := project`
* Included dependency and license reporting tests. 
  * these are merely checking the existence of output. The correctness is not verified beyond _should this be empty or not empty_
* all tests and their table driven test cases should be able to run in parallel
* caches the `fossa-cli-fixtures` directory for use in subsequent runs, speeds up the tests


analyzers/nodejs/integration_test.go
```diff
@@ -59,15 +65,25 @@ func TestTestSetup(t *testing.T) {
...
+		proj := project
+		t.Run("Analysis:"+proj.Name, func(t *testing.T) {
+			t.Parallel()
+			shouldAllowNpmErr := false
+			for _, arg := range proj.Args {
+				if arg == "allow-npm-err" {
+					shouldAllowNpmErr = true
+				}
+
+				break
```
golangcibot: the surrounding loop is unconditionally terminated

.circleci/config.yml
```diff
@@ -87,7 +87,7 @@ jobs:
             source /home/fossa/.bashrc
             # Run native integration tests
             # as long as -short is not being called, integration tests get run
-            go test ./... -v 
+            make -s integration-test > $TEST_RESULTS/go-test-report.xml
             #Run shell based tests
```
liftM: Let's get rid of these old commented-out lines for shell tests.

Makefile
```diff
@@ -90,11 +90,13 @@ unit-test:
...
-	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
+	go test ./... -short
 
 .PHONY: integration-test
-integration-test: docker-test
-	echo "native integration tests are a WIP"
+integration-test: 
+	# integration tests rely on calling shell isntances of fossa-cli, so ensure the binary is compiled
+	make
+	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) | go-junit-report
```
liftM: We don't want to submit code coverage unless we're in CI. Let's move this to the Circle CI script.

Makefile
```diff
@@ -90,11 +90,13 @@ unit-test:
 
 .PHONY: junit-test
 junit-test: $(GO_JUNIT_REPORT) $(GOVERALLS)
```
liftM: Let's remove the `$(GOVERALLS)` requirement here.

Makefile
```diff
@@ -90,11 +90,13 @@ unit-test:
...
-	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
+	go test ./... -short
 
 .PHONY: integration-test
-integration-test: docker-test
-	echo "native integration tests are a WIP"
+integration-test: 
+	# integration tests rely on calling shell isntances of fossa-cli, so ensure the binary is compiled
```
liftM: `calling the fossa-cli binary` is how I would word it. Not sure what `shell instances` means. Also, typo.

Also, I would specify `ensure the binary is recompiled before every test`.

analyzers/nodejs/integration_test.go
```diff
@@ -59,15 +65,24 @@ func TestTestSetup(t *testing.T) {
 	assertProjectFixtureExists(t, "request")
 }
 
-func TestAnalysisOutput(t *testing.T) {
+func TestNodejsAnalsis(t *testing.T) {
```
liftM: Typo.

testing/runfossa/runfossa.go
```diff
@@ -3,18 +3,42 @@ package runfossa
...
+}
+
+var licenseReportPrefix = []string{"report", "licenses"}
+
+func LicenseReport(projectDir string, args []string) (string, error) {
+	args = append(licenseReportPrefix, args...)
+	return runfossa(projectDir, args)
+}
+
+var dependencyReportPrefix = []string{"report", "dependencies"}
```
liftM: [nit] Here and above, I'd probably call these e.g. `dependencyReportCmd`.

testing/runfossa/runfossa.go
```diff
@@ -3,18 +3,42 @@ package runfossa
...
 		Name:    "fossa",
 		Dir:     projectDir,
 		Command: "fossa",
 	}
 
-	_, errMsg, err := exec.Run(cmd)
+	stdout, stderr, err := exec.Run(cmd)
 	if err != nil {
-		println(errMsg)
+		return stderr, err
```
liftM: That's a bit alarming. What about the case where `err != nil` and `stdout` is not empty?

That's exactly why `exec.Run` returns 3 arguments. You should probably return all 3 of them.

testing/runfossa/runfossa.go
```diff
@@ -3,18 +3,42 @@ package runfossa
 import "github.com/fossas/fossa-cli/exec"
 
 // Init executes fossa init the provided directory
-func Init(projectDir string) error {
+func Init(projectDir string) (string, error) {
+	output, err := runfossa(projectDir, []string{"init"})
```
liftM: Why is this command name hardcoded when others are not? Let's hardcode either all or none of them. (My preference is for all of them, especially since the commands are not referenced anywhere else.)

analyzers/nodejs/integration_test.go
```diff
@@ -130,76 +197,100 @@ func initializeProjects(testDir string) error {
...
-		URL:    "https://github.com/Marak/faker.js",
-		Commit: "3a4bb358614c1e1f5d73f4df45c13a1a7aa013d7",
-		Env:    map[string]string{},
+	NodejsProject{
+		Project: fixtures.Project{
+			Name:   "fakerjs",
+			URL:    "https://github.com/Marak/faker.js",
+			Commit: "3a4bb358614c1e1f5d73f4df45c13a1a7aa013d7",
+		},
+		Env: map[string]string{},
```
liftM: Why not omit this field?

analyzers/nodejs/integration_test.go
```diff
@@ -130,76 +197,100 @@ func initializeProjects(testDir string) error {
...
 
 	return nil
 }
 
-var projects = []fixtures.Project{
-	fixtures.Project{
-		Name:   "puppeteer",
-		URL:    "https://github.com/GoogleChrome/puppeteer",
-		Commit: "b97bddf8e5750d20c6ba82392eebe2a3fd2dd218",
+type NodejsProject struct {
```
liftM: I'm not convinced that it makes sense to separate `NodeJSProject` from `fixtures.Project`. Many other test fixtures require special arguments to analyse correctly.

analyzers/nodejs/integration_test.go
```diff
@@ -59,15 +65,24 @@ func TestTestSetup(t *testing.T) {
...
+func TestNodejsAnalsis(t *testing.T) {
 	t.Parallel()
-	for _, proj := range projects {
-		t.Run(proj.Name, func(t *testing.T) {
+	for _, project := range projects {
+		proj := project
+		t.Run("Analysis:"+proj.Name, func(t *testing.T) {
+			t.Parallel()
+			shouldAllowNpmErr := false
+			for _, arg := range proj.Args {
```
liftM: If this is the only usage of `proj.Args`, you should really store a field with module options instead.

analyzers/nodejs/integration_test.go
```diff
@@ -59,15 +65,24 @@ func TestTestSetup(t *testing.T) {
...
+				if arg == "allow-npm-err" {
```
liftM: This is never true given the fixtures you've set up below.

testing/fixtures/fixtures.go
```diff
@@ -15,7 +15,7 @@ type Project struct {
 	Name   string
 	URL    string
 	Commit string
-	Env    map[string]string
+	// Env    map[string]string
```
liftM: Just delete this if you're going to remove it.

(Which you shouldn't, as I've mentioned above.)

analyzers/nodejs/integration_test.go
microsoftly: The `fixture.Project` is so that something can be `git clone`d. The extension of that (adding env) is unique to the environment. Its information that is not needed within the cloning context.

That is why I created a `NodejsProject` with `env` to extend the case for when `npm install` is being run.

analyzers/nodejs/integration_test.go
microsoftly: This is not the only usage, it's meant to define what is passed in as `Options` and as params to other commands... that being said, if we're turning off reporting checks for now, I think that's fine

analyzers/nodejs/integration_test.go
liftM: I think `Args` should be moved into the fixture regardless, because there are many non-NodeJS fixtures (e.g. Go) that require special args to analyse.

Re: `Env`:

From an ontological perspective, I agree.

From a developer experience, `NodejsProject` seems like a lot more boilerplate for very little added value. The two questions I'm considering here are:

- Will we ever use `fixture.Projects` in a way that having `Env` is confusing? (I don't think so.)
- Is it foreseeable that non-Nodejs projects will need to set `Env`? (Yes -- consider tests where we'd like to set e.g. `FOSSA_GRADLE_BINARY` for weird `gradlew` wrappers).

analyzers/nodejs/integration_test.go
liftM: That sounds like a great idea, but we're not _actually_ doing option parsing here. I think we should go all the way for one or the other: either pass an `Options map[string]interface{}` to make our lives easy or do actual option parsing to test that code path. This half-way point feels like the worst of both worlds.

(If you decide to go the full-blown option parsing route, that will require a minor refactor of `package config` -- it may make your life easier to do that in a follow-up PR.)

analyzers/nodejs/integration_test.go
microsoftly: I think that's a fair point. I can move `Env` and `Options` into the base and see it be reasonable. I'll take that approach

Makefile
```diff
@@ -111,17 +111,18 @@ unit-test:
 
 .PHONY: ci-unit-test
 ci-unit-test: $(GO_JUNIT_REPORT) $(GOVERALLS)
+	go test -short -v ./... | go-junit-report;
```
liftM: This looks like a bad merge. `ci-unit-test` should still report coverage when possible.

testing/fixtures/fixtures.go
```diff
@@ -12,10 +12,11 @@ import (
 
 // Project describes a git repo for a test fixture and env vars for when it is to be built
 type Project struct {
-	Name   string
-	URL    string
-	Commit string
-	Env    map[string]string
+	Name    string
```
liftM: [off-topic] It is super annoying that realignment from `go fmt` causes such noisy diffs.

test.sh
```diff
@@ -89,80 +89,6 @@ time fossa analyze --output go:./cmd/agent | json
 time fossa report licenses go:./cmd/agent
 time fossa report dependencies go:./cmd/agent | json
 
-# Test known good NodeJS projects:
```
liftM: :tada:

analyzers/nodejs/integration_test.go
```diff
@@ -107,6 +110,16 @@ func initializeProjects(testDir string) error {
 			defer waitGroup.Done()
 
 			projectDir := filepath.Join(testDir, proj.Name)
+			println("initializing " + projectDir)
```
liftM: Let's get rid of all the `println`s. Use `log.Debug` instead if this is intentional debug logging.

analyzers/nodejs/integration_test.go
```diff
@@ -130,15 +143,16 @@ func initializeProjects(testDir string) error {
 			}
 
 			// any key will work to prevent the "NEED KEY" error message
-			err = runfossa.Init(projectDir)
+			stdout, stderr, err := runfossa.Init(projectDir)
 			if err != nil {
 				println("failed to run fossa init on " + proj.Name)
```
liftM: Same here. You might just want to grep for `println`.

Makefile
microsoftly: This was purposeful. I wanted to get coverage reports from the integration tests as they should have greater coverage. That being said, I think we can have multiple hooks for parallel builds with coveralls, so having this should be fine

liftM: I'm not sure I understand. This target is for unit tests.

PR #298, opened 2018-09-27
refactor(install): Use godownloader instead of custom install script

liftM:
This dramatically simplifies our release process, in preparation for automated releases being delivered in CI.


Makefile
```diff
@@ -98,30 +118,9 @@ integration-test: docker-test
...
-	git tag $(RELEASE) || (git reset HEAD^ && make abort-release)
-	rm install.prev.sh
+install.sh: $(GODOWNLOADER)
+	godownloader --repo=fossas/fossa-cli > install.sh
 
 .PHONY: release
-release:
-	if [ -z "$$(git tag -l --points-at HEAD)" ]; then exit 1; fi
-	[ "$$(grep "^  RELEASE='$$(git tag -l --points-at HEAD)'$$" install.sh | wc -l)" = "1" ]
+release: $(GORELEASER) install.sh
```
zlav: Is goreleaser using install.sh here? If not, do we still need install.sh as a prerequisite?

Makefile
```diff
@@ -98,30 +118,9 @@ integration-test: docker-test
...
 	GOVERSION=$$(go version) goreleaser $(GORELEASER_FLAGS)
```
zlav: Is it necessary to separate GORELEASER_FLAGS as a variable when they are only used once?
Use cases I can think of would be to make the makefile easier to modfiy, but I would be curious to hear the reasoning behind this.

liftM: > make the makefile easier to modify

Yup, that's basically it.

Makefile
liftM: It's not, but this ensures that the file exists.

PR #299, opened 2018-09-27
coveralls removed for forked repos and alternate junit test path added

zlav:
Coveralls causes an issue when run without an API token being present on forked repositories. This fix enables forked repositories with no coveralls token to correctly run unit tests with a junit report.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,11 @@ unit-test:
 
 .PHONY: junit-test
 junit-test: $(GO_JUNIT_REPORT) $(GOVERALLS)
-	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -90,7 +90,11 @@ unit-test:
 
 .PHONY: junit-test
 junit-test: $(GO_JUNIT_REPORT) $(GOVERALLS)
-	goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
+	if [ "$($COVERALLS_TOKEN)" != "" ]; then
+	  goveralls -v -service=circle-ci -repotoken=$(COVERALLS_TOKEN) -flags "-short" | go-junit-report
+	else
+	  go test -v ./... | go-junit-report
```
liftM: Don't forget `-short`.

PR #300, opened 2018-09-28
refactor(codegen): Clean up old code, make codegen consistent

liftM:


buildtools/pip/pipdeptree.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/fossas/fossa-cli/exec"
 )
 
-//go:generate go-bindata -pkg bindata -o ./bindata/bindata.go ./bindata/pipdeptree.py
+//go:generate go-bindata -pkg bindata -modtime 1 -o ./bindata/bindata.go ./bindata/pipdeptree.py
```
microsoftly: what difference does this make?

liftM: `git` doesn't store modification time. Whenever somebody new checks out the CLI repository, the mtime of the file is the time when they did the checkout.

`go-bindata` _does_ store modification time (by default). This means every time somebody new checks out the CLI and runs `make`, it'll produce a diff in this generated file.

By forcing `go-bindata` to set the mtime to a specific value, new developers no longer need to deal with this extra noise.

zlav: Fan of this change, makes git status a lot more straightforward as well.

PR #305, opened 2018-10-01
testing (fossa test) use a test server to test fossa test

zlav:
This test ensures that `fossa test` correctly handles all successful API messages. I need to implement the new errors work throughout this code path in order to be able to test failure modes. 
The point of this PR was to create a testing server which could be used throughout the codebase into V2 development. This PR demonstrates how simple it is to create a test server and use it for anything. The most difficult part of this is configuration.

Once better errors are implemented here I am going to return and handle the failure modes and individual unit testing.

I think that a lot of the functions in this package could be extended to create a server test package but I am not going to do that until I build a couple more of these.

### DIFF
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -81,5 +81,6 @@ func testSetup(endpoint, apiKey string) *cli.Context {
 
 	flagSet := &flag.FlagSet{}
 	flagSet.String("endpoint", endpoint, "")
+	flagSet.Int("timeout", 20, "")
 	return cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
 }
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -47,6 +47,7 @@ func TestSuccessfullTest(t *testing.T) {
 	defer ts.Close()
 
 	context := testSetup(ts.URL, "1")
+	test.PollRequestDelay = 2 * time.Second
 	err := test.Run(context)
 	assert.NoError(t, err)
 }
@@ -68,14 +69,12 @@ func testCustomTestServer(t *testing.T, server *mockServer) *httptest.Server {
 }
 
 func testSetup(endpoint, apiKey string) *cli.Context {
-	test.PollRequestDelay = 2 * time.Second
 	apiErr := fossa.SetAPIKey(apiKey)
 	if apiErr != nil {
 		log.Fatalf("Could not set api key: %s", apiErr)
 	}
 
 	flagSet := &flag.FlagSet{}
-	flagSet.Int("timeout", 100, "")
 	flagSet.String("endpoint", endpoint, "")
 	return cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
 }
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -1,0 +1,90 @@
+package test_test
+
+import (
+	"encoding/json"
+	"flag"
+	"math/rand"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/apex/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
+
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/fossas/fossa-cli/cmd/fossa/cmd/test"
+)
+
+var orgID = rand.Intn(1000)
+
+// taskStatus is a struct that imitates the anonymous struct within fossa.Build
+type taskStatus struct {
+	Status string
+}
+
+type mockServer struct {
+	Responses []interface{}
+}
+
+// TestSuccessfullTest tests the `fossa test` test command and ensures that it properly
+// handles all responses.
+func testSuccessfullTest(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	fossa.SetAPIKey("1")
+	testLocator := fossa.Locator{Fetcher: "custom", Project: "tesRun", Revision: "123"}
+	serverTest := mockServer{
+		Responses: []interface{}{
+			fossa.Organization{OrganizationID: 1},
+			fossa.Build{Task: taskStatus{Status: "CREATED"}},
+			fossa.Build{Task: taskStatus{Status: "ASSIGNED"}},
+			fossa.Build{Task: taskStatus{Status: "RUNNING"}},
+			fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}},
+			fossa.Issues{Status: "WAITING"},
+			fossa.Issues{Status: "SCANNED"},
+		},
+	}
+	ts := testCustomTestServer(t, testLocator, &serverTest)
+	defer ts.Close()
+	context := testContext(testLocator, ts.URL, "1")
+
+	test.PollRequestDelay = 2 * time.Second
+	err := test.Run(context)
+	assert.NoError(t, err)
+}
+
+// testCustomTestServer dequeues the first item in server.Responses, marshals it, and responds with it.
+func testCustomTestServer(t *testing.T, locator fossa.Locator, server *mockServer) *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		resp := server.Responses[0]
+		server.Responses = server.Responses[1:]
+		response, err := json.Marshal(resp)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal JSON: %s", err)
+		}
+		_, err = w.Write(response)
+		if err != nil {
+			t.Fatalf("Failed to write response: %s", err)
+		}
+		return
+	}))
+}
+
+func testContext(locator fossa.Locator, endpoint, apiKey string) *cli.Context {
+	flagSet := &flag.FlagSet{}
+	flagSet.Int("timeout", 100, "")
+	rev, err := strconv.Atoi(locator.Revision)
+	if err != nil {
+		log.Warn(err.Error())
+	}
+	flagSet.Int("revision", rev, "")
+	flagSet.String("fetcher", locator.Fetcher, "")
+	flagSet.String("project", locator.Project, "")
+	flagSet.String("endpoint", endpoint, "")
+	return cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+}
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -25,6 +25,12 @@ type mockServer struct {
 	Responses []interface{}
 }
 
+func (server *mockServer) nextResponse() interface{} {
+	resp := server.Responses[0]
+	server.Responses = server.Responses[1:]
+	return resp
+}
+
 // TestSuccessfullTest tests the `fossa test` test command and ensures that it properly
 // handles all responses in the path to a success.
 func TestSuccessfullTest(t *testing.T) {
@@ -55,8 +61,7 @@ func TestSuccessfullTest(t *testing.T) {
 // testCustomTestServer dequeues the first item in server.Responses, marshals it, and responds with it.
 func testCustomTestServer(t *testing.T, server *mockServer) *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		resp := server.Responses[0]
-		server.Responses = server.Responses[1:]
+		resp := server.nextResponse()
 		response, err := json.Marshal(resp)
 		if err != nil {
 			t.Fatalf("Failed to unmarshal JSON: %s", err)
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/fossas/fossa-cli/api/fossa"
 	"github.com/fossas/fossa-cli/cmd/fossa/cmd/test"
+	"github.com/fossas/fossa-cli/errors"
 )
 
 // taskStatus is a struct that imitates the anonymous struct within fossa.Build
@@ -25,13 +26,13 @@ type mockServer struct {
 	Responses []interface{}
 }
 
-func (server *mockServer) nextResponse() interface{} {
+func (server *mockServer) nextResponse() (interface{}, error) {
 	if len(server.Responses) == 0 {
-		return nil
+		return nil, errors.New("no more responses left")
 	}
 	resp := server.Responses[0]
 	server.Responses = server.Responses[1:]
-	return resp
+	return resp, nil
 }
 
 // TestSuccessfullTest tests the `fossa test` test command and ensures that it properly
@@ -64,7 +65,8 @@ func TestSuccessfullTest(t *testing.T) {
 // testCustomTestServer dequeues the first item in server.Responses, marshals it, and responds with it.
 func testCustomTestServer(t *testing.T, server *mockServer) *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		resp := server.nextResponse()
+		resp, err := server.nextResponse()
+		assert.Nil(t, err)
 		response, err := json.Marshal(resp)
 		if err != nil {
 			t.Fatalf("Failed to unmarshal JSON: %s", err)
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"net/http"
 	"net/http/httptest"
-	"strconv"
 	"testing"
 	"time"
 
@@ -33,7 +32,6 @@ func TestSuccessfullTest(t *testing.T) {
 		return
 	}
 
-	testLocator := fossa.Locator{Fetcher: "custom", Project: "tesRun", Revision: "123"}
 	serverTest := mockServer{
 		Responses: []interface{}{
 			fossa.Organization{OrganizationID: 1},
@@ -45,16 +43,16 @@ func TestSuccessfullTest(t *testing.T) {
 			fossa.Issues{Status: "SCANNED"},
 		},
 	}
-	ts := testCustomTestServer(t, testLocator, &serverTest)
+	ts := testCustomTestServer(t, &serverTest)
 	defer ts.Close()
 
-	context := testSetup(testLocator, ts.URL, "1")
+	context := testSetup(ts.URL, "1")
 	err := test.Run(context)
 	assert.NoError(t, err)
 }
 
 // testCustomTestServer dequeues the first item in server.Responses, marshals it, and responds with it.
-func testCustomTestServer(t *testing.T, locator fossa.Locator, server *mockServer) *httptest.Server {
+func testCustomTestServer(t *testing.T, server *mockServer) *httptest.Server {
 	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		resp := server.Responses[0]
 		server.Responses = server.Responses[1:]
@@ -69,7 +67,7 @@ func testCustomTestServer(t *testing.T, locator fossa.Locator, server *mockServe
 	}))
 }
 
-func testSetup(locator fossa.Locator, endpoint, apiKey string) *cli.Context {
+func testSetup(endpoint, apiKey string) *cli.Context {
 	test.PollRequestDelay = 2 * time.Second
 	apiErr := fossa.SetAPIKey(apiKey)
 	if apiErr != nil {
@@ -78,13 +76,6 @@ func testSetup(locator fossa.Locator, endpoint, apiKey string) *cli.Context {
 
 	flagSet := &flag.FlagSet{}
 	flagSet.Int("timeout", 100, "")
-	rev, err := strconv.Atoi(locator.Revision)
-	if err != nil {
-		log.Warn(err.Error())
-	}
-	flagSet.Int("revision", rev, "")
-	flagSet.String("fetcher", locator.Fetcher, "")
-	flagSet.String("project", locator.Project, "")
 	flagSet.String("endpoint", endpoint, "")
 	return cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
 }
--- a/cmd/fossa/cmd/test/test_test.go
+++ b/cmd/fossa/cmd/test/test_test.go
@@ -26,6 +26,9 @@ type mockServer struct {
 }
 
 func (server *mockServer) nextResponse() interface{} {
+	if len(server.Responses) == 0 {
+		return nil
+	}
 	resp := server.Responses[0]
 	server.Responses = server.Responses[1:]
 	return resp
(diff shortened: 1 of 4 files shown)


cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestRunCustomFetcherCustomProject(t *testing.T) {
+	c := make(chan string)
+	locator := "custom+" + strconv.Itoa(orgID) + "%2FtestRun$1000"
+	ts := testServer(locator, c)
+
+	flagSet := testFlags("custom", "testRun", ts.URL, 1000)
+	context := cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+
+	go Run(context)
```
golangcibot: Error return value of `Run` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestRunCustomFetcherGitSSHProject(t *testing.T) {
+	c := make(chan string)
+	locator := "custom+" + strconv.Itoa(orgID) + "%2Fgit@github.com:fossa%2Ffossa-cli.git$1000"
+	ts := testServer(locator, c)
+
+	flagSet := testFlags("custom", "git@github.com:fossa/fossa-cli.git", ts.URL, 1000)
+	context := cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+
+	go Run(context)
```
golangcibot: Error return value of `Run` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+
+func TestRunCustomFetcherGitHTTPProject(t *testing.T) {
+	c := make(chan string)
+	locator := "custom+" + strconv.Itoa(orgID) + "%2Fgithub.com%2Ffossa%2Ffossa-cli$1000"
+	ts := testServer(locator, c)
+
+	flagSet := testFlags("custom", "http://github.com/fossa/fossa-cli.git", ts.URL, 1000)
+	context := cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+
+	go Run(context)
```
golangcibot: Error return value of `Run` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
+package test
```
liftM: I would call this file `integration_test.go`.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"encoding/json"
+	"flag"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"strconv"
+	"testing"
+
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/stretchr/testify/assert"
```
liftM: [nit] Import groups should be: standard library, third party, first party.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/urfave/cli"
+)
+
+const orgID = 3
```
liftM: Let's randomise this.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+func testServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// Path for OrganizationID
+		if r.URL.EscapedPath() == "/api/cli/organization" {
+			newResp := fossa.Organization{OrganizationID: orgID}
+			request, _ := json.Marshal(newResp)
+			fmt.Fprintf(w, string(request))
```
liftM: You can just `w.Write`, here and below.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	ts.Close()
+}
+
+func testServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
+		// Path for OrganizationID
+		if r.URL.EscapedPath() == "/api/cli/organization" {
```
liftM: This would probably be better as a `switch` on `r.URL.EscapedPath()`.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	if msg != "SUCCESS" {
+		assert.Equal(t, locator, msg)
+	}
+
+	ts.Close()
+}
+
+func testServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+
```
liftM: [nit] Whitespace.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	go Run(context)
+	msg := <-c
+	if msg != "SUCCESS" {
+		assert.Equal(t, locator, msg)
+	}
+
+	ts.Close()
+}
+
+func TestRunCustomFetcherGitSSHProject(t *testing.T) {
```
liftM: This looks like a good fit for table-driven tests. See https://github.com/golang/go/wiki/TableDrivenTests.

zlav:
@microsoftly I checked for this and was able to run them in parallel without any problems. I'd be interested in seeing your case where `config.SetContext()` causes issues. Additionaly the tests themeselves only take 0.037s to run, but I understand the issue is with running the whole suite in parallel.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,169 @@
...
+	locator := "custom+" + strconv.Itoa(orgID) + "%2FtestRun$1000"
+	ts := testServer(locator, c)
+
+	flagSet := testFlags("custom", "testRun", ts.URL, 1000)
+	context := cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+
+	go Run(context)
+	msg := <-c
+	if msg != "SUCCESS" {
+		assert.Equal(t, locator, msg)
```
liftM: Why `assert.Equal` instead of `t.Fail`?

zlav: This is one area I had the most difficulty making clear. The output from `t.Fail` running `go test -v` is 
```
=== RUN   TestRunCustomFetcherCustomProject
⣾ Waiting for analysis to complete... --- FAIL: TestRunCustomFetcherCustomProject (0.00s)
```

Using a goroutine and `assert.Equal` I can improve the feedback to something that is much more descriptive of the actual error.
```
=== RUN   TestRunCustomFetcherCustomProject
⣾ Waiting for analysis to complete... --- FAIL: TestRunCustomFetcherCustomProject (0.00s)
    test_test.go:35:
        	Error Trace:	test_test.go:35
        	Error:      	Not equal:
        	            	expected: "ctom+3%2FtestRun$1000"
        	            	actual  : "/api/cli/custom+3%2FtestRun$1000/latest_build"

        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-ctom+3%2FtestRun$1000
        	            	+/api/cli/custom+3%2FtestRun$1000/latest_build
        	Test:       	TestRunCustomFetcherCustomProject
```

cmd/fossa/cmd/test/test_test.go
zlav: Thanks for the tip

cmd/fossa/cmd/test/test_test.go
liftM: I think there are two good options here:

- Use `log.Debug` (or `t.Log`) to describe the locator. This output will show up on `go test -v`.
- Construct the actual expected locator, and then compare against that in `assert.Equal`. That way your `assert.Equal` is actually semantically an equality assertion.

My personal preference is for the latter. I think that provides good feedback while remaining clear.

cmd/fossa/cmd/test/test_test.go
zlav: You're right, I relied on `go fmt`. I wonder why it ordered them wrong in this case.

cmd/fossa/cmd/test/test_test.go
zlav: I agree with this, I've been thinking of better ways to clarify this. The problem with comparing only the locators is that if the test fails it assumes the rest of the path is correct which might not be the case. I'm comfortable going down the route of asserting only against the locator in this case however because that is what this test is varying between cases.

cmd/fossa/cmd/test/test_test.go
zlav: I'm not convinced that this package is a full integration test because it is not connecting with the app.fossa.io API. This makes me feel like this is more in line with an end to end test and I'm hesitant to call it an integration test.

Regardless the package does need to be changed, thank you for pointing this out.

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+func TestTestRunLocators(t *testing.T) {
+	for _, locatorType := range locatorTypes {
+		c := make(chan string)
+		ts := testServer(locatorType.locator, c)
+		defer ts.Close()
+
+		flagSet := testFlags(locatorType.fetcher, locatorType.project, ts.URL, 1000)
+		context := cli.NewContext(&cli.App{}, flagSet, &cli.Context{})
+
+		go test.Run(context)
```
golangcibot: Error return value of `test.Run` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+}
+
+func testServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.EscapedPath() {
+		case "/api/cli/organization":
+			newResp := fossa.Organization{OrganizationID: orgID}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+			return
+		case "/api/cli/" + locator + "/latest_build":
+			newResp := fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+			return
+		case "/api/cli/" + locator + "/issues":
+			newResp := fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,142 @@
...
+	return nil
+}
+
+func testCustomServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.EscapedPath() {
+		case "/api/cli/organization":
+			newResp := fossa.Organization{OrganizationID: orgID}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,142 @@
...
+			return
+		case "/api/cli/" + locator + "/latest_build":
+			newResp := fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,142 @@
...
+			return
+		case "/api/cli/" + locator + "/issues":
+			newResp := fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}}
+			request, _ := json.Marshal(newResp)
+			w.Write(request)
```
golangcibot: Error return value of `w.Write` is not checked

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+	{"custom", "http://github.com/fossas/fossa-cli.git", "custom+" + strconv.Itoa(orgID) + "%2Fgithub.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "git@github.com:fossas/fossa-cli.git", "git+github.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "http://github.com/fossas/fossa-cli.git", "git+github.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "testRun", "git+testRun$1000"},
+}
+
+// This function tests how the cli's test command constructs a locator and sends a request to the desired endpoint.
+// Currently this function implements goroutines and channels to prevent test.Run() from logging fatal and killing
+// the process. In the future, once errors are handled more gracefully the test server can be broken out
+// into its own TestServer package for other commands to access.
```
microsoftly: I would add something along the lines of state will be asserted against the TestServer (e.g. what urls were called, etc ...) to avoid needing the channel to leak state

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+func testCustomServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.EscapedPath() {
+		case "/api/cli/organization":
+			newResp := fossa.Organization{OrganizationID: orgID}
+			request, _ := json.Marshal(newResp)
+			_, err := w.Write(request)
+			log.Debugf("error writing message: %s\n", err)
```
microsoftly: will this print every time or only when `err != nil`?

cmd/fossa/cmd/test/test_test.go
liftM: `go fmt` will do grouping, but isn't smart enough to understand group ordering.

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	if fetcher == "git" {
+		return testGitServer(locator, c)
+	}
+
+	return nil
+}
+
+func testCustomServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
microsoftly: [nit] you can just return on this line

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/apex/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
+
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/fossas/fossa-cli/cmd/fossa/cmd/test"
+)
+
+var orgID = rand.Intn(1000)
```
liftM: [nit] Let's bump this to `2^32-1`.

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	return nil
+}
+
+func testCustomServer(locator string, c chan string) *httptest.Server {
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.URL.EscapedPath() {
+		case "/api/cli/organization":
+			newResp := fossa.Organization{OrganizationID: orgID}
+			request, _ := json.Marshal(newResp)
```
liftM: [nit] Here and below, you should check and panic on error.

liftM: [nit] Here and below, the marshalled data isn't actually a _request_. I'd call it a payload.

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+			_, err := w.Write(request)
+			log.Debugf("error writing message: %s\n", err)
+			return
+		case "/api/cli/" + locator + "/latest_build":
+			newResp := fossa.Build{Task: taskStatus{Status: "SUCCEEDED"}}
```
liftM: Pull these 4 lines out into a function. It looks like they're repeated a lot.

cmd/fossa/cmd/test/e2e_test.go
```diff
@@ -0,0 +1,148 @@
...
+	{"custom", "git@github.com:fossas/fossa-cli.git", "custom+" + strconv.Itoa(orgID) + "%2Fgit@github.com:fossas%2Ffossa-cli.git$1000"},
+	{"custom", "http://github.com/fossas/fossa-cli.git", "custom+" + strconv.Itoa(orgID) + "%2Fgithub.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "git@github.com:fossas/fossa-cli.git", "git+github.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "http://github.com/fossas/fossa-cli.git", "git+github.com%2Ffossas%2Ffossa-cli$1000"},
+	{"git", "testRun", "git+testRun$1000"},
+}
+
+// This function tests how the cli's test command constructs a locator and sends a request to the desired endpoint.
+// Currently this function implements goroutines and channels to prevent test.Run() from logging fatal and killing
+// the process. In the future, once errors are handled more gracefully the test server can be broken out
```
liftM: This comment confuses me. I thought we were using channels to concurrently run the CLI and the test server.

cmd/fossa/cmd/test/e2e_test.go
zlav: It actually doesn't print every time on my local machine but I do think this should be changed to watch for that. Thank you for pointing it out.

cmd/fossa/cmd/test/e2e_test.go
microsoftly: @liftM why is making this random good or bad? I'd prefer having static state (see my object mother example above).

microsoftly: resolved in person

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// TestSuccessfullTest tests the `fossa test` test command and ensures that it properly
+// handles all responses.
+func testSuccessfullTest(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+
+	fossa.SetAPIKey("1")
```
golangcibot: Error return value of `fossa.SetAPIKey` is not checked (from `errcheck`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/apex/log"
+	"github.com/stretchr/testify/assert"
+	"github.com/urfave/cli"
+
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/fossas/fossa-cli/cmd/fossa/cmd/test"
+)
+
+var orgID = rand.Intn(1000)
```
golangcibot: `orgID` is unused (from `varcheck`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	Status string
+}
+
+type mockServer struct {
+	Responses []interface{}
+}
+
+// TestSuccessfullTest tests the `fossa test` test command and ensures that it properly
+// handles all responses.
+func testSuccessfullTest(t *testing.T) {
```
golangcibot: `testSuccessfullTest` is unused (from `deadcode`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/urfave/cli"
+
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/fossas/fossa-cli/cmd/fossa/cmd/test"
+)
+
+var orgID = rand.Intn(1000)
+
+// taskStatus is a struct that imitates the anonymous struct within fossa.Build
+type taskStatus struct {
```
golangcibot: U1000: type `taskStatus` is unused (from `unused`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	Status string
+}
+
+type mockServer struct {
```
golangcibot: U1000: type `mockServer` is unused (from `unused`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	defer ts.Close()
+	context := testContext(testLocator, ts.URL, "1")
+
+	test.PollRequestDelay = 2 * time.Second
+	err := test.Run(context)
+	assert.NoError(t, err)
+}
+
+// testCustomTestServer dequeues the first item in server.Responses, marshals it, and responds with it.
+func testCustomTestServer(t *testing.T, locator fossa.Locator, server *mockServer) *httptest.Server {
```
golangcibot: U1000: func `testCustomTestServer` is unused (from `unused`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+		server.Responses = server.Responses[1:]
+		response, err := json.Marshal(resp)
+		if err != nil {
+			t.Fatalf("Failed to unmarshal JSON: %s", err)
+		}
+		_, err = w.Write(response)
+		if err != nil {
+			t.Fatalf("Failed to write response: %s", err)
+		}
+		return
```
golangcibot: S1023: redundant `return` statement (from `gosimple`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}))
+}
+
+func testContext(locator fossa.Locator, endpoint, apiKey string) *cli.Context {
```
golangcibot: U1000: func `testContext` is unused (from `unused`)

cmd/fossa/cmd/test/test_test.go
```diff
@@ -0,0 +1,86 @@
...
+type taskStatus struct {
+	Status string
+}
+
+type mockServer struct {
+	Responses []interface{}
+}
+
+func (server *mockServer) nextResponse() interface{} {
+	resp := server.Responses[0]
```
cnr: This is going to panic if the client tries to send more requests than we have responses. Let's make this failure mode more explicit somehow

cnr: also: should we care about what's in the requests?

zlav: I started down the path of testing what is in the requests and found that I was mostly testing `url.PathEscape(locator.String())` which requires me to set a lot more information. I do see validity in making another field to store requests, I just didn't think it was necessary here.

PR #306, opened 2018-10-01
feat(circle.yml) aggregate coverage for multiple reports within a single PR

microsoftly:


microsoftly:
@zlav that is correct. That being said, `goveralls` does not currently have support for parallel jobs. I'm closing this for now until we build the needed support

PR #307, opened 2018-10-01
test(python analyzer) add native python analyzer integration tests

microsoftly:


analyzers/python/integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/fossas/fossa-cli/analyzers"
+	"github.com/fossas/fossa-cli/exec"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/testing/fixtures"
+	"github.com/stretchr/testify/assert"
+)
+
+var pythonAnalyzerFixtureDir = filepath.Join(fixtures.Directory(), "python", "analyzer")
```
liftM: Is it necessary to namespace the fixtures? I don't think we currently do, and we have yet to run in to a name collision.

analyzers/python/integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+			module := module.Module{
+				Dir:         projDir,
+				Type:        pkg.Python,
+				Name:        proj.Name,
+				Options:     map[string]interface{}{"strategy": "deptree"},
+				BuildTarget: projDir,
+			}
+
+			if proj.Name == "sentry" {
+				module.Options["strategy"] = "deptree"
```
liftM: It looks like we're using `deptree` anyway.

analyzers/python/integration_test.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+}
+
+var projects = []fixtures.Project{
+	fixtures.Project{
+		Name:   "vibora",
+		URL:    "https://github.com/vibora-io/vibora",
+		Commit: "51d6691a430bbb6ac57df54dc7d8276f62953bb2",
+	},
+	// ansible is failing checkout to a commit with the error: worktree contains unstaged changes, but I cannot figure out why
```
liftM: That's bizarre. I'm alright with this merging anyway, but I'm also down to pair on this.

analyzers/python/integration_test.go
microsoftly: The idea here is that we can cache on a particular language/ecosystem instead of whether or not there has been any integration test run. Additionally, it's a nice organization when inspecting issues that pop up

analyzers/python/integration_test.go
```diff
@@ -0,0 +1,144 @@
...
+		Commit: "20187b45ce4d32146d2a2d493c315924c1cc8287",
+	},
+	fixtures.Project{
+		Name:   "sshuttle",
+		URL:    "https://github.com/sshuttle/sshuttle",
+		Commit: "6dc368bde8128cd27ad80d48772420ff68a92c8f",
+	},
+}
+
+func projectInitializer(proj fixtures.Project, projectDir string) error {
```
golangcibot: `projectInitializer` is unused

PR #309, opened 2018-10-03
build: Fix Makefile dependencies in release process

liftM:


liftM:
Merging TBR to cut a release.

PR #310, opened 2018-10-03
refactor(integration tests) abstract/simplify project initialization post cloning

microsoftly:
Most if not all of the integration test fixtures are going to need additional steps taken after `git clone` that can be optimized by running asynchronously. This abstractions effectively hides the async structure and creates an painless way of reporting errors during initialization

### DIFF
--- a/analyzers/nodejs/integration_test.go
+++ b/analyzers/nodejs/integration_test.go
@@ -27,7 +27,7 @@ func TestMain(m *testing.M) {
 		return
 	}
 
-	fixtures.InitializeProjects(nodeAnalyzerFixtureDir, projects, projectInitializer)
+	fixtures.Initialize(nodeAnalyzerFixtureDir, projects, projectInitializer)
 
 	exitCode := m.Run()
 	defer os.Exit(exitCode)
@@ -100,7 +100,7 @@ func projectInitializer(proj fixtures.Project, projectDir string) error {
 	}
 
 	if nodeModulesExist {
-		log.Debug("node modules already exists for " + proj.Name + "skipping initialization")
+		log.Debug("node_modules already exists for " + proj.Name + "skipping initialization")
 		return nil
 	}
 
@@ -114,9 +114,9 @@ func projectInitializer(proj fixtures.Project, projectDir string) error {
 	if err != nil {
 		log.Error(errOut)
 		log.Error("failed to run npm install on " + proj.Name)
+		return err
 	}
 
-	// save time on local
 	ymlAlreadyExists, err := files.Exists(filepath.Join(projectDir, ".fossa.yml"))
 	if err != nil {
 		return err
@@ -125,7 +125,6 @@ func projectInitializer(proj fixtures.Project, projectDir string) error {
 		return nil
 	}
 
-	// any key will work to prevent the "NEED KEY" error message
 	stdout, stderr, err := runfossa.Init(projectDir)
 	if err != nil {
 		log.Error("failed to run fossa init on " + proj.Name)
--- a/analyzers/nodejs/integration_test.go
+++ b/analyzers/nodejs/integration_test.go
@@ -4,7 +4,6 @@ import (
 	"flag"
 	"os"
 	"path/filepath"
-	"sync"
 	"testing"
 
 	"github.com/apex/log"
@@ -28,14 +27,24 @@ func TestMain(m *testing.M) {
 		return
 	}
 
+	log.SetLevel(log.DebugLevel)
+
 	err := fixtures.Clone(nodeAnalyzerFixtureDir, projects)
 	if err != nil {
 		panic(err)
 	}
 
-	err = initializeProjects(nodeAnalyzerFixtureDir)
-	if err != nil {
-		panic(err)
+	initializationErrors := fixtures.InitializeProjects(nodeAnalyzerFixtureDir, projects, projectInitializer)
+	initializationErrorFound := false
+	for projName, err := range initializationErrors {
+		if err != nil {
+			initializationErrorFound = true
+			log.Error("Error initializing node project" + projName + " " + err.Error())
+		}
+	}
+
+	if initializationErrorFound {
+		os.Exit(1)
 	}
 
 	exitCode := m.Run()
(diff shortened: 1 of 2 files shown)


analyzers/nodejs/integration_test.go
```diff
@@ -102,59 +109,46 @@ func assertProjectFixtureExists(t *testing.T, name string) {
...
+	if err != nil {
+		return err
+	}
 
-			if nodeModulesExist {
-				log.Debug("node modules already exists for " + proj.Name + "skipping initialization")
-				return
-			}
+	if nodeModulesExist {
+		log.Debug("node modules already exists for " + proj.Name + "skipping initialization")
```
liftM: I'd be more comfortable with this if we validated `node_modules` in some way, but I don't think that's worth implementing. A test that passes due to a bad cache would probably fail in CI anyway (unless the CI cache also got polluted -- how likely is that?).

testing/fixtures/fixtures.go
```diff
@@ -73,3 +73,36 @@ func Clone(baseDir string, projects []Project) error {
...
+	}
+
+	for _, project := range projects {
+		go func(proj Project) {
+			defer waitGroup.Done()
+			err := projectInitializer(proj, filepath.Join(testDir, proj.Name))
+
+			// separate the initialization step from locking to decrease time locking out other writers
+			threadsafeErrorsByProject.Lock()
+			threadsafeErrorsByProject.errorsByProject[proj.Name] = err
```
liftM: Why not panic immediately like we do in `Clone`? I think we should do one or the other in both cases.

Personally, my preference is for an immediate panic. Yes, there is value in seeing all the errors, but I don't think the added value is worth the extra code.

testing/fixtures/fixtures.go
```diff
@@ -73,3 +73,36 @@ func Clone(baseDir string, projects []Project) error {
 
 	return nil
 }
+
+// ProjectInitializerFunction defines how a single project should be initialized *after* it has already been cloned
+type ProjectInitializerFunction func(proj Project, projectDir string) error
+
+// InitializeProjects initializes provided projects in parallel using the provided initialization function. Errors are returned as a map of the Project.Name to the error thrown
+func InitializeProjects(testDir string, projects []Project, projectInitializer ProjectInitializerFunction) map[string]error {
```
liftM: Why not fold this into `Clone` and rename that to `Initialize`? For fixtures that should _only_ be cloned, we can pass a no-op.

Which of these choices is more ergonomic depends on the kinds of fixtures you expect to work with. Will the majority of fixtures need initialization? (I think so.)

analyzers/nodejs/integration_test.go
```diff
@@ -33,9 +32,17 @@ func TestMain(m *testing.M) {
...
+	initializationErrorFound := false
+	for projName, err := range initializationErrors {
+		if err != nil {
+			initializationErrorFound = true
+			log.Error("Error initializing node project" + projName + " " + err.Error())
+		}
+	}
+
+	if initializationErrorFound {
+		os.Exit(1)
```
liftM: Why not panic?

analyzers/nodejs/integration_test.go
```diff
@@ -33,9 +32,17 @@ func TestMain(m *testing.M) {
 		panic(err)
 	}
 
-	err = initializeProjects(nodeAnalyzerFixtureDir)
-	if err != nil {
-		panic(err)
+	initializationErrors := fixtures.InitializeProjects(nodeAnalyzerFixtureDir, projects, projectInitializer)
+	initializationErrorFound := false
+	for projName, err := range initializationErrors {
```
liftM: This multi-error handling seems a bit heavy for the task at hand. See my comments below.

analyzers/nodejs/integration_test.go
microsoftly: It makes no difference

testing/fixtures/fixtures.go
microsoftly: for cases like `npm` where there can be some failures that we may want to handwave away, it will be useful to inspect the results and change what happens depending on it. The current implementation ignores npm errors, as an example as to where this can be used

analyzers/nodejs/integration_test.go
microsoftly: I don't think it's likely CI will get polluted. We update the cache every time we add a new integration test project fixture (and whenever a particular project initialization needs an update).

testing/fixtures/fixtures.go
microsoftly: Also, if cloning does not work, we have bigger issues that will likely need to be addressed ASAP, so panicking there seems more fitting to terminate the program as quickly as possible

microsoftly: looking back, I think you're right

analyzers/nodejs/integration_test.go
liftM: Panic has better-defined semantics and produces a nicer exit trace. I'd prefer panicking.

analyzers/nodejs/integration_test.go
```diff
@@ -102,59 +93,46 @@ func assertProjectFixtureExists(t *testing.T, name string) {
...
+	if err != nil {
+		return err
+	}
 
-			if nodeModulesExist {
-				log.Debug("node modules already exists for " + proj.Name + "skipping initialization")
-				return
-			}
+	if nodeModulesExist {
+		log.Debug("node modules already exists for " + proj.Name + "skipping initialization")
```
liftM: [nit] node modules -> `node_modules`.

analyzers/nodejs/integration_test.go
```diff
@@ -102,59 +93,46 @@ func assertProjectFixtureExists(t *testing.T, name string) {
...
 
-			// save time on local
-			ymlAlreadyExists, err := files.Exists(filepath.Join(projectDir, ".fossa.yml"))
-			if err != nil {
-				panic(err)
-			}
-			if ymlAlreadyExists {
-				return
-			}
+	// save time on local
```
liftM: [nit] Comments should be complete sentences.

analyzers/nodejs/integration_test.go
```diff
@@ -102,59 +93,46 @@ func assertProjectFixtureExists(t *testing.T, name string) {
...
-			stdout, stderr, err := runfossa.Init(projectDir)
-			if err != nil {
-				log.Error("failed to run fossa init on " + proj.Name)
-				log.Error(stdout)
-				log.Error(stderr)
-				log.Error(err.Error())
-				panic(err)
-			}
-		}(project)
+	// any key will work to prevent the "NEED KEY" error message
```
liftM: [nit] Comments should be complete sentences.

testing/fixtures/fixtures.go
```diff
@@ -24,20 +24,23 @@ func Directory() string {
 	return filepath.Join(os.TempDir(), "fossa-cli-fixtures")
 }
 
-// Clone executes git clone in target directory and checksout the provided commit, or master if left ""
-func Clone(baseDir string, projects []Project) error {
+// ProjectInitializerFunction defines how a single project should be initialized *after* it has already been cloned
+type ProjectInitializerFunction func(proj Project, projectDir string) error
+
+// InitializeProjects executes git clone in target directory and checksout the provided commit, then runts the initializerFn. This is done asynchronously for each provided project
+func InitializeProjects(baseDir string, projects []Project, initializerFn ProjectInitializerFunction) {
```
liftM: [nit] I'd call this `Initialize` (`fixtures.InitializeProjects(projects)` isn't much more informative than `fixtures.Initialize(projects)`).

testing/fixtures/fixtures.go
```diff
@@ -24,20 +24,23 @@ func Directory() string {
 	return filepath.Join(os.TempDir(), "fossa-cli-fixtures")
 }
 
-// Clone executes git clone in target directory and checksout the provided commit, or master if left ""
-func Clone(baseDir string, projects []Project) error {
+// ProjectInitializerFunction defines how a single project should be initialized *after* it has already been cloned
+type ProjectInitializerFunction func(proj Project, projectDir string) error
+
+// InitializeProjects executes git clone in target directory and checksout the provided commit, then runts the initializerFn. This is done asynchronously for each provided project
```
liftM: [nit] `s/checksout/checks out`.

liftM: [nit] `s/runts/runs`.

liftM: [nit] Period on last sentence.

analyzers/nodejs/integration_test.go
```diff
@@ -102,59 +93,46 @@ func assertProjectFixtureExists(t *testing.T, name string) {
...
+	_, errOut, err := exec.Run(exec.Cmd{
+		Name:    "npm",
+		Argv:    []string{"install", "--production"},
+		Dir:     projectDir,
+		WithEnv: proj.Env,
+		Command: "npm",
+	})
+	if err != nil {
+		log.Error(errOut)
+		log.Error("failed to run npm install on " + proj.Name)
```
liftM: We should return this error too.

PR #314, opened 2018-10-04
feat(nodejs dev deps) filter nodejs dev deps when using npm ls

microsoftly:
Theres a lot of test/fixture work here that effectively amounts to being unnecessary for this fix, but I created it proactively to prove correctness.

The problem is unit tests did not pick up the proper state because they mocked out `npm ls` and shared a single file that was the output of `npm ls`. To make each fixture more unique, I saved the output of `npm ls` to each nodejs/testdata/chai fixture where it would be appropriate. 

I had run it with `npm ls --json --production`, which, as it turns out resulted in the same output for each case, including dev deps. This led me to the solution that found the bug: we were leaving out `--production` in our call.

That being said, the test refactors are still functionally useful as they keep each fixture separate, and also add cases against which tests are run (e.g. multiple fixtures per test type). For that reason, I have kept them in this PR.

Also captured is the cleanup of unneeded files in the fixtures (e.g. `LICENSE` and `CODEOWNERS`


analyzers/nodejs/testdata/chai/dev-deps/package.json
```diff
@@ -0,0 +1,12 @@
+{
+  "name": "testdata",
+  "version": "1.0.0",
+  "main": "index.js",
+  "license": "MIT",
+  "dependencies": {
+    "chai": "4.1.2"
+  },
+  "devDependencies": {
+    "is-odd": "^3.0.1"
```
liftM: See my same comments on the Yarn parsing re: fixtures.

analyzers/nodejs/nodejs_test.go
```diff
@@ -94,8 +94,11 @@ var chaiDirectDep = pkg.Import{
 }
 
 func TestAnalyzeWithNpmLs(t *testing.T) {
-	buildTarget := filepath.Join("testdata", "chai", "installed")
-
+	testAnalyzeWithNpmLs(t, filepath.Join("testdata", "chai", "installed"))
```
liftM: Let's just make this a table-driven `for` loop.

analyzers/nodejs/nodejs_test.go
```diff
@@ -118,8 +121,12 @@ func TestAnalyzeWithNpmLs(t *testing.T) {
 }
 
 func TestUsingNodeModuleFallback(t *testing.T) {
-	buildTarget := filepath.Join("testdata", "chai", "installed")
+	testUsingNodeModuleFallback(t, filepath.Join("testdata", "chai", "installed"))
```
liftM: Same here as above.

microsoftly:
Will provide new fixture strategy in a follow up PR

Issue #132, opened 2018-05-07
Support Pipfile


anseljh:
[Pipenv](https://docs.pipenv.org/) is the latest and greatest in Python packaging, and it uses a new file called `Pipfile` instead of `requirements.txt`.

More on `Pipfile`: https://github.com/pypa/pipfile

ilikebits:
We're planning a refactor of how builders work to better support multiple strategies per ecosystem, and Pipfile support will probably come in afterwards. That said, I'd be happy to review a PR for this and merge it in afterwards.

zlav:
Support for Pipenv has been added, however fossa will not automatically detect a project using Pipenv. In order to analyze a Pipenv project, set the corresponding module type in your fossa config file, `.fossa.yml`, to Pipenv:
```analyze:
  modules:
  - name: fossa
    path: ./cmd/fossa
    type: pipenv
```

Issue [#351] is dedicated to improving the detection of Pipenv projects.

PR #315, opened 2018-10-04
feat(pipenv) add support for pipenv projects

zlav:
[#132]

Create support for users who manage Python dependencies using pipenv.

This pull request does not include proper fallbacks to ensure that the correct buildtool is identified, the operator must specify pipfile within their yaml file. requirements.txt is currently the default method of analysis but this will be changed in a followup PR. Most of the design decisions I made mirror the decisions made in the golang analyzer and buildtool systems to prepare for the fallback strategy. I chose to internalize `FromRequirements` used in requirments.txt analysis within the pipenv `FromFile` function because that is how pip will look in a follow-up PR.


buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,65 @@
...
+	var imports []pkg.Import
+
+	err := files.ReadJSON(&defaultPackages, filename)
+	if err != nil {
+		return imports, err
+	}
+
+	var reqs []Requirement
+	operators := []string{"==", "<=", ">=", ">", "<", "!="}
+	for a, b := range defaultPackages.Default {
```
microsoftly: Can `a`, `b` be named to be more informative?

zlav:
@microsoftly Absolutely, thanks for the lead on the NPM tests.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,65 @@
...
+		b.Name = a
+		for _, op := range operators {
+			sections := strings.Split(b.Revision, op)
```
microsoftly: Is this guaranteed to give a resolved version? It looks like we're just returning the version after the version is defined in the version specifier.

e.g., what happens if the value is "!==3.2.1"?

analyzers/python/python.go
```diff
@@ -154,6 +155,15 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
 			Transitive: fromImports(imports),
 		}, nil
+	case "pipenv":
+		imports, err := pipenv.FromFile("Pipfile.Lock")
+		if err != nil {
+			return graph.Deps{}, err
+		}
+		return graph.Deps{
+			Direct:     imports,
+			Transitive: fromImports(imports),
```
liftM: This can't be right. We use `fromImports` for flat dependency lists, but `pipenv` has a meaningful notion of transitive dependencies.

See `pipenv graph` (in fact, you could probably parse this output as a strategy -- it looks it there is `pipenv graph --json` and `pipenv graph --json-tree`).

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,65 @@
...
+	var defaultPackages DefaultPackages
+	var imports []pkg.Import
+
+	err := files.ReadJSON(&defaultPackages, filename)
+	if err != nil {
+		return imports, err
+	}
+
+	var reqs []Requirement
+	operators := []string{"==", "<=", ">=", ">", "<", "!="}
```
liftM: Note that there are many more operators than this. See [PEP 440](https://www.python.org/dev/peps/pep-0440/) for the full specification.

That said, is the operator ever anything other than `==`? A quick survey of `Pipfile.lock`s across GitHub says no. Can you check the Pipenv manual, spec, or source code to verify?

buildtools/pipenv/pipenv.go
liftM: :+1: Probably something like `name, pkg`.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,65 @@
...
+func (r Requirement) String() string {
+	return r.Name + r.Operator + r.Revision
+}
+
+// FromFile reads from a Pipfile and returns an array of []pkg.Import.
+func FromFile(filename string) ([]pkg.Import, error) {
+	var defaultPackages DefaultPackages
+	var imports []pkg.Import
+
+	err := files.ReadJSON(&defaultPackages, filename)
```
liftM: The comment says this reads a `Pipfile`, but `Pipfile`s are not JSON. Did you mean `Pipfile.lock`?

buildtools/pipenv/pipenv.go
liftM: Yes, if this is `Pipfile.lock`.

buildtools/pipenv/pipenv.go
zlav: Good idea, I think this ties into your point about better understanding the build tool. I haven't seen an example where the operator isn't `==` but I can imagine it exists. That is the main reason I have `==` first is the matching array.

analyzers/python/python.go
zlav: I will take a look more at how pipenv manages dependencies, however, I had no problem extracting the correct information using `fromImports`. I could be wrong about this but I will let you know what I find out.

zlav: Pipfile.lock is a flat dependency list and so thats why I chose this approach to begin with. You're absolutely right though, I wasn't prioritizing transitive dependencies and taking advantage of how fossa core displays them. `pipenv graph` looks like a better solutuion and will give us better information on dependencies.

The only downside with `pipenv graph` is that we can't get dev dependencies. I wasn't intending on implementing in the first place, but I do see a how this could be useful at some point in the future.

buildtools/pipenv/pipenv.go
zlav: My understanding after researching this is that Pipefile.lock will never have anything but `==`. Pipfile allows different operators but upon running `pipenv lock`, Pipfile.lock is created using the currently installed packages.

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,127 @@
...
+func TestNoFile(t *testing.T) {
+	testEnv := newMockPipenv([]byte{}, errors.New("test error"))
+
+	deps, err := testEnv.Deps()
+	assert.Zero(t, deps)
+	assert.EqualError(t, err, "test error")
+}
+
+func TestBadFile(t *testing.T) {
+	file, err := ioutil.ReadFile("bad.json")
```
golangcibot: ineffectual assignment to `err`

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+func TestBadFile(t *testing.T) {
+	file, err := ioutil.ReadFile("testdata/get-deps/bad.json")
+	assert.NoError(t, err)
+
+	testEnv := newMockPipenv(file, nil)
+
+	deps, err := testEnv.Deps()
+	assert.Zero(t, deps)
+	assert.EqualError(t, err, "could not unmarshall JSON into dependency list: json: cannot unmarshal string into Go value of type []pipenv.dependency")
```
liftM: I'm not sure you need to test this, since this is really a test of `json.Unmarshal`. The only exception is if you plan on returning a _specific_ error in the case of bad files. That might be useful for providing the user with useful debug information, but is irrelevant in this case because we're loading a test fixture into a mock tool, and the goodness of this file is irrelevant.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// Pipenv enables python.go to use a custom Deps() function for testing.
+type Pipenv interface {
+	Deps() (graph.Deps, error)
+}
+
+// Executable allows a custom DataAccess function to be used by Deps().
+type Executable interface {
```
liftM: This is overkill. I bet you'll find that you lose no flexibility or functionality if you remove this interface.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,156 @@
...
+	DataAccess(string) (string, error)
+}
+
+// SystemPipenv implements Pipenv.
+type SystemPipenv struct {
```
liftM: I'm still unsure about naming implementations of tools as `SystemFoo`. I think I'd prefer `FooCmd`.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,156 @@
...
+func getDependencies(graphJSONFile string) ([]dependency, error) {
+	var depList []dependency
+	err := json.Unmarshal([]byte(graphJSONFile), &depList)
+	if err != nil {
+		return nil, errors.Wrap(err, "could not unmarshall JSON into dependency list")
+	}
+	return depList, nil
+}
+
+// DataAccess returns the output from `pipenv graph --json-tree`
```
liftM: [nit] Punctuation.

buildtools/pipenv/pipenv.go
microsoftly: This is what we were talking about earlier. This is the mockable dep that manages access to the executable build tool

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,129 @@
...
+func findPackage(packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
+	for id := range packages {
+		if id.Name == name && id.Revision == revision {
+			return packages[id]
+		}
+	}
+	return pkg.Package{}
+}
+
+func assertImport(t *testing.T, imports pkg.Imports, name, revision string) {
```
microsoftly: `assertImport` and `findPackage` seem like they're going to be re-used a lot. Maybe we can move them to a common testing util package (separate spike of work)

buildtools/pipenv/pipenv_test.go
microsoftly: I agree with @liftM here. I think it's reasonable to assume that `pipenv`'s output will be json consistently.

buildtools/pipenv/pipenv.go
zlav: I chose to implement this as an interface to make it easier to extend in the future if need be. However, I do see how this could be too much. Are you envisioning something along the lines of the following?
```
type SystemPipenv struct {
	PipenvExec func(string) (string, error)
}
```

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,156 @@
...
+	return SystemPipenv{
+		PipenvExec: EnvironmentData{},
+	}
+}
+
+// Deps returns the list of imports and associated package graph
+// using the data from p.PipenvExec.DataAccess.
+func (p SystemPipenv) Deps() (graph.Deps, error) {
+	depGraph := graph.Deps{}
+	rawJSON, err := p.PipenvExec.DataAccess("")
```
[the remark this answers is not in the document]
zlav: This was a mistake, the reason was initially to let users access a custom Pipfile location. Once I moved away from this it had no purpose.

buildtools/pipenv/pipenv_test.go
zlav: I agree I think this could live in the graph package.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+import (
+	"encoding/json"
+
+	"github.com/fossas/fossa-cli/errors"
+	"github.com/fossas/fossa-cli/exec"
+	"github.com/fossas/fossa-cli/graph"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// Pipenv enables python.go to use a custom Deps() function for testing.
```
liftM: [nit] `Pipenv defines the interface for all pipenv tool implementations.`

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+type Pipenv interface {
+	Deps() (graph.Deps, error)
+}
+
+// PipenvCmd controls the data access function for Pipenv.
+type PipenvCmd struct {
+	Graph func() (string, error)
+}
+
+// dependency is used to unmarshall the output from `pipenv graph --json-tree`
```
liftM: [nit] Typo: `unmarshal`.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/fossas/fossa-cli/graph"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// Pipenv enables python.go to use a custom Deps() function for testing.
+type Pipenv interface {
+	Deps() (graph.Deps, error)
+}
+
+// PipenvCmd controls the data access function for Pipenv.
```
liftM: [nit] `PipenvCmd implements Pipenv by parsing command output.`

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+// dependency is used to unmarshall the output from `pipenv graph --json-tree`
+// and store an object representing an imported dependency.
+type dependency struct {
+	Package      string `json:"package_name"`
+	Resolved     string `json:"installed_version"`
+	Target       string `json:"required_version"`
+	Dependencies []dependency
+}
+
+// New returns a new Pipenv instance that controls data access.
```
liftM: [nit] `New constructs a Pipenv instance that calls the pipenv build tool.`

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+func New() Pipenv {
+	return PipenvCmd{
+		Graph: GraphJSON,
+	}
+}
+
+// Deps returns the list of imports and associated package graph
+// using the data from p.PipenvExec.DataAccess.
+func (p PipenvCmd) Deps() (graph.Deps, error) {
```
liftM: [nit] `Deps returns the dependencies of a pipenv project.`

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,146 @@
...
+// and store an object representing an imported dependency.
+type dependency struct {
+	Package      string `json:"package_name"`
+	Resolved     string `json:"installed_version"`
+	Target       string `json:"required_version"`
+	Dependencies []dependency
+}
+
+// New returns a new Pipenv instance that controls data access.
+func New() Pipenv {
```
liftM: We should probably pass in the `pipenv` dir here.

analyzers/python/python.go
```diff
@@ -59,6 +61,7 @@ func New(m module.Module) (*Analyzer, error) {
 		PythonCmd:     pythonCmd,
 		PythonVersion: pythonVersion,
 
+		Pipenv: pipenv.New(m.Dir),
 		Pip: pip.Pip{
```
microsoftly: /tangent
We'll need to migrate `pip` similarly

analyzers/python/python.go
```diff
@@ -154,6 +157,12 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 			Direct:     imports,
 			Transitive: fromImports(imports),
 		}, nil
+	case "pipenv":
+		depGraph, err := a.Pipenv.Deps()
```
microsoftly: [nit] you can just return this line and avoid the nil check below

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,148 @@
...
+	if err != nil {
+		return depGraph, err
+	}
+
+	depGraph.Direct = getDirectDeps(deps)
+	depGraph.Transitive = getTransitiveDeps(deps)
+	return depGraph, nil
+}
+
+func getDependencies(graphJSONFile string) ([]dependency, error) {
```
microsoftly: [nit] I'm not sure `graphJSONFile` makes sense. You're passing a string, not a path to a file

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,148 @@
...
+import (
+	"encoding/json"
+
+	"github.com/fossas/fossa-cli/errors"
+	"github.com/fossas/fossa-cli/exec"
+	"github.com/fossas/fossa-cli/graph"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// Pipenv defines the interface for all pipenv tool implementation.
```
liftM: [nit] `s/implementation/implementations`

buildtools/pipenv/testdata/get-deps/bad.json
```diff
@@ -0,0 +1 @@
+"not a real json file"
```
microsoftly: [nit] doesn't look like this is used anywhere

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	deps, error := testEnv.Deps()
+	assert.NoError(t, error)
+
+	imports := deps.Direct
+	assert.NotZero(t, imports)
+	assert.Equal(t, 2, len(imports))
+	assertImport(t, imports, "a", "1.0.0")
+	assertImport(t, imports, "b", "2.0.0")
+
```
microsoftly: [nit] newline

buildtools/pipenv/testdata/get-deps/pipenv-graph-json-tree.json
```diff
@@ -0,0 +1,42 @@
...
+            },
+            {
+                "package_name": "e",
+                "installed_version": "5.0.0",
+                "required_version": "5.0.0",
+                "dependencies": []
+            }
+        ]
+    }
+
```
microsoftly: is this new line part of the actual output or accidental?

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/buildtools/pipenv"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// newMockPipenv constructs a pipenv.PipenvCmd with a custom Graph function.
+func newMockPipenv(file []byte, err error) pipenv.PipenvCmd {
```
microsoftly: [nit] private and helper methods should be below public methods

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// PipenvCmd implements Pipenv by parsing command output.
+type PipenvCmd struct {
+	Dir   string
+	Graph func(dir string) (string, error)
+}
+
+// dependency is used to unmarshal the output from `pipenv graph --json-tree`
+// and store an object representing an imported dependency.
```
liftM: [nit] This second line doesn't add much extra value to the comment. Also, I would argue that struct values are not objects in the classical OOP sense -- they aren't capable of dynamic dispatch or subtyping.

buildtools/pipenv/pipenv.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// Pipenv defines the interface for all pipenv tool implementation.
+type Pipenv interface {
+	Deps() (graph.Deps, error)
+}
+
+// PipenvCmd implements Pipenv by parsing command output.
+type PipenvCmd struct {
```
liftM: [nit] Now that I think about it, this might be better named `Cmd` because `pipenv.PipenvCmd` stutters.

buildtools/pipenv/testdata/get-deps/bad.json
liftM: Remove this file.

buildtools/pipenv/pipenv_test.go
liftM: I'd just call this `Mock`.

buildtools/pipenv/pipenv_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"io/ioutil"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/buildtools/pipenv"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+// newMockPipenv constructs a pipenv.PipenvCmd with a custom Graph function.
```
liftM: [nit] `Mock constructs a pipenv.Cmd using mock build tool output.`

In general, declaration comments are more useful when they describe _why_ rather than _how_ a declaration is used.

analyzers/python/python.go
zlav: You are absolutely right... thanks for pointing this out.

buildtools/pipenv/testdata/get-deps/pipenv-graph-json-tree.json
zlav: Accidental

Issue #316, opened 2018-10-04
vndr analyer does not acknowledge a user provided repository


zlav:
**Describe the bug**
[Vndr package manage](https://github.com/LK4D4/vndr) provides the option to users to provide a custom repository to users for vendoring forks instead of original repositories

**To Reproduce**
1. Create a vendor.conf file
2. Add a package with a custom rep field
3. Run analyze

**Expected behavior**
Fossa-cli should pick up the user provided repository and analyze it

PR #318, opened 2018-10-05
fix(vndr user repos) add support for user specified package locations (#316)

zlav:


buildtools/vndr/vndr_test.go
```diff
@@ -0,0 +1,30 @@
+package vndr_test
+
+import (
+	"testing"
+
+	"github.com/fossas/fossa-cli/buildtools"
+	"github.com/fossas/fossa-cli/buildtools/vndr"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
```
liftM: [nit] Import grouping.

I think GolangCI might catch this for us when configured, but not by default. Might be worth looking into that, but that configuration should be a separate PR.

zlav: This is actually a known [issue](https://github.com/golang/go/issues/20818) in goimports at the moment that is on their roadmap. I always thought I was safe running `goimport` `gofmt` `golint` in my shell before pushing. [Link](https://github.com/aristanetworks/goarista/tree/master/cmd/importsort) to an example that fixes this problem at least before commits.

PR #320, opened 2018-10-05
feat(ruby analyzer integration test) add ruby analyzer integration test

microsoftly:


testing/fixtures/fixtures.go
```diff
@@ -24,24 +24,48 @@ func Directory() string {
...
+
+	return false, nil
+}
+
+// ProjectInitializerFunction defines how a single project should be initialized *after* it has already been cloned
+type ProjectInitializerFunction func(proj Project, projectDir string) error
+
+// Initialize executes git clone in target directory and checksout the provided commit, then runts the initializerFn. This is done asynchronously for each provided project
+func Initialize(baseDir string, projects []Project, initializerFn ProjectInitializerFunction) {
+	ensureBaseDirectoryExists()
```
golangcibot: Error return value of `ensureBaseDirectoryExists` is not checked

.circleci/config.yml
```diff
@@ -81,16 +81,16 @@ jobs:
...
           keys:
-            - v2-integration-test-fixtures
+            - v3-integration-test-fixtures
       - run:
           name: Run integration tests
           command: |
             # Load shell helpers (e.g. sdkman)
             source /home/fossa/.bashrc
-            make -s ci-integration-test > $TEST_RESULTS/go-test-report.xml
+            #make -s ci-integration-test > $TEST_RESULTS/go-test-report.xml
```
liftM: ???

Makefile
```diff
@@ -118,7 +118,7 @@ integration-test:
 
 .PHONY: ci-integration-test
 ci-integration-test: $(GO_JUNIT_REPORT)
-	GO_TEST_FLAGS="-coverprofile=coverage.txt -v" make -s integration-test | go-junit-report;
+	GO_TEST_FLAGS="-p 256 -coverprofile=coverage.txt -v" make -s integration-test | go-junit-report;
```
liftM: Why 256? [More is not always better.](https://en.wikipedia.org/wiki/Amdahl%27s_law)

My intuition says that ideally the `-p` is set to the number of logical cores, which is the default.

If anything, it sounds like we should be setting `GOMAXPROCS` or `-parallel`, but the defaults for those values (also the number of logical cores) also sound sane to me.

I would not adjust these knobs without benchmarks and good evidence that this is better.

See also https://github.com/golang/go/issues/11521 and `go help testflag`.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	"github.com/apex/log"
+	"github.com/fossas/fossa-cli/analyzers"
+	"github.com/fossas/fossa-cli/exec"
+	"github.com/fossas/fossa-cli/files"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/testing/fixtures"
+	"github.com/fossas/fossa-cli/testing/runfossa"
+	"github.com/stretchr/testify/assert"
```
liftM: [nit] Here and above with `apex/log`: import grouping.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var rubyAnalyzerFixtureDir = filepath.Join(fixtures.Directory(), "ruby", "analyzer")
+
+func TestMain(m *testing.M) {
```
liftM: Be aware that `TestMain` is _package_-scoped, not _file_-scoped. This will prevent unit tests from running on `-short`, just like what's happening for Node.JS right now.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+
+		return nil
+	})
+
+	exitCode := m.Run()
+	defer os.Exit(exitCode)
+}
+
+func TestAnalylze(t *testing.T) {
```
liftM: [nit] Typo.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+		if err != nil {
+			panic(err)
+		}
+		if ymlAlreadyExists {
+			return nil
+		}
+
+		args := []string{"install"}
+
+		// we could extend or refactor the fixtures.Project struct, but because this is a single case, this is simpler for the time being
```
liftM: :+1:

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/testing/fixtures"
+	"github.com/fossas/fossa-cli/testing/runfossa"
+	"github.com/stretchr/testify/assert"
+)
+
+var rubyAnalyzerFixtureDir = filepath.Join(fixtures.Directory(), "ruby", "analyzer")
+
+func TestMain(m *testing.M) {
+	// flags are not parsed at this point. In order to have testing.Short() read actually provided values, this must be executed
```
liftM: [nit] Here and elsewhere, capitalisation and punctuation on comments.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+	exitCode := m.Run()
+	defer os.Exit(exitCode)
+}
+
+func TestAnalylze(t *testing.T) {
+	t.Parallel()
+	for _, project := range projects {
+		proj := project
+		t.Run("Analysis:"+proj.Name, func(t *testing.T) {
```
liftM: [nit] This output would be prettier with a space.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		return nil
+	})
+
+	exitCode := m.Run()
+	defer os.Exit(exitCode)
+}
+
+func TestAnalylze(t *testing.T) {
+	t.Parallel()
```
liftM: Why call `t.Parallel()` here?

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+var projects = []fixtures.Project{
+	// branch 5-2-stable
+	fixtures.Project{
+		Name:   "rails",
+		URL:    "https://github.com/rails/rails",
+		Commit: "3a875402189f0d42446963ff0322f413f892d544",
+	},
+	// release tag v2.1.5
```
liftM: Does it make sense to add `Branch` or `Tag` to `fixtures.Project` for checkout?

My intuition says no, because we'd want to make sure we're on the right commit anyway.

analyzers/ruby/integration_test.go
microsoftly: Ignoring for now, `TestMain` will be removed

analyzers/ruby/integration_test.go
microsoftly: The commits listed here are the pointing to what I believe are stable releases. Which commits would you recommend I use instead? I was just noting the in english details of the commit to help a reader understand why that commit was selected

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,102 @@
...
+func TestAnalyze(t *testing.T) {
+	if testing.Short() {
+		return
+	}
+	t.Parallel()
+
+	fixtures.Initialize(rubyAnalyzerFixtureDir, projects, projectInitializer)
+	for _, project := range projects {
+		proj := project
+		t.Run("Analysis:\t"+proj.Name, func(t *testing.T) {
```
liftM: I'd prefer `" "` over `"\t"`.

analyzers/ruby/integration_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/fossas/fossa-cli/files"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/testing/fixtures"
+	"github.com/fossas/fossa-cli/testing/runfossa"
+)
+
+var rubyAnalyzerFixtureDir = filepath.Join(fixtures.Directory(), "ruby", "analyzer")
+
+func TestAnalyze(t *testing.T) {
```
liftM: [nit] I'd call this `TestIntegration`.

Issue #328, opened 2018-10-09
node dependency graph incorrectly built


zlav:
**Describe the bug**
`nodejs.go` analyzer builds the dependency graph incorrectly

**To Reproduce**
1. Build a node project.
2. Run fossa analyze.
3. Observe that direct dependencies are labeled incorrectly.

liftM:
Fixed by #331.

PR #331, opened 2018-10-09
fix(node graph creation) change the way node creates dependency graphs

zlav:
Issue [#328]


zlav:
fossa analyze
Before
screen shot 2018-10-08 at 5 32 13 pm

After
screen shot 2018-10-08 at 5 30 19 pm

Issue #333, opened 2018-10-09
NodeJS unit tests are not actually running


liftM:
The NodeJS analyser tests currently have a `TestMain`, which short-circuits the unit tests.

liftM:
Fixed by #336.

PR #334, opened 2018-10-09
feat(integration tests) remove use of TestMain, replace with a threadsafe single execution ensured initializer

microsoftly:
addresses #333 

tldr; a fixture is only initialized once, any other attempts to initialize the same fixture will wait until the first is complete and not execute redundant work.

By ensuring we can have multiple tests with the same fixture executed at the start, we can have more flexibly structured tests that can run in parallel.

This introduces some fairly complex logic, but it need only be created once.

### DIFF
--- a/testing/fixtures/fixtures.go
+++ b/testing/fixtures/fixtures.go
@@ -27,8 +27,28 @@ func Directory() string {
 // ProjectInitializerFunction defines how a single project should be initialized *after* it has already been cloned
 type ProjectInitializerFunction func(proj Project, projectDir string) error
 
-// Initialize executes git clone in target directory and checksout the provided commit, then runts the initializerFn. This is done asynchronously for each provided project
+// Initialize executes git clone in target directory and checksout the provided commit, then runts the initializerFn. This is done asynchronously for each provided project.
+// If the same fixture (keyed by baseDir) calls this function twice, the first call will run the initialization while the second waits for the first to complete. After the first
+// completes, the second returns without doing any further work. This allows multiple tests for the same fixture to live within the same file without running into concurrency issues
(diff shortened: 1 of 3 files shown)


testing/fixtures/fixtures.go
```diff
@@ -27,8 +27,28 @@ func Directory() string {
...
+// If the same fixture (keyed by baseDir) calls this function twice, the first call will run the initialization while the second waits for the first to complete. After the first
+// completes, the second returns without doing any further work. This allows multiple tests for the same fixture to live within the same file without running into concurrency issues
 func Initialize(baseDir string, projects []Project, initializerFn ProjectInitializerFunction) {
+	globalLockMutex.Lock()
+	// if the lock already exists, then that particular fixture has already begun initialization; wait until initialization is complete, then return
+	if fixtureMutex, mutexExists := fixtureInitializerLocks[baseDir]; mutexExists {
+		globalLockMutex.Unlock()
+		// call Lock to cause the thread to wait until the execution is complete
+		fixtureMutex.Lock()
+		fixtureMutex.Unlock()
```
golangcibot: empty critical section

microsoftly: That is intended, this is a hack to ensure execution synchronization

microsoftly:
There can be multiple test functions/cases for a single integration test that rely on the same fixture. As an example (though not a good one because it just guarantees the setup executed correctly ...), the `nodejs` integration test has two functions `TestTestSetup` and `TestNodejsAnalysis`. We could conceivably add in additional tests such as `TestNodejsTest`. These require the fixtures be initialized for each test, but are best organized in separate functions/scopes within the same integration test file

microsoftly:
closing in favor of using n-subtests instead of separate rooted functions.

PR #336, opened 2018-10-09
refactor(integration test) do not use TestMain in integration tests

microsoftly:
Addresses #333 

Moves `-short` flag checks into the tests themselves

### DIFF
--- a/analyzers/nodejs/integration_test.go
+++ b/analyzers/nodejs/integration_test.go
@@ -1,7 +1,6 @@
 package nodejs_test
 
 import (
-	"flag"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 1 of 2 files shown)


analyzers/nodejs/integration_test.go
```diff
@@ -20,39 +19,14 @@ import (
...
-	assertProjectFixtureExists(t, "standard")
-	assertProjectFixtureExists(t, "sodium-encryption")
-	assertProjectFixtureExists(t, "request")
-}
-
-func TestNodejsAnalysis(t *testing.T) {
-	t.Parallel()
 	for _, project := range projects {
 		proj := project
 		t.Run("Analysis:"+proj.Name, func(t *testing.T) {
```
liftM: [nit] Whitespace would help here.

analyzers/python/integration_test.go
```diff
@@ -18,12 +16,11 @@ import (
 
 var pythonAnalyzerFixtureDir = filepath.Join(fixtures.Directory(), "python", "analyzer")
 
-func TestMain(m *testing.M) {
-	// flags are not parsed at this point. In order to have testing.Short() read actually provided values, this must be executed
-	flag.Parse()
+func TestPythonAnalysis(t *testing.T) {
 	if testing.Short() {
 		return
```
liftM: Actually, here and below: I'd prefer `if testing.Short() { t.Skip() }` over `if testing.Short() { return }`.

Issue #329, opened 2018-10-09
test suite for node dependency graoh


zlav:
**Is your feature request related to a problem? Please describe.**
Issue [#328] should have a test suite built around it to ensure that the problem never resurfaces

PR #337, opened 2018-10-09
test(nodejs): Add tests for checking import graphs

zlav:
Created from issue [#329]

I added a folder called `normal`, welcom to hear name changes but this is normal operation for nodejs analyzation. This holds the node modules and package for the analyzer to grab information from.

I also privitized a previous function called `AssertImport` which was only used in this package and moved it to the end of the file. Github didn't notice this and that is the reason for the large delta at the end of the file.


analyzers/nodejs/nodejs_test.go
```diff
@@ -13,6 +13,46 @@ import (
 	"github.com/fossas/fossa-cli/pkg"
 )
 
+// testNDepsTransitiveImports checks that Analyze() returns the correct transitive deps
+// by asserting on transitive dependencies
+func testNDepsTransitiveImports(t *testing.T) {
```
golangcibot: `testNDepsTransitiveImports` is unused

analyzers/nodejs/nodejs_test.go
```diff
@@ -202,10 +227,35 @@ func TestUsingYarnLockfileFallback(t *testing.T) {
...
-	AssertImport(t, chaiProject.Imports, "type-detect", "4.0.8")
+	assertImport(t, chaiProject.Imports, "assertion-error", "1.1.0")
+	assertImport(t, chaiProject.Imports, "check-error", "1.0.2")
+	assertImport(t, chaiProject.Imports, "get-func-name", "2.0.0")
+	assertImport(t, chaiProject.Imports, "pathval", "1.1.0")
+	assertImport(t, chaiProject.Imports, "deep-eql", "3.0.1")
+	assertImport(t, chaiProject.Imports, "type-detect", "4.0.8")
+}
+
+func assertPackageReturnImports(t *testing.T, packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
```
golangcibot: func `assertPackageReturnImports` is unused

microsoftly: If you're still using this, I would update it to not execute an assertion. Keeping SRP (single responsibility principle) in mind, just fetch the package. Run the assertion on the correctness of the result outside of this scope

analyzers/nodejs/nodejs_test.go
microsoftly: [Nit] Capitalize test functions

analyzers/nodejs/nodejs_test.go
```diff
@@ -13,6 +13,46 @@ import (
 	"github.com/fossas/fossa-cli/pkg"
 )
 
+// testNDepsTransitiveImports checks that Analyze() returns the correct transitive deps
+// by asserting on transitive dependencies
```
microsoftly: [nit] I think this could be more terse. Something along the lines of `verifies the imports of every transitive dep returned from Analyze`. or something to that effect. Right now I'm unsure what you mean by 'correct transitive deps`. What makes them correct?

analyzers/nodejs/testdata/normal/npm-ls-json.json
```diff
@@ -0,0 +1,23 @@
+{
+  "dependencies": {
+    "a": {
+      "version": "1.0",
```
microsoftly: [nit] standard semver has three fields `<major>.<minor>.<patch>`. `npm`/`nodejs` doesn't enforce this, so it should be okay

analyzers/nodejs/testdata/transitive-deps/npm-ls-json.json
```diff
@@ -0,0 +1,23 @@
+{
+  "dependencies": {
+    "a": {
+      "version": "1.0",
```
microsoftly: [nit] standard semver has three fields: <major>.<minor>.<patch>. I don't think `npm` or `nodejs` require, expect, or enforce this, so it should be okay

zlav: I'll change that, good call.

analyzers/nodejs/nodejs_test.go
zlav: Previously we were returning every transitive dep, we just weren't returning them in the correct order or under the correct dependencies. Now we are asserting that the graph is structured correctly, ensuring that each package has the correct list of imports.

analyzers/nodejs/nodejs_test.go
```diff
@@ -13,6 +13,76 @@ import (
...
+
+	deps, err := a.Analyze()
+	assert.NoError(t, err)
+
+	assert.Equal(t, 1, len(deps.Direct))
+	assertImport(t, deps.Direct, "a", "1.0.0")
+
+	assert.Equal(t, 4, len(deps.Transitive))
+
+	babelPolyfillPackage := assertPackageReturnImports(deps.Transitive, "a", "1.0.0")
```
microsoftly: these package names should reflect the actual package names, not the old ones

analyzers/nodejs/testdata/transitive-deps/package.json
```diff
@@ -1,5 +1,5 @@
 {
   "dependencies": {
-    "babel-polyfill": "6.26.0"
+    "a": "a"
```
microsoftly: this should map to some version that's valid given your other files (e.g. `1.0.0` or `^1.0.0`)

zlav: Yea it should, thanks for catching this.

zlav:
@microsoftly the format for `npm ls --json` has not changed throughout past versions of node and npm. However, an interesting note to point out is that npm has gotten increasingly better at recognizing transitive dependencies with newer releases.

analyzers/nodejs/nodejs_test.go
```diff
@@ -202,10 +239,34 @@ func TestUsingYarnLockfileFallback(t *testing.T) {
...
+	assertImport(t, chaiProject.Imports, "assertion-error", "1.1.0")
+	assertImport(t, chaiProject.Imports, "check-error", "1.0.2")
+	assertImport(t, chaiProject.Imports, "get-func-name", "2.0.0")
+	assertImport(t, chaiProject.Imports, "pathval", "1.1.0")
+	assertImport(t, chaiProject.Imports, "deep-eql", "3.0.1")
+	assertImport(t, chaiProject.Imports, "type-detect", "4.0.8")
+}
+
+func assertPackageReturnImports(packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
+	for ID := range packages {
```
liftM: [nit] I'd call this `id`.

analyzers/nodejs/nodejs_test.go
```diff
@@ -202,10 +239,34 @@ func TestUsingYarnLockfileFallback(t *testing.T) {
...
-	AssertImport(t, chaiProject.Imports, "type-detect", "4.0.8")
+	assertImport(t, chaiProject.Imports, "assertion-error", "1.1.0")
+	assertImport(t, chaiProject.Imports, "check-error", "1.0.2")
+	assertImport(t, chaiProject.Imports, "get-func-name", "2.0.0")
+	assertImport(t, chaiProject.Imports, "pathval", "1.1.0")
+	assertImport(t, chaiProject.Imports, "deep-eql", "3.0.1")
+	assertImport(t, chaiProject.Imports, "type-detect", "4.0.8")
+}
+
+func assertPackageReturnImports(packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
```
liftM: I'd call this `findPackage` since it neither asserts nor returns imports.

analyzers/nodejs/nodejs_test.go
```diff
@@ -13,6 +13,58 @@ import (
...
+	deps, err := a.Analyze()
+	assert.NoError(t, err)
+
+	assert.Equal(t, 1, len(deps.Direct))
+	assertImport(t, deps.Direct, "a", "1.0.0")
+
+	assert.Equal(t, 4, len(deps.Transitive))
+
+	packageA := assertPackageReturnImports(deps.Transitive, "a", "1.0.0")
+	assert.NotEqual(t, packageA, pkg.Package{})
```
liftM: Here and below, I believe you can also use `assert.NotZero`.

zlav: You're right, I need to read up on the available assertions

zlav:
@liftM I confirmed the tests fail for the previous analysis that contained the bug we fixed. Merging this commit now.

Issue #343, opened 2018-10-10
Refactoring build uploads


liftM:
We want to defer the responsibility of fetcher type defining and locator definition/normalization to `fossa core`. 

If this change were in effect today, all cli uploaded builds would have a fetcher of `custom`. This means that fetcher definition in `.fossa.yml` should be removed.

Additionally, we should not attempt to normalize locators on uploads, rather we should supply the necessary fields to construct the locators for `fossa core` on each request

microsoftly:
Proposed upload type:
```typescript
type BuildUploadQueryParams = {
  projectName: string;
  revision?: string;
  v: string;
  title: string;
  branch?: string;
  projectURL?: string;
  jiraProjectKey?: string;
  link?: string;
  team?: string;
};
```

liftM:
I'm beginning to think we should add all of these to the body of the payload rather than just the query parameters. Our payload body is shaped weirdly and that's only because of legacy in the backend. My proposal:

```typescript
type Payload = {
  vcs: {
    type: 'git' | 'none'; // Detected VCS type
    reference: string; // VCS reference (branch or tag) name
  };
  project: {
    title: string; // Human-readable name
    id: string; // Project identifier, probably inferred from git remote origin
    url: string; // Project URL (provides that one link on the project page, I think)
    jira: string; // JIRA project key
    team: string; // FOSSA team
  };
  revision: {
    id: string; // Revision identifier, probably inferred from VCS revision hash
    link: string; // Arbitrary link for revision metadata (e.g. link back to CI)
  };
  meta: {
    v: string; // CLI version
  };
  analysis: Analysis; // Existing analysis payload
};
```

We should NOT do any processing of `project.id` -- the backend should be able to infer OrgID from the API key when it maps `project.id` to a project locator.

microsoftly:
I agree, I like it

sirandysaunders:
Save

PR #353, opened 2018-10-15
Feat/remove fetcher for uploads 343

microsoftly:
addresses #343 

The fossa api exposes/will expose an endpoint `/api/cli/supported_versions` that returns an an array of `vX` versions that are currently supported. We can cross reference that list with what the CLI currently knows it can support and select the best version to support (as of now, that selection criterion is simply what is newest). In theory, this should allow us to merge at our leisure 

I'm forgoing programatic testing of selecting which API is selected and the upload payload until we get a stable mock server implementation.


cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -133,6 +133,68 @@ func Do(modules []module.Module) (analyzed []module.Module, err error) {
...
 
 func uploadAnalysis(normalized []fossa.SourceUnit) error {
+	supportedApiVersions, err := fossa.GetSupportedApiVersions()
+	if err != nil {
+		return err
+	}
+
+	// lists out all compatible versions, where versions listed earlier in the list should be preferred to be used over later appearing versions
+	compatibleVersionsByPreference := []string{"v1", "v0"}
+	for _, compatibleVersion := range compatibleVersionsByPreference {
```
liftM: This code is pretty crazy, and we don't need it if we change the interface to `fossa.GetLatestSupportedAPIVersion()`.

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -133,6 +133,68 @@ func Do(modules []module.Module) (analyzed []module.Module, err error) {
...
+	uploadBody.Project.Url = config.ProjectURL()
+	uploadBody.Project.Title = config.Title()
+	uploadBody.Project.Jira = config.JIRAProjectKey()
+
+	uploadBody.Revision.Link = config.Link()
+	uploadBody.Revision.Id = config.Revision()
+
+	uploadBody.Vcs.Reference = config.Branch()
+
+	if config.GitVcsExists() {
```
liftM: This is a good place for `config.TryStrings`.

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,93 @@
...
+	endpoint, err := url.Parse("/api/cli/v1/build")
+	if err != nil {
+		return Locator{}, errors.New("Failed to generate upload URL")
+	}
+	log.WithField("endpoint", endpoint.String()).Debug("uploading build")
+
+	res, statusCode, err := Post(endpoint.String(), payload)
+	log.WithField("response", res).Debug("build upload completed")
+
+	if statusCode == 428 {
```
liftM: Is this behaviour still valid in the `v1` API route?

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+	log.Debug("build upload succeeded")
+
+	var unmarshalled struct {
+		Locator string
+		Error   string
+	}
+	err = json.Unmarshal([]byte(res), &unmarshalled)
+	if err != nil || unmarshalled.Error != "" {
+		return Locator{}, errors.Errorf("Error response from API: %s", res)
```
zlav: We should probably return the error here as well. The error response here is very similar to line 89 without including the error, although I would be interested to know if you have a reason for this.

cmd/fossa/cmd/analyze/analyze.go
zlav: I agree. It looks like we are looping through `compatibleVersionsByPreference` and then later switching on the same values. Is there a reason to check for a supported version twice?

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -133,6 +133,68 @@ func Do(modules []module.Module) (analyzed []module.Module, err error) {
...
+		uploadBody.Vcs.Type = "git"
+	} else {
+		uploadBody.Vcs.Type = "none"
+	}
+
+	locator, err := fossa.UploadV1(uploadBody)
+	display.ClearProgress()
+	if err != nil {
+		log.Fatalf("Error during upload: %s", err.Error())
+		return err
```
zlav: This code is unreachable if the line above it logs fatal. I'm not a fan of logging fatal and would prefer to return the error here instead, same with line 216 where this was done before.

api/fossa/uploadv1.go
microsoftly: Yes, as of now the goal of the v1 API route was to provide an equivalent output for the a different input. The goal was to keep the 428 response.

cmd/fossa/cmd/analyze/analyze.go
microsoftly: I'd be okay with moving this to be scoped within the fossa API package. 

My goal in having this loop was to handle situations where users have an older version of `cli`   or on prem users that may have an older version of core.

@zlav Its not checking supported version twice, it's determining which versions that core supports are also supported by this build of `cli`

api/fossa/apiversion.go
```diff
@@ -0,0 +1,42 @@
...
+
+func GetSupportedApiVersions() ([]string, error) {
+	endpoint, err := url.Parse("/api/cli/supported_versions")
+	if err != nil {
+		return []string{}, err
+	}
+
+	res, statusCode, err := Get(endpoint.String())
+	if err != nil {
+		return []string{}, errors.Wrap(err, "Could not get supported API versions.")
```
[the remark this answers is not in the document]
microsoftly: I've verified that `404` response results in `err == nil`

cmd/fossa/cmd/analyze/analyze.go
microsoftly: I don't see how that would help here. This is a clear `if` `else` case for setting a value where there is no default or initial value to override

api/fossa/upload.go
```diff
@@ -1,114 +1,94 @@
...
-		q.Add("jiraProjectKey", options.JIRAProjectKey)
-	}
-	if options.Link != "" {
-		q.Add("link", options.Link)
-	}
-	if options.Team != "" {
-		q.Add("team", options.Team)
-	}
+func uploadAnalysisV1(normalized []SourceUnit) (Locator, error) {
+	display.InProgress("Uploading analysis...")
```
liftM: The `display.InProgress` calls should go in the `cmd` packages, not the `api`.

api/fossa/upload.go
```diff
@@ -1,114 +1,94 @@
...
 	}
-	log.WithField("endpoint", endpoint.String()).Debug("uploading build")
 
-	res, statusCode, err := Post(endpoint.String(), payload)
-	log.WithField("response", res).Debug("build upload completed")
+	uploadBody.Project.Id = config.Project()
+	uploadBody.Project.Team = config.Team()
+	uploadBody.Project.Url = config.ProjectURL()
+	uploadBody.Project.Title = config.Title()
+	uploadBody.Project.Jira = config.JIRAProjectKey()
```
liftM: [nit] `s/Jira/JIRA`

api/fossa/upload.go
```diff
@@ -1,114 +1,94 @@
...
-	if err != nil {
-		return Locator{}, errors.New("Failed to generate upload URL")
+	uploadBody := V1UploadBody{
+		Analysis: normalized,
 	}
-	log.WithField("endpoint", endpoint.String()).Debug("uploading build")
 
-	res, statusCode, err := Post(endpoint.String(), payload)
-	log.WithField("response", res).Debug("build upload completed")
+	uploadBody.Project.Id = config.Project()
```
liftM: [nit] `s/Id/ID`

api/fossa/upload.go
```diff
@@ -1,114 +1,94 @@
...
-		// TODO: handle "Managed Project" workflow
-		return Locator{}, errors.New("invalid project or revision; make sure this version is published and FOSSA has access to your repo (to submit a custom project, set Fetcher to `custom` in `.fossa.yml`)")
-	} else if statusCode == 403 {
-		return Locator{}, errors.New("you do not have permission to upload builds for this project")
-	} else if err != nil {
-		return Locator{}, errors.Wrap(err, "could not upload")
-	}
-	log.Debug("build upload succeeded")
+	uploadBody.Revision.Link = config.Link()
+	uploadBody.Revision.Id = config.Revision()
```
liftM: [nit] `s/Id/ID`

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type V1UploadBody struct {
+	VCS struct {
+		Type      string
+		Reference string
+	}
+	Project struct {
+		Title string
+		Id    string
```
liftM: [nit] `s/Id/ID`

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,86 @@
...
+		Url   string
```
liftM: [nit] `s/Url/URL`

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,86 @@
...
+		Jira  string
```
liftM: [nit] `s/Jira/JIRA`

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,86 @@
...
+		Team  string
+	}
+	Revision struct {
+		Id   string
```
liftM: [nit] `s/Id/ID`

api/fossa/uploadv1.go
```diff
@@ -0,0 +1,86 @@
...
+		Link string
+	}
+	Meta struct {
+		v string
```
liftM: This should be `V`, otherwise `json.Marshal` will not pick up an unexported field.

PR #355, opened 2018-10-16
fix(npm) npm analyze incorrectly finds module from .fossa.yml

zlav:
This PR changes the logic that nodejs Analyze uses to determine where a module is located. This introduces a breaking change to a few customers with manually edited .fossa.yml files.

I introduced the concept of `nodejs_discover_test.go` to keep test functionality clear because `nodejs_test.go` is a large file and only tests `analyze()`. I also built functionality to correctly test for ignored modules.

### DIFF
--- a/analyzers/nodejs/nodejs.go
+++ b/analyzers/nodejs/nodejs.go
@@ -215,7 +215,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 
 	// if npm as a tool does not exist, skip this
 	if a.NPM.Exists() {
-		pkgs, err := a.NPM.List(filepath.Dir(a.Module.BuildTarget))
+		pkgs, err := a.NPM.List(a.Module.BuildTarget)
 		if err == nil {
 			// TODO: we should move this functionality in to the buildtool, and have it
 			// return `pkg.Package`s.
--- a/analyzers/nodejs/nodejs_discover_test.go
+++ b/analyzers/nodejs/nodejs_discover_test.go
@@ -1,0 +1,52 @@
+package nodejs_test
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/fossas/fossa-cli/analyzers/nodejs"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+/* Nested Module Order
+	└─┬ nested-modules
+      └─┬ module-1
+       	├── module-2
+      	└── module-3
+*/
+func TestDiscoverNestedModules(t *testing.T) {
+	var expectedModules []module.Module
+	expectedModules = append(expectedModules, newModule("module-2", "module-1/module-2"))
+	expectedModules = append(expectedModules, newModule("module-3", "module-1/module-3"))
+	expectedModules = append(expectedModules, newModule("module-1", "module-1"))
+	expectedModules = append(expectedModules, newModule("nested-modules", "."))
+
+	modules, err := nodejs.Discover("testdata/nested-modules", make(map[string]interface{}))
+	assert.NoError(t, err)
+	assert.Equal(t, expectedModules, modules)
+}
+
+/* Ignored Module Order
+	└─┬ ignored-modules
+      ├── node_modules
+      └── bower_components
+*/
+func TestNodeModulesAndBowerIgnored(t *testing.T) {
+	var expectedModules []module.Module
+	expectedModules = append(expectedModules, newModule("ignored-modules", "."))
+
+	modules, err := nodejs.Discover("testdata/ignored-modules", make(map[string]interface{}))
+	assert.NoError(t, err)
+	assert.Equal(t, expectedModules, modules)
+}
+
+func newModule(name, location string) module.Module {
+	return module.Module{
+		Name:        name,
+		Type:        pkg.NodeJS,
+		BuildTarget: location,
+		Dir:         location,
+	}
+}
--- a/cmd/fossa/cmd/upload/upload.go
+++ b/cmd/fossa/cmd/upload/upload.go
@@ -26,10 +26,10 @@ import (
 	"os"
 	"strings"
 
+	"github.com/apex/log"
 	"github.com/pkg/errors"
 	"github.com/urfave/cli"
 
-	"github.com/apex/log"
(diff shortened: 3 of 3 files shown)


analyzers/nodejs/nodejs_discover_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+/* Nested Module Order
+    └─┬ nested-modules
+      └─┬ module-1
+	  ├── module-2
+	  └── module-3
+*/
+func TestNestedModules(t *testing.T) {
```
microsoftly: were these tests failing before you made the logic change? I'm not seeing how this tests the trailing `/` issue

zlav: This test doesn't specifically test for trailing `/`'s or analyze's reaction to them. What this does test for is that we are correctly building .fossa.yml files. I decided to implement this after talking with @liftM to prevent future degrading changes.

zlav:
At the moment this PR is tested through `integration_test.go` by removing the logic which added `package.json` to the end of a build target. By removing `package.json` we are testing the logic that removes the trailing directory or file. I am not confident that this tests monorepos or nested submodule behavior and I am thinking of ways to do this.

analyzers/nodejs/nodejs.go
```diff
@@ -215,7 +215,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 
 	// if npm as a tool does not exist, skip this
 	if a.NPM.Exists() {
-		pkgs, err := a.NPM.List(filepath.Dir(a.Module.BuildTarget))
+		pkgs, err := a.NPM.List(a.Module.BuildTarget)
```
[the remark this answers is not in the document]
zlav: To ease some of your concerns this won't break catostrophically on old config files with a trailing slash, however it will break with a `package.json`.

I agree with this and think both of these cases are the only times we should issue a warning or change the input. If you have any other cases let me know.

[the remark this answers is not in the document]
zlav: Updated the logic to sanitize `a.Module.BuildTarget` in `Analyze()` as well as `IsBuilt()`. Looking for feedback on the warning messages in particular. I think they are a little verbose but I am highly favoring descriptive information here to make the changes easier for users.

analyzers/nodejs/nodejs.go
```diff
@@ -261,6 +263,18 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	return yarn.FromProject(filepath.Join(a.Module.BuildTarget, "package.json"), filepath.Join(a.Module.BuildTarget, "yarn.lock"))
 }
 
+// sanitizeBuildTarget ensures that legacy behavior stays intact but users are warned if it is implemented.
+func (a *Analyzer) sanitizeBuildTarget() {
+	if strings.HasSuffix(a.Module.BuildTarget, "/package.json") {
+		log.Warn("Specifiying the package.json file as a module's target in fossa's config file is no longer supported. Only the directory is neccessary.")
```
liftM: Typo: `Specifying`

analyzers/nodejs/nodejs.go
```diff
@@ -261,6 +263,18 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	return yarn.FromProject(filepath.Join(a.Module.BuildTarget, "package.json"), filepath.Join(a.Module.BuildTarget, "yarn.lock"))
 }
 
+// sanitizeBuildTarget ensures that legacy behavior stays intact but users are warned if it is implemented.
+func (a *Analyzer) sanitizeBuildTarget() {
```
liftM: For clarity, I would call this `fixLegacyBuildTarget`.

analyzers/nodejs/nodejs.go
```diff
@@ -261,6 +263,18 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
+	if strings.HasSuffix(a.Module.BuildTarget, "/package.json") {
+		log.Warn("Specifiying the package.json file as a module's target in fossa's config file is no longer supported. Only the directory is neccessary.")
+		a.Module.BuildTarget = strings.TrimSuffix(a.Module.BuildTarget, "/package.json")
```
liftM: Is it feasible for this to return the build target string instead of modifying in-place? I know it's a lot more plumbing, but avoiding mutation makes code easier to reason about (because it makes data and control flow dependencies explicit).

analyzers/nodejs/nodejs.go
liftM: Typo: `necessary`

liftM: Instead of the second sentence, I'd prefer: `Instead, the target should specify the path to the project folder.`

analyzers/nodejs/nodejs.go
```diff
@@ -261,6 +263,18 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
+	}
+	if strings.HasSuffix(a.Module.BuildTarget, "/") {
+		log.Warn("The provided fossa config file is malformed. Ensure there are no trailings /'s in any of the specified modules' targets.")
```
liftM: Telling them it's malformed before specifying _how_ it's malformed is a bit redundant. I'd prefer: `Trailing slashes in module targets are not supported.`

analyzers/nodejs/nodejs.go
zlav: It isn't that much more plumbing at all. This is the way I originally had it structured but decided to get cute. I think that is a better practice and will implement it.

analyzers/nodejs/nodejs.go
```diff
@@ -261,6 +263,19 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	return yarn.FromProject(filepath.Join(a.Module.BuildTarget, "package.json"), filepath.Join(a.Module.BuildTarget, "yarn.lock"))
 }
 
+// fixLegacyBuildTarget ensures that legacy behavior stays intact but users are warned if it is implemented.
+func fixLegacyBuildTarget(target string) string {
+	if strings.HasSuffix(target, "/package.json") {
```
liftM: Here and below, use `filepath.Separator` instead of hardcoding `/`. This will do the right thing on Windows, which uses `\`.

Issue #346, opened 2018-10-11
Better experience with options on multi-module builds


liftM:
Multi-module builds (both from `fossa` and with a preconfigured `.fossa.yml`) currently ignore options passed by command line.

This behaviour was implemented because it would be unclear which options were meant for which modules, and there was no clear design for specifying options per module. Instead, the CLI warns that command line flags are ignored.

Experience has shown that:

- Users don't read this warning.
- They often expect flags to apply to all modules (or are unaware that a project can even have multiple modules).
- Generating and configuring a `.fossa.yml` is too much friction for many use cases.

Fortunately, module options are a `map[string]interface{}`, which makes this implementation pretty easy.

PR #359, opened 2018-10-17
fix(module options) allow command line options for multi module builds

zlav:
Created in response to Issue [#346]
Currently, fossa-cli ignores options provided through the command line if modules are specified inside a fossa config file. This fix ensures that if an option is set manually it will apply to each module. If there is a conflict between a config specified option and a command line option, fossa will prioritize the command line option.

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -9,12 +9,24 @@ import (
 	"github.com/urfave/cli"
 
 	"github.com/fossas/fossa-cli/config"
+	"github.com/fossas/fossa-cli/vcs"
 )
 
 func TestInitWorksWithoutGitRepository(t *testing.T) {
-	err := os.Chdir("/")
+	// Change to a directory without Git repository.
+	cwd, err := os.Getwd()
 	assert.NoError(t, err)
+	err = os.Chdir("/")
+	assert.NoError(t, err)
+	defer func() {
+		os.Chdir(cwd)
+	}()
 
+	// Check that there actually is no Git repository.
+	_, _, err = vcs.Nearest(".")
+	assert.Error(t, err, vcs.ErrNoNearestVCS)
+
+	// Check that config.setContext doesn't break.
 	ctx := cli.NewContext(cli.NewApp(), flag.NewFlagSet("test", 0), nil)
 	err = config.SetContext(ctx)
 	assert.NoError(t, err)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -19,7 +19,8 @@ func TestInitWorksWithoutGitRepository(t *testing.T) {
 	err = os.Chdir("/")
 	assert.NoError(t, err)
 	defer func() {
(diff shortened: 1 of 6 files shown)


config/keys_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+func TestModulesCommandLineOptions(t *testing.T) {
+	// Set the expected flags
+	flagSet := flag.NewFlagSet("test", 0)
+	flags.ConfigF.Apply(flagSet)
+	flags.OptionF.Apply(flagSet)
+
+	// Set the flag values
+	ctx := cli.NewContext(cli.NewApp(), flagSet, nil)
+	ctx.Set(flags.Config, "testdata/test.yml")
```
golangcibot: Error return value of `ctx.Set` is not checked

config/keys_test.go
```diff
@@ -0,0 +1,44 @@
...
+	ctx.Set(flags.Option, "allow-unresolved:true")
```
golangcibot: Error return value of `ctx.Set` is not checked

config/config_test.go
```diff
@@ -9,12 +9,24 @@ import (
...
 
 func TestInitWorksWithoutGitRepository(t *testing.T) {
-	err := os.Chdir("/")
+	// Change to a directory without Git repository.
+	cwd, err := os.Getwd()
 	assert.NoError(t, err)
+	err = os.Chdir("/")
+	assert.NoError(t, err)
+	defer func() {
+		os.Chdir(cwd)
```
golangcibot: Error return value of `os.Chdir` is not checked

config/keys.go
```diff
@@ -189,17 +189,29 @@ func Modules() ([]module.Module, error) {
...
+		return modules, errors.New("No modules provided")
 	}
 
-	// TODO: specifying zero modules should be an error (we should add a test for this)
-	return file.Modules(), nil
+	return modulesWithOptions(modules, options), nil
+}
+
+// modulesWithOptions applies command line options to each module and overwrites conflicting options.
+func modulesWithOptions(modules []module.Module, options map[string]interface{}) []module.Module {
```
liftM: [nit] I'd prefer to inline this, since it's not reused or tested separately.

config/keys.go
```diff
@@ -189,17 +189,29 @@ func Modules() ([]module.Module, error) {
...
+	if len(options) == 0 {
```
liftM: [nit] This check isn't really necessary, since the for loop is cheap.

config/keys.go
```diff
@@ -189,17 +189,29 @@ func Modules() ([]module.Module, error) {
...
+		return modules
+	}
+
+	for index := range modules {
+		if modules[index].Options == nil {
```
liftM: This logic is a bit obfuscated. I think the actual _intended_ semantics here are to `make` a map if `nil`.

PR #360, opened 2018-10-18
[Gradle] Fix root project discovery

cedbossneo:


liftM:
Thanks for the PR! I still need to investigate this further to ensure this behaviour is correct, and add some test fixtures to prevent regressions.

zlav:
Closing this PR because recent updates should have solved this issue. We also implemented configuration filtering and will by default return dependencies from the `:compile` configuration as outlined in this PR. Thank you!

Issue #252, opened 2018-09-05
fossa-cli version 0.7.6 - panic: runtime error: slice bounds out of range - java maven project


Michael-Frank:
**Describe the bug**
```
panic: runtime error: slice bounds out of range
goroutine 1 [running]:
github.com/fossas/fossa-cli/buildtools/maven.ParseDependencyTree
```
**To Reproduce**
1. checkout https://github.com/Michael-Frank/jmh-Introduction.git
2. manually downloaded https://github.com/fossas/fossa-cli/releases/download/v0.7.6/fossa-cli_0.7.6_windows_amd64.zip
3.for first tests extracted zip contents into project root folder 
4. executed `fossa.exe -o`

**Debugging information**
Please provide the following:

1. The output of running the buggy command with the `--debug` flag
```
DEBUG done running callers=[/home/fossa/go/src/github.com/fossas/fossa-cli/exec/run.go:github.com/fossas/fossa-cli/exec.Run:63 /home/fossa/go/src/github.com/fossas/fossa-cli/buildtools/maven/maven.go:github.com/fossas/fossa-cli/buildtools/maven.(*Maven).DependencyTree:97 /home/fossa/go/src/github.com/fossas/fossa-cli/analyzers/maven/maven.go:github.com/fossas/fossa-cli/analyzers/maven.(*Analyzer).Analyze:137 /home/fossa/go/src/github.com/fossas/fossa-cli/cmd/fossa/cmd/analyze/analyze.go:github.com/fossas/fossa-cli/cmd/fossa/cmd/analyze.Do:122 /home/fossa/go/src/github.com/fossas/fossa-cli/cmd/fossa/cmd/analyze/analyze.go:github.com/fossas/fossa-cli/cmd/fossa/cmd/analyze.Run:48 /home/fossa/go/src/github.com/fossas/fossa-cli/cmd/fossa/main.go:main.Run:75 /home/fossa/go/src/github.com/fossas/fossa-cli/vendor/github.com/urfave/cli/app.go:github.com/fossas/fossa-cli/vendor/github.com/urfave/cli.HandleAction:490 /home/fossa/go/src/github.com/fossas/fossa-cli/vendor/github.com/urfave/cli/app.go:github.com/fossas/fossa-cli/vendor/github.com/urfave/cli.(*App).Run:264 /home/fossa/go/src/github.com/fossas/fossa-cli/cmd/fossa/main.go:main.main:47] stderr= stdout=[INFO] Scanning for projects...
[...]

PR #362, opened 2018-10-18
fix(maven) fix line parsing in Windows OS

zlav:
This PR is in response to issue [#252].
This fixes an issue fossa-cli has with parsing windows line endings within the maven buildtool.
To verify that the test files are the correct format run `file dos.out` and see the output as `dos.out: ASCII text, with CRLF line terminators` verifying that it has the correct dos line endings. I am considering adding an assertion for this inside of `maven_test.go`, If you have any ideas comment them.
I also discovered an issue with the maven dependency graph that could potentially be included depending on the scope of the solution.


microsoftly:
I would definitely add an assertion to ensure the line endings. Some editors and git settings can auto change that depending on settings.

.circleci/config.yml
```diff
@@ -52,6 +52,10 @@ jobs:
           command: |
             rm $(which go)
             rm $(which dep)
+      - run:
+          name: Install dependencies
+          command: |
+            sudo apt-get install file
```
microsoftly: This could be moved to the `Dockerfile`

buildtools/maven/maven_test.go
```diff
@@ -0,0 +1,71 @@
...
+	assert.Equal(t, "testdata/dos.out: ASCII text, with CRLF line terminators\n", string(out))
+
+	dat, err := ioutil.ReadFile(dosFile)
+	assert.NoError(t, err)
+	direct, transitive, err := maven.ParseDependencyTree(string(dat))
+	assert.NoError(t, err)
+	assert.NotEmpty(t, direct)
+	assert.NotEmpty(t, transitive)
+}
+func TestParseDependencyTreeDOSCat(t *testing.T) {
```
microsoftly: [nit] newline between functions

Issue #295, opened 2018-09-27
Python package with optional dependency is listed as an unreachable dependency


dkleissa:
**Describe the bug**
In one of our python packages, we have a dependency that installs an optional dependency. The syntax is `packagename[optional]==version`. Fossa fails to resolve this dependency, even though it's public and should work fine.

**To Reproduce**
1. Add `python-jose[cryptography]==3.0.0` to a requirements.txt file that is to be scanned by fossa
2. Run scan
3. See error

**Expected behavior**
I would expect this package to be detected and resolved properly, including the optional dependency that has been included.

**Debugging information**
Please provide the following:

1. The output of running the buggy command with the `--debug` flag
2. The output of `fossa --version`
`fossa-cli version 0.7.7 (revision e874ec2891b456ea3689825b37c278b68301bdf0 compiled with go version go1.10.3 linux/amd64)`
3. The project's configuration file (usually `.fossa.yml`)
4. If possible, a minimal test case reproducing the bug (ideally, a GitHub repository)

**Environment (please complete the following information):**
 - OS: macOS/Ubuntu
 - Are you running in CI?: CircleCI
 - Are you running in a Docker image?: No

**Additional context**

PR #365, opened 2018-10-19
fix(python) requirements parses extras properly

zlav:
In response to issue [#295].
This allows our `requirements.txt` parser to acknowledge extras as defined by [PEP 508](https://www.python.org/dev/peps/pep-0508/#extras).

The code I am most interested in hearing feedback on is the logic inside `checkForExtra()`. I can reduce this logic to a single line `return strings.Split(name, "[")[0]` but wanted to keep it verbose to improve readability.


microsoftly:
This looks like it would work, though my concern would be if the character in search is overloaded `]`/`[`. Additionally, my question would be should we ever handle these differently? 

We could add the field to the package metadata once we add that feature

Issue #357, opened 2018-10-17
`fossa upload` causes `panic: upload flag redefined: l`


liftM:
We've defined a duplicate short-flag somewhere.

Fixing the bug here should be easy. Ideally, we should also add automated tests for these kinds of panics -- they should be easy to write.

PR #366, opened 2018-10-23
fix(upload) remove duplication of flags to fix upload

zlav:
In response to issue [#357]
Change link flag to an uppercase flag because fossa upload is also defining a lowercase `-l` flag.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -116,6 +116,10 @@ jobs:
           command: |
             fossa analyze --debug > $ARTIFACTS/fossa-analyze-stdout 2> $ARTIFACTS/fossa-analyze-stderr
       - run:
+          name: Run FOSSA upload
+          command: |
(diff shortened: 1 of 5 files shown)


zlav:
@liftM invoking `upload.Run` won't test this functionality. The problem arises in `main.go` line 38 when `upload.Cmd` is referenced which adds the API flags and upload specific flags. I think the best way to test for this would be in an integration level test and I will investigate where that should go.

.circleci/config.yml
```diff
@@ -110,11 +110,15 @@ jobs:
       - run:
           name: Install CLI
           command: |
-            curl https://raw.githubusercontent.com/fossas/fossa-cli/$CIRCLE_SHA1/install.sh | bash
```
zlav: `$CIRCLE_SHA1` accurately retrieves the current branch's `install.sh` script, however this does not make a difference because `install.sh` references latest tag. The focus here is to test the current binary which can be built using `make build`.

PR #370, opened 2018-11-01
fix(python) Add options support to requirements.txt parsing

zlav:
The requirements.txt parser does not currently acknowledge options or file references as distinct information. It uploads them as dependencies causing a missing dependency on the backend. This fixes that issue.
[Requirements.txt spec](https://pip.pypa.io/en/stable/reference/pip_install/#requirements-file-format)


liftM:
[nit] I would call this PR "Add options support to requirements.txt parsing"

PR #372, opened 2018-11-07
feat(build tags) analyze go build tags

### LINKED ISSUE #341: Specify multiple build tags for Go analysis
Go packages will bring in different dependencies depending on `GOOS` and `GOARCH`. This request is for use cases where users want their report to include a union of all dependencies for all combinations of build tags, OS, and architecture.

zlav:
Fixes #341 
This feature allows a user of fossa to scan their go packages using different sets of build tags in two different ways
1. Run `fossa analyze --option all-tags:true" or include 
```
    options:
      all-tags: true
```
in your fossa config file. This will analyze fossa's default build tags list which corresponds to all allowed golang build tags.
2. If you are aware of all used build tags you may add them individually in the fossa config file:
```
    options:
      tags:
        - windows
        - linux
        - !windows
        - custom-test-build linux
```
This allows users to add custom build tags and combinations that the default tags won't pick up.

Notes on the code modified in this PR:
- User entered build tags are added to the analyzer in `go.New()`.
- Build tag analysis happens as a for loop over the previous analysis method. This Injects build tags into the go build tool and adds them to `go list` command.
- Test fixtures were added as well as the `analyze_test.go` file that tests from `go.New()` and asserts on the dependencies returned upon `Analyze()`.
- Minor style nits on other files, mostly import ordering.


zlav:
Coverage has gone up but the new feature is not being tested.

buildtools/gocmd/go.go
```diff
@@ -59,8 +57,8 @@ type GoListPackageError struct {
 }
 
 // ListOne runs List for a single package.
-func (g *Go) ListOne(pkg string) (Package, error) {
-	pkgs, err := g.List([]string{pkg})
+func (g *Go) ListOne(flags []string, pkg string) (Package, error) {
```
liftM: [nit] Personally, I think it's nicer to use `(pkg string, flags []string)`.

buildtools/gocmd/go.go
```diff
@@ -71,17 +69,14 @@ func (g *Go) ListOne(pkg string) (Package, error) {
 }
 
 // List runs `go list` to return information about packages.
-func (g *Go) List(pkgs []string) ([]Package, error) {
+func (g *Go) List(flags, pkgs []string) ([]Package, error) {
```
liftM: [nit] Same here: I personally prefer `(pkgs, flags []string)`, because it feels like `pkgs` is the "primary" argument somehow.

buildtools/gocmd/go.go
```diff
@@ -71,17 +69,14 @@ func (g *Go) ListOne(pkg string) (Package, error) {
...
 	// Run `go list -json $PKG` and unmarshal output.
 	var output []GoListOutput
+	flags = append(flags, pkgs...)
```
liftM: I wouldn't modify an argument, especially not a value with reference semantics (like slices). Instead, I would declare a new variable e.g. `argv := append(flags, pkgs...)`.

analyzers/golang/util.go
```diff
@@ -18,7 +18,7 @@ var (
 
 // Dir returns the absolute path to a Go package.
 func (a *Analyzer) Dir(importpath string) (string, error) {
-	pkg, err := a.Go.ListOne(importpath)
+	pkg, err := a.Go.ListOne([]string{}, importpath)
```
liftM: The zero value of a slice is `nil`, which is [preferred over zero-length slices](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

analyzers/golang/golang.go
```diff
@@ -110,7 +124,7 @@ func (a *Analyzer) Build() error {
 func (a *Analyzer) IsBuilt() (bool, error) {
 	m := a.Module
 	log.Debugf("%#v", m)
-	pkg, err := a.Go.ListOne(m.BuildTarget)
+	pkg, err := a.Go.ListOne([]string{}, m.BuildTarget)
```
liftM: [Prefer `nil` over `[]T{}`](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

analyzers/golang/analyze_test.go
```diff
@@ -18,7 +18,7 @@ func TestProjectNotInDeps(t *testing.T) {
 	analyzer, err := golang.New(module.Module{Name: "test", Type: pkg.Go, BuildTarget: buildTarget})
 	assert.NoError(t, err)
 
-	main, err := analyzer.Go.ListOne(buildTarget)
+	main, err := analyzer.Go.ListOne([]string{}, buildTarget)
```
liftM: [Prefer `nil` over `[]T{}`](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

analyzers/golang/discover.go
```diff
@@ -40,12 +40,10 @@ func Discover(dir string, opts map[string]interface{}) ([]module.Module, error)
...
 	g := gocmd.Go{
-		Cmd:  cmd,
-		Dir:  dir,
-		OS:   options.BuildOS,
-		Arch: options.BuildArch,
+		Cmd: cmd,
+		Dir: dir,
 	}
-	found, err := g.List([]string{"./..."})
+	found, err := g.List([]string{}, []string{"./..."})
```
liftM: [Prefer `nil` over `[]T{}`](https://github.com/golang/go/wiki/CodeReviewComments#declaring-empty-slices).

analyzers/golang/analyze.go
```diff
@@ -107,79 +107,94 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
-	for _, gopkg := range deps {
-		log.Debugf("Getting revision for: %#v", gopkg)
-
-		// Resolve dependency.
-		revision, err := a.Revision(project, r, gopkg)
+	var allImports []pkg.Import
+	importMap := make(map[pkg.Import]bool)
+	transitiveDeps := make(map[pkg.ID]pkg.Package)
+
+	for _, buildTag := range a.BuildTags {
```
liftM: [process] This section looks like it's literally what we had before but put into a for-loop. It would be really helpful if you added a GH comment here before review indicating that. In general, it would be really helpful having a description of the high-level approach e.g.:

- Pass flags to the Go buildtool
- Run analysis in a for loop with all build tags
- Add fixtures for testing build tags
- Minor style nits on other files

analyzers/golang/analyze_test.go
```diff
@@ -28,3 +28,41 @@ func TestProjectNotInDeps(t *testing.T) {
...
+	testModule.Options["tags"] = []string{"customtag"}
+	testModule.BuildTarget = "github.com/fossas/fossa-cli/analyzers/golang/testdata/demotags"
+
+	analyzer, err := golang.New(testModule)
+	assert.NoError(t, err)
+	deps, err := analyzer.Analyze()
+	assert.NoError(t, err)
+	assert.Contains(t, deps.Transitive, toml)
+	assert.Contains(t, deps.Transitive, semver)
+}
```
liftM: Can you add a third test case for tag combinations? For example:

`testModule.Options["tags"] = []string{"foo bar", "baz"}`

analyzers/golang/analyze.go
liftM: Does this work for combination build tags? For example, let's say someone sets:

```
options:
  tags:
    - foo bar
```

Then `buildTag` will be `"foo bar"`. Can we pass that to `Go.ListOne`, or do we need to `strings.Fields` to pass `"foo"` and `"bar"` as separate arguments?

zlav: Added a test for this but for your reference per `go help build`, `-tags` looks for `"tag1 tag2"`:
```
-tags 'tag list'
a space-separated list of build tags to consider satisfied during the
build. For more information about build tags, see the description of
build constraints in the documentation for the go/build package.```

zlav: Good point, I updated the top comment to clarify for anyone who observes the PR in the future.

PR #379, opened 2018-11-21
Add machine-readable output format to license report

liftM:
Adds a `--json` flag to `fossa report licenses` which outputs license data in JSON format.

### DIFF
--- a/cmd/fossa/cmd/report/report.go
+++ b/cmd/fossa/cmd/report/report.go
@@ -10,7 +10,7 @@ import (
 	"github.com/fossas/fossa-cli/module"
 )
 
-var (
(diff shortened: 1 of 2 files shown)


cmd/fossa/cmd/report/report.go
```diff
@@ -10,7 +10,7 @@ import (
 	"github.com/fossas/fossa-cli/module"
 )
 
-var (
+const (
 	Unknown = "show-unknown"
```
zlav: [nit] Can we move this into `licenses.go` and put it right next to `JSON`? Both are constants only used by `licenses.go`.

PR #361, opened 2018-10-18
feat(raw analyzer) do not upload tarball on --output analysis

microsoftly:
Addresses #348

PR #380, opened 2018-11-28
feat(buck) add cli support for the Buck package manager

zlav:
This PR adds support for the Buck package manager by taking advantage of the `buck audit` command and the existing archive upload functionality.

Important code changes:
`tar.go` has been modified to allow an archive upload for a list of specific files.
`buck.go` is a basic analyzer but will be expanded to handle more logic regarding `buck audit` and discovery.
`buckaudit.go`'s operational order:
 1. Find all dependency inputs and use the archive uploader to upload and retain the project's locator.
 1. Build the transitive dependency graph.
 1. Build the direct imports list.

Other Changes:
1. `tar.go` refactored and a upload flag was introduced to allow uploads to not happen on every call.
2. Package imports changed in a few files.


api/fossa/tar.go
```diff
@@ -136,11 +139,12 @@ func UploadTarball(dir string) (Locator, error) {
 		return Locator{}, err
 	}
 
-	return Locator{
+	locator := Locator{
```
liftM: This change seems like a no-op.

api/fossa/tar.go
```diff
@@ -244,3 +248,182 @@ func CreateTarball(dir string) (*os.File, []byte, error) {
 
 	return tmp, h.Sum(nil), nil
 }
+
+// UploadTarballFiles runs the logic inside of UploadTarball on a list of files.
```
liftM: This should probably be refactored.

buildtools/buckaudit/buckaudit.go
```diff
@@ -0,0 +1,149 @@
+package buckaudit
```
liftM: [nit] I'd also call this `package buck`.

buildtools/buckaudit/buckaudit.go
```diff
@@ -0,0 +1,149 @@
...
+)
+
+type buckOutput struct {
+	outputMapping map[string][]string
+}
+
+// Deps returns the dependencies of a Buck project using the buck audit command
+func Deps(name string) (graph.Deps, error) {
+	// Upload the dependencies.
+	locatorMap, err := uploadDeps(name)
```
liftM: It probably makes more sense to refactor these steps into:

1. Make all the `buck` calls you need, and then build a struct describing the dependency graph and its files.
2. Traverse the dependency graph, doing the uploading and FOSSA-graph-building/import-setting stuff for each dependency in one go.

This way you don't need to carry around a `locatorMap`, which seems error-prone.

analyzers/buck/buck.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// IsBuilt is not implemented.
+func (a *Analyzer) IsBuilt() (bool, error) {
+	return true, nil
+}
+
+// Analyze analyzers a buck package and its dependencies.
+func (a *Analyzer) Analyze() (graph.Deps, error) {
+	return buckaudit.Deps(a.Module.Name)
```
liftM: Why does this use `Name` and not `BuildTarget`?

analyzers/buck/buck.go
```diff
@@ -0,0 +1,90 @@
...
+func Discover(dir string, opts map[string]interface{}) ([]module.Module, error) {
+	var moduleList []module.Module
+
+	err := filepath.Walk(dir, func(filename string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.WithError(err).WithField("filename", filename).Debug("failed to access path")
+			return err
+		}
+
+		if !info.IsDir() && (info.Name() == ".buckconfig") {
```
liftM: This seems to imply that the `BuildTarget` is a cell, when it should be a Buck build target.

buildtools/buckaudit/buckaudit.go
```diff
@@ -0,0 +1,149 @@
...
+	depList, err := buckAudit("input", pack)
+	if err != nil {
+		return locatorMap, err
+	}
+
+	// Upload files and map the corresponding locator to the full dep path.
+	for dep, files := range depList.outputMapping {
+		locator, err := fossa.UploadTarballFiles(files, dep)
+		if err != nil {
+			log.Debugf("error uploading dependency: %s\n", dep)
```
liftM: This should probably break the analysis, or at least `log.Warn`.

buildtools/buckaudit/buckaudit.go
```diff
@@ -0,0 +1,149 @@
...
+			continue
+		}
+		locatorMap[dep] = locator
+	}
+
+	return locatorMap, nil
+}
+
+func depGraph(pack string, locatorMap map[string]fossa.Locator) (map[pkg.ID]pkg.Package, error) {
```
liftM: Here and elsewhere, `pack string` is not a very informative name.

buildtools/buckaudit/buckaudit.go
```diff
@@ -0,0 +1,149 @@
...
+		for _, transDep := range transDeps.outputMapping[dep] {
+			// Create the secondary level deps ID
+			transID := pkg.ID{
+				Type:     pkg.Raw,
+				Name:     locatorMap[transDep].Project,
+				Revision: locatorMap[transDep].Revision,
+			}
+
+			// Create the package
+			pack := pkg.Import{
```
liftM: Avoid shadowing variables, it causes all kinds of problems.

api/fossa/tar.go
zlav: You're right, I did this so I could print the locator before returning.

buildtools/buck/buck_test.go
```diff
@@ -0,0 +1,90 @@
...
+				return buck.BuckOutput{}, nil
+			}
+		},
+	}
+}
+
+func buckAuditFile(file string) buck.BuckOutput {
+	fileRead, _ := ioutil.ReadFile(file)
+	var output buck.BuckOutput
+	json.Unmarshal([]byte(fileRead), &output.OutputMapping)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

buildtools/buck/buck_test.go
```diff
@@ -0,0 +1,100 @@
...
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(deps.Transitive))
+	packageTwo := findPackage(deps.Transitive, "test-two")
+	assert.NotZero(t, packageTwo)
+	assert.Equal(t, 1, len(packageTwo.Imports))
+	assertImport(t, packageTwo.Imports, "test-three")
+}
+
+// Mock constructs a buck.Cmd using mock build tool output.
+func Mock(file string, err error) buck.Cmd {
```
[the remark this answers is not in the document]
zlav: They aren't, I initially planned to use them but never removed them.

buildtools/buck/buck_test.go
```diff
@@ -0,0 +1,100 @@
...
+	return output
+}
+
+func findPackage(packages map[pkg.ID]pkg.Package, name string) pkg.Package {
+	for id := range packages {
+		if id.Name == name {
+			return packages[id]
+		}
+	}
+	return pkg.Package{}
```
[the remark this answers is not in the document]
zlav: Yea this could return an error and I could assert on the error returning instead of the returned package being empty. I plan to make this function a part of a `graph_test` package and returning `(error, package)` does make more sense so I'm going to change it.

api/fossa/tar.go
```diff
@@ -244,3 +174,199 @@ func CreateTarball(dir string) (*os.File, []byte, error) {
...
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return tmp, h.Sum(nil), nil
+}
+
+// Upload the supplied tarball to the given endpoint.
+// Note: "name" should not have any "/"s to ensure core can parse it.
+func tarballUpload(name string, dependency, upload bool, tarball *os.File, hash []byte) (Locator, error) {
```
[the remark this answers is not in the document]
zlav: I agree that it feels a little messy and I'm struggling with finding a better way. I wish I didn't have to address this problem in this PR but its necessary for testing. 

I had a few reasons to use boolean flags.
1. This will solve a problem that PR #361 will introduce [here](https://github.com/fossas/fossa-cli/pull/361/files#diff-8e14521b11888ce1052f4754257ee035R85) that prevents raw modules from giving any information on `fossa analyze -o`.
1. I also wanted to stay away from creating a singleton the way that context and config do. 

I have thought a lot about this problem because we will be using the tarball uploader more. I'm starting to think that the solution is going to be to introduce another function, `tar.TarballRevision()`, which finds the hash of a directory so that a graph can be constructed. The files to be uploaded can then be stored and uploaded inside of `analyze.uploadAnalysis()` when the graph is uploaded.

buildtools/buck/buck_test.go
```diff
@@ -0,0 +1,102 @@
...
+	return output
+}
+
+func findPackage(packages map[pkg.ID]pkg.Package, name string) (pkg.Package, error) {
+	for id := range packages {
+		if id.Name == name {
+			return packages[id], nil
+		}
+	}
+	return pkg.Package{}, errors.New(fmt.Sprintf("Package %s not found", name))
```
golangcibot: should use fmt.Errorf(...) instead of errors.New(fmt.Sprintf(...)) (from `megacheck`)

analyzers/buck/buck.go
```diff
@@ -0,0 +1,166 @@
...
+		for _, target := range targets {
+			if len(target) > 0 {
+				moduleList = append(moduleList, newModule(target, target, dir))
+			}
+		}
+	}
+	return moduleList, nil
+}
+
+func buckCmd(cmd string, args ...string) (string, error) {
```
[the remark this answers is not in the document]
zlav: I did this to remove the complexity under the hood that is used to get readable results from `buck audit`. I will look into only keeping one `buckCmd` to handle the executable but my original concern is that it will make the code messier, although it introduced the need for `buckRoot` in the `Cmd` struct.

analyzers/buck/buck.go
```diff
@@ -0,0 +1,153 @@
...
+
+		aliases, err := file.GetSection("alias")
+		if err == nil && len(aliases.Keys()) > 0 {
+			for name, target := range aliases.KeysHash() {
+				moduleList = append(moduleList, newModule(name, target, dir))
+			}
+			return moduleList, nil
+		}
+
+		out, err := buckCommand("targets", "//")
```
raviagarwal7: this should be `//...`

PR #384, opened 2018-12-06
fix (ant discovery) ensure valid directories are discovered.

zlav:
Ant is not a traditional package manager and our approach should be to fail gracefully. Currently if we discover a `build.xml` file but not a `lib/` directory we error during analysis. This causes problems for customers who use Ant but do not keep `.jar` files inside of a `lib/` directory for use during the Ant build. You can find a reference to this type of use [here](https://ant.apache.org/manual/running.html) under the header Library Directories.

This PR addresses another issue that set a module's directory to the discovered filepath instead of the directory. This resulted in a fatal error in analysis where `build.xml/lib/` was scanned for `.jar`'s instead of `lib/`.

This PR does 4 things.
1. Check for the existence of a `lib/` directory before adding modules.
2. Set `module.dir` to the directory of a discovered `build.xml` file.
3. Test `Discover()`.
4. Fix import ordering around the cli.

### DIFF
--- a/analyzers/ant/ant_test.go
+++ b/analyzers/ant/ant_test.go
@@ -1,0 +1,21 @@
+package ant_test
+
+import (
+	"testing"
+
+	"github.com/fossas/fossa-cli/analyzers/ant"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDiscovery(t *testing.T) {
+	modules, err := ant.Discover("testdata", make(map[string]interface{}))
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(modules))
+	assert.Contains(t, modules, module.Module{
+		Name: "valid",
+		Type: pkg.Ant,
+		Dir:  "testdata/valid",
+	})
+}
--- a/analyzers/ant/ant_test.go
+++ b/analyzers/ant/ant_test.go
@@ -11,24 +11,19 @@ import (
 )
 
 func TestDiscoveryRoot(t *testing.T) {
-	modules, err := ant.Discover("testdata/valid", make(map[string]interface{}))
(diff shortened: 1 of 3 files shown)


analyzers/ant/ant_test.go
```diff
@@ -0,0 +1,34 @@
...
+	modules, err := ant.Discover("testdata", make(map[string]interface{}))
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(modules))
+	assert.Contains(t, modules, module.Module{
+		Name:        "valid",
+		Type:        pkg.Ant,
+		BuildTarget: "testdata/valid",
+		Dir:         "testdata/valid",
+	})
+}
```
liftM: You probably also want a test for making sure a project _without_ a `lib` does not add an extra module.

zlav: This is what the second test case is doing but I will make that more clear.

zlav:
@liftM The problem this PR solves is that when we saw a `build.xml` file we constructed an ant module, even if this file did not handle any dependencies. The reason I chose to name all of the fixtures `build.xml` is to ensure that an ant module can only be discovered if a `lib` directory exists. This tests against the cases where a `build.xml` file at the root or deep in the file tree get skipped if a `lib` directory doesn't exist.

analyzers/python/python.go
```diff
@@ -98,7 +99,11 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 				"path": filename,
 				"name": moduleName,
 			}).Debug("constructing Python module")
-			relPath, _ := filepath.Rel(dir, filename)
+			relPath, err := filepath.Rel(dir, filename)
+			if err != nil {
+				errors.Wrap(err, "Error discovering python modules")
```
golangcibot: Error return value of `errors.Wrap` is not checked (from `errcheck`)

PR #399, opened 2019-01-18
fix (Gradle analyzer) Fix gradle discovery for root projects and add a flag to scan all submodules

zlav:
This PR fixes an issue with how a module is constructed when a `build.gradle` file is discovered but no subprojects are present. 

This PR also adds support for scanning all of a projects submodules using the flag `--option allsubmodules:true`. This is particularly useful when running `fossa analyze` with the argument `gradle::` to signify a gradle root project.

Steps for merging multiple projects dependency maps:
- Create one common `map[string]graph.Deps` object to hold all projects configuration mappings.
- Retrieve a projects dependency graph and start by looping through all configurations specified.
- Retrieve the associated `graph.Deps` from the specified configuration in the returnable map.
- First, merge Transitive dependencies into this configuration map.
- Second, merge Direct dependencies by looping through the exisiting list and appending only those that are not present.

### DIFF
--- a/config/keys_test.go
+++ b/config/keys_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/fossas/fossa-cli/cmd/fossa/flags"
 	"github.com/fossas/fossa-cli/config"
+	"github.com/fossas/fossa-cli/pkg"
 )
 
 func TestModulesCommandLineOptions(t *testing.T) {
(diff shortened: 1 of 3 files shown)


config/keys_test.go
```diff
@@ -45,6 +46,36 @@ func TestModulesCommandLineOptions(t *testing.T) {
 	assert.Equal(t, true, modules[2].Options["allow-unresolved"])
 }
 
+func TestModulesWithAnArgumentAndOptions(t *testing.T) {
+	// Set the expected flags.
+	flagSet := flag.NewFlagSet("test", 0)
+	flags.ConfigF.Apply(flagSet)
+	flags.OptionF.Apply(flagSet)
+	flagSet.Parse([]string{"gradle:argumentmodule"})
```
golangcibot: Error return value of `flagSet.Parse` is not checked (from `errcheck`)

PR #411, opened 2019-02-08
fix (gradle configurations) add default configurations and change how targets are handled

zlav:
1. Add `implementationDependenciesMetadata` after this configuration was seen in multiple customer's ouput. Documentation on this configuration is hard to find on this feature on Gradle's end.
2. Allow buildtargets formatted like `<project>:<subproject>` to be handled correctly. Currently `<subproject>` is turned into the configuration.
3. Add an option `all-configurations` to allow users to catch all configurations by default.
4. Add backward compatibility to support older users' config files.

### DIFF
--- a/analyzers/gradle/gradle.go
+++ b/analyzers/gradle/gradle.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 
 	"github.com/fossas/fossa-cli/buildtools/gradle"
+	"github.com/fossas/fossa-cli/config"
 	"github.com/fossas/fossa-cli/exec"
 	"github.com/fossas/fossa-cli/files"
 	"github.com/fossas/fossa-cli/graph"
@@ -161,6 +162,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	var configurations []string
 	var depsByConfig map[string]graph.Deps
 	var err error
+	targets := strings.Split(a.Module.BuildTarget, ":")
 
 	if a.Options.AllSubmodules {
 		submodules, err := g.Projects()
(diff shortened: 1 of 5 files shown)


analyzers/gradle/gradle.go
```diff
@@ -31,10 +32,11 @@ type Analyzer struct {
...
 type Options struct {
-	Cmd           string `mapstructure:"cmd"`
-	Task          string `mapstructure:"task"`
-	Online        bool   `mapstructure:"online"`
-	AllSubmodules bool   `mapstructure:"allsubmodules"`
+	Cmd               string `mapstructure:"cmd"`
+	Task              string `mapstructure:"task"`
+	Online            bool   `mapstructure:"online"`
+	AllSubmodules     bool   `mapstructure:"allsubmodules"`
+	AllConfigurations bool   `mapstructure:"all-configurations"`
```
liftM: [nit] I'd like these options to be consistent: either `all-submodules` + `all-configurations` or `allsubmodules` + `allconfigurations`.

For the sake of backwards compatibility, it's probably easier to make it the latter.

analyzers/gradle/gradle.go
```diff
@@ -179,20 +181,26 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	} else {
 		if a.Options.Project != "" {
 			project = a.Options.Project
-		} else {
+		} else if config.Version() < 2 {
```
liftM: I think I'd be more comfortable seeing this factored out into a function. You'll repeat more code this way, but it'll be easier to reason about since version checks won't be peppered throughout the code.

For example, I'd prefer:

```
func parseModuleV1() {
	// logic here
}

func parseModuleV2() {
	// logic here, a lot of this will look very similar
}

func (a *Analyzer) Analyze() {
	// ...
	switch config.Version() {
		case 2: parseModuleV2()
		case 1: parseModuleV1()
	}
	//...
}
```

zlav: I agree with this approach. While I think this will add a significant amount of lines it will reduce the amount of time to understand whats happening.

analyzers/gradle/gradle.go
zlav: Agreed... this was a mistake I made with `allmodules`.

PR #418, opened 2019-03-12
Added the NoFail flag to fossa test

xinding33:


buildtools/pip/bindata/bindata.go
```diff
@@ -1,8 +1,6 @@
-// Code generated by go-bindata.
+// Code generated by go-bindata. (@generated) DO NOT EDIT.
```
zlav: These edits are probably a remnant from a previous `make` on a different branch. Rebasing onto master should solve your problem, although removing it from the PR accomplishes the same thing.

cmd/fossa/cmd/test/test.go
```diff
@@ -19,6 +19,7 @@ import (
 )
 
 var Timeout = "timeout"
+var NoFail = "nofail"
```
zlav: [nit] I think we can do better than `nofail`, I like to make flags explicitly related to what the user is trying to accomplish. Something like `suppress`, `suppress-issues`, or `suppress-failure` sounds a little more in line with what the purpose of the flag is.

cmd/fossa/cmd/test/test.go
```diff
@@ -28,6 +29,7 @@ var Cmd = cli.Command{
 	Action: Run,
 	Flags: flags.WithGlobalFlags(flags.WithAPIFlags(([]cli.Flag{
 		cli.IntFlag{Name: Timeout, Value: 10 * 60, Usage: "duration to wait for build completion (in seconds)"},
+		cli.BoolFlag{Name: NoFail, Usage: "don't exit with error if issues are found"},
 	}))),
 }
```
zlav: Github won't let me comment on the lines directly, but we should consider if we want to be suppressing any of the errors from fossa not failing on line 41 and 46. I think I agree with ignoring these `Fatal` exits like you have, which pushes me towards making the flag more specific.

PR #419, opened 2019-03-12
fix (report dependencies) Change report dependencies to track Fossa.com results

zlav:
This PR modifies the logic inside of `dependencies.go` to retrieve scan information from Fossa.com instead of reanalyzing the project itself. This ensures that the dependency information returned contains information about the dependencies themselves and not only 

Other changes:
1. Sort the outputted dependencies for `dependencies.go` as well as `licenses.go`.
2. Modify the default `fossa report dependencies` output to be more readable.
3. Add a `--json` flag to `fossa report dependencies` which returns all information about all dependencies.
4. Modify the manual.


cmd/fossa/cmd/report/dependencies.go
```diff
@@ -1,45 +1,95 @@
...
 
-	pkgs := make([]pkg.Package, 0)
-	for _, module := range analyzed {
-		for _, pkg := range module.Deps {
-			pkgs = append(pkgs, pkg)
-		}
+	defer display.ClearProgress()
+	display.InProgress(fmt.Sprint("Fetching Dependency Information"))
+
+	revs := make([]fossa.Revision, 0)
```
golangcibot: ineffectual assignment to `revs` (from `ineffassign`)

PR #426, opened 2019-03-26
fix (ruby tests) fix flaky ruby integration tests.

zlav:
Get the ruby tests back into a reliable state. The suspected issue is that the rails repo is too large (179mb) and breaking on circleci after attempting to clone and run bundle.


zlav:
After rerunning the integration test workflow I am confident that this greatly improves the ruby integration tests. After 5 rebuilds I have not seen any further failures. Due to the fact that the failure is related to the size of the repository however, I am not confident that this will be a permanent solution.

PR #442, opened 2019-04-13
fix (api error) return api error messages with bad requests

zlav:
Previously the error returned by the cli would be confusing and debug logs would be needed in order to determine what happened. This PR adds support for including actual API error messages.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -62,9 +62,7 @@ func jsonAPIRequest(method string, endpoint *url.URL, APIKey string, body []byte
 	}
 	err = json.Unmarshal(res, v)
 	if err != nil {
-		apiError := errorResponse{}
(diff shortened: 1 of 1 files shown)


api/api.go
```diff
@@ -58,7 +62,9 @@ func jsonAPIRequest(method string, endpoint *url.URL, APIKey string, body []byte
 	}
 	err = json.Unmarshal(res, v)
 	if err != nil {
-		return code, errors.Wrap(err, "could not unmarshal JSON API response")
+		apiError := errorResponse{}
+		json.Unmarshal(res, &apiError)
```
golangcibot: Error return value of `json.Unmarshal` is not checked (from `errcheck`)

PR #444, opened 2019-04-16
Haskell project support via cabal-install and stack

cnr:
This adds project structure scanning for projects using cabal-install and stack

### Init

- It looks for projects (via `cabal.project` and `stack.yaml`) as well as independent cabal directories (`*.cabal`)
- cabal files with a parent project are considered sub-projects. We may want to parse cabal.project and stack.yaml instead to determine when to prune

### Analyze (cabal-install)

- Find (or generate: `cabal new-build --dry-run`) the solver plan and parse the dependency graph from there

### Analyze (stack)

- Uses `stack ls dependencies {--depth 1}` to determine the list of direct and transitive dependencies
- Edges of the dependency graph are hydrated with the output from `stack exec -- ghc-pkg dot`

### Locators

Resulting locators are of the form `hackage+base$4.12.0.0`

### To do
- [x] Move stack- and cabal-specific code to `buildtools/`
- [x] Unit tests
- [ ] Integration tests?
- [x] General code clean-up
- [ ] Add debug logging in line with the other analyzers
- [x] Hydrate dependency graph for stack (an initial attempt was made with a bunch of `stack exec -- ghc-pkg field {package} depends --simple-output`, but it was far too slow. Going to use `ghc-pkg dot` instead


analyzers/haskell/analyzer/stack.go
```diff
@@ -0,0 +1,90 @@
...
+	localDepsStdout, _, err := exec.Run(exec.Cmd{
+		Name: "stack",
+		Argv: []string{"ls", "dependencies", "--depth", "1"},
+	})
+
+	if err != nil {
+		return graph.Deps{}, err
+	}
+
+	allDepsStdout, _, err := exec.Run(exec.Cmd{
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

analyzers/haskell/types.go
```diff
@@ -0,0 +1,12 @@
+package haskell
+
+type Options struct {
+	Strategy Strategy `mapstructure:"strategy"`
+}
+
+type Strategy string
+const (
+	CabalInstall Strategy = "cabal-install"
```
golangcibot: SA9004: only the first constant in this group has an explicit type (from `staticcheck`)

analyzers/analyzer.go
```diff
@@ -3,6 +3,7 @@ package analyzers
 
 import (
 	"errors"
+	"github.com/fossas/fossa-cli/analyzers/haskell"
```
dchenk: if your IDE has an imports grouping feature, I'd suggest turning it on. Otherwise you may want to use [`goimports`](https://godoc.org/golang.org/x/tools/cmd/goimports) instead. Helps reduce diff churn. E.g., https://github.com/fossas/fossa-cli/blob/master/api/fossa/normalize.go#L3

analyzers/haskell/analyzer.go
```diff
@@ -0,0 +1,54 @@
+
+package haskell
+
+import (
+	"github.com/fossas/fossa-cli/analyzers/haskell/analyzer"
+	"github.com/fossas/fossa-cli/errors"
+	"github.com/fossas/fossa-cli/graph"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/mitchellh/mapstructure"
```
dchenk: same idea here with stdlib vs external vs fossa-cli packages grouping

analyzers/haskell/analyzer/stack.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	depth := int(1)
+	stackImmediateDeps, err := stack.GetStackDependencies(m.Dir, &depth)
+
+	if err != nil {
+		return graph.Deps{}, err
+	}
+
+	return AnalyzeStackPure(stackAllDeps, stackImmediateDeps, ghcPkgDeps), nil
```
zlav: Is this logic we could reasonably contain in the build tool? I would prefer doing it that way so that the analyzer really only thinks about strategy and discovery.

buildtools/stack/stack_test.go
```diff
@@ -0,0 +1,101 @@
+package stack
```
zlav: Ideally we should work towards creating external tests under a package like `stack_test`. You'll see this pattern elsewhere in the codebase. The reason we want to do that is to ensure the focus of the test is on the external contract of the public methods. We shouldn't really care how the internals behave.

analyzers/haskell/analyzer/stack.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"github.com/fossas/fossa-cli/buildtools/stack"
+	"github.com/fossas/fossa-cli/graph"
+	"github.com/fossas/fossa-cli/module"
+	"github.com/fossas/fossa-cli/pkg"
+)
+
+func AnalyzeStack(m module.Module) (graph.Deps, error) {
+	stackAllDeps, err := stack.GetStackDependencies(m.Dir, nil)
```
zlav: I need to set up a Haskell environment to try this out, but my initial question is, is it really necessary to shell out to `stack` here three times? I trust your judgment so I won't block on this, but I am curious to know if there is a simpler way in the future.

cnr: There's a feature landing soon in stack that'll output the dependency graph as a proper tree; in the meantime the commands available don't give the context we need.

`stack ls dependencies (--depth N)` outputs a raw list of dependencies without any information about relationships
`ghc-pkg dot` gives us the information about relationships between packages

analyzers/haskell/analyzer/stack.go
cnr: I can move this function to the build tool if that's a better place. I feel strongly that the "pure" code (deterministic based on input, and cannot perform effects -- including failure) should be separate from the effectful code when possible. It makes the `AnalyzeStackPure` function trivial to test and reason about

analyzers/haskell/discover.go
```diff
@@ -0,0 +1,81 @@
...
+
+		return nil
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("could not find Haskell package manifests: %s", err.Error())
+	}
+
+	// Prune .cabal files where parent projects are present
+	for cabalFilePath := range cabalFiles {
```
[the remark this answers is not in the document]
cnr: `filepath.SkipDir` should do it. I wasn't aware of it when I wrote this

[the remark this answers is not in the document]
cnr: errh -- `filepath.SkipDir` wouldn't cover the cases where, e.g., we have "aaaa.cabal" and "cabal.project" in the same directory, because Walk does it in alphabetical order

analyzers/haskell/types.go
```diff
@@ -0,0 +1,12 @@
+package haskell
+
+type Options struct {
+	Strategy Strategy `mapstructure:"strategy"`
+}
+
+type Strategy string
+const (
```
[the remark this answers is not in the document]
cnr: My instinct is to learn toward type safety when possible -- and while this doesn't quite accomplish that, it's better than writing out "cabal-install" or "stack" in several different places. I'm curious if there's a better way to do this without going overboard (`type Strategy struct { Name: string }`? I use that approach in `buildtools/stack/stack.go` with `Canonical`)

buildtools/cabal/cabal.go
```diff
@@ -0,0 +1,79 @@
...
+// ----- Command invocation
+
+func GetSolverPlan(dir string) (Plan, error) {
+	cabalPlanPath := filepath.Join(dir, PlanRelPath)
+
+	// If plan.json doesn't exist, generate it
+	if exists, _ := files.Exists(cabalPlanPath); !exists {
+		_, _, err := exec.Run(exec.Cmd{
+			Name: "cabal",
+			Argv: []string{"new-build", "--dry-run"},
```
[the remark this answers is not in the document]
cnr: This command merely invokes the dependency solver and produces the lockfile. It doesn't perform building of source files. It does seem like a case that wouldn't come up frequently (or at all), so I'm happy to remove it

buildtools/cabal/cabal.go
```diff
@@ -0,0 +1,79 @@
...
+			Dir: dir,
+		})
+
+		if err != nil {
+			return Plan{}, err
+		}
+	}
+
+	if exists, _ := files.Exists(cabalPlanPath); !exists {
+		// TODO: fallback to another strategy?
```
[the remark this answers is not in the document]
cnr: cabal files have a really odd, whitespace-sensitive format that might be difficult to parse. I didn't know if it would be worth the time investment

buildtools/stack/stack_test.go
cnr: Whoops -- didn't know that convention. I'll fix this

analyzers/analyzer.go
cnr: Is it possible for us to add an auto-formatting tool that does this for us?

analyzers/haskell/analyzer/cabal.go
```diff
@@ -0,0 +1,76 @@
...
+
+	var dependencyGraph = make(map[pkg.ID]pkg.Package)
+	var directDependencies []pkg.Import
+
+	// Build the entire dependency graph, keeping track of our direct
+	// dependencies
+	for _, p := range plan.InstallPlans {
+		var builtPackage = InstallPlanToPackage(installPlans, p)
+
+		if p.Type == "configured" && p.Style == "local" {
```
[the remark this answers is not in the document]
cnr: I'll add better documentation in `buildtools/cabal/cabal.go`, and/or turn this into a function on installplan

cnr:
Rebased and made CR changes.

Something to consider before merging: `stack` recently published its 2.0 version, which has a tree-based output for deps. The output from the new `stack ls dependencies --tree` seems more comprehensive than my current approach.

I think we should continue using the current approach until the 2.x versions are more widely adopted, but it might be worth integrating now for better results(?)

cnr:
Something that came up while testing: the package itself appears as a build dependency when the package is library+executable. Need to find a way to remove our own package(s) from the results

zlav:
@cnr I think we should keep this PR as is and then implement the updated `stack` approach in another PR. Landing support should be our priority and then depending on the accuracy of this work we can prioritize the `stack` 2.0 updates.

Aman0509:
@meghfossa - Could you please tell me how I can generate .fossa.yml file as 'init' subcommand has been deprecated in fossa-cli v3.x.x?

meghfossa:
@Aman0509 you don't need to generate .fossa.yml anymore, you can simply perform: `fossa analyze`. 

If you want to perform analysis for specific targets, you can still manually create `.fossa.yml` using editor. 

Example walkthrough with v3: https://github.com/fossas/fossa-cli/blob/master/docs/walkthroughs/integrating.md
Migration guide: https://github.com/fossas/fossa-cli/blob/master/docs/differences-from-v1.md#how-to-upgrade-to-fossa-3x
Spec for .fossa.yml with v3: https://github.com/fossas/fossa-cli/blob/master/docs/references/files/fossa-yml.md

PR #449, opened 2019-04-24
Improve discovery of Maven modules and dependencies

dchenk:
Two important things had previously been overlooked:
1. The CLI did not recurse into directories below that of any discovered `pom.xml` file, yet some codebases have Maven projects within one repository which are not _Maven modules_ of any module or project higher in the tree. So now we will recurse down but keep track of the modules that we have already looked at.
2. The `BuildTarget` field of FOSSA modules wasn't as useful as it could've been because it didn't consistently hold just the path to the corresponding Maven module (or "project"). We use it with the `--projects` flag when running commands such as `dependency:list` or `dependency:tree`, which should now take the relative path to a module or a module's POM file. _Importantly_, we no longer assume that only a "pom.xml" file can be the manifest for a Maven project, since a `<module>` element can point directly to such an XML file with another name.


buildtools/maven/maven.go
```diff
@@ -59,28 +60,73 @@ func (m *Maven) Compile(dir string) error {
...
 	if err != nil {
-		return nil, errors.Wrap(err, "could not read POM file")
+		return nil, errors.Wrap(err, "could not get CWD")
+	}
+
+	dir := path
+	pomFile := path
+	if strings.HasSuffix(path, ".xml") {
+		// We have the manifest file but still need its directory path.
+		dir, err = filepath.Rel(cwd, filepath.Dir(path))
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

analyzers/maven/maven.go
```diff
@@ -62,6 +63,8 @@ func New(m module.Module) (*Analyzer, error) {
 func Discover(dir string, options map[string]interface{}) ([]module.Module, error) {
 	log.WithField("dir", dir).Debug("discovering modules")
 	var modules []module.Module
+	checked := make(map[string]bool, 7)
```
anuccio1: Why 7?

dchenk: prime number

buildtools/maven/maven.go
```diff
@@ -59,27 +60,70 @@ func (m *Maven) Compile(dir string) error {
...
+	checked[pomFile] = true
+
+	var pom Manifest
+	if err := files.ReadXML(&pom, pomFile); err != nil {
+		return nil, errors.Wrapf(err, "could not read POM file %q", pomFile)
+	}
+
+	groupId := pom.GroupID
+	if groupId == "" {
+		groupId = pom.Parent.GroupID
```
golangcibot: ineffectual assignment to `groupId` (from `ineffassign`)

analyzers/maven/maven.go
anuccio1: What do you mean?

buildtools/maven/maven.go
```diff
@@ -59,27 +60,75 @@ func (m *Maven) Compile(dir string) error {
...
+
+	GroupId    string
+	ArtifactId string
+}
+
+// Modules returns a list specifying the Maven module at path, which may name a file or directory, and all the
+// Maven modules nested below it. The Target field of each MvnModule is set to the directory or manifest file
+// describing the module. The visited manifest files are listed in the checked map.
+func Modules(path string, checked map[string]bool) ([]MvnModule, error) {
+	cwd, err := os.Getwd()
```
cnr: I think we want to keep the base directory as an argument to `Modules`, rather than getting cwd

dchenk: I was considering doing it that way too for performance. On the other hand it makes `Modules` more self-contained. If you also think pass it in as an argument, I'd be happy to refactor that

buildtools/maven/maven.go
```diff
@@ -59,27 +60,75 @@ func (m *Maven) Compile(dir string) error {
...
-		return nil, errors.Wrap(err, "could not read POM file")
+		return nil, errors.Wrap(err, "could not get CWD")
 	}
 
-	var modules []string
-	for _, module := range pom.Modules {
-		children, err := Modules(filepath.Join(dir, module))
+	dir := path
+	pomFile := path
+	if strings.HasSuffix(path, ".xml") {
```
cnr: When does this happen?

analyzers/maven/maven.go
dchenk: to those who are curious (Alex and I talked on Slack)... The `map` type in Go is implemented as a hash map. Although the language spec doesn't require that the "size hint" be used as the number of buckets allocated, it's common for performance reasons, reducing chances of collision, to use a prime number of buckets in a map.

buildtools/maven/maven.go
```diff
@@ -59,27 +60,75 @@ func (m *Maven) Compile(dir string) error {
...
+	if err := files.ReadXML(&pom, pomFile); err != nil {
+		return nil, errors.Wrapf(err, "could not read POM file %q", pomFile)
+	}
+
+	groupId := pom.GroupID
+	if groupId == "" {
+		groupId = pom.Parent.GroupID
+	}
+
+	modules := make([]MvnModule, 1, 1+len(pom.Modules))
```
cnr: [nit] could do an inline initialization with something like `modules := []MvnModule{...}`. Also: no need to specify size/capacity

analyzers/maven/maven.go
```diff
@@ -141,11 +141,16 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
 		if err != nil {
+			// Because this was a custom shell command, we do not fall back to any other strategies.
 			return graph.Deps{}, err
 		}
 		imports, deps, err = maven.ParseDependencyTree(output)
 	}
 	if err != nil {
+		log.Warn("Could not use Maven to determine dependencies. Falling back to parse pom.xml file.")
+
+		//pomData, err := maven.ReadPOM(a.Module)
```
cnr: should kill this comment

buildtools/maven/maven.go
dchenk: Two ways:
1. A `module` can be specified directly in the `.fossa.yml` to have the `target: my-project.xml` or whatever.
2. In the `<modules>` of a POM file you can have paths to an actual such manifest file, not necessarily a directory.

My comment at the top of PR thread discusses this.

buildtools/maven/maven.go
dchenk: no need indeed, but with just one element in a slice literal the slice will be allocated just for one element. The generated code effectively would have `make([]MvnModule, 1, 1)` and then an insertion of the first element. As soon as you add anything, this would mean another allocation and a copy of the first element.

analyzers/maven/maven.go
dchenk: fixed

analyzers/maven/maven.go
cnr: It's not the number of buckets -- it's a capacity hint; in this case, it'd actually make zero buckets 😱 . It starts making buckets at 8

buildtools/maven/maven.go
cnr: Oh okay -- didn't know that <module> could point to an xml file. Rather than using `filepath.Rel`, then, can we use `filepath.Dir`? prevents us from having to pass in the base directory altogether 🙂

buildtools/maven/maven.go
cnr: Makes sense 👍

buildtools/maven/maven.go
dchenk: sorry I'm not following. We still want `path` in `.fossa.yml` to be the relative directory of the module

dchenk:
found something that needs to be fixed now, with the relative dir of modules

analyzers/maven/maven.go
dchenk: @cnr good catch if that's accurate. In any case, I agree with not introducing a speedbump to someone scanning the code for the first time—someone might wonder why it has to be 7. Remove that now :)

buildtools/maven/maven.go
cnr: I think `path` here is already relative to the basedir -- so this can be changed to `dir = filepath.Dir(path)`. See the `filepath.Join` on 125, which applies to both modules pointing to directories and xml files

(If that's not the case, then we'd need a `filepath.Rel(cwd, ...)` in the `else` branch as well)

dchenk: I see what you're saying. We both understand that here `dir` needs to be a path that is valid anywhere in the recursion from the fossa project root (`baseDir`). My code was just being very defensive (the `Dir` field in `MvnModule` is document as being the relative path from the root of the FOSSA project, so what if `Modules` is called with an absolute path for `dir`?)—but perhaps overly defensive as you point out.

analyzers/maven/maven.go
```diff
@@ -141,16 +141,26 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
+		log.Warnf(
+			"Could not use Maven to determine dependencies for %q. Falling back to use manifest file.",
+			a.Module.Name)
+		mvnError := err
+		pom, err := maven.ResolveManifestFromBuildTarget(a.Module.BuildTarget)
+		if err != nil {
+			return graph.Deps{},
+				errors.Wrapf(err, "could not identify dependencies; original error: %v", mvnError)
+		}
+		a.Maven.AddCachedManifest(a.Module.BuildTarget, pom)
```
cnr: I think manifest caching should be transparent to the consumer -- we shouldn't have to manually add the cached manifest here. Can we do it in `ResolveManifestFromBuildTarget` instead?

buildtools/maven/maven.go
```diff
@@ -59,53 +72,154 @@ func (m *Maven) Compile(dir string) error {
 	return err
 }
 
-func Modules(dir string) ([]string, error) {
+// CachedManifest returns the manifest from the internal cache, or nil if it is not set.
+func (m *Maven) CachedManifest(target string) *Manifest {
+	if m.manifestCache == nil {
```
cnr: [nit] This nil check isn't required, because indexing into a nil map will already return nil here

dchenk: thanks—I'm always forgetting whether it's just writing to or reading/writing to nil maps that's not allowed

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,27 @@
+package maven_test
+
+import (
+	"testing"
+
+	"github.com/fossas/fossa-cli/analyzers/maven"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestDiscover(t *testing.T) {
```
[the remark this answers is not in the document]
dchenk: It should work deterministically. I guess it makes it so that if func `Discover` is changed then this test may need to be updated. Is that your concern?

analyzers/maven/maven.go
dchenk: Indeed every invocation of `ResolveManifestFromBuildTarget` is followed by saving the manifest object to the cache. I too thought about bundling the cacheing into `ResolveManifestFromBuildTarget` but then leaned more toward making `ResolveManifestFromBuildTarget` simpler. With your suggestion I'll go ahead and revise. :)

buildtools/maven/maven.go
```diff
@@ -59,53 +72,136 @@ func (m *Maven) Compile(dir string) error {
...
-	err := files.ReadXML(&pom, filepath.Join(dir, "pom.xml"))
-	if err != nil {
-		return nil, errors.Wrap(err, "could not read POM file")
+	if err := files.ReadXML(&pom, pomFile); err != nil {
+		return nil, err
+	}
+	if m.manifestCache == nil {
+		m.manifestCache = make(map[string]*Manifest)
+	}
+	m.manifestCache[buildTarget] = &pom
```
zlav: I'm not convinced the manifest cache is the best way to go about this, it doesn't appear to be working on my end because the manifest cache is never accessed before we return the dependencies. We should discuss whether or not this is the way to proceed, I think this might be optimizing a little early.

analyzers/maven/maven_test.go
```diff
@@ -0,0 +1,27 @@
+package maven_test
```
zlav: We should add a test here that ensures we can accurately get the correct graph from nested pom files.

analyzers/maven/maven.go
```diff
@@ -62,6 +63,8 @@ func New(m module.Module) (*Analyzer, error) {
 func Discover(dir string, options map[string]interface{}) ([]module.Module, error) {
```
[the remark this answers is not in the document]
dchenk: I think you need to refresh your browser tab or something. There's a test already: https://github.com/fossas/fossa-cli/pull/449/files#diff-b9f26dd681f3f09b56b8f03a3f8b2af2R10

analyzers/maven/maven.go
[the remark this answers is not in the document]
dchenk: fixed

buildtools/maven/maven.go
dchenk: please try this out with the latest commits of this branch.

As for whether there should even be a `manifestCache`, the thing is that sometimes we'd need to access the data in the XML file at two separate times though closely in time. These manifest files can be very large, and re-parsing the XML can be expensive. What do you think? If not worth it let me know.

analyzers/maven/maven.go
```diff
@@ -1,6 +1,7 @@
 // Package maven implements Maven analysis.
 //
-// A `BuildTarget` for Maven is the Maven project name.
+// A `BuildTarget` for Maven is either a path to a directory containing a POM manifest or a path to such a
```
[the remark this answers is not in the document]
dchenk: I should re-phrase that; it's either:
1. a path to a directory in which there is a "pom.xml" file; or
2. a path to a "pom.xml" file

Actually I'll need to update that to say that it can also be a `groupId:artifactId` project identifier.

analyzers/maven/maven.go
```diff
@@ -77,24 +80,21 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
...
-					BuildTarget: m,
-					Dir:         path,
+					BuildTarget: m.Target,
+					Dir:         dir,
 				})
 			}
-			log.WithField("path", path).Debug("skipping")
-			// Don't continue recursing, because anything else is probably a
-			// subproject.
-			return filepath.SkipDir
```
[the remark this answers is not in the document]
dchenk: To those who weren't on our call: Removing `SkipDir` doesn't make looking at `<modules>` redundant because of projects (such as [this](https://github.com/google/closure-compiler)) which have a generic `pom.xml` at root but split out their logic into other `.xml` files that are named something other than "pom.xml". The root `pom.xml` file would have, for example: `<module>pom-gwt.xml</module>`. So if we don't look at these `<module>` elements we won't know what the project's depedencies are.

buildtools/maven/maven.go
```diff
@@ -32,13 +34,24 @@ type Parent struct {
...
-	Version string
-	Failed  bool
+	Name string
+
+	GroupId    string `xml:"groupId"`
+	ArtifactId string `xml:"artifactId"`
+	Version    string `xml:"version"`
+
+	// Scope is where the dependency is used, such as "test" or "runtime".
+	Scope string `xml:"scope"`
```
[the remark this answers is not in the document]
dchenk: At first I was trying to not also do this in this PR because of PR #240 which addresses this question. But there's enough that I needed to change that I think that PR may not turn out to be very helpful. What do you think, do it here anyway?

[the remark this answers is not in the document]
anuccio1: FWIW, srclib-java applies scope to each dependency it returns, and by default, the fossa UI filters to ‘compile’ and ‘runtime’ deps. You can turn other scopes on in the project settings as well. Might be worth a discussion before merging this

buildtools/maven/maven.go
```diff
@@ -59,53 +72,141 @@ func (m *Maven) Compile(dir string) error {
...
+	var pomFile string
+	if !strings.HasSuffix(buildTarget, ".xml") {
+		pomFile = filepath.Join(buildTarget, "pom.xml")
+	} else {
+		pomFile = buildTarget
+	}
+	if ok, _ := files.Exists(pomFile); !ok {
+		if strings.Count(buildTarget, ":") == 1 {
+			// This is likely a module ID.
+			return nil, fmt.Errorf("build target %q appears to be a module identifier", buildTarget)
```
anuccio1: To a user, what does this error exactly mean? I’m not sure I would know what course of action to take after receiving an error like this

buildtools/maven/maven.go
```diff
@@ -59,53 +72,141 @@ func (m *Maven) Compile(dir string) error {
...
+
+// Modules returns a list specifying the Maven module at path, which may name a file or directory, and all the
+// Maven modules nested below it. The Target field of each MvnModule is set to the directory or manifest file
+// describing the module. The visited manifest files are listed in the checked map.
+func Modules(path string, baseDir string, checked map[string]bool) ([]MvnModule, error) {
+	dir := path
+	pomFile := path
+	if strings.HasSuffix(path, ".xml") {
+		// We have the manifest file but still need its directory path.
+		dir = filepath.Dir(path)
```
anuccio1: Probably a big edge case, but what if the user defines a  custom Pom manifest file without an xml suffix? Is there maybe a more reliable check we could do here?

buildtools/maven/maven.go
zlav: Ah, so I see that this does update `graph.Deps.Direct`, we also need to update `graph.Deps.Transitive` in order to make sure that that core picks up the correct dependency graph.

analyzers/maven/maven.go
```diff
@@ -130,64 +132,48 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
-			imports = append(imports, pkg.Import{
-				Target: child.Version,
-				Resolved: pkg.ID{
-					Type:     pkg.Maven,
-					Name:     child.Name,
-					Revision: child.Version,
-				},
+		imports, deps = pom.ToGraphDeps()
+	case "maven-tree":
+		imports, deps, err = a.Maven.DependencyTree(a.Module.Dir, a.Module.BuildTarget)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

buildtools/maven/maven.go
dchenk: fixed

buildtools/maven/maven.go
dchenk: fixed

buildtools/maven/maven.go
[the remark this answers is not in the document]
dchenk: Discussed this with @zlav —we decided that at least in this PR we'll focus on just discovery of modules and simply parsing the pom.xml files, and next we'll check out [#240](https://github.com/fossas/fossa-cli/pull/240) to see how best to add this feature

buildtools/maven/maven.go
```diff
@@ -32,9 +37,82 @@ type Parent struct {
...
+				Revision: dep.Version,
+			},
+		})
+	}
+	return i
+}
+
+// ToGraphDeps returns simply the list of dependencies listed within the manifest file.
+func (m *Manifest) ToGraphDeps() graph.Deps {
+	g := graph.Deps{
```
zlav: [nit] We should avoid using one letter variable names. It forces us to find the initialization to understand what it is. This also goes for the method below.

buildtools/maven/maven.go
```diff
@@ -59,56 +137,153 @@ func (m *Maven) Compile(dir string) error {
...
+	cmd.Argv = append(goalArgs, "--projects", buildTarget)
+	output, _, err := exec.Run(cmd)
+	if err != nil {
+		// Now we try to identify the groupId:artifactId identifier for the module and specify the path to
+		// the manifest file directly.
+		pom, err2 := m.ResolveManifestFromBuildTarget(buildTarget)
+		if err2 != nil {
+			// Using buildTarget as a module ID or as a path to a manifest did not work.
+			// Return just the error from running the mvn goal the first time.
+			return "", errors.Wrap(err, "could not use Maven to list dependencies")
```
zlav: Should this fallback instead of return an error?

buildtools/maven/maven.go
```diff
@@ -156,4 +331,8 @@ func ParseDependencyTree(stdin string) ([]Dependency, map[Dependency][]Dependenc
 
 		return level, Dependency{Name: name, Version: revision, Failed: failed}, nil
 	})
+	if err != nil {
+		return graph.Deps{}, err
+	}
+	return graph.Deps{Direct: depsList(imports).toImports(), Transitive: depsMap(deps).toPkgGraph()}, nil
```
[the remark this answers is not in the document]
zlav: I'm not a huge fan of type conversions in order to support methods, although this is another area where I don't have a lot of opinions. I think this would be much more clear if we were to make this:
```
dependencyListToImports(dependencyList)

func dependencyListToImports(imports []Dependency)
```
This still enforces the type of `[]Dependency` but allows us to remove both the type conversion and the type definition. Why is your instinct here to make this conversion and add a method?

analyzers/maven/maven_test.go
dchenk: @zlav  this still valid with the migration of the logic to `buildtools/maven`?

analyzers/maven/maven.go
```diff
@@ -130,64 +132,39 @@ func (a *Analyzer) IsBuilt() (bool, error) {
...
+			log.Warnf(
+				"Could not use Maven to determine dependencies for %q. Falling back to use manifest file.",
+				a.Module.Name)
+			mvnError := err
+			pom, err := a.Maven.ResolveManifestFromBuildTarget(a.Module.BuildTarget)
+			if err != nil {
+				return graph.Deps{},
+					errors.Wrapf(err, "could not identify dependencies; original error: %v", mvnError)
+			}
+			return pom.ToGraphDeps(), nil
```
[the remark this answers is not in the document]
dchenk: we need something like `ResolveManifestFromBuildTarget` to get just the contents of a POM file because we sometimes need things like the `groupId`, `artifactId`, and `parent` from it

buildtools/maven/maven.go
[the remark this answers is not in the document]
dchenk: agreed—this is something I too was going to revise to be as you suggest after thinking about it some more

buildtools/maven/maven.go
dchenk: there's nothing really to fall back to, since this is already the fallback method: when Maven cannot take our original `buildTarget` we try again but with `--project=groupId:artifactId` and `--file=the-pom-file.xml`. If we can't identify the groupId or artifactId from the POM file, then there's nothing else we can do.

buildtools/maven/pom_test.go
```diff
@@ -0,0 +1,54 @@
...
+func TestGraphFromTarget(t *testing.T) {
+	// This test needs just the Dependencies field of a Manifest to be translated to a graph.Deps.
+	projectDir := testdataDir
+
+	id1 := pkg.ID{Type: pkg.Maven, Name: "com.google.code.g:g", Revision: "2.7"}
+	id2 := pkg.ID{Type: pkg.Maven, Name: "javax.annotation:jsr250-api", Revision: "1.0.2"}
+	id3 := pkg.ID{Type: pkg.Maven, Name: "junit:junit", Revision: "3.8.2"}
+
+	wantDirect := []pkg.Import{
+		{"", id1},
```
golangcibot: composites: `github.com/fossas/fossa-cli/pkg.Import` composite literal uses unkeyed fields (from `govet`)

buildtools/maven/pom_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{"", id2},
```
golangcibot: composites: `github.com/fossas/fossa-cli/pkg.Import` composite literal uses unkeyed fields (from `govet`)

buildtools/maven/pom_test.go
```diff
@@ -0,0 +1,54 @@
...
+		{"", id3},
```
golangcibot: composites: `github.com/fossas/fossa-cli/pkg.Import` composite literal uses unkeyed fields (from `govet`)

Issue #453, opened 2019-05-01
Gradle analyzer inaccurately determines the revision for requested versions


zlav:
**Describe the bug**
When `./gradlew <target>:dependencies` is run and we see a requested version the fossa-cli has difficulty creating a valid dependency locator.

**To Reproduce**
Mock gradle output:
```
default - Configuration for default artifacts.
+--- org.springframework.boot:spring-boot-starter -> 2.1.0.RELEASE
|    +--- org.springframework.boot:spring-boot:2.1.0.RELEASE
|    |    +--- org.springframework:spring-core:5.1.2.RELEASE
|    |    |    \--- org.springframework:spring-jcl:5.1.2.RELEASE
|    |    \--- org.springframework:spring-context:5.1.2.RELEASE
```

**Expected behavior**
Fossa should create an accurate locator from this dependency graph.

dchenk:
Fixed in #454

Issue #367, opened 2018-10-26
[Gradle project] panic: runtime error: index out of range


vgaidarji:
**Describe the bug**
`fossa -o` command fails with `panic: runtime error: index out of range` error.

**To Reproduce**
1. Checkout https://github.com/vgaidarji/dependencies-overview
2. [Install fossa-cli on Mac OS](https://github.com/fossas/fossa-cli#installation)
3. Run `fossa init`
4. See generated `.fossa.yml` and correct generated configuration

**Expected behavior**
`fossa -o --debug` command should succeed without errors.

**Debugging information**
This is how `.fossa.yml` file looks like after I modified it:
**.fossa.yml**
```
# Generated by FOSSA CLI (https://github.com/fossas/fossa-cli)
# Visit https://fossa.io to learn more

version: 1
cli:
  server: https://app.fossa.io
  fetcher: custom
  project: git@github.com:vgaidarji/dependencies-overview.git
analyze:
  modules:
  - name: dependencies-overview:compile
    path: .
    type: gradle
```

**fossa --version**

```
➜  dependencies-overview git:(integrate-fossa-license-scan) ✗ fossa --version      
fossa-cli version 0.7.13 (revision 623f658a3882ec0d4e64c43fa973c8088bee54fa compiled with go version go1.11 linux/amd64)
```

**fossa -o --debug**
```
➜  dependencies-overview git:(integrate-fossa-license-scan) ✗ fossa -o --debug
WARNING It looks like you're using a configuration file generated by an old version of `fossa` (pre-0.7.x). You may encounter bugs by using this configuration file. You can fix this warning by re-running `fossa init` to generate a new configuration file.
WARNING Support for old configuration files will be dropped in v0.8.0.
[...]

PR #454, opened 2019-05-01
Harden parsing of Gradle's output

dchenk:


dchenk:
Fixes #453 
but does not address #367

dchenk:
Fixes #453 
Fixes #367

buildtools/gradle/testdata/tasks-output.txt
```diff
@@ -0,0 +1,77 @@
+------------------------------------------------------------
```
zlav: It might be helpful to cut out some lines in this file so that its easier to read. I try to focus on trying to only have as many cases that could trip up the parsing logic. This could also be applied to `complex.txt`.

buildtools/gradle/gradle_test.go
```diff
@@ -75,6 +75,54 @@ func TestAllDependencies(t *testing.T) {
...
+	expectDep2 := gradle.Dependency{
+		Name:             "org.slf4j:slf4j-api",
+		RequestedVersion: "1.6.4",
+		ResolvedVersion:  "1.7.25",
+	}
+	assert.Contains(t, deps, expectDep2)
+}
+
+func TestShellCommand_DependencyTasks(t *testing.T) {
+	// We should be able to identify the projects by running "gradle tasks" even if a project has the
```
zlav: Really happy you added coverage for this command here, are there any other cases you think might trip us up in the future?

buildtools/gradle/gradle.go
```diff
@@ -145,7 +145,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 	var projects []string
 	lines := strings.Split(stdout, "\n")
 	for _, line := range lines {
-		if i := strings.Index(line, ":dependencies"); i != -1 {
+		if i := strings.Index(line, ":dependencies -"); i != -1 {
```
zlav: This feel like a bit of a hack, although I somewhat believe parsing `gradle tasks` is a hack in itself so I don't have a problem with this as long as its tested, which you did. You should be aware of the command `gradle projects` which might be more the  more applicable command to run here.

dchenk: I wonder why `gradle tasks` was preferred when this part was first implemented

buildtools/gradle/gradle_test.go
dchenk: none that I know of

dchenk:
it appears that `TestAllDependencies` is already testing `ProjectDependencies`

buildtools/gradle/gradle.go
zlav: I think it is much easier to parse and assures we get projects with `:dependencies` tasks.

Issue #344, opened 2018-10-10
Ergonomic errors


liftM:
## Problem

Our current error messages leave something to be desired. Some examples of bad error messages:

| Error | Root cause
|-|-
| `Failed to get some revisions: could not unmarshal JSON API response: json: cannot unmarshal object into Go value of type []fossa.Revision` | The user forgot to provide an API key.
| `Could not test revision: could not unmarshal JSON API response: json: cannot unmarshal object into Go value of type []fossa.User` | There was a bug in our request handling code.
| `Could not analyze: open subfoo/Podfile.lock: no such file or directory` | A nested module in a project was not built.
| `Could not analyze: fork/exec : no such file or directory` | A build tool was not found.
| `Could not analyze: could not get dependency list from Composer: fork/exec /usr/local/bin/composer: not a directory` | An old configuration file was parsed incorrectly.
| `Could not get OrganizationID while constructing locator` | The environment did not have the correct API key.
| `Unrecognized locator` | I have no idea, but this should definitely not happen.
| `Could not analyze: no revision found for package name` | The user's build was not correctly configured.
| `panic: runtime error: slice bounds out of range` | We didn't handle Windows line endings correctly (also happens on old configuration file parsing).

## Goals

These messages need to get a lot better.

1. A user should be able to understand what the problem is, and distinguish between a bug on our end vs. an incorrect configuration on their end.
2. Ideally, we should provide enough information that a user can resolve the issue on their own.
3. We should provide easy instructions for filing useful bug reports in case they need us to help.

### The error message UX

Here's a prototype of better error messages:

![2018-10-10-184908_1138x898_scrot](https://user-images.githubusercontent.com/2092654/46775715-367f5580-ccbd-11e8-8c4a-7f81eff52770.png)

### Types of errors

There are really 4 groups of errors:

1. The user has done something wrong (bad invocation, bad configuration file, wrong flags, etc.).
2. A module has something wrong with it (it's not built, required build tools are not available, etc.).
3. The system has done something wrong (out of memory, out of disk space, network failure, etc.)
4. There is a bug in the CLI.

These currently manifest in 3 different ways:

1. An error is returned: this is the ideal way to handle errors.
2. A `log.Fatal` occurs: this kind of error handling is very difficult to make consistent, and should be avoided overall. Unfortunately, we do this in a lot of places (especially in situations that should "never" occur).
3. A panic occurs: this happens especially in our parsing code. We must be able to gracefully handle panics.

### Design

[...]

PR #455, opened 2019-05-02
feat (format fossa test) improve fossa test output for readability

### LINKED ISSUE #216: Improve readability of `fossa test` output
`fossa test` currently outputs JSON which is not particularly readable. Instead, it should have two distinct modes:

- By default, output a human-readable report
- With `--json`, output a JSON report with a well-defined schema

zlav:
This PR closes #216 

Improve `fossa test` output so that instead of outputting:
```
Test failed! 4 issues found
{"Count":4,"Issues":[{"ID":137012,"PriorityString":"low","Resolved":false,"Revision":{"loc":null,"Licenses":null,"Project":null,"Meta":null,"Issues":null},"Type":"policy_flag"},{"ID":170908,"PriorityString":"low","Resolved":false,"Revision":{"loc":null,"Licenses":null,"Project":null,"Meta":null,"Issues":null},"Type":"unlicensed_dependency"},{"ID":171495,"PriorityString":"low","Resolved":false,"Revision":{"loc":null,"Licenses":null,"Project":null,"Meta":null,"Issues":null},"Type":"policy_flag"},{"ID":175711,"PriorityString":"low","Resolved":false,"Revision":{"loc":null,"Licenses":null,"Project":null,"Meta":null,"Issues":null},"Type":"policy_flag"}],"Status":"SCANNED"}
```
We will output a formatted message:
```
Test Failed. 4 issues found:
Flagged by Policy

JSV              4.0.2      Artistic-1.0
ua-parser-js     0.7.18     GPL-2.0-only
mapbox-gl        0.44.2     GPL-3.0-only

Unlicensed Dependency

curry     1.2.0
```

Changes made:
- Improve `fossa.Issues` so that it normalizes issues after they are received by an api request.
- Add the option to output information in JSON.

Expected Future Issues:
- New issue types will need to be manually added so that are printed in a nice format.
- Dependencies with multiple issues will output unnecessary noise. I have determined this is out of scope for this PR.

Tests will be added once the fossa test server PR is completed.

### DIFF
--- a/api/fossa/issues.go
+++ b/api/fossa/issues.go
@@ -10,24 +10,6 @@ import (
 
 const IssuesAPI = "/api/cli/%s/issues"
 
-// An Issue holds the FOSSA API response for the issue API.
-type Issue struct {
-	ID             int    `json:"id"`
-	PriorityString string `json:"priorityString"`
-	Resolved       bool   `json:"resolved"`
-	RevisionID     string `json:"revisionId"`
-	Type           string `json:"type"`
-	Rule           Rule   `json:"rule`
-
-	Name     string
-	Revision string
-}
-
-// Rule holds the representation of an Issue Rule.
-type Rule struct {
-	License string `json:"licenseId"`
-}
-
 // A wrapped list of issues returned by the FOSSA CLI issues endpoint
 // If a push-only API key is used, then only the count is returned
 type Issues struct {
@@ -36,6 +18,24 @@ type Issues struct {
 	Status string
 
 	NormalizedByType map[string][]Issue
+}
+
+// An Issue holds the FOSSA API response for the issue API.
+type Issue struct {
+	ID             int    `json:"id"`
+	PriorityString string `json:"priorityString"`
+	Resolved       bool   `json:"resolved"`
+	RevisionID     string `json:"revisionId"`
+	Type           string `json:"type"`
+	Rule           Rule   `json:"rule"`
+
+	Name     string
+	Revision string
+}
+
+// Rule holds the representation of an Issue's Rule.
+type Rule struct {
+	License string `json:"licenseId"`
 }
 
 // GetIssues loads the issues for a project.
@@ -74,7 +74,7 @@ func formatType(issueType string) string {
 	case "outdated_dependency":
 		return "Outdated Dependency"
 	default:
-		return "Unrecognized Issue Type"
+		return issueType
 	}
 }
 
--- a/api/fossa/issues.go
+++ b/api/fossa/issues.go
@@ -3,6 +3,7 @@ package fossa
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 
 	"github.com/pkg/errors"
 )
@@ -11,11 +12,20 @@ const IssuesAPI = "/api/cli/%s/issues"
 
 // An Issue holds the FOSSA API response for the issue API.
 type Issue struct {
-	ID             int
-	PriorityString string
-	Resolved       bool
-	Revision       Revision
-	Type           string
+	ID             int    `json:"id"`
+	PriorityString string `json:"priorityString"`
+	Resolved       bool   `json:"resolved"`
+	RevisionID     string `json:"revisionId"`
+	Type           string `json:"type"`
+	Rule           Rule   `json:"rule`
+
+	Name     string
+	Revision string
+}
+
+// Rule holds the representation of an Issue Rule.
+type Rule struct {
+	License string `json:"licenseId"`
 }
 
 // A wrapped list of issues returned by the FOSSA CLI issues endpoint
@@ -24,6 +34,8 @@ type Issues struct {
 	Count  int
 	Issues []Issue
 	Status string
+
+	NormalizedByType map[string][]Issue
 }
 
 // GetIssues loads the issues for a project.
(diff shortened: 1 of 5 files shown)


api/fossa/issues.go
```diff
@@ -11,11 +12,20 @@ const IssuesAPI = "/api/cli/%s/issues"
...
-	PriorityString string
-	Resolved       bool
-	Revision       Revision
-	Type           string
+	ID             int    `json:"id"`
+	PriorityString string `json:"priorityString"`
+	Resolved       bool   `json:"resolved"`
+	RevisionID     string `json:"revisionId"`
+	Type           string `json:"type"`
+	Rule           Rule   `json:"rule`
```
golangcibot: structtag: struct field tag `json:"rule` not compatible with reflect.StructTag.Get: bad syntax for struct tag value (from `govet`)

cmd/fossa/display/template.go
```diff
@@ -25,3 +37,20 @@ func Template(tmpl *template.Template, data interface{}) (string, error) {
...
 }
+
+// TemplateFormatTabs renders a template with the desired tab format.
+func TemplateFormatTabs(tmpl string, data interface{}, minWidth, tabWidth, padding int) (string, error) {
+	testTemplate, err := template.New("base").Parse(tmpl)
+	if err != nil {
+		return "", errors.Wrap(err, "Could not parse template data")
+	}
+
+	var buf bytes.Buffer
```
dchenk: please use [strings.Builder](https://godoc.org/strings#Builder) is output is only a string

dchenk: I know you didn't touch `func Template` immediately above, but it too should use `strings.Builder`

api/fossa/issues.go
```diff
@@ -3,27 +3,39 @@ package fossa
...
 	Count  int
 	Issues []Issue
 	Status string
+
+	NormalizedByType map[string][]Issue
+}
+
+// An Issue holds the FOSSA API response for the issue API.
+type Issue struct {
+	ID             int    `json:"id"`
```
dchenk: Doesn't specifying the JSON key this way change the output? Are no customers relying on the current names in the JSON objects? 
https://play.golang.org/p/IO7TleEfPap

cmd/fossa/cmd/test/test.go
```diff
@@ -18,8 +18,20 @@ import (
 	"github.com/fossas/fossa-cli/config"
 )
 
+const defaultTestTemplate = `Test Failed. {{.Count}} {{if gt .Count 1 -}} issues {{- else -}} issue {{- end}} found:
+{{- range $type, $issues := .NormalizedByType}}
+========================================================================
```
dchenk: I'm not a big fan of noisy output—even if it's for humans. I'm not against _one_ such line to divide sections, but do we need two? Perhaps whitespace is just as effective but easier to read.

cmd/fossa/cmd/test/test.go
```diff
@@ -18,8 +18,20 @@ import (
...
+{{$type}}
+========================================================================
+{{range $i, $issue := $issues}}
+{{$issue.Name}}	{{$issue.Revision}}	{{$issue.Rule.License}}
+{{- end}}
+{{end}}
+`
+
 var Timeout = "timeout"
```
dchenk: can we make all of these constants in a group:
```
const (
	Timeout        = "timeout"
	SuppressIssues = "suppress-issues"
	JSON           = "json"
)
```
Also I don't think any of these need to be exported.

cmd/fossa/cmd/test/test.go
```diff
@@ -18,8 +18,20 @@ import (
 	"github.com/fossas/fossa-cli/config"
 )
 
+const defaultTestTemplate = `Test Failed. {{.Count}} {{if gt .Count 1 -}} issues {{- else -}} issue {{- end}} found:
```
dchenk: Here we say "Test Failed." but upon failure we say "Test passed!"

I think the messages should be more consistent at least with capitalization of the second word, e.g.: "Test failed" and "Test passed" — or capitalize second word in both cases.

cmd/fossa/cmd/test/test.go
```diff
@@ -18,8 +18,20 @@ import (
 	"github.com/fossas/fossa-cli/config"
 )
```
dchenk: I ran `fossa test` on a project that returned just:
```
Test Failed. 67 issues found:

```
Apparently it couldn't loop over the issues to display them. I'll investigate and let you know what I find out.

dchenk: It's because we can end up with `issues` in func `Run` looking like:
```
{
  Count: 67,
  Issues: nil,
  Status: "SCANNED",
  NormalizedByType: nil,
}
```

The original API response which gets unmarshalled into an `Issues` struct:
```
{"count":67,"status":"SCANNED"}
```

lackstein: If you're using a push only API key, we don't return any issue data, only pass/fail. This is to avoid leaking information if a key is compromised.

dchenk: I see. The CLI should document this somehow (explain that displaying the results requires a full-access key). And if the user still provides a push-only key—and most users still will at least initially—then the output shown should say something like:
```
Test Failed. 67 issues found. Cannot retrieve issues with a push-only API key.
```

zlav: Agreed, this will be addressed with issue #344 which is the next thing I'm working on.

api/fossa/issues.go
zlav: Good observation, although I am not aware of anyone who is relying on this information at the moment. The information up until now has been very limited so `fossa test` has been used as a pass-fail mechanism. Regarding this change, I am a fan of keeping the JSON consistent between the API response and what `fossa test` outputs.

dchenk: Not saying I actually know of any customers who rely on the current JSON output. You'll know better, though it would be interesting to get some insight somehow.

cmd/fossa/cmd/test/test.go
zlav: Yea you're right, these don't have to be exported. I believe the reason this was done is because the unexported names for a lot of these flags are packages or reserved, such as `json` or `type`. I made these constants, but I'm comfortable leaving them exported for clarity as they are littered around the codebase and want to maintain consistency. Good point though.

cmd/fossa/cmd/test/test.go
zlav: I toyed around with a few options but I do still prefer the look of the double lines, especially as the number of issues and types increases. If you feel strongly about a change here feel free to leave an example.

zlav: Resolved offline, message format has been improved as much as possible.

PR #458, opened 2019-05-08
feat (command timeout) add timeout for gradle analyzer

zlav:
Currently a WIP

When the FOSSA CLI shells out to external commands it is put at the mercy of those commands to succeed. If one of those commands consumes excess resources on the host machine and prevents analysis from succeeding, we should have the ability to timeout, kill the command, and move on. 

The worst behaving command is `gradle` which has shown that unpredictably it can cause analysis to take hours. 

This PR accomplishes the following:
- Rework `fossa/exec` to include optional timeout and retry logic.
- Implement this logic in gradle.
- Add options to gradle to allow users to configure these values on their own.

### DIFF
--- a/analyzers/gradle/gradle.go
+++ b/analyzers/gradle/gradle.go
@@ -35,6 +35,8 @@ type Options struct {
 	Online            bool   `mapstructure:"online"`
 	AllSubmodules     bool   `mapstructure:"all-submodules"`
 	AllConfigurations bool   `mapstructure:"all-configurations"`
+	Timeout           int    `mapstructure:"timeout"`
+	Retries           int    `mapstructure:"retries"`
 	// TODO: These are temporary until v2 configuration files (with proper BuildTarget) are implemented.
 	Project       string `mapstructure:"project"`
 	Configuration string `mapstructure:"configuration"`
@@ -55,8 +57,10 @@ func New(m module.Module) (*Analyzer, error) {
 			log.Warnf("A build.gradle file has been found at %s, but Gradle could not be found. Ensure that Fossa can access `gradle`, `gradlew`, `gradlew.bat`, or set the `FOSSA_GRADLE_CMD` environment variable. Error: %s", m.Dir, err.Error())
 		}
 	}
+	retries := options.Retries
+	timeout := options.Timeout
 
-	shellInput := gradle.NewShellInput(binary, m.Dir, options.Online)
+	shellInput := gradle.NewShellInput(binary, m.Dir, options.Online, timeout, retries)
 	analyzer := Analyzer{
 		Module:  m,
 		Options: options,
@@ -76,7 +80,7 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 	return DiscoverWithCommand(dir, options, gradle.Cmd)
 }
 
-func DiscoverWithCommand(dir string, options map[string]interface{}, command func(string, ...string) (string, error)) ([]module.Module, error) {
+func DiscoverWithCommand(dir string, options map[string]interface{}, command func(string, int, int, ...string) (string, error)) ([]module.Module, error) {
 	log.WithField("dir", dir).Debug("discovering gradle modules")
 	var modules []module.Module
 	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
--- a/analyzers/gradle/gradle.go
+++ b/analyzers/gradle/gradle.go
@@ -35,7 +35,7 @@ type Options struct {
 	Online            bool   `mapstructure:"online"`
 	AllSubmodules     bool   `mapstructure:"all-submodules"`
 	AllConfigurations bool   `mapstructure:"all-configurations"`
-	Timeout           int    `mapstructure:"timeout"`
+	Timeout           string `mapstructure:"timeout"`
 	Retries           int    `mapstructure:"retries"`
 	// TODO: These are temporary until v2 configuration files (with proper BuildTarget) are implemented.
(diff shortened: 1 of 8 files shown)


exec/run.go
```diff
@@ -20,6 +24,8 @@ type Cmd struct {
 	// If neither Env nor WithEnv are set, the environment is inherited from os.Environ().
 	Env     map[string]string // If set, the command's environment is _set_ to Env.
 	WithEnv map[string]string // If set, the command's environment is _added_ to WithEnv.
+
+	exec *exec.Cmd
```
golangcibot: `exec` is unused (from `structcheck`)

exec/run.go
```diff
@@ -14,38 +18,103 @@ type Cmd struct {
...
+	}()
+
+	select {
+	case <-time.After(seconds):
+		err := xc.Process.Kill()
+		if err != nil {
+			return "", "", errors.Wrapf(err, "Error killing the process")
+		}
+
+		return "", "", errors.New(fmt.Sprintf("Operation timed out running `%s %s` after %s", cmd.Name, cmd.Argv, cmd.Timeout))
```
zlav: @dchenk I think this is something I corrected you on a while back. I now realize this is due to the inconsistency of the error package we are using in different packages. I think consolidating this and creating a standard is a task for another PR.

analyzers/gradle/gradle.go
```diff
@@ -35,6 +35,8 @@ type Options struct {
 	Online            bool   `mapstructure:"online"`
 	AllSubmodules     bool   `mapstructure:"all-submodules"`
 	AllConfigurations bool   `mapstructure:"all-configurations"`
+	Timeout           int    `mapstructure:"timeout"`
```
dchenk: Why not let the user supply a duration string such as "5m30s" and parse it using [ParseDuration](https://godoc.org/time#ParseDuration)? This could make the UX a little nicer when otherwise people would need to convert their desired timeout to seconds. But then we'd need to document how a duration is specified—but we should document this configuration anyway. What do you think?

This would probably suffice to explain it:
> A duration string is a sequence of numbers, each with a unit suffix, such as "500ms", "1.5m", or "2h45m". Valid time units are "ms", "s", "m", "h".

exec/run.go
```diff
@@ -14,38 +18,103 @@ type Cmd struct {
...
 
-	stdoutBuf, err := xc.Output()
-	stdout = string(stdoutBuf)
-	stderr = stderrBuf.String()
-
-	log.WithFields(log.Fields{
-		"stdout": stdout,
-		"stderr": stderr,
-	}).Debug("done running")
+	done := make(chan error, 1)
```
dchenk: this can be an unbuffered channel

exec/run_test.go
```diff
@@ -41,3 +42,48 @@ func TestDefaultEnv(t *testing.T) {
...
+		Name:    "sleep",
+		Argv:    []string{"3"},
+		Timeout: 1 * time.Second,
+		Retries: 1,
+	}
+
+	start := time.Now()
+	_, _, err := exec.Run(command)
+	assert.Error(t, err)
+	assert.WithinDuration(t, start, time.Now(), 3*time.Second)
```
dchenk: I think it's a good idea to `assert.WithinDuration` as you do here, but the delta duration should be slightly _less_ than 3 seconds to make it clear that we don't want the "sleep" command to run to completion. Perhaps `2*time.Second + 500*time.Millisecond`

exec/run_test.go
```diff
@@ -41,3 +42,48 @@ func TestDefaultEnv(t *testing.T) {
...
+
+	command.Retries = 4
+	start = time.Now()
+	_, _, err = exec.Run(command)
```
dchenk: the assertions that follow on lines 87 and 88 rely on the time when `time.Since` is invoked. Instead, we should mark the time immediately after `exec.Run` returns and compare the end time to the start time:
```
start = time.Now()
_, _, err = exec.Run(command)
end := time.Now()
assert.Error(t, err)
// 4 retries with a 1 second timeout means that the command should take 5 seconds to error.
assert.True(t, end.Sub(start) > 5*time.Second)
assert.True(t, end.Sub(start) < 6*time.Second)
```

dchenk: this also makes it a little more obvious what is being tested

exec/run.go
```diff
@@ -14,38 +18,103 @@ type Cmd struct {
...
+			break
+		}
+	}
+
+	return stdout, stderr, err
+}
+
+// RunTimeout executes a `Cmd` and waits to see if it times out.
+func runWithTimeout(cmd Cmd) (string, string, error) {
+	seconds := cmd.Timeout
```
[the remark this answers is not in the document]
zlav: Yea good catch here, forgot to change this when I stopped enforcing seconds.

analyzers/gradle/gradle.go
zlav: This is a really good thought and idea with the command ParseDuration. I'm a little conflicted with using this in practice because it adds additional complexity for the user. I'm inclined to disagree with using this for the sake of simplicity and because timeouts should always be set to low values. This is really interesting though and I'm going to play around with it before I make up my mind.

exec/run_test.go
zlav: Good call, I'm changing the sleep argument to `4` instead of modifying the time in the assertion.

analyzers/gradle/gradle.go
zlav: I changed it to this, I think this is a little more future proofed. Good idea.

PR #459, opened 2019-05-09
Update how Maven modules are described in .fossa.yml

dchenk:
This changes the way we describe Maven modules in our `.fossa.yml` to make modules filterable based on their path relative to the root of the FOSSA project by providing more of the path to each module's POM file in the `path` property rather than in `target`. Only discovery of modules is changed here. Existing configurations should continue to work fine.

### DIFF
--- a/analyzers/maven/maven.go
+++ b/analyzers/maven/maven.go
@@ -82,9 +82,9 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 				return nil
 			}
 
-			submodules, err := maven.Modules(path, dir, checked)
+			submodules, err := maven.Modules(filepath.Join(path, "pom.xml"), path, checked)
 			if err != nil {
-				log.WithError(err).Debug("could not get modules at path")
+				log.WithError(err).Debugf("could not get modules at path %s", path)
 				return err
 			}
 
@@ -93,7 +93,7 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 					Name:        m.Name,
 					Type:        pkg.Maven,
 					BuildTarget: m.Target,
-					Dir:         dir,
+					Dir:         m.Dir,
 				})
 			}
 			// Continue recursing because there may be modules that are not declared under the current module.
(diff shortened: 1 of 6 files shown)


buildtools/maven/maven_test.go
```diff
@@ -8,30 +8,60 @@ import (
...
-	fullPath := filepath.Join(testdataDir, "pom.xml")
+	// Here we mostly just test the discovery of POM files, and analyzers/maven tests that we get the correct
+	// list of MvnModules.
+
+	path1 := filepath.Join(testdataDir, "pom.xml")
 	checked := make(map[string]bool)
-	mods, err := maven.Modules(fullPath, testdataDir, checked)
+	mods, err := maven.Modules(path1, testdataDir, checked)
 	if assert.NoError(t, err) {
 		assert.Len(t, mods, 1)
```
zlav: Is there a reason you're nesting errors like this? You really should be able to just run the following? I can see why you wouldn't like to be panicking and that's valid although that doesn't concern me as much while running tests.
```
assert.NoError(t, err)
assert.Len(t, mods, 1)

exists, err := files.Exists(mod.Dir, mod.Target)
assert.NoError(t, err)
assert.True(t, exists)
```

analyzers/maven/maven.go
```diff
@@ -82,9 +82,9 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 				return nil
 			}
 
-			submodules, err := maven.Modules(path, dir, checked)
+			submodules, err := maven.Modules(filepath.Join(path, "pom.xml"), path, checked)
```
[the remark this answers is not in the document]
dchenk: Doing this because `files.Exists(path, "pom.xml")` is checked above (line 77).

Because this is just our attempt to _discover_ Maven modules, this is really all we can do. The official Maven documentation only ever speaks about `pom.xml` being the name of the main POM file in a project. All the real Maven projects I've seen have at least a _minimal_ `pom.xml` at the root of their project and a list of `<modules>` in it pointing to other POM files.

buildtools/maven/maven_test.go
dchenk: I guess I've been bitten by panics when not nesting usage of the returned objects within the `if err` check, which only matters if there is in fact an error. I agree that un-indenting will make this look better. Will fix now.

PR #467, opened 2019-05-29
feat (buckw) discover buck command and prefer buckw

zlav:
Some users of buck prefer to use a buck command wrapper `buckw` instead of the default `buck` command. This PR discovers which one the directory is using by preferring `buckw` and ensures this is used throughout the analyzer. The implementation is very similar to how python discovers the correct command. Once the correct command is discovered, it is used to create a `Cmd()` function which can be used to execute the command.

I validated this by working with okbuck and buck codebases to ensure that a `buckw` file would be located and executed. I see no way to test this without also implementing a full buck integration suite.

### DIFF
--- a/docs/integrations/buck.md
+++ b/docs/integrations/buck.md
@@ -35,7 +35,8 @@ analyze:
 ## Analysis
 
 Buck projects are uploaded with all of their source code broken into dependencies for license analysis by FOSSA. This project involves a few steps:
+1. Determine the correct buck command to run by trying to execute `FOSSA_BUCK_CMD`, `./buckw`, and `buck`.
 1. Run `buck audit input <target> --json` to retrieve a list of all dependencies.
-2. Upload these dependencies and maintain references to their uploaded locator.
-3. Create a dependency graph by recursively running `buck audit dependencies <target>` on each dependency and add it to the graph.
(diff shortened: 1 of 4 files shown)


docs/integrations/buck.md
```diff
@@ -35,7 +35,8 @@ analyze:
 ## Analysis
 
 Buck projects are uploaded with all of their source code broken into dependencies for license analysis by FOSSA. This project involves a few steps:
+1. Determine the correct buck command to run by trying to execute `FOSSA_BUCK_CMD`, `./buckw`, and `buck`.
 1. Run `buck audit input <target> --json` to retrieve a list of all dependencies.
-2. Upload these dependencies and maintain references to their uploaded locator.
-3. Create a dependency graph by recursively running `buck audit dependencies <target>` on each dependency and add it to the graph.
-4. Upload the dependency graph to FOSSA which will match the dependency locators to the raw upload and analysis of each dependency.
\ No newline at end of file
+1. Upload these dependencies and maintain references to their uploaded locator.
```
dchenk: are you exploiting a GitHub markdown feature here in numbering all these items with `1`?

zlav: This isn't unique to GitHub: https://www.markdownguide.org/basic-syntax/#ordered-lists. I like to do this so that I can add new steps wherever I want. I found myself re-ordering the list a few times and it got annoying.

PR #469, opened 2019-06-06
feat (license scan tar) use the rawLicenseScan parameter to run a full license scan

zlav:
This PR adds the ability to set the `prebuilt` parameter when making an archive upload. This allows the cli to ensure that the tarball being uploading is treated as an entirely built project. This prevents FOSSA from trying to run its own dependency analysis on the project (which could be done by the cli) as well as ensures that directories such as `vendor` and `node_modules` are entirely scanned for licenses.

This PR enforces that all `raw` type uploads will only have a license scan run on them. In order to take advantage of the ability to upload files for dependency analysis, I would like to break the tarball uploader out into a separate analyzer to keep configuration under control.


api/fossa/tar.go
```diff
@@ -38,6 +38,16 @@ type SignedURL struct {
 	SignedURL string
 }
 
+// UploadTarballDependency uploads the directory specified to be treated on FOSSA as a dependency.
+func UploadTarballDependency(dir string, upload, rawLicenseScan bool) (Locator, error) {
```
dchenk: it looks like this function is only ever called with `rawLicenseScan` as `true`. Perhaps it shouldn't be a parameter?

api/fossa/tar.go
```diff
@@ -363,12 +368,16 @@ func tarballUpload(name string, dependency, upload bool, tarball *os.File, hash
...
 
-	dependencyParameter := url.Values{}
+	parameters := url.Values{}
 	if dependency {
-		dependencyParameter.Add("dependency", "true")
+		parameters.Add("dependency", "true")
+	}
 
+	if rawLicenseScan {
+		parameters.Add("rawLicenseScan", "true")
```
dchenk: Is the backend wired up for this?

dchenk: Looks like there's a pending PR for that. This PR looks good, but in your place I'd make sure the API on the server side is finalized before you merge this.

api/fossa/tar.go
zlav: I was a little split on that decision here. I amended the PR comment to address this. Ideally, we will give users the option to upload projects for the sake of running full backend dependency analysis on it, but that would require adding some hacky flags on `fossa analyze`. To do this right I need to refactor the tar uploader into its own "analyzer".
>This PR enforces that all raw type uploads will only have a license scan run on them. In order to take advantage of the ability to upload files for dependency analysis, I would like to break the tarball uploader out into a separate analyzer to keep configuration under control.

PR #474, opened 2019-07-01
feat (rust support) Support rust through Cargo.lock parsing.

zlav:
This PR adds support for the rust language through the package manager cargo.

How this PR works:
Discovery:
1. Look for `Cargo.lock` file and create a module.

Analysis:
1. Attempt to read `Cargo.toml` from the modules directory, we can assume this module is the "root crate" because we previously discovered the `Cargo.lock` file here.
1. Read all of the "members" from the root crate `Cargo.toml` file and store the dependency names to be considered as direct dependencies (This approximates the direct dependencies because `Cargo.lock` does not provide that information)
1. Parse to singular `Cargo.lock` file and create a dependency graph and map of dependency name to `pkg.ID`.
1. Use the dependency map and information from the manifest files to construct the list of direct dependencies. 

This PR still needs:
Documentation


buildtools/cargo/cargo.go
```diff
@@ -0,0 +1,110 @@
...
+	var man manifest
+	err = files.ReadTOML(&man, filepath.Join(dir, "Cargo.toml"))
+	if err != nil {
+		return graph.Deps{}, errors.Wrap(err, "No lockfile Gopkg.lock found")
+	}
+	manifests = append(manifests, man)
+	for _, list := range man.Workspace.Members {
+		subMan := manifest{}
+		subFile := filepath.Join(dir, list, "Cargo.toml")
+		err = files.ReadTOML(&subMan, subFile)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

analyzers/rust/rust.go
```diff
@@ -0,0 +1,112 @@
...
+		}
+
+		if !info.IsDir() && info.Name() == "Cargo.lock" {
+			moduleName := filepath.Base(path)
+
+			log.WithFields(log.Fields{
+				"path": path,
+				"name": moduleName,
+			}).Debug("found Rust module")
+			relPath, _ := filepath.Rel(dir, path)
```
dchenk: is the error ignored because `filepath.Rel` will never fail here?

analyzers/rust/rust.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Clean logs a warning and does nothing for Rust.
+func (a *Analyzer) Clean() error {
+	log.Warnf("Clean is not implemented for Rust")
+	return nil
+}
+
+func (a *Analyzer) Build() error {
+	return errors.New("Build not implemented for Rust")
```
dchenk: [not important] For consistency, "Build is not..." would go better with the message "Clean is not implemented for Rust" from func Clean.

buildtools/cargo/cargo.go
```diff
@@ -0,0 +1,111 @@
...
+	err = files.ReadTOML(&man, filepath.Join(dir, "Cargo.toml"))
+	if err != nil {
+		log.Warnf("Manifest file `Cargo.toml` was not found in directory: %s. Direct dependencies may be incorrect.", dir)
+	}
+	manifests = append(manifests, man)
+
+	for _, member := range man.Workspace.Members {
+		memberManifest := manifest{}
+		memberFile := filepath.Join(dir, member, "Cargo.toml")
+		err = files.ReadTOML(&memberManifest, memberFile)
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

docs/integrations/rust.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Support
+
+Rust support relies on the presence of the `Cargo.lock` and `Cargo.toml` files.
+
+## Configuration
+
+### Automatic
+
+Run `fossa init` to detect all python directories that contain `Cargo.lock`.
```
dchenk: "python directories"

docs/integrations/rust.md
```diff
@@ -0,0 +1,32 @@
...
+      type: rust
+      target: rust/root
+      dir:  rust/root
+```
+
+## Analysis
+
+Rust analysis is a simple three step process:
+
+1. Parse `Cargo.lock` and retrieve information about the full dependency tree and each dependencies resolved version.
```
dchenk: dependencies -> dependency's

docs/integrations/rust.md
```diff
@@ -0,0 +1,32 @@
...
+2. Read the [root crate](https://doc.rust-lang.org/cargo/reference/manifest.html#the-workspace-section)'s `Cargo.toml` and detect if any "member" crates exist.
+3. Parse all `Cargo.toml` files to retrieve information about which dependencies the user declared.
```
dchenk: it's not entirely clear who "the user" is. I think "which dependencies are declared" is more precise

analyzers/rust/rust.go
```diff
@@ -0,0 +1,110 @@
...
+// Discover constructs modules in all directories with a `Cargo.lock`.
+func Discover(dir string, options map[string]interface{}) ([]module.Module, error) {
+	var modules []module.Module
+	err := filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			log.WithError(err).WithField("path", path).Debug("error while walking for discovery")
+		}
+
+		if !info.IsDir() && info.Name() == "Cargo.lock" {
+			moduleName := filepath.Base(path)
```
dchenk: I don't think `filepath.Base` is the function you're looking for here. It always returns "Cargo.lock". Consider taking the actual name from [`Cargo.toml`](https://doc.rust-lang.org/cargo/reference/manifest.html#the-manifest-format) or just:
```
abs, err := filepath.Abs(info.Name())
// if err ....
moduleName := filepath.Base(filepath.Dir(abs))
```

pkg/types.go
```diff
@@ -248,6 +256,8 @@ func (t Type) String() string {
 		return "pip"
 	case Ruby:
 		return "gem"
+	case Rust:
+		return "cargo"
```
dchenk: I think this should return "rust" because the docs say that Rust modules have `type: rust` in `.fossa.yml`. When you run `fossa init` and see:
```
analyze:
  modules:
  - name: Cargo.lock
    type: cargo
    target: .
    path: .
```
it might appear a little surprising that `type` isn't `rust` as the docs suggest it should be. On the other hand, it'll be just as easy to change the docs

buildtools/cargo/cargo.go
```diff
@@ -0,0 +1,115 @@
...
+	Dependencies      map[string]interface{} `toml:"dependencies"`
+	BuildDependencies map[string]interface{} `toml:"build-dependencies"`
+	Package           Package
+}
+
+type workspace struct {
+	Members []string
+}
+
+// lockfile contains the Projects in a dep lockfile and a corresponding map for retrieving project data.
```
dchenk: Is "and a corresponding map for retrieving project data" still supposed to be here? Seems that `type lockfile` used to have another field that you then removed?

pkg/types.go
zlav: Thanks for the catch, fetcher type of `cargo` is actually correct, but I will change the docs.

buildtools/cargo/cargo.go
zlav: Yea good catch, that's exactly what happened. Removing this comment entirely now as `lockfile` isn't exported and it only has 1 field.

analyzers/rust/rust.go
zlav: Yea you're right, it will always return `Cargo.lock`. I'm hesitant to parse `Cargo.toml` here as it introduces another point of complexity and failure but doesn't really provide any information to the user. I agree that `Cargo.lock` isn't clear though so I will likely use the absolute path.

AfirstS (commented):
@AfirstS

pkg/types.go
AfirstS: buildtools/cargo/testdata/Cargo.lock

AfirstS (commented):
@AfirstS

PR #477, opened 2019-07-02
feat (dependencyManagement field) add dependencyManagement parsing.

zlav:
Maven pom files have the concept of a dependencyManagement which allows parent poms to specify the dependency versions for child poms. This fix adds support for that. This PR also created the `testing/helpers` package because I am sick of writing helper functions to assert on the internals of a graph.

I also duplicated part of the `testdata` folder from `analyzers/maven` into `buildtools/maven` in order to keep the test files closer to where they are being used. I also had to add a new pom file and it felt odd using two different directories.


buildtools/maven/pom_test.go
```diff
@@ -8,12 +8,14 @@ import (
 
 	"github.com/fossas/fossa-cli/buildtools/maven"
 	"github.com/fossas/fossa-cli/pkg"
+	"github.com/fossas/fossa-cli/testing/helpers"
 )
 
-func TestPomFileGraph(t *testing.T) {
+func testPomFileGraph(t *testing.T) {
```
golangcibot: `testPomFileGraph` is unused (from `deadcode`)

buildtools/maven/pom.go
```diff
@@ -55,13 +56,20 @@ func PomFileGraph(target, dir string) (graph.Deps, error) {
 	if err != nil {
 		return graph.Deps{}, err
 	}
+
+	// Aggregate `dependencies` and `dependencyManagement` fields.
+	dependencyList := pom.Dependencies
+	for _, dep := range pom.DependencyManagement {
```
golangcibot: S1011: should replace loop with `dependencyList = append(dependencyList, pom.DependencyManagement...)` (from `gosimple`)

buildtools/maven/pom_test.go
zlav: Fine... I'll refactor another test file.

testing/helpers/helpers.go
```diff
@@ -0,0 +1,62 @@
...
+		if fixture[i] == '\n' {
+			assert.NotEqual(t, uint8('\r'), fixture[i-1])
+		}
+	}
+}
+
+// PackageInTransitiveGraph searches a map (typically from Graph.Deps.Transitive)
+// for a package and returns it if it exists.
+func PackageInTransitiveGraph(packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
+	for id := range packages {
```
dchenk: Go needs to add methods such as `.Contains`, `.Find`, `.Map`, etc to slices and arrays... 🙄  Alas that's probably never going to happen

dchenk: to the `map` type too

buildtools/gradle/gradle_test.go
```diff
@@ -151,30 +152,6 @@ func MockGradle(t *testing.T, file string) gradle.ShellCommand {
...
-		if i == 0 {
-			continue
-		}
-		if fixture[i] == '\n' {
-			assert.NotEqual(t, uint8('\r'), fixture[i-1])
-		}
-	}
-}
-
 func findPackage(packages map[pkg.ID]pkg.Package, name, revision string) pkg.Package {
```
dchenk: you probably meant to delete this and `assertImport` from here

buildtools/maven/pom.go
```diff
@@ -55,13 +56,17 @@ func PomFileGraph(target, dir string) (graph.Deps, error) {
 	if err != nil {
 		return graph.Deps{}, err
 	}
+
+	// Aggregate `dependencies` and `dependencyManagement` fields.
+	dependencyList := append(pom.Dependencies, pom.DependencyManagement...)
```
dchenk: I think you need to de-dup after this.

dchenk: ...maybe, though maybe not

buildtools/maven/pom_test.go
```diff
@@ -36,15 +38,25 @@ func TestPomFileGraph(t *testing.T) {
...
 
+func TestPomFileDependencyManagement(t *testing.T) {
+	got, err := maven.PomFileGraph("dependencyManagement.xml", testPoms)
+	assert.NoError(t, err)
+	assert.Len(t, got.Direct, 1)
+	helpers.AssertPackageImport(t, got.Direct, "junit:junit", "3.8.2")
+
+	assert.Len(t, got.Transitive, 1)
+	packageOne := helpers.PackageInTransitiveGraph(got.Transitive, "junit:junit", "3.8.2")
+	assert.Len(t, packageOne.Imports, 0)
```
dchenk: this will always pass, even if the package isn't found, won't it?

testing/helpers/helpers.go
zlav: Yea, I think once generics land there will be more libraries for this.

buildtools/gradle/gradle_test.go
zlav: Yea I'll kill this

buildtools/maven/pom_test.go
zlav: Yea, I need to add `assert.NotEmpty(t, packageOne)` here. I'm leaving the length check because I want to ensure that no transitive deps are created.

buildtools/maven/pom.go
zlav: There's a chance that pom files are only allowed to use one of these, in which case I won't need to do any appending. I'm going to look into that.

PR #482, opened 2019-07-11
Discover gradle projects using non-groovy dialects

cnr:
This allows discovery of gradle projects using non-groovy dialects for their buildscripts. Each dialect uses a different file extension, e.g., `build.gradle.kts` for kotlin


cnr:
Added kotlin discovery test

PR #484, opened 2019-07-11
Use runtimeClasspath configuration for resolution

cnr:
When using version wildcards (e.g. `+`), the output of `gradle dependencies` doesn't capture the resolved version in the configurations we look for: `api`, `implementation`, ... .

The Java plugin for gradle internally generates a `runtimeClasspath` configuration, which _does_ include the resolved versions: 

This PR changes the set of default configurations to include `runtimeClasspath` instead of `compile`, `api`, and `implementation`.

### DIFF
--- a/docs/integrations/gradle.md
+++ b/docs/integrations/gradle.md
@@ -77,11 +77,11 @@ This options tells Fossa to scan for all sub projects of the specified module. T
 
 This option takes a comma delimited list of configurations to include in the dependency scan. Fossa includes a few configurations by default but allows the user to specify any specific configurations they are interested in.
 
-The default list of configurations is: `compile, api, implementation, compileDependenciesMetadata, apiDependenciesMetadata, implementationDependenciesMetadata`
+The default list of configurations is: `runtimeClasspath, compileDependenciesMetadata, apiDependenciesMetadata, implementationDependenciesMetadata`
 
 Example:
 ```yaml
-    configuration: implementation,implementationTest,customConfiguration
+    configuration: runtimeClasspath,testRuntimeClasspath,customConfiguration
 ```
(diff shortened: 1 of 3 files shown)


zlav:
@cnr I'm convinced with adding `runtimeClasspath`, but is there a reason for us to remove `compile`, `api`, and `implementation`? I understand the chart you posted but I'm thinking that this might open up the potential to regress.

cnr:
When we keep those, we end up with two dependencies listed: one with the wildcard intact, and one with the wildcard resolved.

It looks like the `runtimeClasspath` machinery was [documented over two years ago](https://github.com/gradle/gradle/blame/889f56d074626dd56b6894e2d287f01845be231b/subprojects/docs/src/docs/userguide/java_library_plugin.adoc#L115), so we should be safe. `runtimeClasspath` always includes `api`, `implementation` and `compile` dependencies

docs/integrations/gradle.md
```diff
@@ -77,11 +77,11 @@ This options tells Fossa to scan for all sub projects of the specified module. T
 
 This option takes a comma delimited list of configurations to include in the dependency scan. Fossa includes a few configurations by default but allows the user to specify any specific configurations they are interested in.
 
-The default list of configurations is: `compile, api, implementation, compileDependenciesMetadata, apiDependenciesMetadata, implementationDependenciesMetadata`
+The default list of configurations is: `runtimeClasspath, compileDependenciesMetadata, apiDependenciesMetadata, implementationDependenciesMetadata`
```
zlav: I think the default list now consists of only `compileClasspath, runtimeClasspath`, although I would still mention that we scan the other configurations

PR #487, opened 2019-07-23
Fix WalkUp to be OS-agnostic

cnr:


cnr:
It would probably require running integration tests on the various platforms, similar to how golang tests `filepath.Walk`

PR #488, opened 2019-07-24
feat (gomodules scanning) scan go.mod and go.sum for dependencies.

zlav:
The precursor to the improved fallbacks PR. This is intended to only give users the strategy option for `go.mod` until we expand the fallback feature in an upcoming PR.

Features:
1. Scanning `go.mod` gives deps.
2. Scanning `go.sum` gives deps.

Tricky things:
- Parsing `go.mod`. Refer to the `testdata/go.mod` file to see all supported cases.
- Parsing `go.sum`. Every other line has a duplicate revision with `/go.mod` appended. This code skips these.

### DIFF
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -23,6 +23,11 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	m := a.Module
 	log.Debugf("%#v", m)
 
+	switch a.Options.Strategy {
+	case "lockfile:gomodules":
+		return gomodules.ModGraph("go.mod")
+	}
+
 	// Get Go project.
 	project, err := a.Project(m.BuildTarget)
 	if err != nil {
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -25,7 +25,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 
 	switch a.Options.Strategy {
 	case "lockfile:gomodules":
-		return gomodules.ModGraph("go.mod")
+		return gomodules.ModGraph("")
 	}
 
 	// Get Go project.
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -23,11 +23,6 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	m := a.Module
 	log.Debugf("%#v", m)
 
-	switch a.Options.Strategy {
-	case "lockfile:gomodules":
-		return gomodules.ModGraph("")
-	}
-
 	// Get Go project.
 	project, err := a.Project(m.BuildTarget)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


buildtools/gomodules/gomodules.go
```diff
@@ -121,8 +229,9 @@ func goModuleList(path string) (string, error) {
 // return it as this means version is v0.0.0.
 func extractRevision(version string) string {
 	split := strings.Split(version, "-")
-	if len(split) < 3 {
-		return strings.TrimSuffix(version, "+incompatible")
+	resolved := version
+	if len(split) > 2 {
+		resolved = split[2]
```
dchenk: Another case... There is `// indirect` following a gitSHA, such as:
```
require (
	golang.org/x/net v0.0.0-20190310074541-c10a0554eabf // indirect
)
```

buildtools/gomodules/gomodules.go
```diff
@@ -55,6 +57,112 @@ func New(dir string) (Resolver, error) {
 	return resolver, nil
 }
 
+// ModGraph returns the dependencies found in a `go.mod` file.
+// We cannot resolve a graph so we make all dependencies direct.
+func ModGraph(filename string) (graph.Deps, error) {
+	modFile := filename
+	if modFile == "" {
+		modFile = "go.mod"
```
golangcibot: ineffectual assignment to `modFile` (from `ineffassign`)

buildtools/gomodules/gomodules.go
```diff
@@ -55,6 +57,112 @@ func New(dir string) (Resolver, error) {
...
+}
+
+// SumGraph returns the dependencies found in a `go.sum` file. We cannot
+// resolve a graph so we make all dependencies direct. Note `go.sum`
+// includes dependencies which are not used and this may be inaccurate.
+// https://github.com/golang/go/wiki/Modules#is-gosum-a-lock-file-why-does-gosum-include-information-for-module-versions-i-am-no-longer-using
+func SumGraph(filename string) (graph.Deps, error) {
+	sumFile := filename
+	if sumFile == "" {
+		sumFile = "go.sum"
```
golangcibot: ineffectual assignment to `sumFile` (from `ineffassign`)

buildtools/gomodules/gomodules.go
dchenk: actually nevermind

analyzers/golang/analyze.go
```diff
@@ -23,6 +23,11 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	m := a.Module
 	log.Debugf("%#v", m)
 
+	switch a.Options.Strategy {
+	case "lockfile:gomodules":
```
dchenk: why "gomodules" and not "gomod" or "go.mod"? Or do you intend to combine the deps listed in both files as the "gomodules" strategy?

zlav: Good question. For one, I could probably have left this part out, I'm not entirely sure what this will look like until we do fallbacks. The only reason I did this is to follow the pattern of the other strategies which are `manifest:gomodules`, `manifest:dep`, etc. I think `go.mod` fits more with the concept of a manifest file and `go.sum` with a lockfile so I will also be changing that.

zlav: Your comment brings up a good point though, I'm going to remove this block.

PR #489, opened 2019-07-24
errors (no API key) Common error when users forget to add an API key. 

zlav:
Add an error that looks like the following when users forget to add an API key:
Screen Shot 2019-07-24 at 3 15 06 PM

What is done here:
1. Rework `setContext` to take a boolean that sets the API key. This keeps code clean.
2. Change all instances of `setContext`.
3. Remove some unused mess inside of `report.go`.

### DIFF
--- a/cmd/fossa/setup/setup.go
+++ b/cmd/fossa/setup/setup.go
@@ -2,15 +2,14 @@
 package setup
 
 import (
-	"github.com/urfave/cli"
-
 	"github.com/fossas/fossa-cli/api/fossa"
 	"github.com/fossas/fossa-cli/cmd/fossa/display"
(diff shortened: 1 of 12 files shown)


cmd/fossa/setup/setup.go
```diff
@@ -26,7 +25,13 @@ func SetContext(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	fossa.SetAPIKey(config.APIKey())
+
+	if setAPIKey {
+		err = fossa.SetAPIKey(config.APIKey())
+		if err != nil {
```
golangcibot: nilness: tautological condition: non-nil != nil (from `govet`)

PR #493, opened 2019-07-26
Add setup.py scanning

cnr:
Adds `setup.py` scanning as a fallback when `requirements.txt` isn't present


analyzers/python/python.go
```diff
@@ -163,15 +163,29 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
...
-			Transitive: fromImports(imports),
-		}, nil
+
+		setupPyPath := filepath.Join(a.Module.Dir, "setup.py")
+		reqs, setupErr := pip.FromSetupPy(setupPyPath)
+		if setupErr == nil {
+			return requirementsToDeps(reqs), nil
+		}
+
+		err.Troubleshooting = err.Troubleshooting + " OR " + setupErr.Troubleshooting
```
cnr: We might want to add something to the errors package for combining two errors for fallback situations like these. It could, for example, combine the `Troubleshooting` with "OR", cause / message with "AND", ...

analyzers/python/python.go
```diff
@@ -163,15 +163,29 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	case "requirements":
 		fallthrough
 	default:
-		reqs, err := pip.FromFile(a.requirementsFile(a.Module))
-		if err != nil {
-			return graph.Deps{}, err
+		requirementsPath := a.requirementsFile(a.Module)
```
zlav: 👍

analyzers/python/python.go
```diff
@@ -163,15 +163,29 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 	case "requirements":
```
zlav: We should add an explicit `setuptools` case here that only runs `FromSetupPy`. Basically:
```go
case "setuptools":		
                 setupPyPath := filepath.Join(a.Module.Dir, "setup.py")
		reqs, setupErr := pip.FromSetupPy(setupPyPath)
		if setupErr == nil {
			return requirementsToDeps(reqs), nil
		}

		err.Troubleshooting = err.Troubleshooting + " OR " + setupErr.Troubleshooting

		return graph.Deps{}, err
```
It can be really helpful to mandate this. We may also want to modify the `requirements` case as well so that it no longer fallsthrough.

analyzers/python/python.go
zlav: Yea I agree with your comment from earlier. This is fine for now... but we should get a solution before this gets out of control.

buildtools/pip/testdata/setup.py
```diff
@@ -0,0 +1,8 @@
+from setuptools import setup, find_packages
+
+setup(
```
zlav: Another accepted and commonly used method to specify dependencies is like this:
```
    install_requires=['Cython>=0.25',
                      'h5py>=2.6',
                      'matplotlib>=3.0',
                      'numpy>=1.15',
                      'pandas>=0.19',
                      'scipy>=0.18',
                      'scikit-learn',
                      'scikit-image==0.14.2',
                      'seaborn>=0.7',
                      'spacy>=2.0',
                      'tqdm',
                      'tabulate>=0.7',
                      'tensorflow==1.12',
                      'PyYAML>=3.12'
                      ],
```
Currently the parser panics when trying to read the above data. Let me know if you want some examples.

buildtools/pip/pip.go
```diff
@@ -12,21 +13,37 @@ import (
 	"github.com/fossas/fossa-cli/files"
 )
 
-// TODO: add a Python sidecar that evaluates `setup.py`.
```
zlav: Finally TODOing!

buildtools/pip/testdata/setup.py
cnr: it actually doesn't panic -- we don't split by lines when loading the file, and just skip the `\n` characters while parsing. I'll add it to the test `setup.py` to make that expectation explicit, though

analyzers/python/python.go
cnr: should the combined requirements-fall-back-to-setup.py be used as a separate default?

buildtools/pip/testdata/setup.py
zlav: Understood, I tested this and found a panic but it appears to be happening in a different place. Thanks for the explanation.

PR #494, opened 2019-07-26
Publish homebrew formula

cnr:
This configures goreleaser to publish to our homebrew tap

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,8 +43,7 @@ The following commands will execute scripts to fetch and install the latest [Git
 
 ### Install with Homebrew (MacOS or Linux):
 ```bash
-brew tap fossas/tap
-brew install fossa
+brew install fossas/tap/fossa
 ```
 
 ### MacOS (Darwin) or Linux amd64:
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@
 
 The following commands will execute scripts to fetch and install the latest [GitHub Releases](https://github.com/fossas/fossa-cli/releases) on the corresponding operating system.
 
+### Install with Homebrew (MacOS or Linux):
+```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -41,6 +41,12 @@
 
 The following commands will execute scripts to fetch and install the latest [GitHub Releases](https://github.com/fossas/fossa-cli/releases) on the corresponding operating system.
 
+### Install with Homebrew (MacOS or Linux):
+```bash
+brew tap fossas/tap
```
zlav: [nit] We can simplify this to `brew install fossas/tap/fossa` so that users only need one line.

README.md
```diff
@@ -41,6 +41,12 @@
 
 The following commands will execute scripts to fetch and install the latest [GitHub Releases](https://github.com/fossas/fossa-cli/releases) on the corresponding operating system.
 
+### Install with Homebrew (MacOS or Linux):
```
zlav: TIL homebrew is on linux. Want to include a hyperlink here for Homebrew or is that overkill?

cnr:
We could construct a formula that follows [their guidelines](https://docs.brew.sh/Formula-Cookbook), which adds some maintenance overhead: we'd have to submit a PR to homebrew-core each time we publish a version, and bottle/upload compatible binaries. [Here's `goreleaser`'s formula on homebrew-core](https://github.com/Homebrew/homebrew-core/blob/master/Formula/goreleaser.rb)

It's also not clear whether they'd accept it:
> We frown on authors submitting their own work unless it is very popular.

I think having our own tap reaches the right tradeoffs

zlav:
@cnr yea that sounds like a nightmare especially at the rate we should be releasing. Tests finally passed!

PR #498, opened 2019-07-31
Init: allow use of the --project flag

cnr:


cmd/fossa/cmd/init/init.go
```diff
@@ -28,7 +28,7 @@ var Cmd = cli.Command{
 	Usage:  "Initialize a .fossa.yml configuration file",
 	Action: Run,
 	Flags: flags.WithGlobalFlags(flags.WithOptions([]cli.Flag{
```
zlav: How do you feel about making this `Flags: flags.WithGlobalFlags(flags.WithAPIFlags(flags.WithOptions([]cli.Flag{` like we do with analyze at `https://github.com/fossas/fossa-cli/blob/master/cmd/fossa/cmd/analyze/analyze.go#L26`?
I feel like this could future proof us for when someone asks for other config options.

PR #500, opened 2019-08-05
Fix parsing of gradle dependency tree

pascal-hofmann:
In one of my projects I have a gradle dependency tree that looks like this:

```
-api (n)
No dependencies

-runtime (n)
\--- org.jetbrains.kotlin:kotlin-stdlib:1.3.30 (n)
[…]
```

This breaks the gradle analyzer. It bails out with the error message 

> FATAL Could not analyze: bad depth: 1 -api (n) []string{"-api (n)", "-", "api (n)"}

This PR improves the regex used to match gradle dependencies based on the possible output generated by gradle (see https://github.com/gradle/gradle/blob/e975872ae2cf1d521b73f1107cae772ca09c63cd/subprojects/core/src/main/java/org/gradle/internal/graph/GraphRenderer.java)

### DIFF
--- a/buildtools/gradle/gradle.go
+++ b/buildtools/gradle/gradle.go
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
-	r := regexp.MustCompile("^([ `+\\\\|-]+)([^ `+\\\\|-].+)$")
+	r := regexp.MustCompile(`^((?:[ |\|]{1,1}    )*(?:[\\+|\\]{1,1}--- ))(.*)$`)
 	splitReg := regexp.MustCompile("\r?\n")
 	// Skip non-dependency lines.
 	var filteredLines []string
--- a/buildtools/gradle/gradle.go
+++ b/buildtools/gradle/gradle.go
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
-	r := regexp.MustCompile(`^((?:[ |\|]{1,1}    )*(?:[\\+|\\]{1,1}--- ))(.*)$`)
+	r := regexp.MustCompile(`^((?:[|+] +)*[\\+]--- )(.*)$`)
 	splitReg := regexp.MustCompile("\r?\n")
 	// Skip non-dependency lines.
 	var filteredLines []string
--- a/buildtools/gradle/gradle.go
+++ b/buildtools/gradle/gradle.go
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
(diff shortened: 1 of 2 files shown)


buildtools/gradle/gradle.go
```diff
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
-	r := regexp.MustCompile("^([ `+\\\\|-]+)([^ `+\\\\|-].+)$")
+	r := regexp.MustCompile("^((?:[ |\\|]{1,1}    )*(?:[\\+|\\\\]{1,1}--- ))(.*)$")
```
golangcibot: S1007: should use raw string (`...`) with regexp.MustCompile to avoid having to escape twice (from `gosimple`)

pascal-hofmann: This has been addressed now.

pascal-hofmann:
cla-assistant.io is down right now. Will sign the CLA once it is back up…

buildtools/gradle/gradle.go
```diff
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
-	r := regexp.MustCompile("^([ `+\\\\|-]+)([^ `+\\\\|-].+)$")
+	r := regexp.MustCompile(`^((?:[ |\|]{1,1}    )*(?:[\\+|\\]{1,1}--- ))(.*)$`)
```
dchenk:
```suggestion
	r := regexp.MustCompile(`^((?:[ |\|]{1,1}    )*(?:[\\+|\\]{1,1}--- ))`)
```
Since this reg expression is used only for a single line at a time, we don't need to add `(.*)$` at the end

dchenk:
```suggestion
	r := regexp.MustCompile(`^([ +\\|\x60-]{2,})([^ +\\|\x60-].+)$`)
```
We can just make sure that there are at least two characters at the beginning of the line that can't be a dependency name. Haven't tried this with the complete integration tests, but the the Gradle tests in this PR are currently failing.

pascal-hofmann: This will only work until someone accidentally uses a configuration name that matches the new regex.

`go test -count=1 -run TestGradleIntegration  ./...` succeeds on my local machine.  I assume there is something wrong with the CI system. See also the lower coverage of mercurial/svn, php, python/pip, etc.

dchenk: It's sort of unlikely someone will name something with more than one "-" or "/" at the beginning, but lest we go down a rabbit hole about this needlessly, I'm okay with the idea behind your regex, with a couple of cleanups as shown in my suggestion above. In particular, the quantifier `{1,1}` is essentially meaningless and redundant, and in the two character sets the pipe character appears twice (as `|` and `\|`). You can just have it there once as `|`. And to keep things easier to read, we'll express the four consecutive spaces using ` {4}` please. Thank you!

pascal-hofmann:
Hey, do I still need to do something to get this merged?

dchenk:
Apologies for the delayed response. I was going to investigate why the CI tests failed. I'd like to get this merged soon.

buildtools/gradle/gradle.go
```diff
@@ -160,7 +160,7 @@ func (s ShellCommand) DependencyTasks() ([]string, error) {
 //go:generate bash -c "genny -in=$GOPATH/src/github.com/fossas/fossa-cli/graph/readtree.go gen 'Generic=Dependency' | sed -e 's/package graph/package gradle/' > readtree_generated.go"
 
 func ParseDependencies(stdout string) ([]Dependency, map[Dependency][]Dependency, error) {
-	r := regexp.MustCompile("^([ `+\\\\|-]+)([^ `+\\\\|-].+)$")
+	r := regexp.MustCompile(`^((?:[|+] +)*[\\+]--- )(.*)$`)
```
dchenk:
```suggestion
	r := regexp.MustCompile(`^((?:[|+]? +)*[\\+]--- )(.*)$`)
```
If you could just commit this one small fix, we should be good :)

dchenk: @pascal-hofmann

PR #501, opened 2019-08-09
feat (clojure support) clojure support through leiningen

zlav:
This PR adds support for Clojure by shelling out to leiningen and reading `project.clj` files. 

Fallbacks: We first attempt to run `lein deps :tree` and then if that fails we parse a `project.clj` file.

I have left comments explaining less obvious or unintuitive parts of the code. Please refer to the test files to understand the parsing logic.

File Parsing:
This part is a little bit confusing, looking at the test example provides a very good insight into the design decision and what `brackets` does (It checks if we are in a deps block and then if we are in an individual deps block).


testing/helpers/helpers.go
```diff
@@ -60,3 +62,19 @@ func AssertPackageImport(t *testing.T, imports pkg.Imports, name, revision strin
...
+				if module.BuildTarget == target {
+					if module.Dir == directory {
+						return
+					}
+				}
+			}
+		}
+
+	}
+	assert.Fail(t, fmt.Sprintf("missing module `name: %s, type: %s, dir: %s, target: %s, in modules: %s", name, modType, directory, target, modules))
```
golangcibot: printf: Sprintf format %s has arg modules of wrong type []github.com/fossas/fossa-cli/module.Module (from `govet`)

errors/errors.go
```diff
@@ -68,13 +68,13 @@ func (e *Error) Error() string {
 	message = e.Message
 	if message == "" {
 		switch e.Type {
+		case NotImplemented:
```
zlav: This prevents the other error types from falling back to the notimplemented error. This was a bug from the errors PR.

analyzers/clojure/clojure.go
```diff
@@ -0,0 +1,126 @@
...
+	if err == nil && len(dependencies.Direct) > 0 {
+		return dependencies, nil
+	}
+	log.Warnf("FALLING BACK to parse clojure project file.\nLeiningen analysis could not be run, address the following error for more accurate results:\n%s", err.Troubleshooting)
+	log.Debug(err.Error())
+
+	// 3. Parse `project.clj` as best as we can.
+	// https://github.com/technomancy/leiningen/blob/master/sample.project.clj
+	dependencies, err = leiningen.ProjectFileDependencies(a.Module.Dir, a.Module.BuildTarget)
+	if err == nil {
```
zlav: The reason I do if err == nil is because it is more descriptive of the fallback process. It also makes adding another fallback method friendlier.

buildtools/leiningen/leiningen.go
```diff
@@ -0,0 +1,224 @@
...
+	Version    string `xml:"version"`
+}
+
+// id returns a maven locator from dependency information.
+func (d Dependency) id() string {
+	return d.GroupID + ":" + d.ArtifactID
+}
+
+// Shell implements Output and uses Cmd to generate the DependencyGraph.
+type Shell struct {
```
cnr: This feels like something that would be useful elsewhere, along with `ShellOutput`, to make behavior more uniform across analyzers

analyzers/clojure/clojure.go
```diff
@@ -0,0 +1,126 @@
...
+		return leiningen.ProjectFileDependencies(a.Module.Dir, a.Module.BuildTarget)
+	}
+
+	// 2. Use the lein command to find a dependency graph.
+	// https://cljdoc.org/d/leiningen/leiningen/2.9.1/api/leiningen.deps
+	dependencies, err := a.Lein.DependencyGraph(a.Module.BuildTarget)
+	if err == nil && len(dependencies.Direct) > 0 {
+		return dependencies, nil
+	}
+	log.Warnf("FALLING BACK to parse clojure project file.\nLeiningen analysis could not be run, address the following error for more accurate results:\n%s", err.Troubleshooting)
```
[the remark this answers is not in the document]
zlav: Agreed, it would be nice if this was a method in the error's package.

PR #504, opened 2019-08-14
feat (golang fallbacks) break go strategies down and fallback easier

zlav:
This PR adds support for cascading fallbacks in the go analyzer. It also breaks out the primary method of go analysis into its own file. As a result of adding fallbacks we also now have support for distinct strategies such as `dep` and `gomodules`.

Changes are pretty straightforward.
Fallbacks currently supported
- `go list`
- `gomodules`
- `dep`

During this PR I found that the code which handles running `go list` and comparing that information to the analyzer is so complex that it does not make sense to break it out into its own build tool today. I think we should do that down the road but due to the lack of testing, I am comfortable breaking it down into its own function.

Testing: This would require us to mock the go command which has been discussed offline and is not worth the time right now.

### DIFF
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -39,7 +39,7 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 		}
 		return dep.ManifestGraph("")
 	case "list":
-		return ListResolution(a)
+		return ListLockfileResolution(a)
 	}
 
 	// 2. Run `go list` and resolve using the lockfile.
--- a/analyzers/golang/analyze.go
+++ b/analyzers/golang/analyze.go
@@ -1,6 +1,8 @@
 package golang
 
 import (
+	"path/filepath"
+
 	"github.com/apex/log"
 
 	"github.com/fossas/fossa-cli/analyzers/golang/resolver"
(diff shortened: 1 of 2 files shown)


analyzers/golang/analyze.go
```diff
@@ -13,16 +15,86 @@ import (
...
+		depGraph, err := dep.ManifestGraph("")
+		if returnGraph(filepath.Join(a.Module.Dir, "Gopkg.toml"), err, len(depGraph.Direct)) {
+			return depGraph, nil
+		}
+	}
+
+	return graph.Deps{}, err
+}
+
+func returnGraph(file string, err error, dependencyCount int) bool {
```
zlav: @cnr I have found myself writing this function a lot. Lets somehow make it possible to only write this once in V2.

analyzers/golang/analyze.go
```diff
@@ -90,11 +162,6 @@ func (a *Analyzer) Analyze() (graph.Deps, error) {
 			return graph.Deps{}, err
 		}
 
-	// Resolve revisions by traversing the local $GOPATH and calling the package's
```
zlav: Removed because the code is listed as not implemented and we don't plan on implementing this.

analyzers/golang/analyze.go
cnr: agreed 👌

docs/integrations/golang.md
```diff
@@ -81,7 +81,10 @@ Predefined build tags:
 #### `strategy: <string>`
 
 Manually specify the golang package manager being used. If this option is set, it is recommended [lockfile](#LockfilePath:-<string>) and [manifest](#ManifestPath:-<string>) be set as well. A list of supported strategies is as follows:
-- ```manifest:gomodules, manifest:dep, manifest:gdm, manifest:glide, manifest:godep, manifest:govendor, manifest:vndr, gopath-vcs```
```
cnr: These are still valid strategies -- though I think I agree with removing them from the docs

PR #510, opened 2019-08-20
improve error messages when running commands

BurntSushi:
This PR attempts to make some failure modes in `fossa` a bit better. Specifically, we were running into an issue where `fossa` would report  only this error:

```
FATAL Could not analyze: could not run go list:  (exit status 1)
```

It wasn't quite clear, as a user, how to go about resolving this. This PR reflects the patch I added to `fossa` in order to diagnose the error. (In my case, it turned out to be a repository that had been removed, and `go list -json a/b/c` was failing because an underlying `git fetch` command was failing.) Specifically, this adds the command that was run to the error message, and also fixes a bug that was preventing `fossa`'s internal command helpers from correctly capturing stderr. The latter comes with a regression test.

### DIFF
--- a/buildtools/gocmd/go.go
+++ b/buildtools/gocmd/go.go
@@ -4,6 +4,7 @@ package gocmd
 
 import (
 	"encoding/json"
+	"fmt"
 	"strings"
 
 	"github.com/fossas/fossa-cli/errors"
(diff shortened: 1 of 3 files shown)


buildtools/gocmd/go.go
```diff
@@ -75,18 +75,25 @@ func (g *Go) List(pkgs, flags []string, vendor bool) ([]Package, error) {
...
+	}
+	stdout, stderr, err := exec.Run(cmd)
 	if err != nil && stdout == "" {
 		if strings.Contains(stderr, "build constraints exclude all Go files") {
 			// TODO: add better documentation around this error, and rename it to be
 			// more useful.
 			return nil, errors.New("bad OS/architecture target")
 		}
-		return nil, errors.Errorf("could not run go list: %s (%s)", strings.TrimSpace(stderr), err)
+		return nil, errors.Errorf(
```
zlav:
```suggestion
		return nil, &errors.Error{
			Cause: err,
			Type:  errors.Exec,
			Troubleshooting: fmt.Sprintf("Fossa could not run command `%v` in directory `%s`. Ensure that golang is installed and try running this command on your own.\nstdout: %s\nstderr: %s",
				append([]string{cmd.Name}, cmd.Argv...), cmd.Dir, stdout, strings.TrimSpace(stderr)),
			Link: "https://github.com/fossas/fossa-cli/blob/master/docs/integrations/golang.md#analysis",
		}
```
Feel free to edit this message as you see fit. We just introduced an improved errors package so that we can avoid this exact situation in the future and would like to use it wherever we can.

zlav:
This PR looks great! Thank you for also addressing the issue with `stderr` and for including tests. I added a comment about using our newly created errors package and once that change has been made this should be good to merge.

buildtools/gocmd/go.go
BurntSushi: Ah great! I just pushed this change. I pretty much used your exact suggestion with one tweak: I left `stdout` out of the message since this error is only triggered when `stdout` is empty.

zlav: Good call, thanks!

PR #520, opened 2019-08-23
feat (rpm scanning) support for system level and individual rpms.

### LINKED ISSUE #342: Runtime Dependency Checking
**Is your feature request related to a problem? Please describe.**
In some of our projects we are making use of Cgo. This bring in several runtime dependencies that must be met in order for the binary to be run. Currently, `fossa` isn't able to find these dependencies.

Providing runtime dependencies is solved by specifying packages that we depend on in our DEB/RPM packaging. While the dependencies are dynamically linked at runtime, we still need to understand the licenses of these dependencies.

**Describe the solution you'd like**
Support scanning of `debian/control` files and/or RPM `spec` files.
License information is available in via a dependencies [copyright file](https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/) or via `rpm -q $dep --queryformat '%{license}'`

**Describe alternatives you've considered**
Using `ldd` to resolve dynamically linked dependencies from a binary.
This wouldn't require knowledge of packaging *but* would require a way to query a license of a shared object file.

Using `nm` to resolve statically linked dependencies (if symbols haven't been stripped) from a binary.
You'd need a mechanism to resolve from symbols -> library -> license

**Additional context**
This solution will work for any binaries complied for Linux that have runtime dependencies.
It obviously won't work for:
 - Windows
 - Statically linked dependencies

zlav:
Closes #342 

This PR adds an RPM analyzer which finds all system-level RPMs, uploads them, and associates their project with them.

How this works.
System Level:
1. Run `rpm -qa --queryformat %{name},%{version},%{license}\n` to get all installed rpm deps.
2. Look through `/usr/share/licenses` for the corresponding license files and upload it.
3. If no files exists, then take the RPM declared license and upload it.

Package Level:
1. Run `rpm -q --queryformat %{name},%{version},%{license}\n package.rpm`.
2. TODO: Find all transitive dependencies.
3. Mirror steps 2 & 3 from system-level analysis to get dependencies.

Changes to tar package.
The changes implemented here to `tar.go` are minimal, but allow fossa to upload any string as an archive. This allows us to upload declared license information that can only be found from the command output. This required manually creating a file header and creating a reader from the string, besides that most of the code is duplicated from the rest of the tar package.

TODO: 
- Errors
- Docs
- Single RPM work and deps.


api/fossa/tar.go
```diff
@@ -84,6 +84,84 @@ func UploadTarball(dir string, dependency, rawLicenseScan, upload bool) (Locator
...
+	}
+	_, err = io.Copy(h, s)
+	if err != nil {
+		return nil, nil, err
+	}
+	// Close again to force a disk flush. Closing an *os.File twice is
+	// undefined, but safe in practice.
+	// See https://github.com/golang/go/issues/20705.
+
+	if err != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

buildtools/rpm/rpm.go
```diff
@@ -0,0 +1,330 @@
...
+	Upload           bool
+	RPM              func(...string) (string, string, error)
+	Yum              func(...string) (string, string, error)
+}
+
+type dependency struct {
+	name         string
+	revision     string
+	license      string
+	dependencies []dependency
```
golangcibot: `dependencies` is unused (from `structcheck`)

buildtools/rpm/rpm.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		log.Warn(err.Error())
+		return map[string][]string{}
+	}
+
+	deps := make(map[string][]string)
+	deps[target] = parseTransitive(output)
+	for _, dep := range deps[target] {
+		transitiveDependencies := recursiveInstall(dep, s)
+		if err != nil {
```
golangcibot: nilness: impossible condition: nil != nil (from `govet`)

buildtools/rpm/rpm.go
```diff
@@ -0,0 +1,363 @@
...
+			}
+		}
+		log.Warnf("license files for `%s` could not be found in `%s`. The declared license `%s` will be used: %s", dep.name, licenseDir, dep.license, err.Error())
+	}
+
+	return locator, nil
+}
+
+// Parse dependency from output format dep-name,version,license.
+func dependencyFromOutput(line string) dependency {
```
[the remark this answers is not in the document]
zlav: I'm not actually aware of a case where a dependency won't have a field filled, but I made sure to handle the event it does not. In the event it doesn't have one of these it will result in us logging a warning when we aren't able to find any dependencies.

PR #525, opened 2019-09-09
errors (wrong arguments) better errors when users manually specify modules.

zlav:
Example error when a user accidentally runs `fossa analyse`.
Screen Shot 2019-09-09 at 3 15 21 PM

Ideally this command will be deprecated in the future, but at the moment this is a good fix until we overhaul the commands.

### DIFF
--- a/config/keys.go
+++ b/config/keys.go
@@ -1,7 +1,7 @@
 package config
 
 import (
-	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strconv"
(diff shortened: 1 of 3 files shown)


michaelcoyote:
I'd really like it part of the error line.  e.g. `FATAL Could not parse modules: [analyse]` would be very clear.  

I agree it should be part of the troubleshooting output as well.

zlav:
@michaelcoyote done. I can't make it look exactly like you mentioned due to how the errors are added when it bubbles up.
Screen Shot 2019-09-09 at 3 53 58 PM

michaelcoyote:
LGTM_cat_02

PR #533, opened 2019-09-14
WIP feat (ruby graphtagging) basic support for graph tagging

zlav:
This is currently in an experimental state. I believe the changes required to introduce graph tagging are going to be significantly smaller than previously anticipated. I do however believe that these changes are starting to signal significant areas where we should try to modify the existing package graph data model.

Points of interest: 
- We are currently uploading the contents of `[]pkg.Import` and `map[pkg.ID]pkg.Package`. With graph tagging we can include a boolean field `Direct` in `pkg.Package` which would eliminate the need for `[]pkg.Import` altogether.


zlav:
Closing because everything we need is already supported.

PR #534, opened 2019-09-17
Implement new python analyzer

cnr:


analyzers/python/python.go
```diff
@@ -138,7 +138,7 @@ func (a *Analyzer) IsBuilt() (bool, error) {
 func (a *Analyzer) Analyze() (graph.Deps, error) {
 	switch a.Options.Strategy {
 	case "deptree":
-		tree, err := a.Pip.DepTree()
+		tree, err := a.Pip.DepTree(".")
```
cnr: This is clearly a bug in the original analyzer, but I don't know what the ramifications are when fixing it

PR #537, opened 2019-09-19
fix (rpm install) return stdout instead of zero value

zlav:
RPM commands often return `exit code 1` to signal an error even though nothing is output on `stderr` and the expected output is returned on `stdout`. Due to this, we must log that an error occurred but attempt to proceed with the information that we do have.

### DIFF
--- a/buildtools/rpm/rpm_test.go
+++ b/buildtools/rpm/rpm_test.go
@@ -1,11 +1,11 @@
 package rpm_test
 
 import (
-	"fmt"
 	"io/ioutil"
 	"testing"
 
 	"github.com/fossas/fossa-cli/buildtools/rpm"
+	"github.com/fossas/fossa-cli/errors"
 	"github.com/fossas/fossa-cli/testing/helpers"
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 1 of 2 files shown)


buildtools/rpm/rpm_test.go
```diff
@@ -99,7 +100,8 @@ func mockSingleRPM() rpm.Shell {
 				}
 				switch args[len(args)-1] {
 				case "dep-one":
-					return stringFile("testdata/licenses/dep-one-license")
+					stdout, stderr, _ := stringFile("testdata/licenses/dep-one-license")
```
golangcibot: Error return value of `stringFile` is not checked (from `errcheck`)

buildtools/rpm/rpm_test.go
```diff
@@ -115,7 +117,8 @@ func mockSingleRPM() rpm.Shell {
 			case "-qR":
 				switch args[1] {
 				case "dep-one":
-					return stringFile("testdata/dep-one-transitive")
+					stdout, stderr, _ := stringFile("testdata/dep-one-transitive")
```
golangcibot: Error return value of `stringFile` is not checked (from `errcheck`)

PR #539, opened 2019-09-26
fix (log file sync) defer file syncing until after the command finishes

zlav:
Currently, we are running `file.Sync()` after every log line. This PR ensures that `file.Sync()` is only called once when the program finishes. This speeds up all FOSSA CLI analysis, especially golang.

### DIFF
--- a/cmd/fossa/main.go
+++ b/cmd/fossa/main.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/apex/log"
 	"github.com/urfave/cli"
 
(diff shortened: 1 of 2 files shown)


cmd/fossa/main.go
```diff
@@ -42,6 +44,16 @@ var App = cli.App{
...
 func main() {
+	// Write the temp log file to disk after the command finishes.
+	defer func() {
+		err := display.SyncFile()
+		if err != nil {
+			log.Warnf("error writing to the log file: %s", err.Error())
+		}
+	}()
+
+	defer display.SyncFile()
```
golangcibot: Error return value of `display.SyncFile` is not checked (from `errcheck`)

PR #554, opened 2019-11-13
[FC-1212] create integration with new report endpoint

microsoftly:
Instead of doing report generation on the client, the backend now creates a sane report structure. This can be either printed to console (in the table format) or as json with `--json` command. This was tested locally with the tentative implementation for the BE report generation. The output is from the cli for the console printed reports
## Screenshots
Screen Shot 2019-11-13 at 2 00 41 PM
Screen Shot 2019-11-13 at 2 00 52 PM
## New Dependency
github.com/olekukonko/tablewriter


cmd/fossa/cmd/report/attribution.go
```diff
@@ -0,0 +1,116 @@
...
+func renderDepLicensesToString(licenses []fossa.AttributionDependencyLicense) string {
+	lics := make([]string, 0, len(licenses))
+	for _, lic := range licenses {
+		lics = append(lics, lic.Name)
+	}
+
+	return convertSliceOfStringsToBullets(lics)
+}
+
+func renderLicensesToString(licenses map[string]string) string {
```
golangcibot: `renderLicensesToString` is unused (from `deadcode`)

cmd/fossa/cmd/report/attribution.go
```diff
@@ -0,0 +1,107 @@
...
+		Project:  config.Project(),
+		Revision: config.Revision(),
+	}
+	attrReport, err := fossa.GetReport(locator)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to build report for for project %s:", locator)
+	}
+
+	fmt.Fprintf(os.Stdout, `Generated by fossa-cli (https://github.com/fossas/fossa-cli).
+Attribution for project %s, revision %s
```
zlav: This is a little difficult to read and would look better if we broke the formatted string out as a constant similar to how we've done with the templates in the other report files. maybe something like 
```
const attributionTemplate = `Generated by fossa-cli (https://github.com/fossas/fossa-cli).
Attribution for project %s, revision %s
Direct Dependencies
%s
Deep Dependencies
%s`
```

cmd/fossa/cmd/report/attribution.go
```diff
@@ -0,0 +1,107 @@
...
+		Fetcher:  config.Fetcher(),
+		Project:  config.Project(),
+		Revision: config.Revision(),
+	}
+	attrReport, err := fossa.GetReport(locator)
+	if err != nil {
+		return errors.Wrapf(err, "Unable to build report for for project %s:", locator)
+	}
+
+	fmt.Fprintf(os.Stdout, `Generated by fossa-cli (https://github.com/fossas/fossa-cli).
```
zlav: Because you are outputting to `os.Stdout` you can use `fmt.Printf` here.

cmd/fossa/cmd/report/attribution.go
```diff
@@ -0,0 +1,107 @@
...
+
+	return tableString.String()
+}
+
+func sortStringArray(strs []string) {
+	sort.StringSlice(strs).Sort()
+}
+
+func convertSliceOfStringsToBullets(strs []string) string {
+	sortStringArray(strs)
```
zlav: We should stay away from directory modifying the arguments themselves and treat them as constants. Instead here if we did the following we can remove the `sortStringArray` method and keep `strs` constant.
```suggestion
	sortedStrings := strs
	sort.Strings(sortedStrings)
```

microsoftly: Is this not just the same thing, but referring to a new variable? It still references the same arr.

PR #558, opened 2019-12-02
fix(json report) print json reports when provided json flag

microsoftly:
Fixes a bug where `report attribution --json` would print a formatted table. It now prints json as expected


liftM:
Can we get an integration test for this? We can pipe the output of calling this command on an example project to `jq` or something and see if it exits 1.

PR #560, opened 2019-12-07
feat (pipenv discovery)

zlav:
This PR adds discovery for pipenv which will be prioritized over requirements.txt files. This is a safe assumption to make as any project with a Pipfile should have a fully resolved graph that gives better information than requirements.txt.

This PR also takes care of documented behavior that was never implemented.

### DIFF
--- a/analyzers/python/python.go
+++ b/analyzers/python/python.go
@@ -96,7 +96,11 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 				"path": filename,
 				"name": moduleName,
 			}).Debug("constructing Python module")
-			relPath, _ := filepath.Rel(dir, filename)
+			relPath, err := filepath.Rel(dir, filename)
+			if err != nil {
+				return nil
+			}
+
 			modules[moduleDir] = module.Module{
 				Name:        moduleName,
 				Type:        pkg.Python,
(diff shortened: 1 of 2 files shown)


analyzers/python/python.go
```diff
@@ -84,6 +84,28 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
...
+				return nil
+			}
+
+			moduleName := filepath.Base(moduleDir)
+
+			log.WithFields(log.Fields{
+				"path": filename,
+				"name": moduleName,
+			}).Debug("constructing Python module")
+			relPath, _ := filepath.Rel(dir, filename)
```
golangcibot: Error return value of `filepath.Rel` is not checked (from `errcheck`)

analyzers/python/python.go
```diff
@@ -84,6 +84,32 @@ func Discover(dir string, options map[string]interface{}) ([]module.Module, erro
 			return err
 		}
 
+		if !info.IsDir() && (info.Name() == "Pipfile.lock" || info.Name() == "Pipfile") {
+			moduleDir := filepath.Dir(filename)
+			if modules[moduleDir].Options["strategy"] == "pipenv" {
```
cnr: this looks like it can panic

PR #566, opened 2020-01-08
archive -> archives

zlav:
https://goreleaser.com/deprecations/#archive


cnr:
I think there might be more to it than just changing `archive` to `archives` -- or at least their example does more:

> Change this:
> 
>     archive:
>       format: zip
>
> to this:
>
>     archives:
>       - id: foo
>         format: zip

zlav:
@cnr I thought the same thing when I saw the update but after looking at the [archives section](https://goreleaser.com/archive/) the id has a default value of `default` so I chose to leave it as that.

cnr:
looks like it still needs to be in a list, though:
```
archives:
  - format_overrides:
      - goos: windows
        format: zip```

PR #570, opened 2020-01-28
feat (better bazel) support for bazel deps command

zlav:
This PR enables users to specify a bazel target to run `bazel deps(target)` with and parse the results. This results in a significantly more accurate bazel integration that can selectively parse certain directories (vendor, third_party, etc...)


buildtools/bazel/bazel.go
```diff
@@ -1,16 +1,198 @@
...
+// Handle the npm_req field formats. Name and version are separated by "@" and "@" can be present in the name.
+func npmReqFromString(req string) (string, string) {
+	npmSpec := strings.Split(req, "@")
+
+	name, version := "", ""
+	switch len(npmSpec) {
+	case 3: // Handle requirements in the format "types/react-router-dom@1.0.0"
+		name = "@" + npmSpec[0] + npmSpec[1]
+		version = npmSpec[2]
+	case 2:
```
cnr: if the package name contains "@" but the package doesn't have a version, e.g.,
`@types/blah`
this is going to return
```
name = ""
version="types/blah"
```

buildtools/bazel/bazel.go
```diff
@@ -1,86 +1,199 @@
...
 
-	if id.Type == 0 {
-		log.Debugf("Unable to find dependency information for file %s", file)
-		return graph.Deps{}, nil
+	return packages, nil
+}
+
+func goNameFromRule(ruleName string, target string) string {
+	trimmedTarget := strings.TrimSuffix(target, "...")
+	trimmedRuleName := strings.TrimPrefix(ruleName, trimmedTarget)
```
golangcibot: ineffectual assignment to `trimmedRuleName` (from `ineffassign`)

PR #589, opened 2020-03-31
feat (npm dev) include npm development dependencies

zlav:
This PR adds support for a `fossa analyze --dev` flag that includes development dependencies. This flag is currently implemented only for nodejs projects.

The changes made here are that we now handle `npm ls` with `--production`, `yarn list` without `NODE_ENV=production`, and reading `devDependencies` from `package.json`.


cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -317,7 +319,7 @@ func Do(modules []module.Module, upload, rawModuleLicenseScan bool) (analyzed []
 			continue
 		}
 
-		analyzer, err := analyzers.New(m)
+		analyzer, err := analyzers.New(m, devDeps)
```
golangcibot: shadow: declaration of "err" shadows declaration at line 287 (from `govet`)

zlav: This was already here and intentional.

PR #597, opened 2020-04-21
Fall back to parent's version in ant analyzer

cnr:
We were already doing the fallback for `groupId`, and this adds an additional fallback for `version`


zlav:
Closing this PR for now. I believe we determined that the work required to do this correctly was vastly more complicated, please correct me if I am wrong here.

PR #602, opened 2020-05-19
fix: go.mod modules break when not in GOPATH

liftM:
When we analyze a module, we call `ListLockfileResolution` to attempt to do a recursive `go list` resolution. If this fails (e.g. if we're working outside of a `$GOPATH` because the project uses `go.mod`), we fall back to searching for specific tool manifests and analyzing those manifests.

Currently, the analyzer searches for the existence of these manifests in the `Module.Dir`. However, in general, the manifest file is not always in the same directory of the module - it may also be in a parent directory. This fix searches the project directory instead of the module directory, which correctly locates manifest files in parent directories.


analyzers/golang/testdata/gomod/.fossa.yml
```diff
@@ -0,0 +1,14 @@
+# Generated by FOSSA CLI (https://github.com/fossas/fossa-cli)
```
zlav, line 1: I don't think you meant to keep this file here.

liftM: This is intentional, although not used in the test. Running `fossa analyze` with this configuration file will reproduce the bug.

zlav: Ok, that's fine, I don't feel strongly about removing it.

PR #607, opened 2020-06-15
Add CLI Flag to Skip TLS Verification

rcsalome:
This adds a simple '--insecure' flag that allows the cli to interact
with a Fossa server using self-signed certificate.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -3,6 +3,7 @@ package api
 
 import (
 	"bytes"
+	"crypto/tls"
 	"encoding/json"
 	"io"
 	"io/ioutil"
@@ -26,6 +27,14 @@ var defaultClient = http.Client{
 		DisableKeepAlives: true,
 		Proxy:             http.ProxyFromEnvironment,
 	},
+}
+
+var (
+	insecureTLS bool = false
+)
+
+func SetInsecureTLS(insecureTLSflag bool) {
+	insecureTLS = insecureTLSflag
 }
 
 // Get is a convenience method for MakeAPIRequest.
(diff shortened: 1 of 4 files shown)


zlav:
Hi @rcsalome does this PR attempt to fix an issue that you are seeing with the fossa-cli? We would prefer to keep the configuration options limited unless there is a necessary reason to expand them.

rcsalome:
Hey @zlav. This is a workaround for when a `fossa analyze` occurs on a Fossa Server using an un-trusted TLS certificate.

### Problem
Here is an example of an error:
```sh
~/go/bin/fossa analyze  --endpoint https://MY_HOSTNAME/
FATAL Error during upload: could not upload: could not send API HTTP request: Post "https://MY_HOSTNAME/api/builds/custom?branch=master&locator=custom%2Bhttps%3A%2F%2F......REDACTED-OUTPUT.......": x509: certificate signed by unknown authority
```

Here is the same command using the `--insecure` flag.
```sh
~/go/bin/fossa analyze --insecure --endpoint https://MY_HOSTNAME/

    View FOSSA Report:
    https://MY_HOSTNAME/api/builds/custom?branch=master&locator=custom%2Bhttps%3A%2F%2F......REDACTED-OUTPUT.......

```

### Examples
There are many ways around this, including using a valid certificate, but this flag was useful in testing a Fossa Server installation over HTTPS before it was fronted with a valid certificate.

1. The inspiration came from `curl` to use the same flag `--insecure`: https://curl.haxx.se/docs/sslcerts.html
> Tell libcurl to not verify the peer. With libcurl you disable this with curl_easy_setopt(curl, CURLOPT_SSL_VERIFYPEER, FALSE);
With the curl command line tool, you disable this with -k/--insecure.

2. However other tools like the `awscli` use `--no-verify-ssl`: https://awscli.amazonaws.com/v2/documentation/api/latest/reference/index.html
> By default, the AWS CLI uses SSL when communicating with AWS services. For each SSL connection, the AWS CLI will verify SSL certificates. This option overrides the default behavior of verifying SSL certificates.
--no-verify-ssl (boolean)


### Next Steps
I found the above option useful in my scenario, so I decided to open a PR. However, I'm not too tied to this PR, so I'm ok with whatever you decide 🙂

meghfossa:
Thank you for PR submission @rcsalome. 

I'm closing this PR as, we are migrating https://github.com/fossas/spectrometer source code at tip of master, as we move towards having 3.0 release. I've added the need to address "--no-verify-ssl or --insecure" workflow in our backlog.

PR #611, opened 2020-06-30
Git contrib count

scruffystuffs:
A few assumptions are made here:

- We always shell out from the working directory.  As far as I can tell, this is required because of the config file.
- We only upload once per run, assuming that the git repository is in the current folder.  It would also find a higher-level git repo, but that seems unlikely to encounter normally.
- A failing contributor check is logged, but only at the debug level, and does not hurt the process.
- `git log ...` should not take longer than 10s.  In a stress test, 42,580 commits were processed in under 6 seconds, so 10 seconds seems reasonable even for extreme cases.

### DIFF
--- a/api/fossa/upload.go
+++ b/api/fossa/upload.go
@@ -116,3 +116,25 @@ func Upload(title string, locator Locator, options UploadOptions, data []SourceU
 
 	return ReadLocator(unmarshalled.Locator), nil
 }
+
+// UploadContributors posts contributor data to the project decribed by locator
+func UploadContributors(data map[string]string, locator Locator) {
+	payload, err := json.Marshal(data)
+	if err != nil {
+		log.Debugf("Failed to marshal contributor map: %s", err)
+	}
+	q := url.Values{}
+	q.Add("locator", locator.String())
+	endpoint, err := url.Parse("api/contributors?" + q.Encode())
+	if err != nil {
+		log.Debugf("Failed to generate contributor URL: %s", err)
+	}
+	body, statusCode, err := Post(endpoint.String(), payload)
+	if err != nil {
+		log.Debugf("Failed to POST contributor data: %s", err)
+	}
+	if statusCode != 200 {
+		log.Debugf("FOSSA reported an issue while uploading contributor data: %s", body)
+	}
+	return
+}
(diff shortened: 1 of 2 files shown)


api/fossa/upload.go
```diff
@@ -116,3 +116,25 @@ func Upload(title string, locator Locator, options UploadOptions, data []SourceU
...
+	}
+	body, statusCode, err := Post(endpoint.String(), payload)
+	if err != nil {
+		log.Debugf("Failed to POST contributor data: %s", err)
+	}
+	if statusCode != 200 {
+		log.Debugf("FOSSA reported an issue while uploading contributor data: %s", body)
+	}
+	return
+}
```
kitified, line 140: Do we really want to continue this function on these errors? Seems strange to eg fail to encode data but then still try to POST it.

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -363,8 +375,53 @@ func uploadAnalysis(normalized []fossa.SourceUnit) error {
...
+		return fossa.Locator{}, err
 	}
 	fmt.Println(locator.ReportURL())
-	return nil
+	return locator, nil
+}
+
+func fetchGitContibutors() map[string]string {
+	year, month, day := time.Now().UTC().AddDate(0, 0, -90).Date()
+	fmtSince := fmt.Sprintf("%d-%d-%d", year, month, day)
```
kitified, line 386:
```suggestion
	fmtSince := time.Now().UTC().AddDate(0, 0, -90).Format("2006-01-02")
```

cmd/fossa/cmd/analyze/analyze.go
```diff
@@ -363,8 +375,53 @@ func uploadAnalysis(normalized []fossa.SourceUnit) error {
...
+	output, _, err := exec.Run(cmd)
+	if err != nil {
+		log.Debugf("Failed to run 'git log': %s", err.Error())
+		return nil
+	}
+
+	contribList := string(output)
+
+	contributorDates := make(map[string]string)
+	for _, line := range strings.Split(contribList, "\n") {
```
kitified, line 405: Splitting on `\n` leaves any `\r` in the output (if present). It's recommended to instead use `bufio.Scanner`, which scans line-by-line by default. Example:
```go
scanner := bufio.NewScanner(bytes.NewReader(output))
for scanner.Scan() {
  line := scanner.Text()
  // ...
}
```

api/fossa/upload.go
scruffystuffs: Oops... forgot to add the early returns.

PR #623, opened 2020-08-10
fix (wrong config file path field)

zlav:
## Description
Some of the language docs reccomend using a `dir` field instead of a `path` field 

I'm not entirely sure how this happened but I suspect it was a combination of:
- `dir` being used in the code to describe this field
- `path` rarely being used in a critical way so users' projects would not break if they did use `dir`.


scruffystuffs (approved):
Easy peasy.

PR #629, opened 2020-09-29
Update gradle.md

rlfagan:
## Description
[ticketID](https://fossa.atlassian.net/browse/ticketID)

_... add some insightful info here about the problem/and or changes_

## Acceptance Criteria
_... bulleted list of criteria that *you* used to determine that the work was correct and finished. This may differ from the AC in the ticket_

## Testing plan
_... list of steps for verifying the AC. If there are edge cases, please add headers with each named case with context_

## Screenshots
_... add screenshots here. Explicitly name before and after titling if relevant_

## Risks
_... enumerate list of known or suspected risks that may come from this change_

## Notes
_... anything that couldn't be covered in the other sections_

### DIFF
--- a/docs/integrations/gradle.md
+++ b/docs/integrations/gradle.md
@@ -31,6 +31,8 @@ analyze:
       path: path-to-build.gradle
       target: subproject
       type: gradle
+       options:
+          online: true
 ```
 
 ## Options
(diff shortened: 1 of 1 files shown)


zlav:
Addressed in PR #632

PR #631, opened 2020-11-05
Handle multi module maven builds without distinct pom files

zlav:
## Description
This PR addresses an issue that we have seen where the FOSSA CLI is unable to scan Maven builds that do not have distinct pom files. Specifically if the output from `mvn dependency:tree` has multiple dependency trees the maven parser will panic.

## Acceptance Criteria
Accepted on being able to successfully parse the problematic output.

## Testing plan
I tested this on a working build of https://github.com/biojava/biojava and included a test fixture for this issue to ensure that it doesn't happen again going forward.

Regression potential: I don't see any potential for regressions because the functionally is strictly an addition to existing functionality, but we should keep an eye on the parser.

## Limitations
None that I am aware of.

### DIFF
--- a/analyzers/ruby/integration_test.go
+++ b/analyzers/ruby/integration_test.go
@@ -96,6 +96,6 @@ var projects = []fixtures.Project{
 	fixtures.Project{
 		Name:   "vagrant",
 		URL:    "https://github.com/hashicorp/vagrant",
-		Commit: "c1b695c47119f4294d92152f0c8f8da27f657660",
+		Commit: "f36ece40e227539c0f07e52c56c617bae6586d6c",
 	},
 }
--- a/buildtools/maven/maven.go
+++ b/buildtools/maven/maven.go
@@ -163,6 +163,8 @@ func ParseDependencyTree(stdin string) (graph.Deps, error) {
 	r := regexp.MustCompile("^\\[INFO\\] ([ `+\\\\|-]*)([^ `+\\\\|-].+)$")
 	moduleBreak := regexp.MustCompile("^\\[INFO\\] -+<.+>-+")
 	splitReg := regexp.MustCompile("\r?\n")
+
+	// This loop is parsing maven output and separating each individual module.
 	for _, line := range splitReg.Split(stdin, -1) {
 		if line == "[INFO]" || line == "[INFO] " || line == "[INFO] ------------------------------------------------------------------------" {
 			started = false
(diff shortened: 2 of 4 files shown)


buildtools/maven/maven.go
```diff
@@ -171,6 +172,14 @@ func ParseDependencyTree(stdin string) (graph.Deps, error) {
 		if strings.HasPrefix(line, "Download") {
 			continue
 		}
+
+		if strings.Contains(line, "--------<") && len(filteredLines) > 0 {
```
liftM, line 176: Does `--------<` always work as an indicator? I believe the formatting of the output is `--< whatever the module name is >--` and the longer the module name the shorter the surrounding `--` markers.

liftM (changes requested):
Parsing output is a nightmare in Go.

Can you walk me through what this PR is actually doing mechanically?

analyzers/ruby/integration_test.go
```diff
@@ -96,6 +96,6 @@ var projects = []fixtures.Project{
 	fixtures.Project{
 		Name:   "vagrant",
 		URL:    "https://github.com/hashicorp/vagrant",
-		Commit: "c1b695c47119f4294d92152f0c8f8da27f657660",
+		Commit: "f36ece40e227539c0f07e52c56c617bae6586d6c",
```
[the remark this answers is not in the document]
zlav: It's not relevant to the PR, but it's relevant to fixing the ruby acceptance tests so that I can merge it. I don't know the root cause, but for some reason cloning the old commit, hash started to only download directories and not the files needed to scan such as `Gemfile`. I don't know why this was happening in CI, but it was consistently failing.

buildtools/maven/maven.go
zlav: You're correct that the longer the module name is the shorter the surrounding markers are. This solution worked for the 3 or 4 examples that each had 10-100 modules. It might be worth creating my own maven module with a larger name.

liftM: What if we replace this with a prefix regexp that matches something like `^\[INFO\] -+<`?

buildtools/maven/maven.go
```diff
@@ -179,39 +188,72 @@ func ParseDependencyTree(stdin string) (graph.Deps, error) {
...
-	imports, deps, err := ReadDependencyTree(filteredLines, func(line string) (int, Dependency, error) {
-		log.WithField("line", line).Debug("parsing output line")
-		matches := r.FindStringSubmatch(line)
-		depth := len(matches[1])
-		if depth%3 != 0 {
-			// Sanity check
-			log.WithField("depth", depth).Fatal("bad depth")
+
+	totalImports := make(map[Dependency]int)
+	depGraphMap := make(map[Dependency]map[Dependency]int)
```
liftM, line 197: For a set of `T`, the idiom is to use a `map[T]bool`.

liftM (approved):
LGTM after nits from the call.

Notes:
- Break out the aggregation of modules into a separate loop
- Add some high-level comments re: what the aggregation is doing

PR #632, opened 2020-11-10
Update docs and help output for flags that only effective on project creation.

zlav:
## Description
This PR clarifies the `team`, `title`, and `policy` flags which only have affect a project when it is created.

Other changes have been made as well:
- Fix to the sbt documentation that incorrectly told users to specify the `buildt.sbt` file as their path.
- Update to the maven docs to include an example of using the `cmd` option.
- Remove the `License` field from the `fossa test` output. A breaking change in the FOSSA API was made that prevents us from getting the data. In the future, we can work with a new endpoint but for now, I believe it is best to remove the confusing text.

### DIFF
--- a/cmd/fossa/cmd/test/test.go
+++ b/cmd/fossa/cmd/test/test.go
@@ -24,7 +24,7 @@ Detailed issue data cannot be read. This could be related to using a push-only A
 ========================================================================
 {{$type}}
 ========================================================================
-Dependency	Revision	{{if or (eq $type "Flagged by Policy") (eq $type "Denied by Policy") -}} License {{- end}}
+Dependency	Revision	{{if or (eq $type "Flagged by Policy") (eq $type "Denied by Policy") -}} {{- end}}
 {{- range $i, $issue := $issues}}
 {{$issue.Name}}	{{$issue.Revision}}
 {{- end}}
--- a/cmd/fossa/cmd/test/test.go
+++ b/cmd/fossa/cmd/test/test.go
@@ -26,7 +26,7 @@ Detailed issue data cannot be read. This could be related to using a push-only A
 ========================================================================
 Dependency	Revision	{{if or (eq $type "Flagged by Policy") (eq $type "Denied by Policy") -}} License {{- end}}
 {{- range $i, $issue := $issues}}
-{{$issue.Name}}	{{$issue.Revision}}	{{$issue.Rule.License }}
+{{$issue.Name}}	{{$issue.Revision}}
 {{- end}}
 {{end}}
 `
--- a/cmd/fossa/flags/flags.go
+++ b/cmd/fossa/flags/flags.go
@@ -49,7 +49,7 @@ var (
 	Endpoint        = "endpoint"
 	EndpointF       = cli.StringFlag{Name: Short(Endpoint), Usage: "the FOSSA server endpoint (default: 'https://app.fossa.com')"}
 	Title           = "title"
-	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project. (default: the project name)"}
+	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project, valid ONLY on project creation (default: the project name)"}
 	Fetcher         = "fetcher"
 	FetcherF        = cli.StringFlag{Name: Short(Fetcher), Usage: "type of fetcher to use for fossa. (default: 'custom')"}
 	Project         = "project"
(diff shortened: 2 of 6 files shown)


cmd/fossa/flags/flags.go
```diff
@@ -49,7 +49,7 @@ var (
 	Endpoint        = "endpoint"
 	EndpointF       = cli.StringFlag{Name: Short(Endpoint), Usage: "the FOSSA server endpoint (default: 'https://app.fossa.com')"}
 	Title           = "title"
-	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project. (default: the project name)"}
+	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project, valid ONLY on project creation (default: the project name)"}
```
liftM, line 52: Be more precise than "valid on". What does it mean for a title to be _valid?_

I would suggest "the title of the FOSSA project (applies only to new projects)".

cmd/fossa/flags/flags.go
```diff
@@ -65,9 +65,9 @@ var (
 	Link            = "link"
 	LinkF           = cli.StringFlag{Name: ShortUpper(Link), Usage: "a link to attach to the current build"}
 	Team            = "team"
-	TeamF           = cli.StringFlag{Name: ShortUpper(Team), Usage: "this repository's team inside your organization"}
+	TeamF           = cli.StringFlag{Name: ShortUpper(Team), Usage: "this repository's team inside your organization, valid ONLY on project creation"}
 	Policy          = "policy"
-	PolicyF         = cli.StringFlag{Name: Policy, Usage: "the policy to assign to this project in FOSSA"}
+	PolicyF         = cli.StringFlag{Name: Policy, Usage: "the policy to assign to this project in FOSSA, valid ONLY on project creation"}
```
liftM, line 70: Similarly here, I would prefer the suffix to be " (applies only to new projects)".

docs/integrations/sbt.md
```diff
@@ -14,14 +14,14 @@ In addition, SBT requires the following plugin to be installed:
...
-Manual: Add an `sbt` module with the path to the `build.sbt` file in your project. Set target to `<project>:<configuration>` where project is the desired sbt project and configuration is the configuration analysis is desired for.
+Manual: Add an `sbt` module with the path to the directory where a `build.sbt` file is located in your project. Set target to `<project>:<configuration>` where project is the desired sbt project and configuration is the configuration analysis is desired for.
 
 ```yaml
 analyze:
   modules:
     - name: your-sbt-project
       type: sbt
-      path: build.sbt
+      path: .
```
liftM, line 24: Can you also double-check that the analyzer correctly sets this to the directory (rather than the `build.sbt` file) when `fossa init` runs?

docs/user-guide.md
```diff
@@ -213,18 +204,22 @@ Analyzes the project for a list of its dependencies, optionally uploading the re
...
+| `--revision`    | `-r`  | Configuration value for [revision](/docs/config-file.md/#revision-optional).            |
+| `--endpoint`    | `-e`  | Configuration value for [endpoint](/docs/config-file.md/#endpoint-optional).            |
+| `--output`      | `-o`  | Output `fossa analyze` results to stdout.                                               |
+| `--server-scan` |       | Run a server side dependency scan on raw modules.                                       |
+| `--dev`         |       | Include development dependencies. CAUTION: valid only for nodejs projects.              |
+| `--debug`       |       | Print debugging information to stderr.                                                  |
+| `--help`        | `-h`  | Print a help message.                                                                   |
+| `--team`        | `-T`  | Connect the project with the specified team in FOSSA. Valid only on project creation.   |
+| `--policy`      |       | Connect the project with the specified policy in FOSSA. Valid only on project creation. |
+| `--title`       | `-t`  | Set the title that appears in the FOSSA UI. Valid only on project creation              |
```
liftM, line 220: See above comments re: phrasing. I'd rephrase as "Applies only to new projects.".

cmd/fossa/cmd/test/test.go
```diff
@@ -26,7 +26,7 @@ Detailed issue data cannot be read. This could be related to using a push-only A
 ========================================================================
 Dependency	Revision	{{if or (eq $type "Flagged by Policy") (eq $type "Denied by Policy") -}} License {{- end}}
```
liftM, line 27: Don't forget to remove the `License` column in the table header as well.

liftM (changes requested):
Nits, then ready.

cmd/fossa/flags/flags.go
zlav: I like this change! Thanks for the rec.

cmd/fossa/flags/flags.go
```diff
@@ -49,7 +49,7 @@ var (
 	Endpoint        = "endpoint"
 	EndpointF       = cli.StringFlag{Name: Short(Endpoint), Usage: "the FOSSA server endpoint (default: 'https://app.fossa.com')"}
 	Title           = "title"
-	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project. (default: the project name)"}
+	TitleF          = cli.StringFlag{Name: Short(Title), Usage: "the title of the FOSSA project (applies ONLY to new projects) (default: the project name)"}
```
liftM, line 52: [nit] I would not emphasize "only" here - I think the emphasis makes this look more alarming than it actually is.

liftM (approved):
Nits in help flags but LGTM.

PR #633, opened 2020-11-14
Send CLI version in upload

scruffystuffs:
## Description
Send the existing version string in the upload API call.

## Acceptance Criteria
version is sent to API


zlav (approved):
This looks good. Good call adding the new `Version` function instead of relying on `ShortString`.

PR #635, opened 2020-11-19
feat: Use name field to name modules

liftM:
Use the `name` field to name raw modules when available.


zlav (approved):
Approving with one small nit.

PR #638, opened 2020-12-17
Update documentation for newer sbt versions

bmjsmith:
## Description

Some small documentation updates based on experience of using the CLI with scala projects using newer sbt versions.


zlav (approved):
Thank you for the very detailed documentation! We will also consider what needs to be done to support SBT >1.4

PR #639, opened 2020-12-18
fossa upload-project command

zlav:
## Description

Add support for a `fossa upload-project` command that is intended to behave identically to the UI archive scanning feature.

This PR also adds support to assign a project to a release group version from the CLI by supplying `--release-group <name>` and `--release-group-version <version>` to the command.

Example UI:
```
fossa archive temp --title new-title --release-group test --release-group-version 2.0

    FOSSA Reports:
    https://app.fossa.com/projects/archive+testing%2F/refs/%2Ffossa-archive/6b15f78939019bba6ead53cd2d3a4b8d

```

## Notes
I considered using one flag and splitting on a special character such as `--release-group <name>@<version>` but I decided to implement two separate flags. The main reason is to avoid unintended issues with splitting the strings on the backend and disallowing a certain special character. The second reason is to reduce confusion for the user who needs to manually add the character. I am favoring enforcing explicit declaration over the expense of the command looking a bit messy.

### DIFF
--- a/api/fossa/locator.go
+++ b/api/fossa/locator.go
@@ -61,9 +61,20 @@ func (l Locator) URL() string {
 	if branch == "" {
 		branch = "master"
 	}
+
+	project := l.Project
+
+	if l.Fetcher == "archive" {
+		orgID, err := GetOrganizationID()
+		if err != nil {
+			log.Warnf("Could not get OrganizationID while constructing locator")
+		}
+		project = orgID + "/" + project
+	}
+
 	url, err := url.Parse(
 		"/projects/" +
-			url.PathEscape(l.Fetcher+"+"+l.Project) +
+			url.PathEscape(l.Fetcher+"+"+project) +
 			"/refs/branch/" +
 			url.PathEscape(branch) +
 			"/" +
--- a/api/fossa/tar.go
+++ b/api/fossa/tar.go
@@ -37,15 +37,27 @@ type Component struct {
 type SignedURL struct {
 	SignedURL string
 }
+type UploadTarballOptions struct {
+	Name           string
+	Revision       string
+	Directory      string
+	IsDependency   bool
+	RawLicenseScan bool
+	Upload         bool
+	UploadOptions  UploadOptions
+}
 
 // UploadTarballDependency uploads the directory specified to be treated on FOSSA as a dependency.
 func UploadTarballDependency(dir string, upload, rawLicenseScan bool) (Locator, error) {
-	return UploadTarball("", "", dir, true, rawLicenseScan, upload, UploadOptions{})
-}
-
-// UploadTarballProject uploads the directory specified to be treated on FOSSA as a project.
-func UploadTarballProject(name, revision, dir string, rawLicenseScan bool, options UploadOptions) (Locator, error) {
-	return UploadTarball(name, revision, dir, false, rawLicenseScan, true, options)
+	return UploadTarball(UploadTarballOptions{
+		Name:           "",
+		Revision:       "",
+		Directory:      dir,
+		IsDependency:   true,
+		RawLicenseScan: rawLicenseScan,
+		Upload:         upload,
+		UploadOptions:  UploadOptions{},
+	})
 }
 
 // UploadTarball archives, compresses, and uploads a specified directory. It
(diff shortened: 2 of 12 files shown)


api/fossa/locator.go
```diff
@@ -61,9 +61,20 @@ func (l Locator) URL() string {
...
 	}
+
+	project := l.Project
+
+	if l.Fetcher == "archive" {
+		orgID, err := GetOrganizationID()
+		if err != nil {
+			log.Warnf("Could not get OrganizationID while constructing locator")
+		}
+		project = orgID + "/" + project
```
liftM, line 72: Huh. How did we not do this before? Was this just broken before?

api/fossa/tar.go
```diff
@@ -64,7 +64,7 @@ func UploadTarballProject(dir string, rawLicenseScan bool) (Locator, error) {
 // Since this will be running within CI machines, this is probably not a good
 // idea. (See https://circleci.com/docs/2.0/configuration-reference/#resource_class
 // for an example of our memory constraints.)
-func UploadTarball(name string, dir string, dependency, rawLicenseScan, upload bool) (Locator, error) {
+func UploadTarball(name, revision, dir string, dependency, rawLicenseScan, upload bool, uploadOptions UploadOptions) (Locator, error) {
```
liftM, line 67: Non-blocking given urgency, but it would be better style here for `UploadTarball` to take an options struct e.g.

```go
type UploadTarballOptions struct {
  Name string
  Revision string
  // ...
}
```

This helps people avoid accidentally mix up the order of arguments with the same type (e.g. `name` and `revision`, or `rawLicenseScan` and `upload`). Besides, you're taking an `UploadOptions` struct anyway.

api/fossa/tar.go
```diff
@@ -439,7 +440,7 @@ func tarballUpload(name string, dependency, rawLicenseScan, upload bool, tarball
 	// Queue the component build.
 	build := ComponentSpec{
 		Archives: []Component{
-			Component{PackageSpec: name, Revision: revision},
+			{PackageSpec: name, Revision: revision},
```
liftM, line 443: Why this remove the name here? I would prefer that struct values in slices are named, to reduce ambiguity. Is there an argument for not naming them?

config/file.v1/file.go
```diff
@@ -21,18 +21,20 @@ type File struct {
...
+	Title               string `yaml:"title,omitempty"`
+	Revision            string `yaml:"revision,omitempty"`
+	Branch              string `yaml:"branch,omitempty"`           // Only used with custom fetcher
+	ProjectURL          string `yaml:"project_url,omitempty"`      // Only used with custom fetcher
+	JIRAProjectKey      string `yaml:"jira_project_key,omitempty"` // Only used with custom fetcher
+	Link                string `yaml:"link,omitempty"`
+	Team                string `yaml:"team,omitempty"`
+	Policy              string `yaml:"policy,omitempty"`
+	ReleaseGroup        string `yaml:"release_group,omitempty"`
+	ReleaseGroupVersion string `yaml:"release_group_version,omitempty"`
```
liftM, line 37: [nit] Hmm, I remember deciding that we preferred the compact syntax `--release_group name@version`. I don't think I have a strong opinion either way. If this is done and works and is simpler, let's go for it.

Just for posterity's sake, was there a particular reason you decided against the compact syntax?

cmd/fossa/cmd/upload_project/upload_project.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/fossas/fossa-cli/api/fossa"
+	"github.com/fossas/fossa-cli/cmd/fossa/display"
+	"github.com/fossas/fossa-cli/cmd/fossa/flags"
+	"github.com/fossas/fossa-cli/cmd/fossa/setup"
+	"github.com/fossas/fossa-cli/config"
+)
+
+var Cmd = cli.Command{
+	Name:      "upload-project",
+	Usage:     "Uploads source code to fossa for server side scanning",
```
liftM, line 18: I would be more explicit: "Uploads a module's source code as a top-level FOSSA project rather than a dependency."

liftM (approved):
LGTM. Some comments. Non-blocking due to urgency - please respond before merging, but don't feel like you need to block on a re-approval before merge.

api/fossa/locator.go
zlav: No, we never did this previously but I don't think it was really "broken" because we've never supported top-level projects with `archive` locators through the CLI until now.

api/fossa/tar.go
zlav: I changed this because I got a linting warning and didn't have a preference either way, but if you see a benefit in keeping it I'm happy to do that. I believe the linter warns because the explicit `Component` is redundant as we are already defining the slice as `[]Component`.

config/file.v1/file.go
zlav: Yea good question, this is a note I left in the PR
```
I considered using one flag and splitting on a special character such as --release-group <name>@<version> but I decided to implement two separate flags. The main reason is to avoid unintended issues with splitting the strings on the backend and disallowing a certain special character. The second reason is to reduce confusion for the user who needs to manually add the character. I am favoring enforcing explicit declaration over the expense of the command looking a bit messy.
```
I understand some of the issues with having two flags, but I think being explicit is better in the long run as long as we fail when both aren't supplied.

api/fossa/tar.go
zlav: Yea I like this, I'll implement this here.

PR #647, opened 2021-02-08
SAML build link

scruffystuffs:
## Description
Build SAML links if org uses SAML, if they don't or saml fails, use normal link

## Acceptance Criteria
* New link redirects thru SAML correctly (manually tested)
* New and old tests pass

## Testing plan
* Verify existing build links (if you don't already have SAML set up it can be difficult)

## Risks
Edge cases in v1 make this more risky than v2, but this is a best guess at handling them correctly.

### DIFF
--- a/api/fossa/locator.go
+++ b/api/fossa/locator.go
@@ -3,6 +3,7 @@ package fossa
 import (
 	"net/url"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/apex/log"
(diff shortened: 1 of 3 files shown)


api/fossa/locator.go
```diff
@@ -82,9 +84,26 @@ func (l Locator) URL() string {
...
+	}
+
 	return server.ResolveReference(url).String()
 }
 
+// SamlURL constructs a SAML redirect URL for a given FOSSA URL
+func SamlURL(uri url.URL, org Organization) *url.URL {
+	samlBase, err := url.Parse("/account/saml/" + strconv.Itoa(org.OrganizationID))
+	if err != nil {
+		log.Fatalf("Invalid SAML URL construction: %s", err.Error())
```
zlav, line 99: I think this should be `log.Warnf` to follow line 70. The project is still uploaded at this point if we get an error correct? If so I think we should make this a warning because `analyze` technically succeeded

scruffystuffs: Technically, it doesn't matter, since this can't fail.  It's parsing a constant value and any integer.  But I can change it.

zlav: Thats a good point... this is going to be nitpicky but can we change it `log.Warn` to follow our convention of not logging fatal unless absolutely necessary?

zlav (approved):
Looks great! +1 for the tests

PR #648, opened 2021-02-09
Use 'go list' to determine transitive dependencies for gomodules projects

cnr:
Our golang dependency analysis is quite complex, and best-case analysis with `go list` [is predicated on everything taking place within the `GOPATH`](https://github.com/fossas/fossa-cli/blob/83303918cc2ea8cb833d60a4a1114c6ce841395b/analyzers/golang/project.go#L76-L86). This was fine until golang introduced gomodules for package management, which can exist outside of the `GOPATH`.

This PR adds a special case to golang project creation for gomodules projects. Instead of determining the module import path and project dir via a combination of VCS and `GOPATH`, we rely on the location and contents of `go.mod` to give us that information.

### DIFF
--- a/analyzers/golang/project.go
+++ b/analyzers/golang/project.go
@@ -1,9 +1,12 @@
 package golang
 
 import (
+	"path/filepath"
+
 	"github.com/apex/log"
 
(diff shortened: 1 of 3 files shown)


zlav (approved):
Great fix and testing!

PR #652, opened 2021-02-16
Ant: use pom.xml's <parent> version if one isn't declared at the top level

cnr:
Fixes https://github.com/fossas/issues/issues/386 . Context available in that ticket


zlav (approved):
Interesting... good fix!

PR #653, opened 2021-02-18
Use gomodules instead of dep

cnr:
Go 1.16 broke our build by defaulting `GO111MODULES` to `on` instead of auto

In this PR, we give into golang's demands and switch over to gomodules


zlav (approved):
Built on my own machine and this all works great! You're now the destroyer of the fossa-cli with a -392,934 diff

CloudKid1273:
nprf

Issue #651, opened 2021-02-14
fossa v1.1.4 Homebrew install - No such file or directory - fossa-cli


dpjtaylor:
**Describe the bug**
When running `brew upgrade fossas/tap/fossa` the upgrade fails with an error `No such file or directory - fossa-cli`

<img width="974" alt="Screenshot 2021-02-14 at 08 57 55" src="https://user-images.githubusercontent.com/47084657/107872587-356d1d00-6ea3-11eb-8c70-eb645ed1c9ab.png">

It looks like this is because the tap [here](https://github.com/fossas/homebrew-tap/blob/master/fossa.rb) is configured to install a binary called `fossa-cli`, but the binary in the release 1.1.4 tar file [here](https://github.com/fossas/fossa-cli/releases/tag/v1.1.4) is called `fossa`.

<img width="340" alt="Screenshot 2021-02-14 at 09 09 54" src="https://user-images.githubusercontent.com/47084657/107872770-774a9300-6ea4-11eb-895c-657c73859ba4.png">

fossa-bot` updated the tap with [this commit](https://github.com/fossas/homebrew-tap/commit/06452263353f5438f84b8b1a7487f9481f99d50a#diff-a74fefab164fb062b1a5603e1a828450a0d5c32fa29382f80e3035ee3a7f9037), but not quite sure how that all joins up (not being familiar with fossa-bot)

```
  def install
    bin.install "fossa-cli"
  end
```

**To Reproduce**
Run `brew upgrade fossas/tap/fossa`

**Expected behavior**
Installs fossa client v1.1.4

**Debugging information**
See screenshot above

**Environment (please complete the following information):**
 - OS: Mac
 - Are you running in CI?: Both on CI and developer Macbook
 - Are you running in a Docker image?: No

**Additional context**
N/A

commjoen:
I get a similar issue when trying to run `brew install fossas/tap/fossa` It gives:

```sh
==> Installing fossa from fossas/tap
==> Downloading https://github.com/fossas/fossa-cli/releases/download/v1.1.5/fossa-cli_1.1.5_darwin_amd64.tar.gz
==> Downloading from https://github-releases.githubusercontent.com/116995448/5f8e4a00-71d7-11eb-8422-fdf5950db0a6?X-Amz-Algorithm=AWS4-HMAC-SHA
######################################################################## 100.0%
Error: An exception occurred within a child process:
  Errno::ENOENT: No such file or directory - fossa-cli```

[...]

PR #655, opened 2021-02-27
fix: brew upgrade failing since v1.1.4

dpjtaylor:
## Description

As described in #651 `brew upgrade fossas/tap/fossa` is failing with an error `No such file or directory - fossa-cli` since v1.1.4 was released

This is because the binary in the zipped files is called `fossa` but the brew tap refers to a binary called `fossa-cli`

The issue seems to have been introduced when fossa-bot updated the Homebrew tap with [this commit](https://github.com/fossas/homebrew-tap/commit/06452263353f5438f84b8b1a7487f9481f99d50a#diff-a74fefab164fb062b1a5603e1a828450a0d5c32fa29382f80e3035ee3a7f9037).

```
  def install
    bin.install "fossa-cli" # refers to fossa-cli, but the binary being shipped is called "fossa"
  end
```

If I understand right, fossa-bot is using GoReleaser to update [homebrew-tap](https://github.com/fossas/homebrew-tap).

This change adds a custom install script step in the GoReleaser config, to set the home-brew script to use "fossa" for the binary name instead of "fossa-cli" when fossa-cli is next released:
```
  def install
    bin.install "fossa"
  end
```

GoReleaser Documentation: https://goreleaser.com/customization/homebrew/
```
    # Custom install script for brew.
    # Default is 'bin.install "program"'.
    install: |
      bin.install "program"
```

## Acceptance Criteria

- Users can successfully upgrade fossa cli using `brew upgrade fossas/tap/fossa`

## Testing plan

I assume testing will require a fresh release of the fossa cli.

### Before state

Screenshot 2021-02-27 at 14 16 30

## Risks

Risk of breaking the home brew tap, causing problems on brew upgrade. This is already broken as far as I can tell, so worst case it stays broken.

## Notes

N/A

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -39,3 +39,6 @@ brews:
     commit_author:
       name: fossabot
       email: engineering@fossa.com
+
+    install: |
+      bin.install "fossa"


dpjtaylor:
Is there anyone who can help with this?

I'm switching to using the curl approach for installing for our CI jobs at work, but I think it would be good to resolve this to help any users trying to install via Homebrew.

```
curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install.sh | bash
```

meghfossa:
Thank you for PR submission @dpjtaylor. 

I'm closing this PR as, we are migrating https://github.com/fossas/spectrometer source code at tip of master, as we move towards having 3.0 release. 

We as of now, only support using installation method via install script. Of course, you can also get binary releases from Github release. Nonetheless, I've added homebrew support in our backlog for 3.x release.

PR #660, opened 2021-04-08
Fix Maven parsing error with "Downloaded" line

zlav:
## Description
This PR addresses an issue with the Maven parser that would fail when it encountered lines that began with `[INFO] Downloaded`. 

I am unsure if this is new maven behavior since we have never come across this before, but it appears to be affecting a few users in scenarios where it was not before. 

## Acceptance Criteria
- The fossa-cli can parse the full user test output that it was previously failing on.
- The fossa-cli can handle `[INFO] Downloaded` lines anywhere the dependency graph was created.

## Testing plan
- Ran the user's full output through the CLI to validate that it could successfully parse the list and get a dependency graph.
- I also wrote a test to drive the solution which contained the problematic `[INFO] Downloaded` line.
- Validated that the fix does not change any behavior with existing Maven parsing.

## Risks
No known risks that I can think of.

### DIFF
--- a/buildtools/maven/maven.go
+++ b/buildtools/maven/maven.go
@@ -169,7 +169,7 @@ func ParseDependencyTree(stdin string) (graph.Deps, error) {
 		if line == "[INFO]" || line == "[INFO] " || line == "[INFO] ------------------------------------------------------------------------" {
 			started = false
 		}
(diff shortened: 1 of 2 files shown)


taliamccormick (approved):
LGTM! Thanks.

PR #664, opened 2021-06-16
fix ant subdirectoy analysis

zlav:
## Description
This ticket addresses an issue with Ant analysis not being able to scan projects with lib directories located in different directories than the `build.xml` file.

Closes https://github.com/fossas/team-analysis/issues/539

## Acceptance Criteria
`lib-directories` that are located in subdirectories of an ant project, when the ant project is located in a subdirectory of the root, are properly scanned. 

## Testing plan

- Construct a project with a `build.xml` file in a subdirectory. Example `test/build.xml`
- Add a `lib` directory with at least 1 dependency to a deeper subdirectory `test/tmp/lib`
- Configure the CLI properly with the `lib-directory` option and run `fossa analyze`

## Risks
There are no risks with this. I considered the fact that maybe this is not backward compatible, however, due to the fact we were previously checking if the lid directory existed (in the correct location) we are not breaking any workflow that is not already broken.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,7 +90,7 @@ jobs:
     <<: *defaults
     working_directory: /go/src/github.com/fossas/fossa-cli
     docker:
-      - image: circleci/golang:1
+      - image: circleci/golang:1.16.0
     steps:
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


scruffystuffs (approved):
LGTM

PR #665, opened 2021-06-24
Fix node_modules strategy

zlav:
## Description
A customer is seeing an issue where after the command `npm ls --production` fails, we infinitely recurse through their `node_modules` directory.

The strategy appears to be infinitely recursing if any manifest in the `node_modules` dir refers to another manifest in the `node_modules`. I'm not sure if this strategy ever worked.

This PR is fixing the recursion by exiting when we reach a node that we have already seen before.

## Acceptance Criteria
Running analysis with `strategy: package.json` succeeds and does not infinitely recurse.

## Testing plan
I tested this on the FOSSA core repository with and without my changes. Without the changes, I saw an infinite recursion. 

## Risks
The risks are very low. This removes the potential for infinite recursion and does not add any new failure cases to the CLI. The largest risk is that recursing without this check somehow provided superior data, however, I am unable to see how that would be the case.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,7 +17,7 @@ jobs:
       # gets better performance. The trade-off is that the CircleCI image is
       # cached but the FOSSA CLI image has build tools baked into it (but also
       # is correspondingly larger).
-      - image: circleci/golang:1
+      - image: circleci/golang:1.16
     steps:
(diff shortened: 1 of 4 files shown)


scruffystuffs (approved):
LGTM, nice fix.

PR #666, opened 2021-07-07
fix: Unit test failures should cause CI failure

liftM:
## Description

This PR fixes up our CI tests in preparation for another bug fix I want to make.

The main changes here:

- Set `pipefail` to fix CI runs reporting as green even on test failures.
- Fix Ruby integration test by updating to a compatible commit.
- Fix Go unit tests for build tags.
- Update bindata.
- Update `buildtools` Dockerfile to fix Composer installation issue.
- Fix a small typo in an error message.

The Go build tag test fixes are the most complicated part of this PR. Here's how they work:

The build tag tests check that certain dependencies are or are not included when certain build tags are passed as we shell out to `go`. In Go 1.16, `GO111MODULE` now defaults to `on` ([release notes](https://golang.org/doc/go1.16#go-command)), so the Go command runs in module mode by default. In #653, we switched this repository to use Go modules.

This means that the `go` command was now running in module mode in the build tag tests. This caused the tests to break because `go list` now exited 1 since the dummy dependencies (e.g. `import/combo`) were now invalid because they were not contained in FOSSA CLI's `go.mod` (they were originally specified in a `Gopkg.{toml,lock}` and located in `vendor`, but `go` no longer looked at `vendor` once it began running in module mode).

I fixed this by bypassing the need to have use dummy dependencies by using standard library packages as dependencies. After all, the only thing we needed to test here is that dependency exclusion based on tags worked correctly.

In the long term, the right fix for this is probably to break these tests out into their own folder. At the end of the day, this problem is caused by our Go tests being contained within a Go repository, and this causing weird interactions when we shell out to `go`.

## Acceptance Criteria

- CI should be red if tests fail.
- CI should be green if tests succeed.
- Fix tests.

## Testing plan

Break the unit tests, run `make ci-unit-test`, make sure exit is non-zero. Do the same with `make ci-integration-test`.


analyzers/golang/testdata/demotags/tags/combo.go
```diff
@@ -2,8 +2,8 @@
 
 package tags
 
-import "imports/combo"
+import "fmt"
```
zlav, line 5: Good call adding standard lib here.

zlav (approved):
Looks great. I verified that the tests failed when I broke something and that the fixes for go build tags were correct.

analyzers/golang/testdata/demotags/tags/combo.go
liftM: Yeah, the alternative is we could've picked non-overlapping dependencies in `go.mod`, but I didn't want that to cause more mysterious breaks down the line when we change the CLI's actual dependencies.

PR #668, opened 2021-07-08
Update genny calls to not use gopath

scruffystuffs:
## Description
If a dev does not have a clone in their `GOPATH` (and they shouldn't need to after we switched to gomod), then `genny` will fail, as it was explicitly using `$GOPATH`.  This has been fixed by using relative file paths.

## Acceptance Criteria
* Run `go generate ./...` at the project root, no files named `readtree_generated.go` should be altered.

## Testing plan
Checkout this pr, Run `make clean && make build` in your local environment, confirm no significant changes to generated files.

## Risks
Might break builds if working directory from genny invocation is not guaranteed.

## Notes
Tried using `$GOMOD` (from `go env`), but that was somehow not set during `go:generate`

### DIFF
--- a/buildtools/composer/composer.go
+++ b/buildtools/composer/composer.go
@@ -11,7 +11,7 @@ import (
 	"github.com/fossas/fossa-cli/exec"
 )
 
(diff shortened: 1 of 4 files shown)


liftM:
Re: testing plan - testers should also make sure to delete `fossa-cli` from `$GOPATH` as part of verification.

liftM (approved):
Builds locally for me.

Merging since Wes is out.

Issue #672, opened 2021-08-18
Deep dependencies are not found with NPM7+ (Nodejs16+)


zenonhun:
**Describe the bug**
In the latest nodejs (v16), the npm also upgraded to 7+.
In npm 7+ the behavior of `npm ls` is changed, now by default it lists only the direct dependencies.
To get the deep dependencies the `--all` option is required. 

https://docs.npmjs.com/cli/v7/commands/npm-ls

Fossa-cli uses the `npm ls --json --production` command to get the list of packages.
Now it shall be changed to add the `--all` option. In legacy npm versions it does nothing, so it is safe to add permanently.

**To Reproduce**
1. install nodejs 16.5.0
2. Run a fossa-cli analysis
3. See that it lists only direct dependencies

**Expected behavior**
Fossa-cli analysis shall find all direct and deep dependencies. 

**Debugging information**
Please provide the following:

1. The output of running the buggy command with the `--debug` flag
2. The output of `fossa --version`
3. The project's configuration file (usually `.fossa.yml`)
4. If possible, a minimal test case reproducing the bug (ideally, a GitHub repository)

**Environment (please complete the following information):**
 - OS: [e.g. Windows]
 - Are you running in CI?: [e.g. No / Yes: CI_PROVIDER_NAME]
 - Are you running in a Docker image?: [e.g. No / Yes: DOCKER_IMAGE_NAME]

**Additional context**

zlav:
HI @zenonhun, the latest version of the FOSSA CLI should no longer have the issue you are listing here.

PR #673, opened 2021-08-18
NodeJS - Add support for NPM7

zenonhun:
NPM7 changed npm ls command. '--all' is required to get deep dependencies.

## Description
As in described in Issue #672, NPM 7 changed npm ls behavior. `--all` option is required to have deep dependencies listed. 

Fossa cli uses npm ls internally to get the list of dependencies, so to add support to the latest version, the --all option shall be added.

### DIFF
--- a/buildtools/npm/npm.go
+++ b/buildtools/npm/npm.go
@@ -40,10 +40,10 @@ func (n SystemNPM) List(dir string, devDeps bool) (Output, error) {
 	listCmd := exec.Cmd{
 		Name: n.Cmd,
 		Dir:  dir,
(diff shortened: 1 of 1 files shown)


meghfossa:
Thank you for PR submission @zenonhun. 

I'm closing this PR as, we are migrating https://github.com/fossas/spectrometer source code at tip of master, as we move towards having 3.0 release. Npm support with 3.0 does not use dynamic analysis at the moment, but rather parses package.json, lock files and hydrates the dependency graph. So moving forth, we won't run into aforementioned concern.

You can open the PR targeting v1-master branch to patch of 1.x releases moving forth.

PR #691, opened 2021-11-15
Relegates Circle CI build for only v1

meghfossa:
# Overview

It relegates circle CI build workflow to only v1 prefix on git tag, or branch


meghfossa:
This is so we can avoid the circle ci build failures even when we are not using it!
 
image

kitified (approved):

PR #701, opened 2021-11-15
Removes orphan spectrometer references

meghfossa:
# Overview

Removes orphaned spectrometer references.


zlav (approved):
Good catch here!

PR #704, opened 2021-11-17
Fixes .net graphing for project.assets.json target type

meghfossa:
# Overview

This PR introduces a change that ensures for `projectassetsjson` targets, 

- Project type dependencies are shrinked (and not reported)
- It correctly labels direct dependency based on `project.[framework]` data

## Acceptance criteria

- For target `projectassetsjson` type, dependencies of type project are not reported
- For target `projectassetsjson` type, dependencies are differentiated between direct and deep

## Testing plan

- Create .net project (or refer to any open source project.assets.json from github)
- Perform `fossa analyze -o | jq`, ensure produced dependency graphing,
   - reports all dependencies of package type
   - does not report dependencies of project type (if it has any transitive dependencies of package type, they are reported)
   - direct dependencies are correctly identified

## Risks

I did not like hack included with framework name matching, but I can't really move forward without having access to windows pc for complete testing. 

I do think, what I have should address 99+% cases. We still maintain previous behaviour (which was ignoring project's framework) when we are not able to find matching framework. 

## References

Closes https://github.com/fossas/team-analysis/issues/791
Closes https://github.com/fossas/team-analysis/issues/790

## Checklist

### DIFF
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -47,6 +47,7 @@ module Graphing (
   fromList,
   toList,
   unfold,
+  unfoldDeep,
 ) where
 
 import Algebra.Graph.AdjacencyMap (AdjacencyMap)
@@ -229,6 +230,31 @@ deeps = Graphing . AM.vertices . map Node
 -- __Unfold does not work for recursive inputs__
 unfold :: forall dep res. Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
 unfold seed getDeps toDependency = directs directNodes <> edges builtEdges
+  where
+    directNodes :: [res]
+    directNodes = map toDependency seed
+
+    builtEdges :: [(res, res)]
+    builtEdges = map (bimap toDependency toDependency) (edgesFrom seed)
+
+    edgesFrom :: [dep] -> [(dep, dep)]
+    edgesFrom nodes = do
+      node <- nodes
+      let children = getDeps node
+      map (node,) children ++ edgesFrom children
+
+-- | @unfoldDeep dep getDeps toDependency@ unfolds a graph, given:
+--
+-- - The @dep@ dependencies in the graph
+--
+-- - A way to @getDeps@ for a dependency
+--
+-- - A way to convert a dependency @toDependency@
+--
+-- __unfoldDeep does not work for recursive inputs__
+-- __unfoldDeep marks all dependencies as deeps__
+unfoldDeep :: forall dep res. Ord res => [dep] -> (dep -> [dep]) -> (dep -> res) -> Graphing res
+unfoldDeep seed getDeps toDependency = deeps directNodes <> edges builtEdges
   where
     directNodes :: [res]
     directNodes = map toDependency seed
--- a/src/Graphing.hs
+++ b/src/Graphing.hs
@@ -245,7 +245,7 @@ unfold seed getDeps toDependency = directs directNodes <> edges builtEdges
 
 -- | @unfoldDeep dep getDeps toDependency@ unfolds a graph, given:
 --
--- - The @dep@ dependencies in the graph
+-- - The @deep@ dependencies in the graph
 --
 -- - A way to @getDeps@ for a dependency
 --
--- a/src/Strategy/NuGet/ProjectAssetsJson.hs
+++ b/src/Strategy/NuGet/ProjectAssetsJson.hs
@@ -7,21 +7,31 @@ module Strategy.NuGet.ProjectAssetsJson (
   mkProject,
   buildGraph,
   ProjectAssetsJson (..),
+
+  -- * for testing
+  approxEql,
 ) where
 
 import App.Fossa.Analyze.Types (AnalyzeProject, analyzeProject)
-import Control.Effect.Diagnostics
+import Control.Effect.Diagnostics hiding (fromMaybe)
 import Data.Aeson
+import Data.Aeson.Types (Parser)
+import Data.HashMap.Strict qualified as HM
 import Data.Map.Strict qualified as Map
 import Data.Maybe
+import Data.Set (Set)
+import Data.Set qualified as Set
+import Data.String.Conversion (toString)
 import Data.Text (Text)
 import Data.Text qualified as Text
+import Data.Traversable (for)
 import DepTypes
 import Discovery.Walk
 import Effect.ReadFS
 import GHC.Generics (Generic)
-import Graphing (Graphing, unfold)
+import Graphing (Graphing, empty, gmap, promoteToDirect, shrink, unfoldDeep)
 import Path
+import Text.Read (readMaybe)
 import Types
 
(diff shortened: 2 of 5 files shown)


src/Graphing.hs
```diff
@@ -242,6 +243,31 @@ unfold seed getDeps toDependency = directs directNodes <> edges builtEdges
       let children = getDeps node
       map (node,) children ++ edgesFrom children
 
+-- | @unfoldDeep dep getDeps toDependency@ unfolds a graph, given:
+--
+-- - The @dep@ dependencies in the graph
```
cnr, line 248:
```suggestion
-- - The @deep@ dependencies in the graph
```

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -68,52 +78,165 @@ analyze' file = do
...
+      map
+        (graphOfFramework $ projectFramework project)
+        (Map.toList $ targets project)
+
+graphOfFramework :: Map.Map Text (Set Text) -> (Text, Map.Map Text DependencyInfo) -> Graphing Dependency
+graphOfFramework projectFrameworkDeps (targetFramework, targetFrameworkDeps) =
+  withoutTags
+    . withoutProjects
+    . promoteToDirect isDirectDep
+    $ unfoldDeep allResolvedDeps getTransitiveDeps toDependency
```
cnr, line 146: The way I view the `Dependency` type is that once we've built something of type `Dependency`, we generally shouldn't touch it ever again

The fields on `Dependency` have very few guarantees, and depending on the tags field here to communicate type information indirectly feels brittle

When we need to modify `Dependency` values, it's usually a sign that we can model the domain more closely by using an intermediate type

We almost do that here with the `NuGetDep` type, but we eagerly convert it to `Dependency` with `toDependency`. May I recommend instead:

```hs
unfoldDeep allResolvedDeps getTransitiveDeps id :: Graphing NuGetDep
```

do the `Graphing.shrink withoutProjects` on that, slightly changing `withoutProjects`

```hs
withoutProjects :: Graphing NuGetDep -> Graphing NuGetDep
```

_then_ turn it into `Graphing Dependency` with `Graphing.gmap toDependency`

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -68,52 +78,165 @@ analyze' file = do
...
+  deriving (Show, Eq, Ord)
 
-instance FromJSON ProjectAssetsJson where
-  parseJSON = withObject "ProjectAssetsJson" $ \obj ->
-    ProjectAssetsJson <$> obj .: "targets"
+data ProjectAssetsJson = ProjectAssetsJson
+  { targets :: Map.Map Text (Map.Map Text DependencyInfo)
+  , projectFramework :: Map.Map Text (Set Text)
+  }
+  deriving (Show, Eq, Ord)
```
cnr, line 93: [nit] there are a lot of `Text` values floating around; it's probably worthwhile to `newtype` wrapper some (or all?) of these. it was hard to follow some of the type signatures below and make sure the right text values were used

for json decoding purposes, you can derive `FromJSON` on the newtypes:

```hs
newtype FrameworkName = FrameworkName { unFrameworkName :: Text }
  deriving (Eq, Ord, Show, FromJSON)
```

test/NuGet/ProjectAssetsJsonSpec.hs
```diff
@@ -54,16 +57,161 @@ dependencyFour =
...
+  Dependency
+    { dependencyType = NuGetType
+    , dependencyName = "five"
+    , dependencyVersion = Just (CEq "5.0.0")
+    , dependencyLocations = []
+    , dependencyEnvironments = mempty
+    , dependencyTags = Map.empty
+    }
+
+-- | Reference: https://github.com/NuGet/NuGet.Client/blob/820bfe9d5e5e4e7bf0b6310c285db7b07aff3087/test/NuGet.Core.Tests/NuGet.Frameworks.Test/NuGetFrameworkParseTests.cs
```
cnr, line 71: [no action] wtf

test/NuGet/ProjectAssetsJsonSpec.hs
```diff
@@ -54,16 +57,161 @@ dependencyFour =
...
+  , ("net452", ".NETFramework,Version=v4.5.2/win7-x86")
+  , ("net472", ".NETFramework,Version=4.7.2")
+  ]
+
 spec :: Spec
 spec = do
   testFile <- runIO (BS.readFile "test/NuGet/testdata/project.assets.json")
 
+  describe "approxEqlFramework" $ do
+    it "should return false when provided with equivalent framework" $ do
```
cnr, line 201: is this test description correct?

cnr (approved):
LGTM generally; a few minor comments

test/NuGet/ProjectAssetsJsonSpec.hs
meghfossa: good catch!

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -68,46 +82,94 @@ analyze' file = do
...
-    convertDep :: (Text, DependencyInfo) -> Maybe NuGetDep
-    convertDep (depString, dep) = case Text.splitOn "/" depString of
+    isProjectDep :: NuGetDep -> Bool
+    isProjectDep NuGetDep{..} = completeDepType == "project"
+
+    allResolvedDeps :: [NuGetDep]
+    allResolvedDeps = mapMaybe toNugetDep $ Map.toList targetFrameworkDeps
+
+    toNugetDep :: (Text, DependencyInfo) -> Maybe NuGetDep
+    toNugetDep (depString, dep) = case Text.splitOn "/" depString of
```
cnr, line 165: comment I forgot to add yesterday: we can do this check at JSON-decode-time, which would prevent us from silently ignoring invalid dependency keys -- and would also correctly propagate the error

there's a typeclass [called `FromJSONKey`](https://hackage.haskell.org/package/aeson-2.0.2.0/docs/Data-Aeson.html#t:FromJSONKey) that's used by `Map`'s `FromJSON` instance:

```hs
instance (FromJSONKey k, Ord k, FromJSON v) => FromJSON (Map k v) where
  ...
```

we can take advantage of this:

```hs
data NuGetDepKey = NuGetDepKey
  { depKeyName :: Text
  , depKeyVersion :: Text
  } deriving (Eq, Ord, Show)

instance FromJSONKey NuGetDepKey where
  fromJSONKey = FromJSONKeyTextParser $ \depString ->
    case Text.splitOn "/" depString of
      [name, ver] -> pure $ NuGetDepKey name ver
      other -> fail $ "Unexpected dependency key: " <> show depString
```

and change the type of `targets` in `ProjectAssetsJson`:

```hs
data ProjectAssetsJson = ProjectAssetsJson
  { targets :: Map.Map FrameworkName (Map.Map NuGetDepKey DependencyInfo)
  ...
```

PR #705, opened 2021-11-17
Ignore debug output

kitified:
# Overview

Adds `fossa.debug.json.gz` to `.gitignore`.

## Acceptance criteria

If we execute `cabal run fossa -- analyze --debug`, this file is generated at the project root, but it shouldn't be accidentally committed.


meghfossa (approved):
Good addition! I also run into this once in while.

PR #706, opened 2021-11-17
Removes cliv1 reference from the cabal project dependency

meghfossa:
# Overview

Removes redundant cli-v1 file dependency.


spatten (approved):
It turns out that this does not fix cabal install, but I still think it's a good change and we should merge it

@meghfossa I think we should rename the PR to a more correct description and then merge

PR #707, opened 2021-11-18
Go: `go mod graph` and `go list` tactic (pass #2)

meghfossa:
# Overview

Adds golang (go mod graph) tactic, which uses `go mod graph`, and `go list -m -json all` to construct dependency graph!

## Acceptance criteria

- When `go mod graph` command can be executed, `go mod graph` tactic is favored!
- `go mod graph` tactic does not include golang's stdlib
- `go mod graph` tactic does not include main module
- `go mod graph` tactic includes only selected version of the module (same as build list)
- `go mod graph` tactic labels direct dependencies
- `go mod graph` tactic reports all build list dependencies

## Testing plan

1. Clone https://github.com/opstrace/opstrace.git or any open source go project (must use go module)
2. Perform `go mod tidy`
3. Perform `fossa analyze -o | jq`, ensure direct dependencies matches that of go.mod file after performing step (2)
4. Perform `go list -m all`, ensure all dependencies listed are included in the graphing produced by `fossa analyze`

## Risks

I'm adding this as preferred tactic, and as consequence of that, 
- Users will not see stdlib anymore in the dependency graphing results

Go's dependency tooling has varying support per different golang version, and performs resolution differently (which is problematic), but from my testing - I did not run into any issues.

## References

Closes https://github.com/fossas/team-analysis/issues/813
Closes https://github.com/fossas/team-analysis/issues/659

## Checklist

### DIFF
--- a/src/App/Fossa/Analyze/Project.hs
+++ b/src/App/Fossa/Analyze/Project.hs
@@ -42,4 +42,5 @@ data ProjectResult = ProjectResult
 
 shouldKeepUnreachableDeps :: Text -> Bool
 shouldKeepUnreachableDeps "swift" = True
+shouldKeepUnreachableDeps "gomod" = True
 shouldKeepUnreachableDeps _ = False
--- a/src/Strategy/Go/GoModGraph.hs
+++ b/src/Strategy/Go/GoModGraph.hs
@@ -10,12 +10,11 @@ module Strategy.Go.GoModGraph (
   buildGraph,
 ) where
 
-import Control.Effect.Diagnostics hiding (fromMaybe)
+import Control.Effect.Diagnostics
 import Control.Monad (void)
 import Data.Aeson.Internal (formatError)
 import Data.Foldable (find)
 import Data.Map qualified as Map
-import Data.Maybe (fromMaybe)
 import Data.SemVer qualified as SemVer
 import Data.SemVer.Internal (Version (..))
 import Data.Set (Set, fromList, member, notMember)
@@ -44,6 +43,7 @@ import Text.Megaparsec (
   MonadParsec (eof),
   Parsec,
   empty,
+  errorBundlePretty,
   optional,
   parse,
   sepEndBy,
@@ -86,10 +86,9 @@ goListJsonCmd =
     , cmdAllowErr = Never
     }
 
-data GoGraphMod = GoGraphMod
-  { goGraphModName :: Text
-  , goGraphModVersion :: Maybe PackageVersion
-  }
+data GoGraphMod
+  = MainMod Text
+  | OtherMod Text PackageVersion
   deriving (Show, Eq, Ord)
 
 -- * Parsers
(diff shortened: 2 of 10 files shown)


meghfossa:
NOT merging as fetcher is behaving _very_ _very_ poorly as of now!

src/App/Fossa/Analyze/Project.hs
```diff
@@ -42,4 +42,5 @@ data ProjectResult = ProjectResult
 
 shouldKeepUnreachableDeps :: Text -> Bool
 shouldKeepUnreachableDeps "swift" = True
+shouldKeepUnreachableDeps "gomod" = True
```
cnr, line 45: I'd love to kill this logic by moving dependency pruning into the tactics that need it (namely, the golang tactics that call `fillInTransitive`). matching on strings is brittle and `fillInTransitive` is the only reason we prune unreachable dependencies anyway

src/Strategy/Go/GoModGraph.hs
```diff
@@ -0,0 +1,189 @@
...
+  name <- toText <$> lexeme (some (alphaNumChar <|> char '/' <|> char '-' <|> char '.' <|> char '_' <|> char '~'))
+  version <- optional $ do
+    _ <- try $ symbol "@"
+    lexeme (parsePackageVersion lexeme)
+  pure $ GoGraphMod name version
+
+toGoModVersion :: Text -> Maybe PackageVersion
+toGoModVersion modVersion = case parse (parsePackageVersion lexeme) "go module version" modVersion of
+  Left _ -> Nothing
+  Right vc -> Just vc
```
cnr, line 109: is it safe to ignore this parse error?

src/Strategy/Go/GoModGraph.hs
```diff
@@ -0,0 +1,189 @@
...
+    parseGoModPair = (,) <$> parseGoGraphMod <*> parseGoGraphMod
+
+-- Builds graph from edges, main module, and selected module version set.
+buildGraph :: [(GoGraphMod, GoGraphMod)] -> Maybe GoGraphMod -> Set GoGraphMod -> Set GoGraphMod -> Bool -> Graphing.Graphing Dependency
+buildGraph fromToMods mainMod selectedMods directMods applyMVS =
+  Graphing.gmap toDependency
+    . Graphing.induceJust
+    . Graphing.gmap withSelection
+    . Graphing.promoteToDirect (`member` directMods)
+    . Graphing.shrink (\m -> m /= fromMaybe impossibleMod mainMod)
```
cnr, line 125: We always expect to find a main module, right? If so, we should error out if we don't find one, rather than silently ignoring (i.e. pass `mainMod :: GoGraphMod` to `buildGraph` rather than `mainMod :: Maybe GoGraphMod`)

src/Strategy/Go/GoModGraph.hs
```diff
@@ -0,0 +1,189 @@
...
+  where
+    parseGoModPair :: Parser (GoGraphMod, GoGraphMod)
+    parseGoModPair = (,) <$> parseGoGraphMod <*> parseGoGraphMod
+
+-- Builds graph from edges, main module, and selected module version set.
+buildGraph :: [(GoGraphMod, GoGraphMod)] -> Maybe GoGraphMod -> Set GoGraphMod -> Set GoGraphMod -> Bool -> Graphing.Graphing Dependency
+buildGraph fromToMods mainMod selectedMods directMods applyMVS =
+  Graphing.gmap toDependency
+    . Graphing.induceJust
+    . Graphing.gmap withSelection
```
cnr, line 123: [nit] use `Graphing.filter`/`Graphing.induce` instead?

cnr (changes requested):
See comment on the `GoGraphMod` type; I think we can avoid `go list` here altogether

src/App/Fossa/Analyze/Project.hs
meghfossa: Addressing in https://github.com/fossas/fossa-cli/pull/721

cnr (approved):
Looks great. Can we deprecate/remove the `GoList` tactic or do we still need it? (if we still need it, a comment describing why would be helpful for future maintainers)

PR #708, opened 2021-11-19
Add the README banner back

zlav:
# Overview

During the merge, we forgot to migrate the banner from the old readme. I also changed some references of fossa-cli to FOSSA CLI which is the term we use publicly when marketing the CLI.

## Testing

The banner link looks for master, so the banner will not appear until we merge the banner asset.

## Acceptance criteria

The banner appears and "fossa-cli" is not referenced on the readme page.


meghfossa (approved):
LGTM

PR #713, opened 2021-11-23
Reformats fossa.io so lint check for links can pass

meghfossa:
# Overview

Fossa.io seems to be 404ing, I'm not sure why. 

Reformatting so markdown link checker test can pass, to unblock dev workflows.


spatten (approved):
:+1:

PR #716, opened 2021-11-25
Fixes Yarn.Lock V1 missing direct dependency (sometimes)

meghfossa:
# Overview

This PR, fixes YarnV1 analysis defect, in which we were missing direct dependency sometimes.
 
## Acceptance criteria

- YarnV1 analysis always graphs all valid direct dependencies

## Testing plan

Given following `package.json`, and `yarn.lock` file

```json
{
  "name": "example-project",
  "dependencies": {
      "pngjs": "6.0.0",
      "once": "1.4.0"
  },
  "devDependencies": {
      "mocha": "*"
  }
}
```

Running, `fossa analyze` will produce:

Previously: dependency once will be missing from dependency graph
Now: dependency once will be included in dependency graph

## Risks

- N/A

## References

Closes  https://github.com/fossas/team-analysis/issues/824

## Checklist

### DIFF
--- a/src/Strategy/Node/YarnV1/YarnLock.hs
+++ b/src/Strategy/Node/YarnV1/YarnLock.hs
@@ -85,12 +85,14 @@ buildGraph ::
   FlatDeps ->
   m (Graphing Dependency)
 buildGraph lockfile FlatDeps{..} = fmap hydrateDepEnvs . withLabeling toDependency $
-  for_ (map firstKey $ MKM.toList lockfile) $ \(key, pkg) -> do
+  for_ (map allkeys $ MKM.toList lockfile) $ \(firstKey, keys, pkg) -> do
+    -- Since all keys have same resolved version, and same extracted name
+    -- we can use any key to infer parent.
     let parent :: YarnV1Package
-        parent = pairToPackage key pkg
+        parent = pairToPackage firstKey pkg
 
-        keyAsNodePackage :: NodePackage
-        keyAsNodePackage = toNodePackage key
+        allKeysAsNodePackages :: [NodePackage]
+        allKeysAsNodePackages = toNodePackage <$> keys
 
         childrenSpecs :: [YL.PackageKey]
         childrenSpecs = YL.dependencies pkg
@@ -106,7 +108,7 @@ buildGraph lockfile FlatDeps{..} = fmap hydrateDepEnvs . withLabeling toDependen
     -- Add edges from current parent
     traverse_ (edge parent) children
     let promote env pkgSet =
-          if keyAsNodePackage `Set.member` pkgSet
+          if any (`Set.member` pkgSet) allKeysAsNodePackages
             then do
               direct parent
               label parent $ NodeEnvironment env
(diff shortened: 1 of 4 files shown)


src/Strategy/Node/YarnV1/YarnLock.hs
```diff
@@ -173,8 +175,8 @@ missingResolvedVersionErrorMsg key =
 pairToPackage :: YL.PackageKey -> YL.Package -> YarnV1Package
 pairToPackage key pkg = YarnV1Package (extractFullName key) (YL.version pkg)
 
-firstKey :: (NE.NonEmpty a, b) -> (a, b)
-firstKey (neList, pkg) = (NE.head neList, pkg)
+allkeys :: (NE.NonEmpty a, b) -> (a, [a], b)
+allkeys (neList, pkg) = (NE.head neList, NE.toList neList, pkg)
```
scruffystuffs, line 179: [suggest]: I think naming and type might need to change here, in it's current state we're kind of just putting a band-aid on it.  A complete fix (i.e.: doing it right the first time) would probably look more like this:
```suggestion
extractPkgTriple :: (NE.NonEmpty YL.PackageKey, YL.Package) -> (Text, [YL.PackageKey], YL.Package)
extractPkgTriple (neList, pkg) = (extractFullName $ NE.head neList, NE.toList neList, pkg)
```

src/Strategy/Node/YarnV1/YarnLock.hs
```diff
@@ -85,12 +85,14 @@ buildGraph ::
   FlatDeps ->
   m (Graphing Dependency)
 buildGraph lockfile FlatDeps{..} = fmap hydrateDepEnvs . withLabeling toDependency $
-  for_ (map firstKey $ MKM.toList lockfile) $ \(key, pkg) -> do
+  for_ (map allkeys $ MKM.toList lockfile) $ \(firstKey, keys, pkg) -> do
```
scruffystuffs, line 88: [suggest]: If we use a better extraction function like the suggested `extractPkgTriple`, then this needs to change accordingly:
```suggestion
  for_ (map extractPkgTriple $ MKM.toList lockfile) $ \(pkgName, keys, pkg) -> do
```

src/Strategy/Node/YarnV1/YarnLock.hs
```diff
@@ -85,12 +85,14 @@ buildGraph ::
...
+    -- Since all keys have same resolved version, and same extracted name
+    -- we can use any key to infer parent.
     let parent :: YarnV1Package
-        parent = pairToPackage key pkg
+        parent = pairToPackage firstKey pkg
```
scruffystuffs, line 92: [suggest]: If we use a better extraction function like the suggested `extractPkgTriple`, then this needs to change accordingly:
```suggestion
        parent = YarnV1Package pkgName $ YL.version pkg
```

src/Strategy/Node/YarnV1/YarnLock.hs
```diff
@@ -173,8 +175,8 @@ missingResolvedVersionErrorMsg key =
 pairToPackage :: YL.PackageKey -> YL.Package -> YarnV1Package
 pairToPackage key pkg = YarnV1Package (extractFullName key) (YL.version pkg)
```
scruffystuffs, line 177: [suggest]: If we use a better extraction function like the suggested `extractPkgTriple`, then this can be removed:
```suggestion
```

scruffystuffs (approved):
Left an optional suggestion across multiple comments.

src/Strategy/Node/YarnV1/YarnLock.hs
meghfossa: sigh - I def needed a coffee that day. Applied all suggestions!

PR #724, opened 2021-11-29
add cabal-fmt to CI, separate lints

scruffystuffs:
# Overview

Add `cabal-fmt -i` check to CI, move linter/formatter checks to separate CI job, so we don't have to re-run builds in all cases.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -169,47 +169,6 @@ jobs:
         name: ${{ runner.os }}-binaries
         path: release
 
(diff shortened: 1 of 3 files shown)


scruffystuffs:
After this PR merges, I'll make the `link-check` job required.

meghfossa (approved):
LGTM! Do we need to fmt-check cabal.project.* file as well?

scruffystuffs:
Nope.  That file is a slightly different format (I think), and rarely changes.  The `.cabal` file is our actual project file.  I also don't know if the tool actually works on that file.

PR #728, opened 2021-12-03
Updates makefile to include cabal-fmt

meghfossa:
# Overview

- Adds `cabal-fmt` on `make fmt`, and `make check-fmt`
- Adds `make install-dev` option for install `fossa-dev` on local user's path


scruffystuffs (approved):
LGTM

PR #730, opened 2021-12-04
VSI Rewrite: Comment Stripped Fingerprint

kitified:
# Overview

Adds the ability to generate comment stripped fingerprints for files.
Ported from the [functionality here (private link)](https://github.com/fossas/basis/blob/6b0a1ce7ca5d88d033732f6dcfebd90b8f143038/sherlock/pkg/lib/indexer/cleaned/strip_comments.go#L14-L99). 

## Acceptance criteria

This functionality is currently unused, but is part of fossas/team-analysis#825.
As such my AC is largely "do tests pass".

## Testing plan

* I created the set of test data (inside `test/App/Fossa/VSI/testdata`) that I believe covers the expected cases.
* I used the `debug fingerprint` binary in Basis to generate the fingerprints we expect to see: 

```shell
$ basis run sherlock debug fingerprint ~/projects/fossa-cli/test/App/Fossa/VSI/testdata/

binary
	[sha_256]: a57d87677b2b67c573c870aff75d8c92c23beaf9f8682e7749ba7b421e84c991
empty.c
	[sha_256]: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
	[comment_stripped:sha_256]: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
multi_line.c
	[sha_256]: 8bd20ac94646806aec6dadb9325218718d1ba8abc039da6abc9aa8159658671d
	[comment_stripped:sha_256]: 2f0ac453340cab3a92d68297964634c8a80ca5828f760c588b80a69f963cea68
multi_line_comment.c
	[sha_256]: 9976b2fda660ce9a9084992881df35ec17bc78b39a2bf24434d8fed875af3a2d
	[comment_stripped:sha_256]: 44fc8f68ab633c7ca0240a66e4ff038c0f2412fe69d14b6f052556edaa1b9160
single_line.c
	[sha_256]: deac66ccb79f6d31c0fa7d358de48e083c15c02ff50ec1ebd4b64314b9e6e196
	[comment_stripped:sha_256]: 80a7161009ffaf868641acac3f5e49bc5f86021ee1d177f3b1cbb47573513649
single_line_comment.c
	[sha_256]: 7bf434baac2af7cf06fea2125d453f7b14cf942d5b3cef2a4a2a9281eec2534b
	[comment_stripped:sha_256]: 80a7161009ffaf868641acac3f5e49bc5f86021ee1d177f3b1cbb47573513649
```

* I then set up the test with the values.
* I then relied on automated tests to demonstrate that we're working properly.

## Risks

If implemented incorrectly, VSI results can degrade.

## References

Closes https://github.com/fossas/team-analysis/issues/838

## Checklist


src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,177 @@
...
+import Path (Abs, Dir, File, Path, toFilePath)
+
+-- | Fingerprint deterministically idenfies a file and is derived from its content.
+--
+-- The type variable is:
+--
+--   @k@ - Kind, the kind of fingerprint computed.
+--
+-- For ease of implementation, the backing representation of a @Fingerprint@ instance is a @Base16@ encoded @Text@.
+newtype Fingerprint k = Fingerprint Text
```
scruffystuffs, line 37: Let's add a type role here:
```suggestion
newtype Fingerprint k = Fingerprint Text
type role Fingerprint nominal
```

I wrote a explanation for this reasoning in the [`Data.Tagged` module](https://github.com/fossas/fossa-cli/blob/8e7777e3fb07c1632639d406b1e3c67530eae9c1/src/Data/Tagged.hs#L12) (though it's worth noting that that module is actually wrong right now).

In short, this prevents `coerce :: Fingerprint Raw -> Fingerprint CommentStripped`.

scruffystuffs (commented):
Left an initial note.  I'm not yet familiar with `Conduit`, so I'll take a dive into this after checking that out.

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,181 @@
...
+-- It is very basic:
+--
+-- * Only works for C-style comments
+-- * Only catches @\\n@ newlines
+-- * Doesn't handle edge cases (escaped comments for example)
+-- * Also omits any blank lines
+--
+-- Despite these drawbacks, we have to reimplement it the same way so that fingerprints line up correctly in the VSI analysis service.
+basicCStyleCommentStripC :: Monad m => ConduitT Text Text m ()
+basicCStyleCommentStripC = splitOnUnboundedE (== '\n') .| process .| mapC Text.strip .| filterC (not . Text.null) .| bufferedNewline Nothing
```
scruffystuffs, line 135: Looks like `linesUnboundedC` is implemented as `splitOnUnboundedE (== '\n')`.  So it's not causing the windows issues.  We should keep this as `linesUnboundedC`.

https://hackage.haskell.org/package/conduit-1.3.4.2/docs/src/Data.Conduit.Combinators.Unqualified.html#linesUnboundedC

which leads to 

https://hackage.haskell.org/package/conduit-1.3.4.2/docs/src/Data.Conduit.Combinators.html#linesUnbounded

kitified: Good point, probably it's the git checkout thing like you mentioned.

Thinking about it, we should probably just convert any `\r\n` to `\n` to keep it consistent with our indexers (which all ran on Linux). While this'll cause us to not match files that truly contain `\r\n`, it's probably an overall more correct thing to do. I'll make that change.

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -195,6 +170,30 @@ basicCStyleCommentStripC =
...
+
+        -- Here, we know we have a new line coming down the pipe.
+        -- If we had a previous line buffered, write it with a newline.
+        -- If not, store this new line in the buffer first.
+        Just line -> case buf of
+          Nothing -> bufferedNewline (Just line)
+          Just bufferedLine -> do
+            yield $ bufferedLine <> "\n"
+            bufferedNewline (Just line)
+
```
scruffystuffs, line 196: [optional]: I think this could be made a little clearer by writing it as a `case (chunk, buf) of` expression.  As it currently is, this jumps around a lot and was hard to follow in terms of readability (programmatically, it's fine).

I'm thinking it could look a little more like this (note the ordering being consistent with actual execution)
```haskell
case (chunk, buf) of
  -- First line will land here, we always buffer the first line
  (Just line, Nothing) -> bufferedNewline (Just line)
  -- All lines between first and last will be yielded with newlines, and the NEXT line will be buffered
  (Just line, Just bufferedLine) -> do
    yield $ bufferedLine <> "\n"
    bufferedNewline (Just line)
  -- No next line, so the buffered line is the last one.
  (Nothing, Just bufferedLine) -> yield bufferedLine
  -- All lines have been written, now we can exit.
  -- Technically unreachable, since we don't recurse after yielding the last line.
  (Nothing, Nothing) -> pure ()
```

scruffystuffs: But don't kill your comments about compat with old fingerprint data, those are obviously needed.

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,217 @@
...
+-- | Hashes a stream of 'B.ByteString'@s@ and creates a digest @d@.
+-- Adapted from @sinkHash@ in https://hackage.haskell.org/package/cryptonite-conduit-0.2.2/docs/src/Crypto-Hash-Conduit.html
+sinkHash :: (Monad m, HashAlgorithm hash) => ConduitT B.ByteString Void m (Digest hash)
+sinkHash = sink hashInit
+  where
+    sink ctx = do
+      b <- await
+      case b of
+        Nothing -> return $! hashFinalize ctx
```
scruffystuffs, line 79: [nit]: Prefer `pure` over `return`
```suggestion
        Nothing -> pure $! hashFinalize ctx
```

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,217 @@
...
+        Just bs -> sink $! hashUpdate ctx bs
+
+-- | Hashes the whole contents of the given file in constant memory.
+hashBinaryFile :: (Has (Lift IO) sig m, Has Diagnostics sig m, HashAlgorithm hash) => FilePath -> m (Digest hash)
+hashBinaryFile fp =
+  sendIO (runConduitRes (sourceFile fp .| sinkHash))
+    `catch` (\(e :: IOException) -> fatalText ("unable to hash file: " <> toText (show e)))
```
scruffystuffs, line 86: Can we extract this to something like `catchingIOExc :: m a -> m a`?  It's used multiple times.  This might even belong in `Control.Effect.Diagnostics` (with a parameterized type and generic message).

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,217 @@
...
+data Raw
+
+-- | Represents a 'Fingerprint' derived from the content of a file with all C-style comments removed.
+data CommentStripped
+
+instance ToText (Fingerprint k) where
+  toText (Fingerprint k) = k
+
+instance ToJSON (Fingerprint k) where
+  toJSON = toJSON . toText
```
scruffystuffs, line 53: [nit]: We can probably kill this and just add `deriving (ToJson)` to the newtype declaration.

src/App/Fossa/VSI/Fingerprint.hs
```diff
@@ -0,0 +1,217 @@
...
+
+fingerprintRaw :: (Has ReadFS sig m, Has (Lift IO) sig m, Has Diagnostics sig m) => Path Abs File -> m (Fingerprint Raw)
+fingerprintRaw file = contentIsBinary file >>= doFingerprint
+  where
+    doFingerprint True = do
+      (fp :: Digest SHA256) <- hashBinaryFile $ toFilePath file
+      pure $ encodeFingerprint fp
+    doFingerprint False = do
+      (fp :: Digest SHA256) <- hashTextFile $ toFilePath file
+      pure $ encodeFingerprint fp
```
scruffystuffs, line 122: [nit]: reduce duplication here:
```suggestion
    doFingerprint isBinary = do
      let hasher = if isBinary then hashBinaryFile else hashTextFile
      (fp :: Digest SHA256) <- hasher $ toFilePath file
      pure $ encodeFingerprint fp
```

scruffystuffs (approved):
Few nits, but LGTM otherwise.

src/App/Fossa/VSI/Fingerprint.hs
kitified: Oooh, I like this. Good idea!

src/Control/Effect/Diagnostics.hs
```diff
@@ -89,6 +92,12 @@ fatal = send . Fatal
 fatalText :: Has Diagnostics sig m => Text -> m a
 fatalText = fatal
 
+-- | Throw a generic error message on IO error
+fatalOnIOException :: (Has (Lift IO) sig m, Has Diagnostics sig m) => m a -> m a
+fatalOnIOException go = catch go die'
+  where
+    die' (e :: IOException) = fatalText ("io exception: " <> toText (show e))
```
scruffystuffs, line 99: I think this might be a bit of a debugging footgun, since it's easy to get a diagnostic error with no real context for what caused it.

What if we required using `context` to wrap this catching function?

`fatalOnIOException msg go = context msg $ catch go die'`

scruffystuffs (approved):
One last optional bit, but other than that, this is ready to go.

src/Control/Effect/Diagnostics.hs
kitified: I like it! Done.

PR #732, opened 2021-12-06
Adds bundler integration test

meghfossa:
# Overview

Adds bundler integration test.

See passing integration test: https://github.com/fossas/fossa-cli/runs/4432992412?check_suite_focus=true

### DIFF
--- a/integration-test/Analysis/RubySpec.hs
+++ b/integration-test/Analysis/RubySpec.hs
@@ -1,0 +1,36 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE QuasiQuotes #-}
+
+module Analysis.RubySpec (spec) where
+
+import Analysis.FixtureExpectationUtils
+import Analysis.FixtureUtils
+import Path
+import Strategy.Bundler qualified as Bundler
+import Test.Hspec
+import Types (GraphBreadth (Complete))
+
+rails :: AnalysisTestFixture (Bundler.BundlerProject)
+rails =
+  AnalysisTestFixture
+    "rails"
+    Bundler.discover
+    LocalEnvironment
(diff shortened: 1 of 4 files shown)


integration-test/Analysis/RubySpec.hs
```diff
@@ -0,0 +1,35 @@
...
+      $ Just [reldir|rails-6.1.4/|]
+
+testRailSrcRepo :: Spec
+testRailSrcRepo =
+  aroundAll (withAnalysisOf rails) $ do
+    describe "rails" $ do
+      it "should find targets" $ \(result, extractedDir) -> do
+        shouldFindProjectOf ("bundler", extractedDir) result
+      it "should have expected dependency results" $ \(result, extractedDir) -> do
+        (DependencyResultsSummary 210 70 293 1 Complete) `expectDepResultsSummary` (getDepResultsOf result ("bundler", extractedDir))
```
zlav, line 31: Do you think testing for the exact number of deps is going to be brittle? I'm very much in favor of testing for finer details to start, but I'm somewhat concerned that this could change for reasons out of our control.

zlav (approved):
Amazing, love how simple this is.

integration-test/Analysis/RubySpec.hs
meghfossa: I agree, it is not ideal. Trade off is however, is time saved. I would like to get the tests in ASAP, ideally close to same state as spectrometer-test, and iterate from there on, with very specific story/backlog priority.  

I would like to eventually, make FOSSA API itself effectual, so we can assert entire POST graph response (and it's properties), as opposed to component test on analyzer we are doing like spectrometer-test repo.

PR #733, opened 2021-12-07
Nuget: Project Assets Json - fixes infinite loop

meghfossa:
# Overview

This PR, changes graphing method to ensure, we build graph iteratively.

## Acceptance criteria

- Nuget analyzer for `project.assets.json` does not lead to infinite loop

## Testing plan

- Perform analysis against provided debug bundle in referenced ticket.

Expected: `fossa analyze` should complete within seconds.

## Risks

N/A

## References

Closes https://github.com/fossas/team-analysis/issues/850

## Checklist

### DIFF
--- a/src/Strategy/NuGet/ProjectAssetsJson.hs
+++ b/src/Strategy/NuGet/ProjectAssetsJson.hs
@@ -28,10 +28,12 @@ import Data.Text qualified as Text
 import Data.Traversable (for)
 import DepTypes
 import Discovery.Walk
+import Effect.Grapher (LabeledGrapher, deep, direct, edge, withLabeling)
 import Effect.ReadFS
 import GHC.Generics (Generic)
-import Graphing (Graphing, empty, gmap, promoteToDirect, shrink, unfoldDeep)
+import Graphing (Graphing)
 import Path
+import Data.Foldable (for_)
 import Text.Read (readMaybe)
 import Types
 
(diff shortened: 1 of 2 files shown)


meghfossa:
Added https://github.com/fossas/team-analysis/issues/851 as tech debt to avoid this scenario completely in future.

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -148,49 +150,54 @@ instance FromJSON ProjectAssetsJson where
...
+        Dependency
+          { dependencyType = NuGetType
+          , dependencyName = depName
+          , dependencyVersion = Just (CEq depVersion)
+          , dependencyLocations = []
+          , dependencyEnvironments = mempty
+          , dependencyTags = Map.empty
+          }
+
+graphOfFramework :: Has (LabeledGrapher NuGetDep NugetLabel) sig m => (Map.Map FrameworkName (Set Text)) -> (FrameworkName, Map.Map NuGetDepKey DependencyInfo) -> m ()
```
scruffystuffs, line 175: We aren't using labels at all here.  Can we just use the grapher effect, and then run `Graphing.gmap` to convert to dependency later?

src/Strategy/NuGet/ProjectAssetsJson.hs
```diff
@@ -148,49 +150,54 @@ instance FromJSON ProjectAssetsJson where
...
 
-graphOfFramework :: Map.Map FrameworkName (Set Text) -> (FrameworkName, Map.Map NuGetDepKey DependencyInfo) -> Graphing Dependency
-graphOfFramework projectFrameworkDeps (targetFramework, targetFrameworkDeps) =
-  Graphing.gmap toDependency
-    . withoutProjects
-    . promoteToDirect isDirectDep
-    $ unfoldDeep allResolvedDeps getTransitiveDeps id
+    toDependency :: NuGetDep -> Set NugetLabel -> Dependency
+    toDependency NuGetDep{..} =
+      foldr (\_ d -> d) $
```
scruffystuffs, line 165: No need to use foldr without accumulation:
```suggestion
    toDependency NuGetDep{..} _ =
```

scruffystuffs (approved):
I'd like to avoid `LabeledGrapher` when we don't use labels.  I'm not gonna block the PR in case this is a time-sensitive fix, but I'd like to get this cleaned up once we deliver any critical fixes.

meghfossa:
@scruffystuffs are you ok with integration test, as opposed to unit test for this? That's the only bit remaining from CR.

scruffystuffs:
Unless there's a good reason not to put it in unit tests, I think it makes the most sense there.

PR #736, opened 2021-12-10
VSI Rewrite: Add API endpoints

kitified:
# Overview

Adds functions to use VSI API endpoints.

## Acceptance criteria

Feature parity with VSI

## Testing plan

Unused, so not really tested. These are 1:1 with the implementations in the VSI plugin though, so unlikely to require significant change.

## Risks

Nothing really risky, especially since they're currently unused.

## References

Closes https://github.com/fossas/team-analysis/issues/843

## Checklist

### DIFF
--- a/src/App/Fossa/FossaAPIV1.hs
+++ b/src/App/Fossa/FossaAPIV1.hs
@@ -623,8 +623,8 @@ vsiCreateScan apiOpts ProjectRevision{..} = fossaReq $ do
   let projectID = renderLocator $ Locator "custom" projectName Nothing
   let reqBody = VSICreateScanRequestBody orgId projectRevision projectID
 
-  (resBody :: VSICreateScanResponseBody) <- responseBody <$> req POST (vsiCreateScanEndpoint baseUrl) (ReqBodyJson reqBody) jsonResponse baseOpts
-  pure $ unVSICreateScanResponseBody resBody
+  body <- responseBody <$> req POST (vsiCreateScanEndpoint baseUrl) (ReqBodyJson reqBody) jsonResponse baseOpts
+  pure $ unVSICreateScanResponseBody body
 
 newtype VSIAddFilesToScanRequestBody = VSIAddFilesToScanRequestBody {vsiAddFilesToScanRequestBodyFiles :: Map (Path Rel File) Fingerprint.Combined}
 
@@ -632,7 +632,7 @@ instance ToJSON VSIAddFilesToScanRequestBody where
   toJSON VSIAddFilesToScanRequestBody{..} = object ["ScanData" .= vsiAddFilesToScanRequestBodyFiles]
 
 vsiAddFilesToScanEndpoint :: Url scheme -> VSI.ScanID -> Url scheme
-vsiAddFilesToScanEndpoint baseurl scanID = baseVsiUrl baseurl /: "scans" /: toText scanID /: "files"
+vsiAddFilesToScanEndpoint baseurl (VSI.ScanID scanID) = baseVsiUrl baseurl /: "scans" /: scanID /: "files"
 
 vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts -> VSI.ScanID -> Map (Path Rel File) Fingerprint.Combined -> m ()
 vsiAddFilesToScan apiOpts scanID files = fossaReq $ do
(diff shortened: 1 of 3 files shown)


src/App/Fossa/FossaAPIV1.hs
```diff
@@ -580,3 +587,100 @@ resolveProjectDependencies apiOpts locator = fossaReq $ do
...
+    (status :: Text) <- obj .: "Status"
+    pure . VSIScanAnalysisStatusBody $ VSI.parseAnalysisStatus status
+
+vsiScanAnalysisStatusEndpoint :: Url scheme -> VSI.ScanID -> Url scheme
+vsiScanAnalysisStatusEndpoint baseurl scanID = baseVsiUrl baseurl /: "scans" /: toText scanID /: "status" /: "analysis"
+
+vsiScanAnalysisStatus :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts -> VSI.ScanID -> m VSI.AnalysisStatus
+vsiScanAnalysisStatus apiOpts scanID = fossaReq $ do
+  (baseUrl, baseOpts) <- useApiOpts apiOpts
+  (resBody :: VSIScanAnalysisStatusBody) <- responseBody <$> req GET (vsiScanAnalysisStatusEndpoint baseUrl scanID) NoReqBody jsonResponse baseOpts
```
scruffystuffs, line 669: Is this type annotation needed?  I think the newtype unwrapping function below hints the type well enough.

src/App/Fossa/VSI/Types.hs
```diff
@@ -24,6 +27,27 @@ import Effect.Logger (Pretty (pretty), viaShow)
 import Srclib.Converter (depTypeToFetcher, fetcherToDepType)
 import Srclib.Types qualified as Srclib
 
+-- | The VSI backend returns a scan ID when a scan is created, which is then used to add files to the scan and get inferred OSS dependencies.
+newtype ScanID = ScanID {unScanID :: Text} deriving (ToJSON, FromJSON)
+
+instance ToText ScanID where
+  toText = unScanID
```
scruffystuffs, line 34: I think we should avoid creating `ToText` instances for `Text`-wrapping newtypes.  It kind of breaks the semantic safety that they offer.  See my comment on `vsiAddFilesToScanEndpoint` for an example.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -580,3 +587,100 @@ resolveProjectDependencies apiOpts locator = fossaReq $ do
...
+  (resBody :: VSICreateScanResponseBody) <- responseBody <$> req POST (vsiCreateScanEndpoint baseUrl) (ReqBodyJson reqBody) jsonResponse baseOpts
+  pure $ unVSICreateScanResponseBody resBody
+
+newtype VSIAddFilesToScanRequestBody = VSIAddFilesToScanRequestBody {vsiAddFilesToScanRequestBodyFiles :: Map (Path Rel File) Fingerprint.Combined}
+
+instance ToJSON VSIAddFilesToScanRequestBody where
+  toJSON VSIAddFilesToScanRequestBody{..} = object ["ScanData" .= vsiAddFilesToScanRequestBodyFiles]
+
+vsiAddFilesToScanEndpoint :: Url scheme -> VSI.ScanID -> Url scheme
+vsiAddFilesToScanEndpoint baseurl scanID = baseVsiUrl baseurl /: "scans" /: toText scanID /: "files"
```
scruffystuffs, line 635: Using a `toText` here kind of breaks the type safety that we gain with the `ScanID` newtype.
I would usually prefer to see something like this:
```suggestion
vsiAddFilesToScanEndpoint baseurl (VSI.ScanID scanID) = baseVsiUrl baseurl /: "scans" /: scanID /: "files"
```

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -580,3 +587,100 @@ resolveProjectDependencies apiOpts locator = fossaReq $ do
...
+
+vsiCreateScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts -> ProjectRevision -> m VSI.ScanID
+vsiCreateScan apiOpts ProjectRevision{..} = fossaReq $ do
+  (baseUrl, baseOpts) <- useApiOpts apiOpts
+
+  Organization orgId _ <- getOrganization apiOpts
+  let projectID = renderLocator $ Locator "custom" projectName Nothing
+  let reqBody = VSICreateScanRequestBody orgId projectRevision projectID
+
+  (resBody :: VSICreateScanResponseBody) <- responseBody <$> req POST (vsiCreateScanEndpoint baseUrl) (ReqBodyJson reqBody) jsonResponse baseOpts
```
scruffystuffs, line 626: Is this type annotation needed?  I think the newtype unwrapping function below hints the type well enough.

scruffystuffs (approved):
Left a few minor comments, nothing blocking.

Also, I didn't check that the URL's were actually the correct ones, since I'm not sure what the correct ones are.  If we're able to integration test this at some point, that would be a good idea, but I think we don't have a mechanism for that today.

src/App/Fossa/FossaAPIV1.hs
kitified: Oh! I didn't even think of that. Nice point!

src/App/Fossa/FossaAPIV1.hs
kitified: Fair enough, I buy that argument. Thanks!

PR #738, opened 2021-12-10
Refactor `fossa analyze` entrypoint

scruffystuffs:
# Overview

Closes fossas/team-analysis#828

Implements the new entrypoint model (from fossas/team-analysis#796) for `fossa analyze`.

Notable side changes

- Legacy filters have been removed entirely.
- `FilterCombination` is now strongly typed with `Exclude` or `Include`, and `AllFilters` is specific about accepting one of each.
- The old `App.Fossa.Main` module is dead, and the binary entrypoint refers exclusively to the new one.  As a result, **THIS PR IS NOT SAFE TO BE MERGED DIRECTLY INTO MASTER!  DON'T MERGE THIS INTO MASTER!  DON'T APPROVE THIS PR IF IT'S MERGING INTO MASTER!**
- More `MonadIO` constraints have been removed.

## Acceptance criteria

- The business-logic-module of `fossa analyze` sees configuration as a immutable fact, rather than actively participating in the configuration process.
- The configuration module of `fossa analyze` only exposes types which are either opaque or representative of the final configuration "fact".
- Config errors are collected, and reported together.  i.e.: an invalid basedir and a missing API key (assuming no `--output` flag) are both reported together, rather than aborting at the first error.  This is done using `Validation`, from the `validation-selective` package.
- No semantic changes except for:
  - Missing subcommands (`fossa test`, `fossa report`, etc. those are follow-up tickets)
  - Monorepo is temporarily broken.
  - Config file accept/ignore/warn/abort semantics are change via [this slack conversation](https://teamfossa.slack.com/archives/G01DUC2JW3B/p1639173134260800)

## Testing plan

Run `fossa analyze`, without any monorepo options.

## Risks

- This is not safe to merge into master.
- It's very possible that the configuration rewrite has mistakes that ignore some config values from certain places.  Reviewers, please pay close attention to those configuration resolving functions.
- This PR lays out a massive change to our CLI frontend, and its conclusion will become the basis for the future of the CLI and configuration.  We should not merge/approve it until we are convinced that this model is the best way forward for the CLI.

## References

Task Ticket: fossas/team-analysis#828
Project Ticket: fossas/team-analysis#796

## Checklist

- [x] ~I added tests for this PR's change (or confirmed tests are not viable).~
- [x] ~If this PR introduced a user-visible change, I added documentation into `docs/`.~
- [x] ~I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.~
[...]

PR #742, opened 2021-12-16
Refactor `fossa test` entrypoint

scruffystuffs:
# Overview

closes fossas/team-analysis#829

## Acceptance criteria

Same as  #738, but for `fossa test`.

Additionally, `timeout` now has an effectful counterpart: `timeout'`.  To use it, correctly, you must also use the new functions in the buildwait module: `waitForIssues'`, `waitForScanCompletion'`, `waitForSherlockScan'`.  The old functions are unsafe, since they infinitely loop and must be killed from the outside.

Note that `fossa test` is feature-complete, even though `fossa analyze` does not support monorepo, `fossa test` does.

## Testing plan

Run `fossa test`, in any scenario.

## Risks

- New timeout module, infinite loops are possible if stuff is done wrong.
- As with all of the tickets in the fossas/team-analysis#796 project, configuration is being rewritten, so that needs to be checked.

## References

Task Ticket: fossas/team-analysis#829
Project Ticket: fossas/team-analysis#796

## Checklist

- [x] ~I added tests for this PR's change (or confirmed tests are not viable).~
- [x] ~I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.~

### DIFF
--- a/src/App/Fossa/API/BuildWait.hs
+++ b/src/App/Fossa/API/BuildWait.hs
@@ -20,8 +20,8 @@ import App.Types (ProjectRevision (projectName, projectRevision))
 import Control.Concurrent (threadDelay)
 import Control.Concurrent.Async qualified as Async
 import Control.Effect.Diagnostics (
-  Has,
   Diagnostics,
+  Has,
   ToDiagnostic (..),
   fatal,
   fatalText,
--- a/src/App/Fossa/API/BuildWait.hs
+++ b/src/App/Fossa/API/BuildWait.hs
@@ -9,7 +9,7 @@ module App.Fossa.API.BuildWait (
   waitForScanCompletion',
   waitForIssues',
   waitForSherlockScan',
-  shouldCancel,
+  shouldCancelRightNow,
   Cancel,
 ) where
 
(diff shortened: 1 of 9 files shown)


src/Control/Timeout.hs
```diff
@@ -0,0 +1,93 @@
+-- Provide tools for internal-cancellation functions, built for the funcitons in 'App.Fossa.API.BuildWait'
+
+module Control.Timeout (
+  Cancel,
+  shouldCancelRightNow,
+  checkForCancel,
+  Duration (..),
+  durationToMicro,
+  timeout',
```
meghfossa, line 9: [nit] would prefer unit test on public methods (at-least happy path)

src/App/Fossa/API/BuildWait.hs
```diff
@@ -131,3 +159,122 @@ timeout ::
...
+  Fossa.Organization orgId _ <- Fossa.getOrganization apiOpts
+  let locator = VPSCore.createLocator (projectName revision) orgId
+
+  logSticky' "[ Getting latest scan ID ]"
+  scan <- ScotlandYard.getLatestScan apiOpts locator (projectRevision revision)
+
+  logSticky' "[ Waiting for monorepo scan... ]"
+  waitForSherlockScan apiOpts locator (ScotlandYard.responseScanId scan)
+
+waitForIssues' ::
```
zlav, line 231: Leaving a note that this should be changed `waitForIssues` before you merge and remove the other `waitForIssues`

scruffystuffs: Already covered in the following PR.  See [this commit](https://github.com/fossas/fossa-cli/pull/754/commits/573860d9c41652955a85ce16a007442131f2c640).

scruffystuffs: I wish I had done it in this PR, but I had already based too much work on it after I did it.  Also, I couldn't remove these too early, since other places in the codebase depend on these, and that would prevent this PR from building successfully.

PR #754, opened 2021-12-30
Add all remaining commands

scruffystuffs:
# Overview

Implement phases 3 and 4 of the Entrypoint refactor project (fossas/team-analysis#796)

Closes fossas/team-analysis#830
Closes fossas/team-analysis#831
Closes fossas/team-analysis#832
Closes fossas/team-analysis#833
Closes fossas/team-analysis#834
Closes fossas/team-analysis#835
Closes fossas/team-analysis#836

## Acceptance criteria

All commands have been moved over (except `fossa analyze --monorepo` (not yet implemented), `fossa log4j` (to be added during the merge) and `fossa vps ninja-graph` (removed)).

## Testing plan

(NOT YET TESTED)

* Run each command, using every option/argument at least once.
* Run the various `--help` commands.

## Risks

* CLI parsing needs close review, make sure our frontend hasn't changed.
* As always, it's possible to ignore options in the config file and the cli, and just provide some data to the configuration struct.  We should make sure that the data is flowing through the config validation pipeline correctly.  This is why many validators/collectors/utilities are defined in the common module.

## References

Project Ticket: fossas/team-analysis#796

## Checklist

- ~I added tests for this PR's change (or confirmed tests are not viable).~
- ~If this PR introduced a user-visible change, I added documentation into `docs/`.~
- ~I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.~
- ~I updated `*schema.json` if I have made changes for `.fossa.yml`, `fossa-deps.{json, yaml, yml}`. You may also need to update these if you have added/removed new dependency (e.g. pip) or analysis target type (e.g. poetry).~


src/App/NewFossa/Config/Container.hs
```diff
@@ -0,0 +1,350 @@
...
+data ContainerTestConfig = ContainerTestConfig
+  { apiOpts :: ApiOpts
+  , timeoutDuration :: Duration
+  , outputFormat :: OutputFormat
+  , testImageLocator :: ImageText
+  , testRevisionOverride :: OverrideProject
+  }
+  deriving (Eq, Ord, Show)
+
+data ContainerDumpScapConfig = ContainerDumpScapConfig
```
kitified, line 265:
```suggestion
data ContainerDumpScanConfig = ContainerDumpScanConfig
```

kitified (commented):
Stopping here for now but one small comment.

kitified (approved):
I primarily looked over VSI and VPS stuff, but also went through the rest, this lgtm!

The main (non-blocking) comment I have is similar to the other PR, I'd personally prefer using wording that isn't "global" for these args that are opt-in (and therefore not truly global).

test/Control/TimeoutSpec.hs
```diff
@@ -0,0 +1,71 @@
...
+  timeout',
+ )
+import Control.Timeout.Internal (Cancel (Cancel))
+import Data.Functor (void)
+import Prettyprinter (pretty)
+import Test.Effect (expectationFailure', it', shouldBe')
+import Test.Hspec (Spec, describe, it, shouldBe)
+
+tinyDuration :: Duration
+tinyDuration = MicroSeconds 200
```
meghfossa, line 28: [optional] [nit] prefer var name `twoHundredMicroseconds`

meghfossa (approved):
LGTM code wise, nothing stood out as red/yellow flag. 

Will perform e/e testing on PR targeting main branch.

test/Control/TimeoutSpec.hs
scruffystuffs: The amount of time is arbitrary, the only relevant fact is that it's small.  Being specific about the amount of time actually reduces the value of the variable name.  e.g.: This variable would serve the exact same purpose if it were 175 _us_.

PR #757, opened 2022-01-06
Remove the name field from package-lock

zlav:
## Overview

Currently, our `package-lock.json` parser fails when the optional top-level field `name` is missing. This PR removes all references to `name` since we don't use it for analysis.

## Acceptance criteria

- Running `fossa analyze` on a project with a `project-lock.json` file without the top-level `name` field succeeds. 

## Testing plan

Tested by running on a `package-lock.json` file without the `name` field before and after the changes in this PR.

## Risks

None that I'm aware of.

## References

Deepak had an issue analyzing a project.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Fossa CLI Changelog
 
+## v3.0.13
+
+- `package-lock.json` parser ignores name field. ([#757](https://github.com/fossas/fossa-cli/pull/757)) 
+
 ## v3.0.12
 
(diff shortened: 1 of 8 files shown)


scruffystuffs:
Code is fine, but the docs are a little off.  The way we handle workspaces makes the directory relationship between `package.json` and `package-lock.json` a little more complicated.  Let's chat offline to make sure we're on the same page about how it works.

scruffystuffs (approved):

PR #762, opened 2022-01-13
Composer json license detection

csasarak:
# Overview

The intent of this change is to add support for finding licenses in PHP projects with a `composer.json` file to the `pathfinder` sub-command of fossa-cli.

## Acceptance criteria

If a `composer.json` file is present, fossa-cli should parse it and extract the contents of its `license` key.

## Testing plan

I added several unit tests which verify that the new `LicenseAnalyzeProject` for `ComposerProject`s will read and find single or multiple licenses listed when a directory has a `composer.json` file. In addition, there are test cases for when no `composer.json` file is present, when the `license` key does not exist, or when the `license` key is out of schema.

As an additional test, I ran pathfinder in a composer project with licenses and verified using the debug output that the licenses were discovered properly.

## Risks

I would like reviewers to pay particular attention to the tests, as well as comment if they think there are necessary integration tests that should be added.

## References

This closes fossas/team-analysis#861

## Checklist

### DIFF
--- a/src/Strategy/Composer.hs
+++ b/src/Strategy/Composer.hs
@@ -5,11 +5,12 @@ module Strategy.Composer (
   buildGraph,
   ComposerLock (..),
   CompDep (..),
-  ComposerProject(..),
+  ComposerProject (..),
 ) where
 
 import App.Fossa.Analyze.Types (AnalyzeProject, analyzeProject)
 import App.Pathfinder.Types (LicenseAnalyzeProject (licenseAnalyzeProject))
+import Control.Applicative ((<|>))
 import Control.Effect.Diagnostics hiding (fromMaybe)
 import Data.Aeson.Types
 import Data.Foldable (traverse_)
@@ -124,15 +125,35 @@ instance FromJSON ComposerJson where
     do
       licenses <- obj .:? "license"
       case licenses of
-        Just licenses' -> case licenses' of
-          Array licenseVec -> pure $ ComposerJson $ Just (License LicenseSPDX <$> mapMaybe f (Vec.toList licenseVec))
-          String licenseStr -> pure $ ComposerJson $ Just [License LicenseSPDX licenseStr]
-          _ -> fail "Invalid schema for key 'license' in composer.json"
+        Just licenses' ->
+          parseArrayLicense licenses'
+            <|> parseStrLicense licenses'
+            <|> fail "Invalid schema for key 'license' in composer.json"
         Nothing -> pure $ ComposerJson Nothing
     where
-      f :: Value -> Maybe Text
-      f (String t) = Just t
-      f _ = Nothing
+      parseStrLicense s =
+        withText
+          "licenseString"
+          ( \licenseStr ->
+              pure . ComposerJson . Just $ [License LicenseSPDX licenseStr]
+          )
+          s
+
+      parseArrayLicense s =
+        withArray
+          "licenseArray"
+          ( pure
+              . ComposerJson
+              . Just
+              . map (License LicenseSPDX)
+              . mapMaybe maybeString
+              . Vec.toList
+          )
+          s
+
+      maybeString :: Value -> Maybe Text
+      maybeString (String t) = Just t
+      maybeString _ = Nothing
 
 newtype CompPkg = CompPkg {pkgName :: Text}
   deriving (Eq, Ord, Show)
--- a/src/Strategy/Composer.hs
+++ b/src/Strategy/Composer.hs
@@ -5,6 +5,7 @@ module Strategy.Composer (
   buildGraph,
   ComposerLock (..),
   CompDep (..),
+  ComposerProject(..),
 ) where
 
 import App.Fossa.Analyze.Types (AnalyzeProject, analyzeProject)
@@ -83,7 +84,7 @@ analyzeLicenses :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> 
 analyzeLicenses path =
   maybe [] mkLicenseResult . license <$> readContentsJson licenseFileName
   where
-    licenseFileName = path </> $(mkRelFile "composer.lock")
+    licenseFileName = path </> $(mkRelFile "composer.json")
     mkLicenseResult = pure . LicenseResult (toFilePath licenseFileName)
 
 data ComposerLock = ComposerLock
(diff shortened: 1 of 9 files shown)


test/Composer/ComposerJsonSpec.hs
```diff
@@ -0,0 +1,75 @@
...
+  currentDir <- runIO getCurrentDir
+  describe "composer.json LicenseAnalyzeProject" $ do
+    it' "Reads a composer.json with a single license string" $
+      let baseDir = currentDir </> $(mkRelDir "test/Composer/testdata/composer_json_single")
+          project =
+            ComposerProject
+              { composerDir = baseDir
+              , composerLock = currentDir </> $(mkRelFile "not_tested_here")
+              }
+       in do
```
scruffystuffs, line 24: [nit]: We avoid `let ... in ...` blocks in the codebase as much as possible.  We stick to using `let` within do notation, or using `where` clauses.

test/Composer/ComposerJsonSpec.hs
```diff
@@ -0,0 +1,75 @@
...
+
+    it "Fails to read a composer.json which is out of schema" $
+      let baseDir = currentDir </> $(mkRelDir "test/Composer/testdata/composer_json_bad_schema")
+          project =
+            ComposerProject
+              { composerDir = baseDir
+              , composerLock = currentDir </> $(mkRelFile "not_tested_here")
+              }
+       in do
+            shouldThrow (runTestEffects $ licenseAnalyzeProject project >> pure ()) anyException
```
scruffystuffs, line 75: Ideally, we should use our normal error handling to test this.  I believe that `readContentsJson` returns `Diagnostics.fatal` if there is a parsing error.  We can use `Diagnostics.recover` to catch that, which performs `m a -> m (Maybe a)`, returning `Nothing` if there was some error.  Checking for `Nothing` here would catch this properly.

test/Composer/ComposerJsonSpec.hs
```diff
@@ -0,0 +1,75 @@
...
+                                      { licenseType = LicenseSPDX
+                                      , licenseValue = "GPL"
+                                      }
+                                  , License
+                                      { licenseType = LicenseSPDX
+                                      , licenseValue = "BSD"
+                                      }
+                                  ]
+                              }
+                          ]
```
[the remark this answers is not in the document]
scruffystuffs: This should probably written as:
```
do
  licenses <- licenseAnalyzeProject project
  licenses `shouldBe'` arrayLicenses
```

with `arrayLicenses` (or whatever name you choose) defined at the top level.

src/Strategy/Composer.hs
```diff
@@ -99,6 +115,46 @@ instance FromJSON CompDep where
       <*> obj .:? "require"
       <*> obj .:? "require-dev"
 
+newtype ComposerJson = ComposerJson
+  { license :: Maybe [License]
```
scruffystuffs, line 119: Part of what's made this parser so complicated is that we're trying to combine the `composer.json` schema with our internal data model.  We should separate these, and simply model the `composer.json` schema as-is.

scruffystuffs: My instinct here is to write something like this:

```
newtype ComposerJson = ComposerJson { composerLicenses :: [Text] }

instance FromJSON ComposerJson where
  parseJSON = "much simpler parser which converts strings to singleton lists and missing keys to empty lists"

analyzeLicenses = do
  composerJson <- readContentsJson foo
  pure . map (licenseToResult . textToLicense) . filter Text.isEmpty $ composerLicenses composerJson 
```

test/Composer/ComposerJsonSpec.hs
```diff
@@ -0,0 +1,83 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Composer.ComposerJsonSpec (spec) where
+
+import App.Pathfinder.Types (LicenseAnalyzeProject (licenseAnalyzeProject))
+import Control.Effect.Diagnostics qualified as Diagnostics
+import Path (mkRelDir, mkRelFile, toFilePath, (</>))
+import Path.IO (getCurrentDir)
+import Path.Posix (Abs, Dir, Path)
```
scruffystuffs, line 9: This line is causing your windows build errors.  You need to import these from `Path`, since that will properly dispatch to the correct platform-specific module at build time.

csasarak: Thanks!

scruffystuffs (approved):
LGTM, make sure you sign the CLA so that your builds actually complete.

src/Strategy/Composer.hs
```diff
@@ -99,6 +151,23 @@ instance FromJSON CompDep where
...
+  parseJSON (Array t) = ComposerLicenses <$> parseJSON (Array t)
+  parseJSON (String t) = pure $ ComposerLicenses [t]
+  parseJSON _ = fail "Invalid schema for key 'license' in composer.json"
+
+newtype ComposerJson = ComposerJson
+  { license :: ComposerLicenses
+  }
+  deriving (Eq, Ord, Show)
+
+instance FromJSON ComposerJson where
```
meghfossa, line 167: [nit] Add reference doc as comment somewhere: https://getcomposer.org/doc/04-schema.md#license

src/Strategy/Composer.hs
```diff
@@ -72,6 +113,17 @@ instance ToJSON ComposerProject
...
   analyzeProject _ = getDeps
 
+instance LicenseAnalyzeProject ComposerProject where
+  licenseAnalyzeProject = analyzeLicenses . composerJson
+
+analyzeLicenses :: (Has ReadFS sig m, Has Diagnostics sig m) => Maybe (Path Abs File) -> m [LicenseResult]
+analyzeLicenses Nothing = pure []
+analyzeLicenses (Just licenseFileName) = do
+  let mkLicenseResults l = [LicenseResult (toFilePath licenseFileName) l]
+      textToLicense = License LicenseSPDX
```
meghfossa, line 123: Is this guaranteed that licences will always be in SPDX?

Ideally we can check against https://github.com/spdx/license-list-data, but this is likely out of scope for PR. Either way would prefer tiny comment.

meghfossa (approved):
🚀  Added some optional nits.

src/Strategy/Composer.hs
scruffystuffs: Probably a link to the composer docs and a short comment should suffice, the docs are pretty clear.

csasarak: The recommendation is that they be SPDX. I have added a link to the composer documentation to the code.

meghfossa: 👍

PR #763, opened 2022-01-14
Add list functions I wish I had

scruffystuffs:
List-based functions I find myself consistently checking for.

## Checklist

- [x] ~I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.~
- [x] ~I linked this PR to any referenced GitHub issues, if they exist.~

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -23,11 +23,11 @@
 - suggest: {lhs: "(Data.Set.size x) == 0" , rhs: "Data.Set.null x"}
 - suggest: {lhs: "(Data.Set.size x) /= 0" , rhs: "not $ Data.Set.null x"}
 
-# Forbidden items, only allowed in compile-time code, or test code.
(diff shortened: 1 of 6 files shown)


meghfossa (approved):
I would prefer to use these
 
- https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-Maybe.html#v:listToMaybe. 
- https://hackage.haskell.org/package/base-4.16.0.0/docs/Data-List.html#v:singleton

Can you add `head -> head'` and singleton function to hlint's `suggest`, if you want to follow Data.List.Extra module?

scruffystuffs:
`singleton` is only available for GHC >= 9.0, which is why we annoyingly have to define our own.
`head -> head'` as a lint doesn't really work, since they have different signatures, I can check about adding a note to the `head` error.

I changed `head'` to be an alias for `listToMaybe`, since I can never remember the existence of `listToMaybe`.

meghfossa (approved):
🚀

PR #765, opened 2022-01-15
VSI Rewrite: Drop plugin usage

kitified:
# Overview

Finalizes the VSI rewrite, bring all VSI functionality into pure Haskell.

## Acceptance criteria

The goal of this project was:

* Better error reporting
* Better progress reporting for users
* Easier for other analysis team members to contribute or maintain this functionality
* Merge work streams (for example, archive expansion)

The one place this doesn't quite deliver is on performance: this version of VSI is about 6x slower at fingerprinting than the Go version. However, overall scan times remain roughly consistent with or better than what they were before:

* Most of the end-to-end scan performance bottleneck is in the forensics process, which hasn't changed.
* A substantial server side issue was fixed, improving ingestion speed 166x. This more than covers for the slower client.

## Testing plan

I have successfully scanned the following projects, which display the expected results:

- [x] VPS Demo: "standard" 
- [x] VPS Demo: "large"
- [x] VSI "example internal" demo projects
- [x] Chromium (can't actually complete in cloud analysis, but completed fingerprinting!)

## Risks

Not risky: VSI isn't in the critical path of any users yet.

## References

Closes https://github.com/fossas/team-analysis/issues/844

## Checklist

### DIFF
--- a/src/App/Fossa/VSI/Analyze.hs
+++ b/src/App/Fossa/VSI/Analyze.hs
@@ -28,7 +28,7 @@ import Effect.ReadFS (ReadFS)
 import Fossa.API.Types (ApiOpts)
 import Path (Abs, Dir, File, Path, Rel, SomeBase (Abs, Rel), isProperPrefixOf, (</>))
 import Path qualified as P
-import Path.Extra (tryMakeRelative)
+import Path.Extra (renderRelative, tryMakeRelative)
 
 type FingerprintEffs sig m =
   ( Has (Lift IO) sig m
@@ -165,8 +165,9 @@ runFingerprint :: FingerprintEffs sig m => PathFilters -> Path Abs Dir -> m (Map
 runFingerprint filters root = flip walk' root $ \dir _ files -> do
   if filters `allow` dir
     then do
-      logDebug . pretty $ "Fingerprint: " <> toText (show dir)
-      fingerprints <- traverse (fingerprintFile filters root) files
+      fingerprints <-
+        context ("Fingerprint directory contents: " <> renderRelative root dir) $
+          traverse (fingerprintFile filters root) files
       pure (Map.unions fingerprints, WalkContinue)
     else do
       logDebug . pretty $ "Excluded by filter: " <> toText (show dir)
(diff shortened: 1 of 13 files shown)


src/App/Fossa/VSI/Analyze.hs
```diff
@@ -0,0 +1,352 @@
...
+  ) =>
+  -- | Parallelization factor
+  Int ->
+  -- | Channel on which new files are collected
+  TBMChan (Path Rel File, Combined) ->
+  Path Abs Dir ->
+  AllFilters ->
+  m ()
+runFingerprintDiscovery capabilities files dir filters = do
+  withTaskPool capabilities (updateProgress "Fingerprint files") . runAtomicCounter $ do
```
cnr, line 101: [optional] if we "create" a new sticky logger here:

```hs
  runStickyLogger SevInfo . runTaskPool [...]
```

..rather than sending the constraint upward, we can eliminate the contention on the sticky logger output between this `updateProgress` and the `updateProgress` of the parent taskpool.

this will result in two separate "sticky" progress lines at the bottom of the terminal

src/Control/Effect/Diagnostics.hs
```diff
@@ -98,6 +99,12 @@ fatalOnIOException ctx go = context ctx $ catch go die'
   where
     die' (e :: IOException) = fatalText ("io exception: " <> toText (show e))
 
+-- | Throw a generic error message on any exception, wrapped in a new 'context' using the provided @Text@.
+fatalOnSomeException :: (Has (Lift IO) sig m, Has Diagnostics sig m) => Text -> m a -> m a
+fatalOnSomeException ctx go = context ctx $ catch go die'
```
cnr, line 104: use `Control.Exception.Extra.safeCatch` here

the short explanation is that we don't want to catch "async exceptions", like `Ctrl-C` -- because we'll end up continuing/ignoring it, which stops us from being able to terminate early with `Ctrl-C`

`safeCatch` is a wrapper around `catch` that ignores async exceptions

cnr (approved):
This is really great. A few minor comments

PR #769, opened 2022-01-16
Refractor discovered project type's datatype

meghfossa:
# Overview

This PR modifies,  Discovered projectType's datatype. It changes types from string to sum type. 

## Testing plan

N/A

## Risks

N/A

## References

N/A

## Checklist

- [ ] ~I added tests for this PR's change (or confirmed tests are not viable).~
- [ ] ~If this PR introduced a user-visible change, I added documentation into `docs/`.~
- [ ] ~I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.~
- [ ] ~I updated `*schema.json` if I have made changes for `.fossa.yml`, `fossa-deps.{json, yaml, yml}`. You may also need to update these if you have added/removed new dependency (e.g. pip) or analysis target type (e.g. poetry).~
- [ ] ~I linked this PR to any referenced GitHub issues, if they exist.~

### DIFF
--- a/src/App/Pathfinder/Scan.hs
+++ b/src/App/Pathfinder/Scan.hs
@@ -22,6 +22,7 @@ import Control.Monad.IO.Class (MonadIO)
 import Data.Aeson
 import Data.Bool (bool)
 import Data.ByteString.Lazy qualified as BL
+import Data.String.Conversion (ToText (toText))
 import Data.Text (Text)
 import Discovery.Projects (withDiscoveredProjects)
 import Effect.Exec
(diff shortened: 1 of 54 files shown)


src/App/Pathfinder/Scan.hs
```diff
@@ -133,8 +134,8 @@ instance ToJSON CompletedLicenseScan where
 mkLicenseScan :: DiscoveredProject n -> [LicenseResult] -> ProjectLicenseScan
 mkLicenseScan project licenses =
   ProjectLicenseScan
-    { licenseStrategyType = projectType project
-    , licenseStrategyName = projectType project
+    { licenseStrategyType = toText (show $ projectType project)
+    , licenseStrategyName = toText (show $ projectType project)
```
scruffystuffs, line 138: If we keep the `ToText` instance, this should just be `toText`.  Otherwise, we can use `showT` here.

scruffystuffs (approved):
Looks good, one question about the `ToText` instance.

PR #773, opened 2022-01-19
Remove fdescribe.

csasarak:
# Overview

A number of our unit tests weren't running because of an `fdescribe` which focuses the test runner on a specific set of tests and ignores the others. This PR puts them back to `describe`.

## Acceptance criteria

Remove `fdescribe` in unit tests.

## Testing plan

When running unit tests I found that only some of them were running. I tested this by removing fdescribe then running again. 

## Risks

None

## References

None

## Checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -43,8 +43,8 @@
   - {name: "Data.Map.!", within: []}
     # TODO: remove the need to special-case this test module.
   - {name: fromJust, within: [App.DocsSpec]}
-  - {name: fdescribe, within: []}
-  - {name: fit, within: []}
-  - {name: focus, within: []}
-  - {name: fspecify, within: []}
-  - {name: fcontext, within: []}
+  - {name: Test.Hspec.fdescribe, within: []}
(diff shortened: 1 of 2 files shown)


meghfossa (approved):
LGTM

.hlint.yaml
```diff
@@ -43,3 +43,8 @@
   - {name: "Data.Map.!", within: []}
     # TODO: remove the need to special-case this test module.
   - {name: fromJust, within: [App.DocsSpec]}
+  - {name: fdescribe, within: []}
+  - {name: fit, within: []}
+  - {name: focus, within: []}
+  - {name: fspecify, within: []}
+  - {name: fcontext, within: []}
```
scruffystuffs, line 50: Can we prefix all of these with `Test.Hspec`?  I want to make sure that a function called `focus` is allowed if provided elsewhere.

scruffystuffs (approved):

.hlint.yaml
csasarak: I've added that, thanks!

PR #777, opened 2022-01-20
Fixes the build in master

meghfossa:
# Overview

Fixes the build in master branch, this was due to recent merge making import explicits, which I hadn't rebase prior to merging #769.


scruffystuffs (approved):
This is why I don't like merging (as opposed to rebasing).  3-way merge sometimes (though not exactly often) just silently does the wrong thing.  With rebasing, this case is incredibly rare.

PR #779, opened 2022-01-21
Pkglock nested deps

scruffystuffs:
# Overview

When a `package-lock.json` has the following structure (simplified for clarity), we were not reporting the edge between `js-yaml` and `argparse 2.0.1`:

```json
{
  "dependencies": {
    "argparse": {
      "version": "1.0.10",
      "requires": {
        "sprintf-js": "~1.0.2"
      }
    },
    "js-yaml": {
      "version": "4.1.0",
      "requires": {
        "argparse": "^2.0.1"
      },
      "dependencies": {
        "argparse": {
          "version": "2.0.1",
        }
      }
    },
    "sprintf-js": {
      "version": "1.0.3",
    }
  }
}
```

This was causing `argparse 2.0.1` to not be reported, since it was then pruned in the `SourceUnit` conversion (it was an orphan node in the graph without that edge).

To resolve this, we carry the parent into the recursive dep analysis, and add an edge when analyzing the child.
When implementing, another oversight was revealed: When adding edges from packages under the `requires` key, we try to detect the package's version from the top-level deps.  Now, we look at the local `dependecies` block first (at the same level of the `requires` that we're currently working on), before falling back to the top-level `dependencies` package list.

## Acceptance criteria

- Tests cover the failing case described in the linked issue
  - Specifically, we test that an edge is created between `js-yaml` and `argparse 2.0.1` in the example above.  This causes `pruneUnreachable` to not delete `argparse 2.0.1`, which would otherwise be an orphan node.
- Tests pass

## Testing plan

Unit testing should be sufficient

## References

Fixes fossas/team-analysis#874

## Checklist

### DIFF
--- a/.hlint.yaml
+++ b/.hlint.yaml
@@ -44,7 +44,8 @@
     # TODO: remove the need to special-case this test module.
   - {name: fromJust, within: [App.DocsSpec]}
   - {name: Test.Hspec.fdescribe, within: []}
-  - {name: Test.Hspec.fit, within: []}
+    # We define fit' in terms of fit,so this is necessary
+  - {name: Test.Hspec.fit, within: [Test.Effect]}
   - {name: Test.Hspec.focus, within: []}
   - {name: Test.Hspec.fspecify, within: []}
   - {name: Test.Hspec.fcontext, within: []}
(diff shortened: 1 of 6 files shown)


cnr (approved):
🙌

PR #778, opened 2022-01-21
VSI + Dynamic Linking: Parse `ldd` output

kitified:
# Overview

Parse `ldd` output into a set of local file paths, which represent dynamically linked dependencies.

_I'm leaving this PR out of the changelog as it's literally irrelevant for users; I'll include this PR in a merged changelog message when I'm done (if anyone disagrees let me know!)_

## Acceptance criteria

We now have the ability to parse `ldd` output.

## Testing plan

I relied on automated tests passing.

## Risks

Not really risky.

## References

Closes https://github.com/fossas/team-analysis/issues/885

## Checklist

- [x] I added tests for this PR's change (or confirmed tests are not viable).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] I updated `Changelog.md` if this change is externally facing. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] I updated `*schema.json` if I have made changes for `.fossa.yml`, `fossa-deps.{json, yaml, yml}`. You may also need to update these if you have added/removed new dependency (e.g. pip) or analysis target type (e.g. poetry).
- [x] I linked this PR to any referenced GitHub issues, if they exist.

PR #780, opened 2022-01-21
VSI + Dynamic Linking: Resolve dynamic deps to `SourceUnit`

kitified:
# Overview

Provides the ability to resolve dynamic dependencies into a `SourceUnit`.

_I'm leaving this PR out of the changelog as it's literally irrelevant for users; I'll include this PR in a merged changelog message when I'm done (if anyone disagrees let me know!)_

## Acceptance criteria

We now have the ability to resolve dynamic dependencies obtained by inspecting the local package manager into a `SourceUnit` that'll be reported to FOSSA.

## Testing plan

I relied on automated testing.

## Risks

None

## References

Closes https://github.com/fossas/team-analysis/issues/882

## Checklist

### DIFF
--- a/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
+++ b/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
@@ -37,7 +37,7 @@ toSourceUnit root dependencies = do
   let (resolved, unresolved) = sortResolvedUnresolved dependencies
   binaries <- traverse (analyzeSingleBinary root) unresolved
 
-  let project = toProject root $ Graphing.fromList (map toDependency resolved)
+  let project = toProject root $ Graphing.directs (map toDependency resolved)
   let unit = Srclib.toSourceUnit False project
   pure $ unit{additionalData = fmap toDepData (Just binaries)}
   where
--- a/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
+++ b/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
@@ -12,7 +12,7 @@ import Control.Algebra (Has)
 import Control.Effect.Diagnostics (Diagnostics)
 import Control.Effect.Lift (Lift)
 import Data.Set (Set, toList)
-import Data.Text (intercalate)
+import Data.Text (Text, intercalate)
 import DepTypes (DepType (LinuxAPK, LinuxDEB, LinuxRPM), Dependency (..), VerConstraint (CEq))
 import Effect.Logger (Logger)
 import Effect.ReadFS (ReadFS)
@@ -50,6 +50,7 @@ toDependency ResolvedLinuxPackage{..} = case resolvedLinuxPackageManager of
   LinuxPackageManagerRPM -> renderRPM resolvedLinuxPackageMetadata
   LinuxPackageManagerAPK -> renderAPK resolvedLinuxPackageMetadata
   where
+    render :: DepType -> [Text] -> [Text] -> Dependency
     render fetcher projectParts revisionParts =
       Dependency
         { dependencyType = fetcher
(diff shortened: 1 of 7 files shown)


src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
```diff
@@ -0,0 +1,86 @@
+{-# LANGUAGE RecordWildCards #-}
+
+module App.Fossa.VSI.DynLinked.Internal.Resolve (
```
scruffystuffs, line 3: Any reason why this is named `Internal`?  This doesn't seem break any sort of module encapsulation, which is normally what haskell modules with `Internal` are made for.

kitified: I plan to have `App.Fossa.VSI.DynLinked.Analyze`, which uses this and other `App.Fossa.VSI.DynLinked.Internal` modules. 

From there, I wanted to make it clear that `App.Fossa.VSI.DynLinked.Analyze` is the correct entry point, not any of the other stuff, while still exporting a bunch of internal functions that are not really needed except for testing (this effect is much more pronounced in [my ldd parsing PR](https://github.com/fossas/fossa-cli/pull/778)). It's my understanding `Internal` is the correct way to do that.

scruffystuffs: Ok, I just couldn't tell from this PR, but it seems like it's all going to work together.

kitified:
Wes has a lot going on, reassigning to someone else for review please!

src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
```diff
@@ -0,0 +1,86 @@
...
+import Types (DiscoveredProjectType (VsiProjectType), GraphBreadth (Complete))
+
+-- | Resolves a set of dynamic dependencies into a @SourceUnit@.
+-- Any @DynamicDependency@ that isn't resolved to a Linux package dependency is converted to an unknown binary dependency.
+toSourceUnit :: (Has (Lift IO) sig m, Has Logger sig m, Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> Set DynamicDependency -> m SourceUnit
+toSourceUnit root dependencies = do
+  let (resolved, unresolved) = sortResolvedUnresolved dependencies
+  binaries <- traverse (analyzeSingleBinary root) unresolved
+
+  let project = toProject root $ Graphing.fromList (map toDependency resolved)
```
meghfossa, line 32: [nit] I know `fromList` is alias for directs, but would prefer `directs` to signify they are direct deps.

src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
```diff
@@ -0,0 +1,86 @@
...
+--
+-- This function reverses the order of dependencies for performance reasons,
+-- but dependency order isn't significant so we don't reverse it back.
+sortResolvedUnresolved :: Set DynamicDependency -> ([ResolvedLinuxPackage], [Path Abs File])
+sortResolvedUnresolved = buckets [] [] . toList
+  where
+    buckets resolved unresolved [] = (resolved, unresolved)
+    buckets resolved unresolved (dep : remaining) = case dynamicDependencyResolved dep of
+      Nothing -> buckets resolved (dynamicDependencyDiskPath dep : unresolved) remaining
+      Just linuxPackage -> buckets (linuxPackage : resolved) unresolved remaining
```
cnr, line 86: [optional] in the same way that general `for` loops make me think too hard, general recursion also makes me think too hard, even for really simple snippets like this one

this case is really simple and perfectly fine as-is, so this is genuinely optional. here's an alternative implementation that might make it ever so slightly clearer / more-"obviously correct":

We can break this up into two smaller steps we can compose:
```hs
step1 :: [DynamicDependency] -> [Either ResolvedLinuxPackage (Path Abs File)]
step1 = undefined

step2 :: [Either ResolvedLinuxPackage (Path Abs File)] -> ([ResolvedLinuxPackage], [Path Abs File])
step2 = undefined

sortResolvedUnresolved :: Set DynamicDependency -> ([ResolvedLinuxPackage], [Path Abs File])
sortResolvedUnresolved = step2 . step1 . Set.toList
```

`step1` is fairly straightforward, where we'd map some function `DynamicDependency -> Either ResolvedLinuxPackage (Path Abs File)` into the list:

```hs
step1 :: [DynamicDependency] -> [Either ResolvedLinuxPackage (Path Abs File)]
step1 = map _something
```

`step2` is already defined for us in `Data.Either`:
```hs
partitionEithers :: [Either a b] -> ([a],[b])
```

..and putting it all together:

```hs
sortResolvedUnresolved :: Set DynamicDependency -> ([ResolvedLinuxPackage], [Path Abs File])
sortResolvedUnresolved = partitionEithers . map something . Set.toList
  where
    something :: DynamicDependency -> Either ResolvedLinuxPackage (Path Abs File)
    something dep = case ... of ...
```

---

There are a couple of reasons I say this is more "obviously-correct":

1. We know that `map`/`partitionEithers` will preserve the same number of elements, so we don't need to check our general-recursive version to make sure we're not dropping/adding elements. And their very-general types ensure they're not doing anything tricky with the provided values
2. We've narrowed the interesting logic into this `something` function with a much narrower type than our `buckets` from before. There are fewer things that can go wrong

---

Lastly, it's worth noting that this `partitionEithers . map something` implementation retains the same order of elements in the provided list.

`buckets` is essentially a hand-rolled `foldl` over the list, whereas `partitionEithers` uses a `foldr`. A useful exercise might be to implement with both `foldl` and `foldr` versions of something like `partitionEithers` -- possibly evaluating by hand -- to see where and why they differ

cnr (approved):

meghfossa (approved):
🚀

src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
kitified: This is such good feedback, thank you!

PR #781, opened 2022-01-25
Entrypoint refactor [Final]

scruffystuffs:
# Overview

This is the collection of all non-followup tickets in fossas/team-analysis#796.

## Acceptance criteria

- All entrypoints and config gathering procedures are standardized and refactored.

## Testing plan

Not yet tested fully, a test plan will be written and executed soon.

## Risks

- Many of the changes have gone through 10+ git rebases, due to the size and complexity of this change.  Please be aware that copy-paste errors are likely, though there should not be a large number.
- We have no automated testing for our frontend, and as a result, we do not concretely know that there were no unintended changes to the CLI UI.  We should probably have some automated UI tests.  For now, manual testing will cover this.
- There is no way to enforce conflicting config semantics through types (i.e.: types can't tell us when we accidentally prefer config file values to CLI values), so that should be checked as well.

## References

closes fossas/team-analysis#837

## Checklist


src/App/Fossa/Subcommand.hs
```diff
@@ -0,0 +1,50 @@
...
+  , configLoader :: cli -> EffStack (Maybe ConfigFile) -- CLI args can control where we look for config file
+  , optMerge :: Maybe ConfigFile -> EnvVars -> cli -> EffStack cfg
+  , perform :: cfg -> EffStack ()
+  }
+
+type EffStack = ExecIOC (ReadFSIOC (DiagnosticsC (LoggerC IO)))
+
+class GetSeverity a where
+  getSeverity :: a -> Severity
+  getSeverity = const SevInfo
```
cnr, line 35: I think the way we kill `GetSeverity` might be to add a `Logger` operation that allows changing the filter severity, e.g., `withSeverityFilter :: Has Logger sig m => Severity -> m a -> m a`

for the purposes of `runSubCommand`, we would just start with `withDefaultLogger SevInfo`

src/Control/Timeout.hs
```diff
@@ -0,0 +1,103 @@
...
+    threadDelay $ durationToMicro duration
+    -- We fill the MVar here, which is the cancel signal.
+    -- SAFETY: putMVar can block infinitely if we do it twice, since we never
+    --   empty the MVar. Since we never expose the underlying MVar to the
+    --   consumer, we only have to validate that this putMVar call is safe,
+    --   and that all other usage in this module is non-blocking.
+    putMVar cancel ()
+  -- Once 'start' is taken, we know that the fork has begun the timer
+  -- We have to actually wait for this value, which is why we use $!.
+  sendIO $! takeMVar start
```
cnr, line 90: can we eliminate the `start` var entirely / just assume the other thread is doing the thing?

src/Discovery/Filters.hs
```diff
@@ -44,20 +51,49 @@ filterIsVSIOnly AllFilters{..} = do
       _ -> False
 
 data AllFilters = AllFilters
-  { legacyFilters :: [TargetFilter]
-  , includeFilters :: FilterCombination
-  , excludeFilters :: FilterCombination
+  { includeFilters :: FilterCombination Include
+  , excludeFilters :: FilterCombination Exclude
```
cnr, line 55: [no action] good refactor 🚀

src/App/Fossa/Analyze.hs
```diff
@@ -332,33 +301,40 @@ analyze ::
...
-  ModeOptions ->
-  AllFilters ->
   m ()
-analyze (BaseDir basedir) destination override unpackArchives jsonOutput includeAll ModeOptions{..} filters = Diag.context "fossa-analyze" $ do
+analyze = Diag.context "fossa-analyze" $ do
   capabilities <- sendIO getNumCapabilities
 
+  -- TODO: refactor other code to use config reader, rather than passing
+  -- values directly and defining them upfront like this.
+  cfg <- ask
```
cnr, line 312: This is isomorphic to passing in the config as an argument to `analyze`. Let's do that instead, because the `Reader` isn't used anywhere else

cnr (commented):
These changes look good. Most of my comments are nits. Would like to hear more about a test plan before approving

src/App/Fossa/Config/Container.hs
```diff
@@ -0,0 +1,343 @@
...
+  ) =>
+  Maybe ConfigFile ->
+  EnvVars ->
+  ContainerCommand ->
+  m ContainerScanConfig
+mergeOpts cfgfile envvars = \case
+  ContainerAnalyze opts -> AnalyzeCfg <$> mergeAnalyzeOpts cfgfile envvars opts
+  ContainerTest opts -> TestCfg <$> mergeTestOpts cfgfile envvars opts
+  ContainerParseFile fp -> ParseCfg <$> mergeParseFileOpts cfgfile envvars fp
+  ContainerDumpScan opts -> DumpCfg <$> mergeDumpOpts cfgfile envvars opts
```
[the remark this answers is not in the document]
scruffystuffs: Not sure I agree here (or rather, not sure **yet**, since I haven't tried it).  These are small enough that combining them into one module is significantly easier than splitting them up.  I can take a swing at this and see if I like it, but my initial reaction is that it might be more confusing by being too spread out.

src/App/Fossa/Subcommand.hs
scruffystuffs: I'd have to check if we would be losing debug info from some of the operations that were moved into the configuration phase, e.g. revision inferrence.  If we aren't missing anything, then this is a good move.

src/App/Fossa/Analyze.hs
scruffystuffs: Yeah, that was from the original pass, but I think that's not nearly as useful now.  I'll change it.

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -1,37 +1,146 @@
...
+
+defaultConfigFileLocation :: Path Rel File
+defaultConfigFileLocation = $(mkRelFile ".fossa.yml")
+
+data ConfigLocation = ConfigLocation
+  { -- | Did the user specify the path (True/required), or is this default (False/optional)?
+    isRequired :: Bool
+  , -- | What is the absolute file path?
+    configFilePath :: Path Abs File
+  }
```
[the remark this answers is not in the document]
scruffystuffs: Addressed [here](https://github.com/fossas/fossa-cli/pull/781/files#diff-22466a7d10db58d3535e227faf597eb7fdbdc2ed572a3fde3830e221a64d003bR89), but I don't know if that's much clearer than before.

src/App/Fossa/Subcommand.hs
scruffystuffs: So, if we did this, we wouldn't lose any debug info, I've checked thoroughly.

However, I'm not sure this works out, for 2 reasons:

- I'm not sure how we would run `withSeverityFilter` from `runSubCommand` without having some uniform way of accessing the `Severity` anyways. I'd rather have this (admittedly not ideal) typeclass than allow the generic subcommand handler to NOT properly start up the logger.  Notably, the type system can't tell us that we haven't started the "correct" logger effect handler yet.
- How would we actually define `withSeverityFilter`?  
  - Do we just run a new logger underneath?  Are there concerns with running the console handling functions more than once?
  - Do we use the fact that it's a simple effect and somehow create a `local`-like for `SimpleC` that replaces simple effects?

Unless those questions have really simple answers, I'd like to leave this as-is for now.  Happy to leave a TODO and tackle it later, though.

cnr: oh right -- I forgot `Logger` is a `Simple` effect. that makes things harder. we can just keep this as-is

src/Control/Timeout.hs
cnr: unreliable on what timescale? if the thread is sleeping for anything more than a few ms, I don't expect this wait-for-start to substantially impact the timing

scruffystuffs: I was testing at 1ms scales, that's a fair point, our timing is never nearly that fine-grained.  I'll retest it at 200ms-2s scales.  If it's not that substantial, we can just remove this.

src/App/Fossa/Config/ConfigFile.hs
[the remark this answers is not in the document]
scruffystuffs: nvm, Looked again and this is more straightforward by virtue of the two separate code paths.

src/Control/Timeout.hs
scruffystuffs: Actually, even that is too small for what we normally do.  I can do manual testing at a 10s scale, and leave unit tests for about 200ms-1s.  At that point, there's almost no way this is actually needed.

scruffystuffs:
CI is running into [this issue](https://gitlab.haskell.org/ghc/ghc/-/issues/12576).
Looks like it's fixed in GHC `9.0.1`, but not sure if there's anything we can do about it here.  I'm trying some random things to see if it's really consistent, but unchanged rebuilds seem to be consistently hitting this issue.

src/App/Fossa/Test.hs
```diff
@@ -68,11 +68,9 @@ testMain config = runStickyLogger SevInfo $
             TestOutputJson -> logStdout . decodeUtf8 . Aeson.encode $ issues
             TestOutputPretty -> pure ()
         n -> do
-          logError $ "Test failed. Number of issues found: " <> pretty n
           if null (issuesIssues issues)
-            then logError "Check the webapp for more details, or use a full-access API key (currently using a push-only API key)"
+            then logError "Cannot display issues in the terminal. Check the webapp for more details, or use a full-access API key (currently using a push-only API key)."
```
cnr, line 72: [nit] wording: consider something like "A push-only API key was used, so we can't display issue details. Check the webapp for issue details, or run this command again with a full-access API key"

src/App/Fossa/Test.hs
```diff
@@ -68,11 +68,9 @@ testMain config = runStickyLogger SevInfo $
...
             else case outputType of
               TestOutputPretty -> logError $ pretty issues
               TestOutputJson -> logStdout . decodeUtf8 . Aeson.encode $ issues
-
-          sendIO exitFailure
+          fatalText $ "The scan has revealed issues, test failed. Number of issues found: " <> showT n
```
cnr, line 76: [nit] wording: "test failed" is scary; let's kill it from this message
```suggestion
          fatalText $ "The scan has revealed issues. Number of issues found: " <> showT n
```

cnr (approved):
🚀

PR #785, opened 2022-01-27
Settings gradle detection

csasarak:
# Overview

In newer versions of gradle, there may only be a `settings.gradle` file in a project root directory. We would not be able to discover subprojects in this case and would report no dependencies. 

## Acceptance criteria

Detect when a project has a top-level file starting with `settings.gradle` without a `build.gradle` and also run discovery.

## Testing plan

I tested with some sample repositories that had `settings.gradle` only in their root directories. These included projects in Java/Kotlin. I also wrote an integration test which tests both a real repository (spring-boot) with both files in its root and one of the samples with only `settings.gradle`.

[This](https://docs.gradle.org/7.3.3/samples/zips/sample_building_java_applications-groovy-dsl.zip) project will fail currently on master. On this branch it succeeds and reports dependencies.

## Risks

_Highlight any areas that you're unsure of, or want reviewers to pay particular attention to._

As part of the integration testing I extended some code in `TestFixtureUtils.hs` to support more archive formats than just `.tar.gz`. Please look at how it detects which extraction function to use and also how it reports errors.

## References

Closes fossas/team-analysis#846

## Checklist

### DIFF
--- a/integration-test/Analysis/GradleSpec.hs
+++ b/integration-test/Analysis/GradleSpec.hs
@@ -13,7 +13,7 @@ import Analysis.FixtureUtils (
   FixtureArtifact (FixtureArtifact),
   FixtureEnvironment (NixEnv),
  )
-import Path (mkRelDir, reldir)
+import Path (reldir)
 import Strategy.Gradle qualified as Gradle
 import Test.Hspec (Spec, aroundAll, describe, it, shouldBe)
 import Types (DiscoveredProjectType (..))
@@ -31,7 +31,7 @@ springBoot =
     $ FixtureArtifact
       "https://github.com/spring-projects/spring-boot/archive/refs/tags/v3.0.0-M1.tar.gz"
       [reldir|gradle/sample/|]
-      $(mkRelDir "spring-boot-3.0.0-M1")
+      [reldir|spring-boot-3.0.0-M1|]
 
 gradleSettingsOnly :: AnalysisTestFixture (Gradle.GradleProject)
 gradleSettingsOnly =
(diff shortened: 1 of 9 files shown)


scruffystuffs:
Looks like the dep-scan job is stuck.  Cancelling doesn't work.  Can you push a new commit so that it resubmits?

integration-test/Analysis/GradleSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+
+gradleSettingsOnly :: AnalysisTestFixture (Gradle.GradleProject)
+gradleSettingsOnly =
+  AnalysisTestFixture
+    "gradle-java"
+    Gradle.discover
+    goEnv
+    Nothing
+    $ FixtureArtifact
+      "https://docs.gradle.org/current/samples/zips/sample_building_java_applications-groovy-dsl.zip"
```
scruffystuffs, line 44: [nit]: Use a link that's less likely to rot.
```suggestion
      "https://docs.gradle.org/7.3.3/samples/zips/sample_building_java_applications-groovy-dsl.zip"
```

integration-test/Analysis/GradleSpec.hs
```diff
@@ -0,0 +1,67 @@
...
+  FixtureArtifact (FixtureArtifact),
+  FixtureEnvironment (NixEnv),
+ )
+import Path (mkRelDir, reldir)
+import Strategy.Gradle qualified as Gradle
+import Test.Hspec (Spec, aroundAll, describe, it, shouldBe)
+import Types (DiscoveredProjectType (..))
+
+goEnv :: FixtureEnvironment
+goEnv = NixEnv ["gradle"]
```
scruffystuffs, line 22: [nit]: copy paste error?
```suggestion
gradleEnv :: FixtureEnvironment
gradleEnv = NixEnv ["gradle"]
```

scruffystuffs (approved):
We can probably mark a release here, go ahead and update the changelog with `3.0.17` and ping me for details on releasing if you need to.

integration-test/Analysis/GradleSpec.hs
csasarak: Good idea, thanks!

PR #788, opened 2022-02-01
VSI + Dynamic Linking: lookup linux distro

kitified:
# Overview

Provides the ability to look up the linux distro.

## Acceptance criteria

We need to be able to identify the linux distro in which we are running in order to report linux locators.

## Testing plan

I relied on automated testing.

## Risks

No risk, this code is unreachable.

## References

Related to https://github.com/fossas/team-analysis/issues/883

## Checklist

### DIFF
--- a/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
+++ b/src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
@@ -5,7 +5,7 @@ module App.Fossa.VSI.DynLinked.Internal.Resolve (
   toSourceUnit,
   toDependency,
   environmentDistro,
-  readLinuxDistro,
+  parseLinuxDistro,
 ) where
 
 import App.Fossa.Analyze.Project (ProjectResult (..))
@@ -13,11 +13,10 @@ import App.Fossa.BinaryDeps (analyzeSingleBinary)
 import App.Fossa.VSI.DynLinked.Types (DynamicDependency (..), LinuxDistro (..), LinuxPackageManager (..), LinuxPackageMetadata (..), ResolvedLinuxPackage (..))
 import App.Fossa.VSI.DynLinked.Util (fsRoot, runningLinux)
 import Control.Algebra (Has)
-import Control.Applicative.Permutations (intercalateEffect, toPermutation)
-import Control.Effect.Diagnostics (Diagnostics, fatal, recover, (<||>))
+import Control.Effect.Diagnostics (Diagnostics, recover, (<||>))
 import Control.Effect.Lift (Lift)
-import Data.Char (isSpace)
 import Data.Either (partitionEithers)
+import Data.Map qualified as Map
 import Data.Set (Set, toList)
 import Data.String.Conversion (toText)
 import Data.Text (Text, intercalate)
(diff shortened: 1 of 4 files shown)


src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
```diff
@@ -105,3 +117,69 @@ sortResolvedUnresolved = partitionEithers . map forkEither . toList
...
+    prefixes :: [Text]
+    prefixes = map (<> "=") ["ID", "VERSION_ID"]
+
+    filteredLines :: Text
+    filteredLines = filterLinePrefixes contents
+
+    filterLinePrefixes :: Text -> Text
+    filterLinePrefixes = Text.unlines . filter lineIsUsed . fmap Text.strip . Text.lines
+      where
+        lineIsUsed line = any (`Text.isPrefixOf` line) prefixes
```
kitified, line 156: I would absolutely _love_ to handle this inside `parseLinuxDistro`, but I couldn't figure out how to make `toPermutation` work while ignoring fields it doesn't care about within my timebox. Interested in any suggestions if you have them.

src/App/Fossa/VSI/DynLinked/Util.hs
```diff
@@ -31,6 +37,10 @@ hasSetUID file = do
   stat <- fatalOnSomeException "get file status" . sendIO . getFileStatus $ P.toFilePath file
   pure $ fileMode stat .&. setUserIDMode /= 0
 
+-- | The root of the primary file system.
+fsRoot :: Path Abs Dir
+fsRoot = $(mkAbsDir "/")
+
 #endif
 
 runningLinux :: Bool
```
kitified, line 46: I'm probably going to extract `fsRoot` and a set of `runningLinux`-like helpers into a higher-level package in a future PR. They seem really helpful for cross-platform work.

src/App/Fossa/VSI/DynLinked/Util.hs
```diff
@@ -16,6 +18,10 @@ import System.Info qualified as SysInfo
 hasSetUID :: Monad m => Path t File -> m Bool
 hasSetUID _ = pure False
 
+-- | The root of the primary file system.
+fsRoot :: Path Abs Dir
+fsRoot = $(mkAbsDir "C:/")
```
cnr, line 23: IIRC it's not safe to assume `C:\` is the system root on windows, but I'm not going to block on this edge case

src/App/Fossa/VSI/DynLinked/Internal/Resolve.hs
```diff
@@ -105,3 +117,69 @@ sortResolvedUnresolved = partitionEithers . map forkEither . toList
...
+
+type Parser = Parsec Void Text
+
+parseLinuxDistro :: Parser LinuxDistro
+parseLinuxDistro =
+  intercalateEffect sc $
```
cnr, line 162: While I applaud the clever use of `toPermutation` here, another way to approach this might be to parse the file into a `Map Text Text`, and specifically `Map.lookup` the `ID` and `VERSION_ID` fields afterward

src/App/Fossa/VSI/DynLinked/Util.hs
scruffystuffs: I don't know of a way to change the primary system root from `C:\`, so this is likely a very far out edge case.  We can let someone run into this and then figure out a fix from there.

cnr (approved):
🚀

kitified:
@cnr I was reading over this again prior to merging and noticed I left a little logic bug in there... 8922945, if you want to take a look.

cnr:
@kitified lgtm 👍

PR #792, opened 2022-02-03
Uses Fossa API's error messages for api related errors

meghfossa:
# Overview

This PR uses provided FOSSA API error messages, instead of mangling error (and presuming) intent. This keeps the error messages consistent across API, UI, and CLI related to FOSSA API.  

#### Example Screenshot Change

CleanShot 2022-02-03 at 13 56 07@2x

## Acceptance criteria

You agree that, CLI should use provided error message from API instead of performing interpretation of them. This means the error messages related to API will always have to be updated upstream. Likewise, we forego coupling between CLI version and rendered API related error messages. 

- Fossa API errors messages are consistent across CLI. 
- CLI error messages provide ERROR uuid (provided via API) for faster support resolution. 
- CLI error messages are reflective, when team is not found.
- CLI error messages are reflective, when release group is not found.

## Backend work in progress

Improve error message wording to include potential corrective actions in following: 

- ~[ ] Improve NotFoundError message for project~
- ~[ ] Improve NotFoundError message for revision~
- ~[ ] Improve NotFoundError message for team~
- ~[ ] Improve NotFoundError message for Release Group~

I intent to have sibling PR on backend for these. 
Sibling PR: https://github.com/fossas/FOSSA/pull/7065

## Testing plan

```
make install-dev
fossa-dev analyze ../some-project/ --fossa-api-key incorrect-key
fossa-dev analyze ../some-project/ --team team-that-does-not-exist
fossa-dev test --project project-that-does-not-exist
fossa-dev test --project some-project --revision revision-that-does-not-exist
fossa-dev analyze --link --fossa-api-key push-only-api-key
```

## Risks

Not risky. 

## References

- Closes https://github.com/fossas/team-analysis/issues/754
- Works on https://github.com/fossas/team-analysis/issues/854

## Checklist

### DIFF
--- a/src/App/Fossa/FossaAPIV1.hs
+++ b/src/App/Fossa/FossaAPIV1.hs
@@ -69,9 +69,23 @@ import Network.HTTP.Req.Extra (httpConfigRetryTimeouts)
 import Network.HTTP.Types qualified as HTTP
 import Path (File, Path, Rel)
 import Srclib.Types
-import Text.URI (URI)
 import Text.URI qualified as URI
 import Prelude
+
+-- | Represents error emitted via FOSSA instance.
+-- This data-shape corresponds to 'PublicFacingError' type in backend,
+-- used for rendering errors from API endpoints, as well as FOSSA UI.
+data FossaPublicFacingError = FossaPublicFacingError
+  { fpeMessage :: Text
+  , fpeUuid :: Text
+  }
+  deriving (Show, Eq, Ord)
+
+instance FromJSON FossaPublicFacingError where
+  parseJSON = withObject "FossaPublicFacingError" $ \v ->
+    FossaPublicFacingError
+      <$> v .: "message"
+      <*> v .: "uuid"
 
 newtype FossaReq m a = FossaReq {unFossaReq :: m a}
   deriving (Functor, Applicative, Monad, Algebra sig)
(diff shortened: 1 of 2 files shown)


src/App/Fossa/FossaAPIV1.hs
```diff
@@ -69,10 +69,24 @@ import Network.HTTP.Req.Extra (httpConfigRetryTimeouts)
 import Network.HTTP.Types qualified as HTTP
 import Path (File, Path, Rel)
 import Srclib.Types
-import Text.URI (URI)
 import Text.URI qualified as URI
 import Prelude
 
+-- | Represents error emitted via FOSSA instance.
```
zlav, line 75: We're probably going to need to refactor a lot of these types out of this file since it's starting to get pretty large. I don't think we need to do that now, but I'm just noticing that this file is now around 800 lines and holds most of these types and their `FromJSON` definitions.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -228,11 +252,10 @@ mkMetadataOpts ProjectMetadata{..} projectName = mconcat $ catMaybes maybes
 
 mangleError :: HttpException -> FossaError
 mangleError err = case err of
-  VanillaHttpException (HTTP.HttpExceptionRequest _ (HTTP.StatusCodeException resp _)) ->
-    case HTTP.responseStatus resp of
-      HTTP.Status 404 _ -> InvalidProjectOrRevision err
```
zlav, line 233: Are we no longer tracking the HTTP status? I feel like the status would be good to at least log if we can.

zlav (approved):
Code looks good, one question about HTTP status tracking.

src/App/Fossa/FossaAPIV1.hs
scruffystuffs: Yeah, this is definitely the biggest target on my personal agenda for a module split.  I don't think it's unreasonable for each API and it's types, instances, and functions to live together in a small module.  We have the same split between entrypoints in the entrypoint refactor

meghfossa: Agreed, but I prefer not to do it in this PR's scope. 

I've added techdebt ticket in to review for now - https://github.com/fossas/team-analysis/issues/917

src/App/Fossa/FossaAPIV1.hs
meghfossa: I don't think we were tracking the status previously (ever?). This includes the debug messages, and debug bundle. Different messages were all proxy for some HTTP status.

From support perspective, error UUID, will allow to isolate the request stack trace in backend with respective http status code - we can do this by copy pasting UUID in datadog search.

For end-users, I can see some value having the http status code in error message. I can update error messaging format to reflect this.

zlav: > Different messages were all proxy for some HTTP status.
I forgot about this behavior, thanks for reminding me. I don't feel strongly about outputting the HTTP status code to the user, with improved API errors I think HTTP errors may just cause confusion.

I do feel strongly that we should include HTTP status codes in the debug messages but I'm not sure this PR is the right place to do that.

PR #794, opened 2022-02-04
Initial Cargo.toml SPDX license parsing.

csasarak:
# Overview

Adds code to parse Cargo.toml and extract license information.

## Acceptance criteria

When we run pathfinder, it should read Cargo.toml files and parse both the`license` and `license-file` fields. Documentation for the license fields can be found [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields). 

## Testing plan

I added unit tests for each case that I programmed for. I also manually tested against several popular crates I downloaded from crates.io

## Risks

There isn't any testing at the level of the pathfinder executable yet, so to verify that the parsing code was actually run required manual testing.

## References

Closes fossas/team-analysis#910

## Checklist


src/Strategy/Cargo.hs
```diff
@@ -156,6 +201,69 @@ instance ToJSON CargoProject
 instance AnalyzeProject CargoProject where
   analyzeProject _ = getDeps
 
+data CargoPackage = CargoPackage
+  { unLicense :: Maybe Text.Text
+  , unLicenseFile :: Maybe Text.Text
```
scruffystuffs, line 206: [nit]: Generally, `un`-prefixed names are reserved for newtypes, and only then when a field name can't really provide extra semantic meaning.  I'd prefer to give these normal field names.

src/Strategy/Cargo.hs
```diff
@@ -156,6 +201,69 @@ instance ToJSON CargoProject
...
+
+-- |Analyze a Cargo.toml for license information. The format is documented
+-- (here)[https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields]
+analyzeLicenses :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m [LicenseResult]
+analyzeLicenses tomlPath = do
+  CargoPackage
+    { unLicenseFile = maybeLicenseFile
+    , unLicense = maybeLicense
+    } <-
+    cargoPackage <$> readContentsToml cargoTomlCodec tomlPath
```
scruffystuffs, line 241: [nit]: This `do` syntax is really confusing to read.  It took me way too long to realize that this is where `maybeLicense` and `maybeLicenseFile` are _defined_, rather than _referenced_, which is what I originally thought.

I think we should use more obvious naming/syntax here (feel free to adjust this however you like it):
```haskell
pkg <- cargoPackage <$> readContentsToml cargoTomlCodec tomlPath
```

src/Strategy/Cargo.hs
```diff
@@ -156,6 +201,69 @@ instance ToJSON CargoProject
...
+
+  -- The license-file field in Cargo.toml is relative to the dir of the
+  -- Cargo.toml file. Generate an absolute path to license-file.
+  let licensePathText = fmap prependTomlDir . parseRelFile . Text.unpack =<< maybeLicenseFile
```
scruffystuffs, line 245: [suggestion]: Notably, we never actually use the `Text` version of this value, so we should consider parsing it into a `Filepath` instead of a `Text`.

It also deeply bothers me that `hlint` isn't telling you to use `toString` instead of `Text.unpack`, even though I know we've configured that rule.

scruffystuffs: Also, we could probably parse it into a `Maybe (Path b File)` as part of the toml decoding.  It should work, since `parseRelFile` requires a monad with `MonadThrow`, like `Maybe`, and that's already on our datatype.

scruffystuffs (changes requested):
Few small comments, but since this isn't a rush PR, I'd like to see these change before merging

src/Strategy/Cargo.hs
csasarak: Please let me know if this is any easier.

src/Strategy/Cargo.hs
csasarak: I have made changes to make this a `Path Rel File`. Re: hlint, I think that the rules we had weren't permissive enough to catch my usage of `unpack` here. It was configured to match a usage like `Data.Text.unpack x` but wouldn't catch something like `Data.Text.unpack . foo`. I've made changes to .hlint.yaml that will hopefully remedy this.

csasarak: One other thing I didn't mention - having a `Maybe (Path Rel File)` for the license path means we could have a `Nothing` value for one of two reasons:

1. There was no `license-file` field.
2. There was a `license-file` field, but it could not be parsed.

Do we want to try to make this type more expressive so we can distinguish between these cases? Or is there some other part of the product that would catch this and is a more logical place to put it?

src/App/Fossa/Container.hs
```diff
@@ -47,7 +47,7 @@ import Path.IO (getCurrentDir)
 newtype ImageText = ImageText {unImageText :: Text} deriving (Show, Eq, Ord)
 
 imageTextArg :: Parser ImageText
-imageTextArg = ImageText . pack <$> argument str (metavar "IMAGE" <> help "The image to scan")
+imageTextArg = ImageText . toText @String <$> argument str (metavar "IMAGE" <> help "The image to scan")
```
scruffystuffs, line 50: I think you should be able to remove the `String`-handling part of this entirely:

```suggestion
imageTextArg = ImageText <$> argument str (metavar "IMAGE" <> help "The image to scan")
```

scruffystuffs: Generally, using string conversion with type application should NEVER need to happen.  Usually, one or the other is enough, and when it's not, there's almost certainly a hidden problem elsewhere causing this.  The original `pack` was unneeded as well.

src/App/Fossa/Container.hs
```diff
@@ -247,7 +247,7 @@ runSyft image = runExecIO . withSyftBinary $ \syftBin -> do
 syftCommand :: BinaryPaths -> ImageText -> Command
 syftCommand bin (ImageText image) =
   Command
-    { cmdName = pack . toFilePath $ toExecutablePath bin
+    { cmdName = toText . toFilePath $ toExecutablePath bin
```
scruffystuffs, line 250: [nit]: Double conversions are usually unnecessary
```suggestion
    { cmdName = toText $ toExecutablePath bin
```

src/Strategy/Cargo.hs
scruffystuffs: Frustratingly, `tomland` doesn't actually support emitting custom errors.  I think it's probably best to use `Filepath` instead here, and just parse it under better error-reporting conditions.

src/Strategy/Cargo.hs
```diff
@@ -156,6 +202,68 @@ instance ToJSON CargoProject
...
+        { licenseFile = toFilePath tomlPath
+        , licensesFound =
+            catMaybes
+              [ License <$> licenseCon <*> maybeLicense
+              , License LicenseFile <$> licensePathText
+              ]
+        }
+    ]
+  where
+    prependTomlDir = toText . toFilePath . (parent tomlPath </>)
```
scruffystuffs, line 258: [nit]: No double conversion:
```suggestion
    prependTomlDir = toText . (parent tomlPath </>)
```

test/Cargo/CargoTomlSpec.hs
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+-- |The license-related fields for a Cargo.toml file are documented
+-- [here](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields)
+licenseSpecs :: Spec
+licenseSpecs = do
+  currentDir <- runIO getCurrentDir
+  describe "Cargo.toml LicenseAnalyzeProject" $ do
+    it' "Reads a Cargo.toml with an SPDX license field" $ do
+      let baseDir = currentDir </> $(mkRelDir "test/Cargo/testdata/spdx_cargo_toml")
```
scruffystuffs, line 61: [optional]: I'd recommend splitting this into a function which takes `$(mkRelDir "filename")` and returns `baseDir`, since you're copying the testdata directory.

src/Strategy/Cargo.hs
```diff
@@ -156,6 +202,68 @@ instance ToJSON CargoProject
...
+    textElem c = isJust . Text.findIndex (== c)
+    -- Old versions of Cargo allow '/' as a separator between SPDX values in
+    -- 'license'. In that case 'license' can't be treated as a LicenseSPDX.
+    selectLicenseCon licenseText =
+      if textElem '/' licenseText
```
scruffystuffs, line 263: Use `Data.Text.elem` instead.
```suggestion
    -- Old versions of Cargo allow '/' as a separator between SPDX values in
    -- 'license'. In that case 'license' can't be treated as a LicenseSPDX.
    selectLicenseCon licenseText =
      if Text.elem '/' licenseText
```

csasarak: `Data.Text.elem` only exists starting in text version 1.2.5 as far as I can tell, and fossa-cli uses `^>= 1.2.3`, I did this to avoid upgrading. I can add that upgrade to this PR if you think it's reasonable.

scruffystuffs: Damn.  We should probably upgrade some of our deps soon, this has hit us a few times recently.  Please leave it for now.

scruffystuffs (commented):
Everything looks good, but I'm going to block the merge of this PR until the entrypoint refactor is merged in, and then the conflicts are handled.  Left a note explaining.

PR #800, opened 2022-02-11
Ruby: Improve error/warnings (1 of Many)

meghfossa:
# Overview

- Adds contextual warn/error for Ruby.
- Prefers lock analysis over bundler show.

## Acceptance criteria

- Shows warning when bundler tactic is selected.

## Screenshot

CleanShot 2022-02-11 at 16 42 58@2x

## Testing plan

When Gemfile.lock does not exist - warning is rendered showing limitations. 

```
make install-dev
echo '' > example/Gemfile
fossa-dev analyze ./example/ -o 
```

## Risks

N/A

## References

Works on: https://github.com/fossas/team-analysis/issues/854

## Checklist

### DIFF
--- a/src/Strategy/Ruby/Errors.hs
+++ b/src/Strategy/Ruby/Errors.hs
@@ -45,4 +45,4 @@ instance ToDiagnostic RubyMissingEdges where
 data RubyMissingDepClassification = RubyMissingDepClassification
 instance ToDiagnostic RubyMissingDepClassification where
   renderDiagnostic (RubyMissingDepClassification) =
-    "Could not classify dependencies between direct or deep. All dependencies will be classified as direct."
+    "Could not differentiate between direct and deep dependencies. All dependencies will be classified as direct."
--- a/src/Strategy/Ruby/Errors.hs
+++ b/src/Strategy/Ruby/Errors.hs
@@ -24,10 +24,13 @@ newtype BundlerMissingLockFile = BundlerMissingLockFile (Path Abs File)
 instance ToDiagnostic BundlerMissingLockFile where
   renderDiagnostic (BundlerMissingLockFile path) =
     vsep
-      [ "We could not find or parse Gemfile.lock corresponding to Gemfile: " <> viaShow path
+      [ "We could not perform Gemfile.lock analysis for Gemfile: " <> viaShow path
       , ""
-      , "Ensure your ruby project is built, and has lockfile prior to running fossa."
-      , "If you are using bundler, you can perform: `bundler install` to generate Gemfile.lock."
+      , indent 2 $
+          vsep
+            [ "Ensure valid Gemfile.lock exists, and is readable by user."
+            , "If you are using bundler, you can perform: `bundler install` to generate Gemfile.lock."
+            ]
       , ""
       , "Refer to:"
       , indent 2 $ pretty $ "- " <> bundlerLockFileRationaleUrl
--- a/src/Strategy/Ruby/Errors.hs
+++ b/src/Strategy/Ruby/Errors.hs
@@ -1,0 +1,45 @@
+module Strategy.Ruby.Errors (
+  BundlerMissingLockFile (..),
+  RubyMissingEdges (..),
+  RubyMissingDepClassification (..),
+
+  -- * Reference docs
+  bundlerLockFileRationaleUrl,
+  rubyFossaDocUrl,
+) where
+
+import App.Docs (strategyLangDocUrl)
+import Data.Text (Text)
+import Diag.Diagnostic (ToDiagnostic, renderDiagnostic)
+import Path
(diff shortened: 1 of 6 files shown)


src/Strategy/Ruby/Errors.hs
```diff
@@ -0,0 +1,45 @@
...
+      , ""
+      , "Refer to:"
+      , indent 2 $ pretty $ "- " <> bundlerLockFileRationaleUrl
+      , indent 2 $ pretty $ "- " <> rubyFossaDocUrl
+      ]
+
+data RubyMissingEdges = RubyMissingEdges
+instance ToDiagnostic RubyMissingEdges where
+  renderDiagnostic (RubyMissingEdges) =
+    "Could not infer edges between dependencies."
```
scruffystuffs, line 40: These errors could use some context.  As a user, I would ask "**Why** can't edges be inferred?"  Same question for the error below: "**Why** can't deps be classified as direct/deep?"

scruffystuffs: It's not clear that the missing `Gemfile.lock` is what causes this.

scruffystuffs:
In the screenshot, the missing `Gemfile.lock` is represented as an error, but we succeeded anyway.  It should probably be rendered as a warning instead.

scruffystuffs:
Rethinking my last comment...

If I (a user) see a warning `Cannot foo`, and then separately I see an error `bar failed: baz is incorrectly formatted`, I might not know that they are connected.  But if the warning says `Cannot foo because bar failed`, then it's more clear that the later error is related to this warning.

src/Strategy/Ruby/Errors.hs
meghfossa: It's not necessarily missing Gemfile.lock, which may cause this. It could be a parser on Gemfile.lock, some ReadFS  error (no permission etc).

src/Strategy/Ruby/Errors.hs
```diff
@@ -0,0 +1,45 @@
...
+
+newtype BundlerMissingLockFile = BundlerMissingLockFile (Path Abs File)
+instance ToDiagnostic BundlerMissingLockFile where
+  renderDiagnostic (BundlerMissingLockFile path) =
+    vsep
+      [ "We could not find or parse Gemfile.lock corresponding to Gemfile: " <> viaShow path
+      , ""
+      , "Ensure your ruby project is built, and has lockfile prior to running fossa."
+      , "If you are using bundler, you can perform: `bundler install` to generate Gemfile.lock."
+      , ""
```
meghfossa, line 31: - [ ] Fossa must have read permission

src/Strategy/Ruby/Errors.hs
meghfossa: Maybe - `Could not infer edges between dependencies because Gemfile.lock analysis failed.` can suffice?

src/Strategy/Ruby/Errors.hs
```diff
@@ -0,0 +1,48 @@
...
+
+data RubyMissingEdges = RubyMissingEdges
+instance ToDiagnostic RubyMissingEdges where
+  renderDiagnostic (RubyMissingEdges) =
+    "Could not infer edges between dependencies."
+
+data RubyMissingDepClassification = RubyMissingDepClassification
+instance ToDiagnostic RubyMissingDepClassification where
+  renderDiagnostic (RubyMissingDepClassification) =
+    "Could not classify dependencies between direct or deep. All dependencies will be classified as direct."
```
spatten, line 48: Maybe change this to something like this?

"Could not differentiate between direct and deep dependencies. All dependencies will be classified as direct"

src/Strategy/Ruby/Errors.hs
spatten: I think your suggestion is better than the current text, but I think it would be worth making this wordier to explain those possible causes.

"Could not infer edges between dependencies because Gemfile.lock analysis failed. Two possible reasons this can happen are that Gemfile.lock does not exist or we do not have read permission for Gemfile.lock."

src/Strategy/Ruby/Errors.hs
spatten: Also, it would be nice to add some info about why we couldn't do the differentiation. Does this happen when we can't read Gemfile.lock?

spatten (approved):
I'm probably not competent to review the code changes, but they look pretty straightforward to me.

I think my suggestions for changes to the error messages make sense, but it's possible that they'll look really ugly in context as they're definitely longer.

src/Strategy/Ruby/Errors.hs
meghfossa: It is anything downstream of Gemfile.lock analysis - file not found, file could not be read, parser failed etc.

PR #803, opened 2022-02-14
Pipenv and Poetry: Improve errors (3 of Many)

meghfossa:
# Overview

Improves warn/errors for pipenv and poetry errors. 

CleanShot 2022-02-14 at 12 49 38@2x

## Acceptance criteria

- When poetry performs suboptimal analysis warnings are shown with context
- When pipenv performs suboptimal analysis warnings are shown with context

## Testing plan

```
make install-dev
fossa-dev analyze ../example/poetry-without-lock -o
```

```
make install-dev
# uninstall pipenv 
fossa-dev analyze ../example/pipenv-proj -o
```

## Risks

N/A

## References

Works on: https://github.com/fossas/team-analysis/issues/854

## Checklist

### DIFF
--- a/src/Diag/Common.hs
+++ b/src/Diag/Common.hs
@@ -1,6 +1,7 @@
 module Diag.Common (
   MissingDeepDeps (..),
   MissingEdges (..),
+  AllDirectDeps (..),
 ) where
 
 import Diag.Diagnostic (ToDiagnostic (renderDiagnostic))
@@ -14,3 +15,8 @@ data MissingEdges = MissingEdges
 instance ToDiagnostic MissingEdges where
   renderDiagnostic (MissingEdges) =
     "Could not analyze edges between dependencies."
+
+data AllDirectDeps = AllDirectDeps
+instance ToDiagnostic AllDirectDeps where
+  renderDiagnostic (AllDirectDeps) =
+    "Could not differentiate between direct and deep dependencies, all dependencies will be reported as direct."
--- a/src/Effect/Exec.hs
+++ b/src/Effect/Exec.hs
@@ -34,8 +34,9 @@ import Data.Aeson
 import Data.Bifunctor (first)
 import Data.ByteString.Lazy qualified as BL
 import Data.String (fromString)
-import Data.String.Conversion (decodeUtf8, toString, toText)
+import Data.String.Conversion (ToText, decodeUtf8, toString, toText)
 import Data.Text (Text)
+import Data.Text qualified as Text
 import Data.Void (Void)
 import GHC.Generics (Generic)
 import Path
@@ -60,6 +61,9 @@ instance ToJSON Command
 instance RecordableValue Command
 instance FromJSON Command
 instance ReplayableValue Command
+
+instance ToText Command where
+  toText (Command name args _) = Text.intercalate " " $ [name] <> args
 
 data CmdFailure = CmdFailure
   { cmdFailureCmd :: Command
--- a/src/Effect/Exec.hs
+++ b/src/Effect/Exec.hs
@@ -16,6 +16,7 @@ module Effect.Exec (
   execJson,
   ExecIOC,
   runExecIO,
+  renderCommand,
   module System.Exit,
   module X,
 ) where
@@ -34,7 +35,7 @@ import Data.Aeson
 import Data.Bifunctor (first)
 import Data.ByteString.Lazy qualified as BL
 import Data.String (fromString)
-import Data.String.Conversion (ToText, decodeUtf8, toString, toText)
+import Data.String.Conversion (decodeUtf8, toString, toText)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Void (Void)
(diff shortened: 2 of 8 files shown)


src/Strategy/Python/Errors.hs
```diff
@@ -0,0 +1,44 @@
...
+      [ pretty $ "We could not perform pipenv graph analysis using, command:" <> toText cmd
+      , ""
+      , indent 2 $
+          vsep
+            [ "Ensure pipenv executable is in your PATH."
+            ]
+      ]
+
+newtype MissingPoetryLockFile = MissingPoetryLockFile (Path Abs File)
+instance ToDiagnostic MissingPoetryLockFile where
```
meghfossa, line 32: I would be doing refactor of this type later, once I have seen all types of error cases into a some ActionableDiagCtx. But for now, I would like to get this in.

src/Diag/Common.hs
```diff
@@ -0,0 +1,22 @@
...
+instance ToDiagnostic MissingDeepDeps where
+  renderDiagnostic (MissingDeepDeps) =
+    "Could not analyze deep dependencies."
+
+data MissingEdges = MissingEdges
+instance ToDiagnostic MissingEdges where
+  renderDiagnostic (MissingEdges) =
+    "Could not analyze edges between dependencies."
+
+data MissingClassificationBetweenDirectAndDeepDeps = MissingClassificationBetweenDirectAndDeepDeps
```
scruffystuffs, line 19: [nit]: Let's avoid massive names like this.  How about `AllDirectDeps`/`MissingRootEdge`?  The message can stay the same, but code-wise, this is pretty awkward.

src/Strategy/Python/Pipenv.hs
```diff
@@ -24,13 +24,15 @@ import Data.Set (Set)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import DepTypes
+import Diag.Common
```
scruffystuffs, line 27: [optional nit]: Can we make the imports for this file explicit while we're here?

src/Strategy/Python/Poetry.hs
```diff
@@ -10,13 +10,14 @@ module Strategy.Python.Poetry (
...
 import Control.Applicative ((<|>))
-import Control.Effect.Diagnostics (Diagnostics, context)
+import Control.Effect.Diagnostics (Diagnostics, context, errCtx, fatalText, recover, warnOnErr)
 import Data.Aeson (ToJSON)
 import Data.Map (Map)
 import Data.Map.Strict qualified as Map
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import DepTypes (DepType (..), Dependency (..))
+import Diag.Common
```
scruffystuffs, line 20: [nit]: Explicit imports please

src/Strategy/Python/Poetry.hs
```diff
@@ -137,6 +141,12 @@ analyze PoetryProject{pyProjectToml, poetryLock} = do
           , dependencyManifestFiles = [poetryLockPath lockPath]
           }
     Nothing -> do
+      _ <-
+        recover
```
scruffystuffs, line 145: [nit]: use `void` instead.  Will require running the formatter.

```suggestion
      void . recover
```

src/Effect/Exec.hs
```diff
@@ -60,6 +61,9 @@ instance RecordableValue Command
 instance FromJSON Command
 instance ReplayableValue Command
 
+instance ToText Command where
+  toText (Command name args _) = Text.intercalate " " $ [name] <> args
+
```
scruffystuffs, line 66: I don't know if I agree with this use of `ToText`.  When used, it will read as `toText cabalCommand`, which doesn't actually tell me how the command is formatted.  I can _kind of_ infer that it's going to be in shell format, but I think this may be better off as something like `renderCommand :: Command -> Text`, instead of using the `ToText` instance.  I think we should only use `ToText` for string-like types.

scruffystuffs: An instance where `ToText` makes more sense is in locators.  Locators are really just strings, but we can access partitions of them via fields like `fetcher` and `revision`.  To do that, we secretly store the string as its component parts, use that to make the field accessors work, and then silently put them back together when we run `ToText`.

scruffystuffs (approved):
I'd like to see the `ToText` block addressed (either via change or counterargument) before merge, and left a few non-blocking nits as well.

src/Effect/Exec.hs
meghfossa: resolved in: https://github.com/fossas/fossa-cli/pull/803/commits/93cdf4ecf7fb004f155f7337c9c850dd29e1189d

src/Strategy/Python/Poetry.hs
meghfossa: resolved in: https://github.com/fossas/fossa-cli/pull/803/commits/93cdf4ecf7fb004f155f7337c9c850dd29e1189d

PR #817, opened 2022-02-22
Run themis on deps

spatten:
# Overview

[Closes ANE-25](https://fossa.atlassian.net/browse/ANE-25) — Use themis to license scan vendored dependencies
[Closes ANE-26](https://fossa.atlassian.net/browse/ANE-26) — Upload cli-side license data to S3 and trigger the build

Run `themis-cli` on the list of vendored dependencies in `fossa-deps.yml`, upload the resulting scan to S3 and then trigger a build on Core.

Things this PR does not do:

- we do not add the license data to the output from the `--output` flag (https://fossa.atlassian.net/browse/ANE-97)
- We do not emit any useful or actionable errors or warnings (https://fossa.atlassian.net/browse/ANE-96)

There are also no tests. I *think* that this code is similar to other code-paths that are not tested, but please let me know if there are areas that are reasonable to test.

## Acceptance criteria

This PR changes the behavior for vendored dependencies found in `fossa-deps.yml`.

* If `archiveOrCLI` in `App.Fossa.ManualDeps` is set to `ArchiveUpload`, then there should be no change: we should run an archive upload on any vendored dependencies
* If `archiveOrCLI` is set to `CLILicenseScan`, then we should run a license scan on any vendored dependencies

Running a license scan means:

* we run themis the  directories pointed to by the vendored dependency
* the results from each of the themis scans are gzipped and uploaded to S3
* Once all of the scans are completed and uploaded, we trigger a `CLILicenseScanBuild` job for each vendored dependency.

- [ ] `archiveOrCLI` must be set to `ArchiveUpload` before this is merged into master

### Core setup

Check out the [CLI license scan endpoint](https://github.com/fossas/FOSSA/pull/7165) PR. The branch name is `cli-license-scan-endpoint`

```
git checkout cli-license-scan-endpoint
```

Do whatever you normally do to get core running on your system. You'll need to have minio, the agent and a faktory worker running, so you might also need to do:

```
docker compose up -d s3 agent faktory-worker
```

You'll want to watch these logs to make sure that the right job is getting queued:

```
docker compose logs --follow faktory-worker
```

### test directory setup

Make a simple yarn project with a single vendored dependency:

```
mkdir ~/fossa/license-scan-dirs/archive-upload-with-target
cd ~/fossa/license-scan-dirs/archive-upload-with-target
mkdir yarn-package
```

Put this in `fossa-deps.yml`:

```yml
vendored-dependencies:
  - name: yarn-package
    path: yarn-package
    version: 1.0.2
```

Put this in `yarn-package/package.json`:

```json
{
  "name": "yarn-testing",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "author": "",
  "license": "MIT",
  "dependencies": {
    "bson": "4.5.1"
  },
}
```

```
cd yarn-package
yarn install
rm -rf node_modules
```

Now do the same for a directory with multiple archives:

```
mkdir ~/fossa/license-scan-dirs/archive-upload-with-multple-targets
cd ~/fossa/license-scan-dirs/archive-upload-with-multiple-targets
mkdir first-yarn-package
mkdir second-yarn-package
```

Put this in `fossa-deps.yml`:

```yml
vendored-dependencies:
  - name: first-yarn-package
    path: first-yarn-package
    version: 1.0.2
  - name: second-yarn-package
    path: second-yarn-package
    version: 2.0.1
```

in first-yarn-package/package.json:

```
{
  "name": "yarn-testing",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "author": "",
  "license": "MIT",
  "dependencies": {
    "bson": "4.5.1"
  },
  "devDependencies": {
    "@storybook/addon-actions": "6.3.7",
    "@storybook/addon-docs": "6.3.7"
  }
}
```

in second-yarn-package/package.json:

```
{
  "name": "yarn-testing-2",
  "version": "1.0.0",
  "description": "",
  "main": "index.js",
  "scripts": {
  },
  "author": "",
  "license": "Apache-2.0",
  "dependencies": {
    "bson": "4.5.1"
  },
  "devDependencies": {
    "@storybook/addon-actions": "6.3.7",
    "@storybook/addon-docs": "6.3.7"
  }
}
```

```
cd first-yarn-package
yarn install
rm -rf node_modules
cd ../second-yarn-package
yarn install
rm -rf node_modules
```

### Run the CLI

Without any changes, run the CLI:

```
export FOSSA_API_KEY='valid api key for your dev setup'
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/license-scan-dirs/archive-upload-with-target
```

This should successfully go through the previously existing archive upload workflow.

Now edit `src/App/Fossa/ManualDeps.hs`. On line 92 change `ArchiveUpload` to `CLILicenseScan`

```haskell
  let archiveOrCLI = CLILicenseScan
```

Edit `fossa-deps.yml` and change the version of the dependency from `1.0.1` to `1.0.2`.

run the scan again:

```
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/license-scan-dirs/archive-upload-with-target
```

This time it should run a license scan on the `yarn-package` directory and upload the results to your local minio.

There should be no difference in the UI. The license scan result from a cli-side license scan should be the same as a license scan done on the servers.

However, what happens on the back end is very different.

The archive upload workflow uploads the contents of the vendored dependency to `archive_components/archive+<org ID>/<locator>`.

So, if your ORG ID is 1, you'll see a file uploaded to `archive_components/archive+1/yarn-package$1.0.1`.

Download that file and gunzip it. Check that you see the contents of the yarn-package directory.

The CLI license scan workflow scans the contents of the vendored dependency and uploads the results of that license scan to `archive_license_data/archive+<org ID>/locator`.

So, if your ORG ID is 1, you'll see a file uploaded to `archive_license_data/archive+1/yarn-package$1.0.2`.

Download that file and gunzip it. Check that you see a JSON file containing the results of a license scan. It should look like this:

```
mv ~/Downloads/yarn-package\$1.0.1 ~/Downloads/yarn-package\$1.0.1.gz; gunzip ~/Downloads/yarn-package\$1.0.1.gz; cat ~/Downloads/yarn-package\$1.0.1
```

```json
{"Name":"yarn-package","Type":"cli-license-scanned","LicenseUnits":[{"Files":["yarn.lock"],"Name":"No_license_found","Type":"LicenseUnit","Dir":"","Info":{"Description":""},"Data":[{"path":"yarn.lock","ThemisVersion":"db4cf6c4b22e29bc4f2a4bd48488b8d3c2a3f5d9","Copyrights":null,"match_data":null,"Copyright":null}]},{"Files":["package.json"],"Name":"mit","Type":"LicenseUnit","Dir":"","Info":{"Description":""},"Data":[{"path":"package.json","ThemisVersion":"db4cf6c4b22e29bc4f2a4bd48488b8d3c2a3f5d9","Copyrights":null,"match_data":[{"index":0,"match_string":"MIT\",","length":4,"location":201}],"Copyright":null}]}]}
```

The contents of the file are just the JSON output.

In both cases, the UI should show an MIT license found for the `yarn-package` dependency.

Do the same thing for the project in `~/fossa/license-scan-dirs/archive-upload-with-multple-targets`:

cd ~/fossa/license-scan-dirs/archive-upload-with-multple-targets

```
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/license-scan-dirs/archive-upload-with-multiple-targets
```

You should see two new files in minio in `archive_components/archive+<org ID>`.

You should also see two `CliLicenseScannedBuild` tasks run in your faktory logs.

You should see an MIT license for `first-package` and an Apache-2.0 license for `second-package`.

## Checklist

- [x] I confirmed tests are not viable
- [ ] I linked this PR to any referenced GitHub issues, if they exist.

### DIFF
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -97,7 +97,7 @@ extractThemisFiles = do
           , binaryFilePath = $(mkRelFile "index.gob")
           }
   sendIO $ extractLzma (toPath compressedThemisIndex) (toPath decompressedThemisIndex)
-  pure $ ThemisBins themisActual (applyTag @ThemisIndex decompressedThemisIndex)
+  pure $ ThemisBins themisActual $ applyTag @ThemisIndex decompressedThemisIndex
 
 withSyftBinary ::
   ( Has (Lift IO) sig m
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -16,7 +16,9 @@ import App.Fossa.Config.DumpBinaries (
   DumpBinsOpts,
   mkSubCommand,
  )
+import Data.String.Conversion (toLazy, toStrict)
 import App.Fossa.Subcommand (SubCommand)
+import Codec.Compression.Lzma qualified as Lzma
 import Control.Effect.Exception (bracket)
 import Control.Effect.Lift (Has, Lift, sendIO)
 import Data.ByteString (ByteString, writeFile)
@@ -24,7 +26,6 @@ import Data.FileEmbed.Extra (embedFileIfExists)
 import Data.Foldable (for_, traverse_)
 import Data.Tagged (Tagged, applyTag, unTag)
 import Data.Time.Clock.POSIX (getPOSIXTime)
-import Discovery.Archive (extractLzma)
 import Path (
   Abs,
   Dir,
@@ -90,14 +91,15 @@ withThemisAndIndex = bracket extractThemisFiles cleanupThemisBins
 -- When running Themis we always need both the themis-cli and the decompressed index.gob
 extractThemisFiles :: Has (Lift IO) sig m => m ThemisBins
 extractThemisFiles = do
-  themisActual <- applyTag @ThemisBinary <$> extractEmbeddedBinary Themis
-  compressedThemisIndex <- extractEmbeddedBinary ThemisIndex
+  themisBin <- extractEmbeddedBinary Themis
+  let themisActual = applyTag @ThemisBinary themisBin
+  -- themisActual <- applyTag @ThemisBinary themisBin
   let decompressedThemisIndex =
         BinaryPaths
-          { binaryPathContainer = binaryPathContainer compressedThemisIndex
+          { binaryPathContainer = binaryPathContainer themisBin
           , binaryFilePath = $(mkRelFile "index.gob")
           }
-  sendIO $ extractLzma (toPath compressedThemisIndex) (toPath decompressedThemisIndex)
+  _ <- pure $ writeFile (show $ toPath decompressedThemisIndex) (toStrict $ Lzma.decompress $ toLazy embeddedBinaryThemisIndex)
   pure $ ThemisBins themisActual $ applyTag @ThemisIndex decompressedThemisIndex
 
 withSyftBinary ::
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -27,6 +27,13 @@ import Control.Effect.Lift (Has, Lift, sendIO)
 import Data.ByteString (ByteString, writeFile)
 import Data.FileEmbed.Extra (embedFileIfExists)
 import Data.Foldable (for_)
+import Effect.Logger (Logger, logDebug, logInfo)
+import Prettyprinter (Pretty (pretty))
+
+import Data.ByteString.Lazy qualified as BL
+import Data.Conduit.Binary (sinkLbs)
+import Data.Conduit.Lzma qualified as Lzma
+import Conduit (runConduit, runResourceT, sourceFileBS, (.|))
 import Path (
   Abs,
   Dir,
@@ -90,10 +97,28 @@ withWigginsBinary = withEmbeddedBinary Wiggins
 
 withThemisBinaryAndIndex ::
   ( Has (Lift IO) sig m
+  , Has Logger sig m
   ) =>
   ([BinaryPaths] -> m c) ->
   m c
-withThemisBinaryAndIndex = bracket (traverse extractEmbeddedBinary [Themis, ThemisIndex]) cleanupMultipleBinaries
+withThemisBinaryAndIndex = bracket (extractThemisBinaryAndIndex [Themis, ThemisIndex]) cleanupMultipleBinaries
+
+extractThemisBinaryAndIndex :: (Has (Lift IO) sig m, Has Logger sig m) => [PackagedBinary] -> m [BinaryPaths]
+extractThemisBinaryAndIndex (themisPackagedBinary : themisIndexPackagedBinary : _) = do
+  themisBinaryPath <- extractEmbeddedBinary themisPackagedBinary
+  indexBinaryPath <- extractEmbeddedBinary themisIndexPackagedBinary
+  decompressedIndexFileContents <- sendIO (runResourceT . runConduit $ sourceFileBS (fromAbsFile $ toExecutablePath indexBinaryPath) .| Lzma.decompress Nothing .| sinkLbs)
+  let decompressedIndexBinaryPaths = BinaryPaths {
+    binaryPathContainer = binaryPathContainer indexBinaryPath
+  , binaryFilePath = $(mkRelFile "index.gob")
+  }
+  let decompressedIndexPath = fromAbsFile $ toExecutablePath decompressedIndexBinaryPaths
+  sendIO $ BL.writeFile decompressedIndexPath decompressedIndexFileContents
+  logDebug $ pretty $ "decompressedIndexFilePath = " ++ show decompressedIndexBinaryPaths
+  pure [themisBinaryPath, decompressedIndexBinaryPaths]
+
+extractThemisBinaryAndIndex [_] = pure []
+extractThemisBinaryAndIndex [] = pure []
 
 withEmbeddedBinary ::
   ( Has (Lift IO) sig m
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -7,7 +7,7 @@ module App.Fossa.EmbeddedBinary (
   withEmbeddedBinary,
   dumpEmbeddedBinary,
   toExecutablePath,
-  BinaryPaths,
+  BinaryPaths (..),
   withWigginsBinary,
   withSyftBinary,
   withThemisBinaryAndIndex,
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -93,13 +93,16 @@ extractThemisFiles :: Has (Lift IO) sig m => m ThemisBins
 extractThemisFiles = do
   themisBin <- extractEmbeddedBinary Themis
   let themisActual = applyTag @ThemisBinary themisBin
-  -- themisActual <- applyTag @ThemisBinary themisBin
+  -- decompress index.gob.gz binary and write it to disk
+  let bin = toStrict $ Lzma.decompress $ toLazy embeddedBinaryThemisIndex
+  container <- sendIO extractDir
   let decompressedThemisIndex =
         BinaryPaths
-          { binaryPathContainer = binaryPathContainer themisBin
+          { binaryPathContainer = container
           , binaryFilePath = $(mkRelFile "index.gob")
           }
-  _ <- pure $ writeFile (show $ toPath decompressedThemisIndex) (toStrict $ Lzma.decompress $ toLazy embeddedBinaryThemisIndex)
+  _ <- sendIO $ ensureDir $ parent $ toPath decompressedThemisIndex
+  _ <- sendIO $ writeFile (fromAbsFile $ toPath decompressedThemisIndex) bin
   pure $ ThemisBins themisActual $ applyTag @ThemisIndex decompressedThemisIndex
 
 withSyftBinary ::
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -16,7 +16,6 @@ import App.Fossa.Config.DumpBinaries (
   DumpBinsOpts,
   mkSubCommand,
  )
-import Data.String.Conversion (toLazy, toStrict)
 import App.Fossa.Subcommand (SubCommand)
 import Codec.Compression.Lzma qualified as Lzma
 import Control.Effect.Exception (bracket)
@@ -24,6 +23,7 @@ import Control.Effect.Lift (Has, Lift, sendIO)
 import Data.ByteString (ByteString, writeFile)
 import Data.FileEmbed.Extra (embedFileIfExists)
 import Data.Foldable (for_, traverse_)
+import Data.String.Conversion (toLazy, toStrict)
 import Data.Tagged (Tagged, applyTag, unTag)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Path (
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -90,14 +90,13 @@ withThemisAndIndex = bracket extractThemisFiles cleanupThemisBins
 extractThemisFiles :: Has (Lift IO) sig m => m ThemisBins
 extractThemisFiles = do
   themisActual <- applyTag @ThemisBinary <$> extractEmbeddedBinary Themis
-  compressedThemisIndex <- applyTag @ThemisIndex <$> extractEmbeddedBinary ThemisIndex
+  compressedThemisIndex <- extractEmbeddedBinary ThemisIndex
   let decompressedThemisIndex =
         BinaryPaths
-          { binaryPathContainer = binaryPathContainer $ unTag compressedThemisIndex
+          { binaryPathContainer = binaryPathContainer compressedThemisIndex
           , binaryFilePath = $(mkRelFile "index.gob")
           }
-  sendIO $ extractLzma (toPath $ unTag compressedThemisIndex) (toPath decompressedThemisIndex)
-
+  sendIO $ extractLzma (toPath compressedThemisIndex) (toPath decompressedThemisIndex)
   pure $ ThemisBins themisActual (applyTag @ThemisIndex decompressedThemisIndex)
 
 withSyftBinary ::
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -93,7 +93,7 @@ withThemisBinaryAndIndex ::
   ) =>
   ([BinaryPaths] -> m c) ->
   m c
-withThemisBinaryAndIndex = bracket (traverse extractEmbeddedBinary [Themis, ThemisIndex]) (traverse cleanupExtractedBinaries)
+withThemisBinaryAndIndex = bracket (traverse extractEmbeddedBinary [Themis, ThemisIndex]) cleanupMultipleBinaries
 
 withEmbeddedBinary ::
   ( Has (Lift IO) sig m
@@ -105,6 +105,12 @@ withEmbeddedBinary bin = bracket (extractEmbeddedBinary bin) cleanupExtractedBin
 
 cleanupExtractedBinaries :: (Has (Lift IO) sig m) => BinaryPaths -> m ()
 cleanupExtractedBinaries (BinaryPaths binPath _) = sendIO $ removeDirRecur binPath
+
+-- Horrible hack to get withThemisBinaryAndIndex working while Wes figures out a better way
+-- to run multiple binaries
+cleanupMultipleBinaries :: (Has (Lift IO) sig m) => [BinaryPaths] -> m ()
+cleanupMultipleBinaries ((BinaryPaths binPath _) : _) = sendIO $ removeDirRecur binPath
+cleanupMultipleBinaries [] = pure ()
 
 extractEmbeddedBinary :: (Has (Lift IO) sig m) => PackagedBinary -> m BinaryPaths
 extractEmbeddedBinary bin = do
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -100,10 +100,11 @@ extractThemisBinaryAndIndex :: (Has (Lift IO) sig m) => m [BinaryPaths]
 extractThemisBinaryAndIndex = do
   themisBinaryPaths <- extractEmbeddedBinary Themis
   compressedIndexBinaryPaths <- extractEmbeddedBinary ThemisIndex
-  let decompressedIndexBinaryPaths = BinaryPaths {
-    binaryPathContainer = binaryPathContainer compressedIndexBinaryPaths
-  , binaryFilePath = $(mkRelFile "index.gob")
-  }
+  let decompressedIndexBinaryPaths =
+        BinaryPaths
+          { binaryPathContainer = binaryPathContainer compressedIndexBinaryPaths
+          , binaryFilePath = $(mkRelFile "index.gob")
+          }
   sendIO $ extractLzma (toExecutablePath compressedIndexBinaryPaths) (toExecutablePath decompressedIndexBinaryPaths)
   pure [themisBinaryPaths, decompressedIndexBinaryPaths]
 
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -101,12 +101,12 @@ withThemisBinaryAndIndex ::
   ) =>
   ([BinaryPaths] -> m c) ->
   m c
-withThemisBinaryAndIndex = bracket (extractThemisBinaryAndIndex [Themis, ThemisIndex]) cleanupMultipleBinaries
+withThemisBinaryAndIndex = bracket extractThemisBinaryAndIndex cleanupMultipleBinaries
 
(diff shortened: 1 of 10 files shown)


src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+  Fossa.Organization orgId _ <- Fossa.getOrganization apiOpts
+
+  let updateArcName :: Text -> Archive -> Archive
+      updateArcName updateText arc = arc{archiveName = updateText <> "/" <> archiveName arc}
+      archivesWithOrganization = updateArcName (toText $ show orgId) <$> archives
+
+  pure $ arcToLocator <$> archivesWithOrganization
+
+-- licenseNoScanSourceUnit exists for when users run `fossa analyze -o` and do not upload their source units.
+licenseNoScanSourceUnit :: [VendoredDependency] -> [Locator]
```
meghfossa, line 118: You should use `-- | some header text` for haddoc method docs, this way vscode/ide/language can show docs on hover. 

Reference: https://haskell-haddock.readthedocs.io/en/latest/markup.html#documenting-a-top-level-declaration

src/App/Fossa/ManualDeps.hs
```diff
@@ -83,9 +86,17 @@ toSourceUnit :: (Has (Lift IO) sig m, Has Diagnostics sig m, Has StickyLogger si
...
   -- If the file exists and we have no dependencies to report, that's a failure.
   when (hasNoDeps manualDeps) $ fatalText "No dependencies found in fossa-deps file"
+  -- TODO: archiveOrCLI will eventually be set by querying core to see if we should archive upload or cli-side license scan
+  -- See https://fossa.atlassian.net/browse/ANE-23
+  -- IMPORTANT: Until ANE-23 is implemented, this should always be `ArchiveUpload` on master
+  let archiveOrCLI = ArchiveUpload
   archiveLocators <- case maybeApiOpts of
-    Nothing -> pure $ archiveNoUploadSourceUnit vendoredDependencies
-    Just apiOpts -> archiveUploadSourceUnit root apiOpts vendoredDependencies
+    Nothing -> case archiveOrCLI of
```
meghfossa, line 94: [Optional] You can pattern match on multiple cases, making it easier to parse 4 branches

```haskell
  -- TODO: archiveOrCLI will eventually be set by querying core to see if we should archive upload or cli-side license scan
  -- See https://fossa.atlassian.net/browse/ANE-23
  -- IMPORTANT: Until ANE-23 is implemented, this should always be `ArchiveUpload` on master
  let defaultArchiveOrCLI = ArchiveUpload

  archiveLocators :: [Locator] <- case (maybeApiOpts, defaultArchiveOrCLI) of
    (Just apiOpts, ArchiveUpload) -> archiveUploadSourceUnit root apiOpts vendoredDependencies
    (Just apiOpts, CLILicenseScan) -> licenseScanSourceUnit root apiOpts vendoredDependencies
    (Nothing, ArchiveUpload) -> pure $ archiveNoUploadSourceUnit vendoredDependencies
    (Nothing, CLILicenseScan) -> pure $ licenseNoScanSourceUnit vendoredDependencies

```

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -430,6 +476,26 @@ archiveUpload signedArcURI arcFile = fossaReq $ do
...
+  LicenseSourceUnit ->
+  m LbsResponse
+licenseScanResultUpload signedArcURI licenseScanResult = fossaReq $ do
+  let arcURL = URI.mkURI $ signedURL signedArcURI
+
+  uri <- fromMaybeText ("Invalid URL: " <> signedURL signedArcURI) arcURL
+  validatedURI <- fromMaybeText ("Invalid URI: " <> toText (show uri)) (useURI uri)
+
+  context ("Uploading license scan result to " <> signedURL signedArcURI) $ case validatedURI of
+    Left (url, options) -> uploadArchiveRequest url options
```
meghfossa, line 493: [Prefer] I know in the codebase this is previously used but lets name Left/Right with scheme for readability.
```haskell
    Left (httpUrl, httpOption) -> uploadArchiveRequest httpUrl httpOption
    Right (httpsUrl, httpsOptions) -> uploadArchiveRequest httpsUrl httpsOptions
```

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+
+import App.Fossa.RunThemis (
+  execThemis,
+ )
+
+import Control.Carrier.Diagnostics qualified as Diag
+import Control.Carrier.StickyLogger (StickyLogger, logSticky)
+import Control.Effect.Lift
+
+import App.Fossa.ArchiveUploader
```
meghfossa, line 16: Avoid implicit imports.

```haskell
import Control.Effect.Diagnostics -- bad
import Control.Effect.Diagnostics (fatal, context)  -- Good
```

src/App/Fossa/ManualDeps.hs
```diff
@@ -25,6 +25,7 @@ import Data.Aeson (
  )
 
 import App.Fossa.ArchiveUploader
+import App.Fossa.LicenseScanner
```
meghfossa, line 28: Avoid implicit imports.

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+  runExecIO $ runThemis themisBins scanDir
+
+runThemis :: (Has Exec sig m, Has Diagnostics sig m) => ThemisBins -> Path Abs Dir -> m [LicenseUnit]
+runThemis themisBins scanDir = do
+  context "Running license scan binary" $ execThemis themisBins scanDir
+
+md5 :: B.ByteString -> Digest MD5
+md5 = hash
+
+scanAndUploadVendoredDep :: (Has Diag.Diagnostics sig m, Has (Lift IO) sig m, Has StickyLogger sig m) => ApiOpts -> Path Abs Dir -> VendoredDependency -> m Archive
```
meghfossa, line 62: [Optional] [Prefer] I would collapse arguments with linebreak like you have for `themisRunner` for consistency.

meghfossa (commented):
I still have to go through uploading logic, and testing. Leaving some optional nits, and other comments for now.

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -86,13 +87,18 @@ cleanupThemisBins (ThemisBins a b) = traverse_ cleanupExtractedBinaries [unTag a
...
-
-  pure $ ThemisBins themisActual themisIndex
+  compressedThemisIndex <- extractEmbeddedBinary ThemisIndex
+  let decompressedThemisIndex =
+        BinaryPaths
+          { binaryPathContainer = binaryPathContainer compressedThemisIndex
+          , binaryFilePath = $(mkRelFile "index.gob")
+          }
+  sendIO $ extractLzma (toPath compressedThemisIndex) (toPath decompressedThemisIndex)
+  pure $ ThemisBins themisActual $ applyTag @ThemisIndex decompressedThemisIndex
```
scruffystuffs, line 101: We have the entire binary file in memory, it seems a little weird to dump it out to a file and then unzip it on-disk.  I think it's best if we create some `extractXZippedBinary` function which does similar work to `extractEmbeddedBinary`, but runs in-memory `lzma` unzipping on the embedded bytestring and dumps that directly to the file.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -390,6 +394,27 @@ archiveBuildUpload apiOpts archiveProjects = runEmpty $
...
+---------- license-scan build queueing. This Endpoint ensures that after a license-scan is uploaded, it is scanned.
+
+licenseScanFinalizeUrl :: Url 'Https -> Url 'Https
+licenseScanFinalizeUrl baseUrl = baseUrl /: "api" /: "license_scan" /: "finalize"
+
+licenseScanFinalize ::
+  (Has (Lift IO) sig m, Has Diagnostics sig m) =>
+  ApiOpts ->
+  ArchiveComponents ->
+  m (Maybe C.ByteString)
```
scruffystuffs, line 406: Why is the return type a bytestring?  Generally, we maintain that the API function should handle all data transformation and parsing from the server before returning that value to the caller.

scruffystuffs: I see that the old archive upload function did that, let's leave a TODO to fix this later.

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+  -- These aren't valid and can't be automatically handled, so fail the scan with them.
+  let duplicates = duplicateNames uniqDeps
+  unless (null duplicates) $ Diag.fatalText $ duplicateFailureBundle duplicates
+
+  -- At this point, we have a good list of deps, so go for it.
+  archives <- traverse (scanAndUploadVendoredDep apiOpts baseDir) uniqDeps
+
+  -- archiveBuildUpload takes archives without Organization information. This orgID is appended when creating the build on the backend.
+  -- We don't care about the response here because if the build has already been queued, we get a 401 response.
+  _ <- Fossa.licenseScanFinalize apiOpts $ ArchiveComponents archives
```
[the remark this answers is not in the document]
scruffystuffs: The build is queued by this call, so either

1. The function returns non-fatal because the build was successfully queued by this call.
2. The function returns non-fatal because the build was already queued, and we ignored the 401 error (legacy core issue) that signals that.
3. The function returns fatal because something went wrong trying to queue the build.

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+
+  -- The organizationID is needed to prefix each locator name. The FOSSA API automatically prefixes the locator when queuing the build
+  -- but not when reading from a source unit.
+  Fossa.Organization orgId _ <- Fossa.getOrganization apiOpts
+
+  let updateArcName :: Text -> Archive -> Archive
```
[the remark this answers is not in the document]
scruffystuffs: Actually, we shouldn't combine `where` clauses with `do` notation, since it takes away the ordered definition of expressions.

src/App/Fossa/RunThemis.hs
```diff
@@ -0,0 +1,46 @@
...
+import Effect.Exec (
+  AllowErr (Never),
+  Command (..),
+  Exec,
+  execJson,
+ )
+
+execThemis :: (Has Exec sig m, Has Diagnostics sig m) => ThemisBins -> Path Abs Dir -> m [LicenseUnit]
+execThemis themisBins scanDir = do
+  execJson @[LicenseUnit] scanDir $ themisCommand themisBins
```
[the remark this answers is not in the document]
scruffystuffs: We are planning to do this, but we may need to make some changes to actually HAVE the index version.  This is definitely required for GA, but we might push this particular requirement until after we finished pre-reqs for the beta.

src/App/Fossa/EmbeddedBinary.hs
scruffystuffs: Looks like `Lzma.decompress` is what we're looking for.

scruffystuffs (commented):
One required change, specifically the `extractXZippedBinary` comment.

src/App/Fossa/FossaAPIV1.hs
scruffystuffs: Yeah, but let's make it a todo to keep this ticket smaller.

src/Discovery/Archive.hs
```diff
@@ -142,3 +144,11 @@ removeTarLinks (Tar.Fail e) = Tar.Fail e
 extractZip :: Has (Lift IO) sig m => Path Abs Dir -> Path Abs File -> m ()
 extractZip dir zipFile =
   sendIO $ Zip.withArchive (fromAbsFile zipFile) (Zip.unpackInto (fromAbsDir dir))
+
+----------- LZMA and XZ files
+-- .lzma and .xz files can only contain a single file, not a directory
+extractLzma :: Has (Lift IO) sig m => Path Abs File -> Path Abs File -> m ()
```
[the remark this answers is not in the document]
scruffystuffs: we should probably be using [`decompressIO`](https://hackage.haskell.org/package/lzma-0.0.0.3/docs/Codec-Compression-Lzma.html#v:decompressIO).  We won't be able to catch any errors in the current form, which is a big problem.

Also, the API for working with `decompressIO` is really unintuitive, so we might need to pair to get something clean out of this.

src/Srclib/Types.hs
```diff
@@ -20,6 +24,121 @@ import Data.Text qualified as Text
 import Path (File, SomeBase)
 import Types (GraphBreadth (..))
 
+data LicenseSourceUnit = LicenseSourceUnit
+  { licenseSourceUnitName :: Text
+  , licenseSourceUnitType :: Text
+  , licenseSourceUnitLicenseUnits :: [LicenseUnit]
```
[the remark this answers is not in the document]
spatten: You can get an empty `licenseSourceUnitLicenseUnits` if you run themis on an empty directory, so I left a comment explaining that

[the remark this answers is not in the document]
meghfossa: Sorry, I should have been more clear - you can do something like:

```haskell 
licenseSourceUnitLicenseUnits :: Maybe (NonEmpty LicenseUnit)
```

This would be less error-prone, as it won't let you have empty [], and subsequent usage of this type by later code will need to explicitly deal with it being Nothing.

src/Discovery/Archive.hs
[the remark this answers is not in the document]
spatten: It turns out that we're not using this function after changing `extractThemisFiles` to decompress in memory, so I just removed it altogether

src/App/Fossa/RunThemis.hs
```diff
@@ -0,0 +1,47 @@
...
+
+themisCommand :: ThemisBins -> Command
+themisCommand ThemisBins{..} = do
+  Command
+    { cmdName = toText $ fromAbsFile $ toPath $ unTag themisBinaryPaths
+    , cmdArgs = generateThemisArgs $ unTag indexBinaryPaths
+    , cmdAllowErr = Never
+    }
+
+generateThemisArgs :: BinaryPaths -> [Text]
```
scruffystuffs, line 41: I would prefer to see `Tagged ThemisIndex BinaryPaths -> [Text]` here, so that we know we didn't use the wrong `BinaryPaths` object.  This is where the type-safety of `Tagged` is actually useful.

src/Srclib/Types.hs
```diff
@@ -8,18 +8,153 @@ module Srclib.Types (
...
 import Data.Text qualified as Text
 import Path (File, SomeBase)
 import Types (GraphBreadth (..))
 
+data LicenseScanType = CliLicenseScanned
+  deriving (Eq, Ord, Show)
+
+instance ToText LicenseScanType where
+  toText :: LicenseScanType -> Text
+  toText _ = "cli-license-scanned"
```
scruffystuffs, line 34: [nit]: Simplify and future-proof:

We don't need the type sig, since that's the only type that this function could possibly have, dictated by the type class.

```suggestion
instance ToText LicenseScanType where
  toText CliLicenseScanned = "cli-license-scanned"
```

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -86,13 +88,22 @@ cleanupThemisBins (ThemisBins a b) = traverse_ cleanupExtractedBinaries [unTag a
...
 extractThemisFiles :: Has (Lift IO) sig m => m ThemisBins
 extractThemisFiles = do
-  themisActual <- applyTag @ThemisBinary <$> extractEmbeddedBinary Themis
-  -- TODO: The gob is still xzipped, unzip it here
-  themisIndex <- applyTag @ThemisIndex <$> extractEmbeddedBinary ThemisIndex
-
-  pure $ ThemisBins themisActual themisIndex
+  themisBin <- extractEmbeddedBinary Themis
+  let themisActual = applyTag @ThemisBinary themisBin
+  -- decompress index.gob.gz binary and write it to disk
```
scruffystuffs, line 96: [nit]: wrong extension in the comment
```suggestion
  -- decompress index.gob.xz binary and write it to disk
```

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -86,13 +88,22 @@ cleanupThemisBins (ThemisBins a b) = traverse_ cleanupExtractedBinaries [unTag a
...
+  let bin = toStrict $ Lzma.decompress $ toLazy embeddedBinaryThemisIndex
+  container <- sendIO extractDir
+  let decompressedThemisIndex =
+        BinaryPaths
+          { binaryPathContainer = container
+          , binaryFilePath = $(mkRelFile "index.gob")
+          }
+  _ <- sendIO $ ensureDir $ parent $ toPath decompressedThemisIndex
+  _ <- sendIO $ writeFile (fromAbsFile $ toPath decompressedThemisIndex) bin
```
scruffystuffs, line 105: The `from(Abs|Rel)(Dir|File)` functions are only really useful when the `b`/`t` of a `Path b t` aren't known at compile time.  That's not the case here, since `toPath binpath` always returns a `Path Abs File`.

```suggestion
  _ <- sendIO $ writeFile (toString $ toPath decompressedThemisIndex) bin
```

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -86,13 +88,22 @@ cleanupThemisBins (ThemisBins a b) = traverse_ cleanupExtractedBinaries [unTag a
...
 extractThemisFiles = do
-  themisActual <- applyTag @ThemisBinary <$> extractEmbeddedBinary Themis
-  -- TODO: The gob is still xzipped, unzip it here
-  themisIndex <- applyTag @ThemisIndex <$> extractEmbeddedBinary ThemisIndex
-
-  pure $ ThemisBins themisActual themisIndex
+  themisBin <- extractEmbeddedBinary Themis
+  let themisActual = applyTag @ThemisBinary themisBin
+  -- decompress index.gob.gz binary and write it to disk
+  let bin = toStrict $ Lzma.decompress $ toLazy embeddedBinaryThemisIndex
```
scruffystuffs, line 97: Instead of forcing this back to strict right away, we could use the lazy version of writeFile.  Generally, using lazy IO with lazy bytestrings is dangerous, but the scope dies right at the end of the function, since we run the IO completely with `sendIO`, so it's likely completely safe here.

If this works, it will probably be better in terms of memory usage, especially since the file is so large once unzipped.  In it's current form, it balloons up to the full size of the file, which is about 100MB, and then drains all at once.  With lazy bytestring, LZMA will chunk the bytestrings however it wants, on-demand (demand coming from the file writer).

scruffystuffs (commented):
Drive-by, I should probably go though this more thoroughly if you want my full review on it.

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+  themisScanResult <- runLicenseScanOnDir cliScanDir
+  let licenseSourceUnit =
+        LicenseSourceUnit
+          { licenseSourceUnitName = vendoredPath
+          , licenseSourceUnitType = "cli-license-scanned"
+          , licenseSourceUnitLicenseUnits = themisScanResult
+          }
+
+  depVersion <- case vendoredVersion of
+    Nothing -> pure $ toText $ show $ md5 $ encodeUtf8 $ show themisScanResult
```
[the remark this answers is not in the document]
spatten: It is consistent on every run, and that is done deliberately in Themis.

Wes, I'm not going to get to this before I go, but I'm fine with sorting it or putting a TODO to sort it later

src/App/Fossa/LicenseScanner.hs
```diff
@@ -0,0 +1,119 @@
...
+  signedURL <- Fossa.getSignedLicenseScanURL apiOpts depVersion vendoredName
+
+  logSticky $ "Uploading '" <> vendoredName <> "' to secure S3 bucket"
+  _ <- Fossa.licenseScanResultUpload signedURL licenseSourceUnit
+
+  pure $ Archive vendoredName depVersion
+
+-- archiveUploadSourceUnit receives a list of vendored dependencies, a root path, and API settings.
+-- Using this information, it license scans each vendored dependency, uploads the license scan results and then queues a build for the dependency.
+licenseScanSourceUnit :: (Has Diag.Diagnostics sig m, Has (Lift IO) sig m, Has StickyLogger sig m) => Path Abs Dir -> ApiOpts -> [VendoredDependency] -> m [Locator]
```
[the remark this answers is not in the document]
spatten: I think we should do this, but in a separate PR where we avoid hitting either the archive upload or the cli license scan endpoints if `vendoredDeps` is empty.

I think this would involve changing the definition of `ManualDependencies` to something like this:

```haskell
data ManualDependencies = ManualDependencies
  { referencedDependencies :: Maybe (NotEmpty ReferencedDependency)
  , customDependencies :: Maybe (NotEmpty CustomDependency)
  , vendoredDependencies :: Maybe (NotEmpty VendoredDependency)
  , remoteDependencies :: Maybe (NotEmpty RemoteDependency)
  }
```

src/App/Fossa/RunThemis.hs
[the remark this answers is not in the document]
spatten: I've added a TODO that we should log these

src/Srclib/Types.hs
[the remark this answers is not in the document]
spatten: Megh and I talked about this.

You can end up with a non-empty `[LicenseUnit]` if you scan an empty directory. I.e. if you point the vendored dep path at a directory that exists but doesn't have anything in it.

This actually works, in that it uploads an empty `[]` for the license scan result, and you end up with an unlicensed dependency issue in the UI.

This is technically correct, but is a really cryptic way of flagging "you tried to scan an empty directory".

So I've changed it like Megh suggested, and left a TODO to output a warning if we get `[]` back from Themis when scanning a directory

scruffystuffs:
Superseded by #847, couldn't rebase on latest master due to merge commits in the history.

PR #821, opened 2022-02-25
Don't ignore 'server' key in config file

scruffystuffs:
# Overview

During the entrypoint refactor, we didn't include new logic that picks up the `server` from the config file.

There was also duplicated code in that area, which was a remnant of an older version of that change.

## Acceptance criteria

- `server` is not ignored, unless `--endpoint` is used.

## Testing plan

Intercept the config object by setting `FOSSA_CONFIG_DEBUG=1`, confirm that the `server` option is picked up.

## Risks

None, I audited the other config file fields to make sure this case wasn't happening for any other fields.

## Checklist

- [ ] I added tests for this PR's change (or confirmed tests are not viable).
- [ ] I linked this PR to any referenced GitHub issues, if they exist.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Fossa CLI Changelog
 
+## v3.1.2
+
+- Fixes a bug which ignored the `server` field in the config file. ([#821](https://github.com/fossas/fossa-cli/pull/821))
+
 ## v3.1.1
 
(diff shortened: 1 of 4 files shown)


spatten (approved):
Yes, perfect. This fixes the problem and the code changes make sense to me

 :shipit:

PR #823, opened 2022-03-01
Refactor Embedded Binary module

scruffystuffs:
Provides a no-footgun API for using themis and its index from the embedded binary data.

### DIFF
--- a/src/App/Fossa/EmbeddedBinary.hs
+++ b/src/App/Fossa/EmbeddedBinary.hs
@@ -21,8 +21,7 @@ import Control.Effect.Exception (bracket)
 import Control.Effect.Lift (Has, Lift, sendIO)
 import Data.ByteString (ByteString, writeFile)
 import Data.FileEmbed.Extra (embedFileIfExists)
-import Data.Foldable (for_)
-import Data.Maybe (fromMaybe)
+import Data.Foldable (for_, traverse_)
 import Data.Tagged (Tagged, applyTag, unTag)
 import Data.Time.Clock.POSIX (getPOSIXTime)
 import Path (
@@ -35,8 +34,7 @@ import Path (
   mkRelDir,
   mkRelFile,
   parent,
-  parseRelDir,
-  (</>),
+  (</>), parseRelDir
  )
 import Path.IO (
   Permissions (executable),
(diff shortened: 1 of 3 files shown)


src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -134,10 +156,21 @@ extractedPath bin = case bin of
   Themis -> $(mkRelFile "themis-cli")
   ThemisIndex -> $(mkRelFile "index.gob.xz")
 
+-- | Extract to @$TMP/fossa-vendor/<timestamp>
+-- We used to extract everything to @$TMP/fossa-vendor@, but there's a subtle issue with that.
+-- Cleanup is just removing the directory where the file resides, which is fine unless there's
+-- more than one active extracted file.  Cleanup could potentially kill both while one is in use.
+-- Extracting to a another subdir meane that the cleanup only cleans the timestamp subdir.
```
spatten, line 163:
```suggestion
-- Extracting to another subdir means that the cleanup only cleans the timestamp subdir.
```

src/App/Fossa/EmbeddedBinary.hs
```diff
@@ -134,10 +156,21 @@ extractedPath bin = case bin of
...
+-- The only downside is that we never cleanup the fossa-vendor directory, which is not an issue,
+-- since it should be empty by the time we finish cleanup.  The tempfile cleaner on the system
+-- should pick it up anyway.
 extractDir :: Has (Lift IO) sig m => m (Path Abs Dir)
 extractDir = do
   wd <- sendIO getTempDir
-  pure (wd </> $(mkRelDir "fossa-vendor"))
+  -- Get some positive "random" number, in this case a timestamp
+  ts <- show @Int . abs . floor <$> sendIO getPOSIXTime
```
spatten, line 171: Do need to worry about the two dirs having the same timestamp? `getPOSIXTime` has a resolution of microseconds, but then applying floor to it, so I think we only have a resolution of seconds.

The easy fix is to multiply it by 1,000,000 before `floor`

spatten (approved):

src/App/Fossa/EmbeddedBinary.hs
scruffystuffs: Nice catch.

PR #824, opened 2022-03-01
Ruby declared license scan

csasarak:
# Overview

This PR adds a parser that looks for assignment statements in ruby gemspec files and parses out the licenses. 

## Acceptance criteria

The goal is to provide a way for fossa-cli to analyze bundler projects and determine what license it has. Gemspec files are ruby code, but usually licenses are written in as static text. The reason to prefer parsing ourselves was to avoid needing to run the ruby code in the gemspec file in order to determine the declared license. 

## Testing plan

Unit tests test both the parser layer as well as the license discovery layer.

I wrote a script to pull down some of the most popular ruby gems and got about 60. I tested pathfinder against all 60 and determined that the detected licenses were correct in each of them, aside from 3 which weren't bundler projects at all.

## Risks

Please pay special attention to the tests and let me know if there are common scenarios I may have missed when writing this.

## References
Closes fossas/team-analysis#565

## Checklist


src/Strategy/Ruby/Gemspec.hs
```diff
@@ -0,0 +1,108 @@
...
+  toText <$> (stringText <* optional (string ".freeze"))
+  where
+    betweenDelim :: Char -> Parsec Void Text String
+    betweenDelim c =
+      let (d1, d2) = selectDelim c
+       in between (char d1) (char d2) (many (anySingleBut d2))
+    pctQ = optional (choice [char 'q', char 'Q'])
+    arbitraryDelim = try $
+      do (char '%' *> pctQ *> lookAhead anySingle) >>= betweenDelim
+    stringText =
```
meghfossa, line 46: [nit] would like to see type signature in where block.

src/Strategy/Ruby/Gemspec.hs
```diff
@@ -0,0 +1,108 @@
...
+-- |Given a single start delimiter, return start/end delimiters
+selectDelim :: Char -> (Char, Char)
+selectDelim = \case
+  '{' -> ('{', '}')
+  '<' -> ('<', '>')
+  '(' -> ('(', ')')
+  '[' -> ('[', ']')
+  c -> (c, c)
+
+rubyString :: Parser Text
```
meghfossa, line 32: [nit] Would like to see comment with example text it ought to be parsing (few cases). Here is good [example](https://github.com/fossas/fossa-cli/blob/master/src/App/Fossa/VSI/DynLinked/Internal/Binary.hs#L108). 

I'm guilty of not doing this too, but given this parser seems bit involved, example case and edge case illustrated would help in reviewing. 

This applies to other applicable parser as well in this PR.

src/Strategy/Ruby/Gemspec.hs
```diff
@@ -0,0 +1,108 @@
...
+rubyString =
+  -- '.freeze' is a ruby idiom that turns a string into an immutable version of
+  -- itself. In this case it's safe to just ignore it and take the string it's
+  -- attached to.
+  toText <$> (stringText <* optional (string ".freeze"))
+  where
+    betweenDelim :: Char -> Parsec Void Text String
+    betweenDelim c =
```
spatten, line 40: I'm not familiar with how this parsing library works, but does this deal with escaped delimiters within the string?

E.g. `"This is a \"string\" with quotes in it"` or `'we don\'t need to worry about escapes'`

There are similar worries for arbitrary delimiters: `%Q{this string \{keeps on going\} until here}`

This isn't too likely for the data-set we're looking at here (hopefully SPDX expressions), but if it's easy to deal with it'd be nice.

spatten: Also, for strings delimited by `"` or strings with arbitrary delimiters that start with `%Q`, you can have string interpolation like this:

```
"hello, #{name}"
%Q{hello, #{name}}
```

Should we just fail if we see string interpolation?

src/Strategy/Ruby/Gemspec.hs
```diff
@@ -0,0 +1,108 @@
...
+  '(' -> ('(', ')')
+  '[' -> ('[', ']')
+  c -> (c, c)
+
+rubyString :: Parser Text
+rubyString =
+  -- '.freeze' is a ruby idiom that turns a string into an immutable version of
+  -- itself. In this case it's safe to just ignore it and take the string it's
+  -- attached to.
+  toText <$> (stringText <* optional (string ".freeze"))
```
spatten, line 37: It's idiomatic to say `.freeze`, but `.freeze()` is also valid Ruby code. Should we deal with that too?

spatten: This might be going too far, but we could probably just ignore anything after a `.` following a closing string delimiter.

This would deal with things like `license = 'mit'.upcase` or `license = '  Apache 2.0 '.trim`

I'd say we ignore this if you didn't see it when you downloaded 1000 gemspecs or whatever the number was.

src/Strategy/Ruby/Gemspec.hs
csasarak: Hi Megh, I've updated with text around each of the parsers. I hope they make it a little more clear what they do

src/Strategy/Ruby/Gemspec.hs
csasarak: I'm going to mark this resolved per our earlier conversation.

src/Strategy/Ruby/Gemspec.hs
csasarak: I've added support for escaping the closing delimiter in strings. In all my tests I didn't see this situation but it's good to have.

src/Strategy/Ruby/Gemspec.hs
```diff
@@ -0,0 +1,175 @@
...
+-- these examples:
+--
+-- > %w(foo bar)
+-- > %W(foo bar baz)
+--
+-- This is interpreted as an array of strings. The 'W' variant also allows
+-- interpolation, but as with 'rubyString' these are treated as ordinary text.
+parseRubyWordsArray :: Parser [Text]
+parseRubyWordsArray =
+  char '%'
```
spatten, line 132: Sorry I didn't notice this the first time through, but array literals can also use the same delimiters as string literals. Yet another case where Ruby took too much inspiration from Perl.

https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Literals#The_.25_Notation

So these all result in an array of words:

```ruby
a = %w(one two three)
a = %w{one two three}
a = %w#one two three#
a = %w<one two three>
```

The exact same thing happens for the `W` variants, but with a capital W you can put string interpolation inside of the words (which we should ignore like we are ignoring it for strings).

I think this might be worth fixing, but only because we already have the machinery you already built for strings that will make this reasonably easy to do.

src/Strategy/Bundler.hs
```diff
@@ -58,6 +84,23 @@ instance ToJSON BundlerProject
...
+instance LicenseAnalyzeProject BundlerProject where
+  licenseAnalyzeProject = fmap mconcat . traverse findLicenses . bundlerGemSpec
+
+findLicenses :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m [LicenseResult]
+findLicenses gemspecPath = do
+  assignments <- readContentsParser (readAssignments rubyLicenseValuesP) gemspecPath
+  let licenses = foldMap value . filter isLicenseKey $ assignments
+
+  -- license keys are recommended to be SPDX, but there isn't any requirement:
+  -- https://guides.rubygems.org/specification-reference/#license=
```
spatten, line 96: Did you ever run into this case from that URL?

> You can also use a custom license file along with your gemspec and specify a LicenseRef-<idstring>, where idstring is the name of the file containing the license text.

I had no idea that that existed. I'm mostly asking for curiousity's sake — unless you saw this in the one of the gemspecs you downloaded, I think we should fix this if and when we see it in the wild.

csasarak: I never encountered that and totally missed it myself. For now I think I'll hold off on implementing it because it seems kind of arcane.

src/Strategy/Ruby/Gemspec.hs
csasarak: I have incorporated this into my latest commit.

spatten (approved):
This is looking good to me from a "how Ruby works" side of things.

I think we'll also want an approval from @meghfossa on the Haskell side of things before merging

meghfossa (approved):

PR #826, opened 2022-03-03
Queue archive builds individually

zlav:
# Overview

We have been seeing that archive uploads using CLI v2 in certain scenarios for on-prem users have been failing for CLI v2 but not for CLI v1. After investigating the differences, I found that the only difference was that CLI v2 was queueing all specified archives in the same request. This is an accepted upload format as specified by the API, but we have verified that removing this behavior fixes the issue in the user's environment.

This PR adds:
- Better debug logging for archive uploads.
- Individual queuing of archives instead of queueing all of them at once.

## Acceptance criteria

- Successful archive uploads in the user's environment.

## Testing plan

Unfortunately, I cannot reproduce the failing upload behavior in the FOSSA cloud instance. In order to test these changes, I validated that the updated binary resulted in successful uploads by testing directly in the user's environment. Additionally, I manually tested a few different archive upload permutations to validate that the behavior with the FOSSA cloud had not changed.

These permutations were:
- 1 archive initially uploaded
- 2 archives uploaded (1 in addition to the 1st)
- 5 archives uploaded (1 queued before the first, 1 before the second, and 1 after the last)

The goal of these tests was to validate that any permutation of build queues would result in a successful upload.

## Risks

Nothing that I can think of, we are essentially reverting behavior to what it was like 2 years ago which had no issues.

## Checklist


Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Fossa CLI Changelog
 
+## v3.1.7
+- Queue archive builds individually. ([#826](https://github.com/fossas/fossa-cli/pull/826))
```
meghfossa, line 4: [Optional] I think we can make this entry more user friendly: 

```
Fossa-deps: Fixes an archive uploading bug for vendor dependency by queuing archive builds individually.
```

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -91,7 +91,8 @@ archiveUploadSourceUnit baseDir apiOpts vendoredDeps = do
 
   -- archiveBuildUpload takes archives without Organization information. This orgID is appended when creating the build on the backend.
   -- We don't care about the response here because if the build has already been queued, we get a 401 response.
-  _ <- Fossa.archiveBuildUpload apiOpts (ArchiveComponents $ NonEmpty.toList archives)
+  res <- traverse (Fossa.archiveBuildUpload apiOpts) (NonEmpty.toList archives)
+  logDebug $ pretty $ show res
```
meghfossa, line 95: [Question] This shows `X-Amz-Credential=...`, but I'm presuming this is a pre-signed URL for put object with a specific scope on such object so this is fine?

meghfossa (approved):
Left one optional, and one question, other than that lgtm 🚀

src/App/Fossa/ArchiveUploader.hs
zlav: Yea this will be the information for the pre-signed URL, which is helpful when figuring out exactly where the CLI tried to upload the archive. I think exposing this only for debug logs is acceptable unless you have a further objection.

meghfossa: No objections - I wanted to make sure this was not a security gotcha!

Issue #189, opened 2018-07-20
Build tool support


ilikebits:
This is a collection of build tools we may potentially support:

- [x] .NET: [NuGet](https://www.nuget.org/)
- [ ] C/C++: [Buckaroo](https://buckaroo.pm/)
- [ ] C/C++: [Conan](https://conan.io/)
- [ ] C/C++: [Hunter](https://docs.hunter.sh/en/latest/)
- [ ] C/C++: [Meson](https://mesonbuild.com/)
- [ ] C++: [Vcpkg](https://github.com/Microsoft/vcpkg)
- [ ] Clojure: [Boot](http://boot-clj.com/)
- [ ] Clojure: [Leiningen](https://leiningen.org/)
- [ ] Cross-language: [Bazel](https://bazel.build)
- [ ] Cross-language: [Buck](https://buckbuild.com/)
- [ ] Cross-language: [Pants](https://www.pantsbuild.org/index.html)
- [ ] Crystal: [Shards](https://crystal-shards-registry.herokuapp.com/shards)
- [ ] D: [DUB](http://code.dlang.org/)
- [ ] Dart: [Pub](https://pub.dartlang.org/)
- [ ] Docker: [Docker](https://www.docker.com/)
- [ ] Elixir: [Mix](https://hex.pm/)
- [ ] Elm: [Elm](http://package.elm-lang.org/)
- [ ] Erlang: [Rebar3](https://www.rebar3.org/)
- [x] Go: [dep](https://github.com/golang/dep)
- [x] Go: [GDM](https://github.com/sparrc/gdm)
- [x] Go: [glide](https://github.com/Masterminds/glide)
- [x] Go: [godep](https://github.com/tools/godep)
- [x] Go: [govendor](https://github.com/kardianos/govendor)
- [x] Go: [GPM](https://github.com/pote/gpm)
- [x] Go: [vndr](https://github.com/LK4D4/vndr)
- [ ] Haskell: [Cabal](https://www.haskell.org/cabal/)
- [ ] Haskell: [Stack](https://www.stackage.org/)
- [ ] Haxe: [Haxelib](https://lib.haxe.org/)
- [ ] iOS: [Carthage](https://github.com/Carthage/Carthage)
- [x] iOS: [Cocoapods](https://cocoapods.org)
- [x] Java: [Ant](http://ant.apache.org)
- [x] Java: [Gradle](https://gradle.org/)
- [ ] Java: [Ivy](http://ant.apache.org/ivy/)
[...]

PR #827, opened 2022-03-04
Handles symlink loops during discovery

periodic:
# Overview

Loops in symlinks could cause the discovery to loop forever.  This change uses the file/device ID instead of the path when determining when to recurse.

## Acceptance criteria

Cyclic directory tree (via symbolic links) can be analyzed by FOSSA CLI successfully.

## Testing plan

Tests were added for an example case provided which triggered this behavior.  That case has been encoded as a unit test.

```
.
└── lib
    ├── pg
    │   └── lib -> ..
    └── sqlite
        ├── example.txt
        └── lib -> ..
```

## Risks

This change relies on the `PosixCompat` library to handle cross-platform execution.  There's a chance that this leads to unexpected behavior.  Our testing suite should catch much of this.  In any case that it may be incorrect we are unlikely to run into issues because the IDs being used are just for checking what was visited and so just have to be unique and stable within a execution.

## References

This structure is typically used by [shards](https://crystal-lang.org/reference/1.3/the_shards_command/index.html#usage) dependency manager. As such, CLI discovery will fail, when CLI encounters projects managed by shards.

Support for Shards is still pending: https://github.com/fossas/fossa-cli/issues/189

## Checklist


periodic:
This seems to have come from an arbitrary ordering on the directory contents.  I fixed it by turning the paths back into a tree. It's not the prettiest solution, but it works.

periodic:
Finally, the windows build passed.

meghfossa (approved):
Tested on couple of crystal projects - no infinite loop 🚀

PR #832, opened 2022-03-09
Skips uploading an empty set of archives.

periodic:
# Overview

A user is seeing that their project is unable to use fossa-deps.json when they only have referenced-dependencies. They are getting a 443 and the error “archives must be a non-empty array” which corresponds to the CLI trying to queue an empty array of archive deps even though the user never specifies any vendored-dependencies.

The user reported that they had no issues on CLI v2.15.6 but had issues with v3.1.2. I validated that both versions no longer work which leads me to believe this is related to a change in the server response.

Note: this issue only impacts users uploading dependency graphs. Running fossa analyze -o succeeds because the build queue step is skipped

## Acceptance criteria

- The CLI no longer attempts to queue empty archive arrays
- A user can specify only referenced-dependencies in the fossa-deps file and fossa analyze will succeed.

## Testing plan

Tested on a local project with the following `fossa-deps.yml`: 

```yaml
referenced-dependencies:
- name: Django
  version: 1.0.0
  type: pypi
```

Before the change there was an error:

```
    Error

      An error occurred when accessing the FOSSA API.

      Error message from API:

          archives must be a non-empty array
```

After the change there was none.

Also tested with a non-empty `vendored-dependencies`

## Risks

This bug comes from coordination with the server.  This doesn't address any other future errors we may get.

This code is hard to test, so right now there are no additional tests.  The change should be small enough that that is acceptable.  There are plans to make this easier in the future.

## References

None

## Checklist

### DIFF
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -26,6 +26,8 @@ import Data.Aeson.Extra
 import Data.ByteString.Lazy qualified as BS
 import Data.Functor.Extra ((<$$>))
 import Data.List (intercalate, nub)
+import Data.List.NonEmpty (NonEmpty)
+import Data.List.NonEmpty qualified as NonEmpty
 import Data.Map.Strict qualified as Map
 import Data.Maybe (fromMaybe)
 import Data.String.Conversion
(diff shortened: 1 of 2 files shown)


periodic:
the `test-windows` check seems to be failing due to a timeout in set-up.  I have no reason to expect that there would be any issues with Windows though.

csasarak (approved):
LGTM!

PR #834, opened 2022-03-10
Feat/cocoapods declared license scanning

csasarak:
# Overview

_Provide an overview of this change. Describe the intent of this change, and how it implements that intent._

This changes implements and adds support for declared license scanning in cocaopods projects. 

_What is this PR trying to accomplish?_

* Discovery should find a podspec file in a cocoapods project directory
* The parser should be able to find license strings within the podspec file and report them out in pathfinder.

## Testing plan

I used unit tests. 

I ran manual tests against a set of about 60 cocoapods projects and found the correct licenses for all the podspecs that had a string literal as a license (one project didn't).

## Risks

Many of the cocoapods projects didn't have a podspec in the same directory as the `Podfile` and `Podfile.lock`. I describe these cases [here](https://github.com/fossas/team-analysis/wiki/Cocoapods-Gemspec-Declared-License-Scanning#cocoapods-notes). I don't think this is a risk for this PR but I think we should be aware of it going forward. 

## References

[Documentation on testing](https://github.com/fossas/team-analysis/wiki/Cocoapods-Gemspec-Declared-License-Scanning#cocoapods-notes)

Closes ANE-102.

## Checklist

### DIFF
--- a/src/Strategy/Ruby/Parse.hs
+++ b/src/Strategy/Ruby/Parse.hs
@@ -8,6 +8,7 @@ module Strategy.Ruby.Parse (
   gemspecLicenseValuesP,
   Symbol (..),
   parseRubySymbol,
+  parseRubyDict,
 ) where
 
 import Control.Applicative ((<|>))
@@ -127,9 +128,9 @@ newtype Symbol = Symbol {unSymbol :: Text}
   deriving (Show, Eq)
 
 -- |Parses a ruby symbol. Ex:
--- :this_is_a_symbole
--- :"this is also a symbol"
--- :'single quote symbol'
+-- > :this_is_a_symbol
+-- > :"this is also a symbol"
+-- > :'single quote symbol'
 parseRubySymbol :: Parser Symbol
 parseRubySymbol =
   Symbol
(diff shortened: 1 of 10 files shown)


src/Strategy/Ruby/Parse.hs
```diff
@@ -29,6 +36,22 @@ selectDelim = \case
...
+      (char d2)
+      ( many
+          ( try delimEscape
+              <|> (toText <$> anySingleBut d2)
+          )
+      )
+  where
+    -- only the close delimiter can stop parsing, so only check
+    -- for escaped versions of that.
+    delimEscape = string $ "\\" <> toText d2
```
kitified, line 53: I think the intent here is to consume an escaped `d2` before `between` can reach it (and terminate the parse), correct?

test/Ruby/ParseSpec.hs
```diff
@@ -51,9 +51,18 @@ stringParseSpec =
 rubyStringArray :: Text
 rubyStringArray = "[ \"hello\",\"world\" ,\t\"foo\",\"bar\"]"
 
+commentedRubyStringArray :: Text
+commentedRubyStringArray = "[ #cmt\n\"hello\"#cmt\n,\"world\" ,\t\"foo\",\"bar\"]"
```
kitified, line 55: Please prefer string literals. Example: https://github.com/fossas/fossa-cli/blob/3251a54302faccacf7a79cce09eb8d861e0faf0e/test/App/Fossa/VSI/DynLinked/Internal/Lookup/RPMSpec.hs#L32-L59

Only commenting here, but applies generally.

kitified (approved):

src/Strategy/Ruby/Parse.hs
csasarak: That is correct. Looking at this after a few days with fresh eyes I'm wondering if there's a simpler way to do it, do you have any ideas?

kitified: I don't; this is how I'd do it too.
I think it would be useful to comment that this is the intent though to make it more clear to future readers.

test/Ruby/ParseSpec.hs
csasarak: Oh thank you, I tripped over myself a few times writing these tests. I will update this and hopefully it will be a bit easier to understand.

PR #835, opened 2022-03-10
Fix/windows scoop install test

meghfossa:


.github/workflows/install-script-test.yml
```diff
@@ -85,7 +85,17 @@ jobs:
...
+          # We are installing scoop as admin for workaround, to: 
+          #   https://github.com/fossas/fossa-cli/runs/5503534701?check_suite_focus=true
+          #
+          # Error:
+          #   Running the installer as administrator is disabled by default, use -RunAsAdmin parameter if you know what you are doing.
+          #
+          # Note, github's windows hosted runners by default use admin user without UAC.
+          Set-ExecutionPolicy -ExecutionPolicy RemoteSigned -Scope CurrentUser
+          iwr -useb 'https://raw.githubusercontent.com/scoopinstaller/install/master/install.ps1' -outfile 'install.ps1'
+          .\install.ps1 -RunAsAdmin
```
meghfossa, line 98: `-RunAsAdmin` is spooky, but 
- We have moved from scoop's domain install script to github hosted script
- After inspecting install script, nothing seems "risky". 

If there is a concern, we can pin this to specific revision, but I don't see particular concern.

csasarak: I'm not very concerned either, is there anything particularly sensitive on this runners anyways?

meghfossa: Nope nothing of sensitive.

PR #839, opened 2022-03-15
Move dump commands to their own module

scruffystuffs:
These were causing circular dependencies due to them just being in the wrong place, so I moved them.  `EmbeddedBinary` shouldn't depend on `SubCommand`


meghfossa (approved):
LGTM 🚀

PR #840, opened 2022-03-16
API Prototype

periodic:
This is a draft of what the API effect might look like, including tests.


src/Control/Carrier/FossaAPIClientIO.hs
```diff
@@ -0,0 +1,28 @@
...
+-- | Runs FossaAPI effects as IO operations
+-- TODO: Put ApiOpts in a reader
+-- TODO: Exception handling
+runFossaAPIClientIO ::
+  ( Has (Lift IO) sig m
+  , Has Diagnostics sig m
+  , Has Logger sig m
+  ) =>
+  FossaAPIClientIOC m a ->
+  m a
```
scruffystuffs, line 24: I think some later work to do is split out an HTTP effect from this, but for now this works fine.

scruffystuffs: And `runFossaApiClientIO` might have a prime version that also terminates the HTTP effect.

test/Test/Effect.hs
```diff
@@ -109,3 +110,11 @@ shouldMatchList' a b = sendIO $ shouldMatchList a b
 
 expectFailure' :: Has (Lift IO) sig m => Result a -> m ()
 expectFailure' res = sendIO $ expectFailure res
+
+-- TODO: Better naming.
+expectFails' :: (Has Diagnostics sig m, Has (Lift IO) sig m) => m () -> m ()
+expectFails' f = do
```
scruffystuffs, line 116: Maybe `expectFatal`, since `fatal` is what `recover` catches.

src/Control/Carrier/FossaAPIClientIO.hs
```diff
@@ -0,0 +1,28 @@
...
+type FossaAPIClientIOC = SimpleC FossaAPIClientF
+
+-- | Runs FossaAPI effects as IO operations
+-- TODO: Put ApiOpts in a reader
+-- TODO: Exception handling
+runFossaAPIClientIO ::
+  ( Has (Lift IO) sig m
+  , Has Diagnostics sig m
+  , Has Logger sig m
+  ) =>
```
scruffystuffs, line 22: Let's add the `ReaderC ApiOpts` here now, and see if we can improve the testing api from that.

scruffystuffs: I think with some work, we can make `withMockApi` a lot easier to use, and more powerful.

periodic:
Closing in favor of #843

PR #841, opened 2022-03-16
Ane 78 go mod replace

csasarak:
# Overview

When a replacement is included in the output of `go list -m --json all` the replaced package isn't shown in the output of `fossa-cli`. This PR ensures that that replacement is performed before giving the report to the user.

## Acceptance criteria

* When a replacement is specified in the output of `go list`, our output should do the same replacement in our output.

## Testing plan

_How did you validate that this PR works? How did you check that the acceptance criteria were fulfilled?_

I've added unit tests and also done a manual test with a minimal project that exhibits the incorrect behavior.

## Risks

Testing - I checked and the go tools don't seem to allow recursive replacements but if there are any odd configurations a reviewing thinks I missed I'd like to know about them. 

## References

Closes ANE-98

## Checklist


Changelog.md
```diff
@@ -1,5 +1,8 @@
 # Fossa CLI Changelog
 
+## Unreleased
```
meghfossa, line 3: Let's do a release for this change!

src/Strategy/Go/GoModGraph.hs
```diff
@@ -179,16 +192,36 @@ analyze dir = do
...
   where
+    -- TODO: need to convert the version from the replacement into something well-typed
     toGoGraphMod :: GoListModule -> GoGraphMod
     toGoGraphMod GoListModule{..} = case (toGoModVersion =<< version) of
       Nothing -> MainMod path
       Just pv -> OtherMod path pv
 
-    withoutMain :: [GoListModule] -> Set GoGraphMod
-    withoutMain = fromList . map toGoGraphMod . filter (not . isMain)
+    toGraphModWithReplacement :: GoListModule -> (GoGraphMod, Maybe ModWithReplacement)
```
meghfossa, line 204: [Prefer] That we order definition of function in order they appear for easy readability (so we don't have to scan back and forth when reading)

E.g 
```haskell
    findSelectedMods :: [GoListModule] -> (Set GoGraphMod, GoModReplacements)
    findSelectedMods =
      bimap fromList (mkReplacementMap . catMaybes)
        . unzip
        . map toGraphModWithReplacement
        . filter (not . isMain)
 
     mkReplacementMap :: [ModWithReplacement] -> GoModReplacements
    mkReplacementMap = GoModReplacements . Map.fromList . map unModWithReplacement
 
     toGraphModWithReplacement :: GoListModule -> (GoGraphMod, Maybe ModWithReplacement)
     toGraphModWithReplacement listMod =
      case toGoGraphMod listMod of
        m@(MainMod _) -> (m, Nothing)
        m@(OtherMod _ _) ->
          ( m
          , do
              GoModuleReplacement{..} <- moduleReplacement listMod
              version <- toGoModVersion versionReplacement
              Just $ ModWithReplacement (m, OtherMod pathReplacement version)
       
```

src/Strategy/Go/GoModGraph.hs
```diff
@@ -179,16 +192,36 @@ analyze dir = do
...
+    toGraphModWithReplacement listMod =
+      case toGoGraphMod listMod of
+        m@(MainMod _) -> (m, Nothing)
+        m@(OtherMod _ _) ->
+          ( m
+          , do
+              GoModuleReplacement{..} <- moduleReplacement listMod
+              version <- toGoModVersion versionReplacement
```
meghfossa, line 212: [Prefer] `replacedVersion`

src/Strategy/Go/GoModGraph.hs
```diff
@@ -179,16 +192,36 @@ analyze dir = do
   goModGraphStdout <- context ("Getting selected dependencies versions using, " <> toText (show goModGraphCmd)) $ execParser parseGoModGraph dir goModGraphCmd
   ggm <- fromMaybeText "expected to find main module, but found no main module!" mainMod
 
-  let graph = buildGraph goModGraphStdout ggm selectedMods directMods filterModsNotUsedInBuild
+  let graph = buildGraph goModGraphStdout ggm selectedMods directMods filterModsNotUsedInBuild modReplacements
   pure (graph, Complete)
   where
+    -- TODO: need to convert the version from the replacement into something well-typed
```
meghfossa, line 198: If this does not require significantly more cycles - I rather we do this as part of this PR.

src/Strategy/Go/GoModGraph.hs
```diff
@@ -117,44 +127,45 @@ parseGoModGraph = parseGoModPair `sepEndBy` (symbol "\n" <|> symbol "\r") <* eof
     parseGoModPair = (,) <$> parseGoGraphMod <*> parseGoGraphMod
 
 -- Builds graph from edges, main module, and selected module version set.
-buildGraph :: [(GoGraphMod, GoGraphMod)] -> GoGraphMod -> Set GoGraphMod -> Set GoGraphMod -> Bool -> Graphing.Graphing Dependency
-buildGraph fromToMods mainMod selectedMods directMods applyMVS =
-  Graphing.gmap toDependency
-    . Graphing.filter (withSelection)
+buildGraph :: [(GoGraphMod, GoGraphMod)] -> GoGraphMod -> Set GoGraphMod -> Set GoGraphMod -> Bool -> GoModReplacements -> Graphing.Graphing Dependency
```
meghfossa, line 130: [Optional] Now that it has 7 args, let's simplify this. 

One suggestion that is quick/dirty is to use a container, it's not the ideal approach but would simplify and improve code readability. 

```haskell 
data GoGraphArgs = GoGraphArgs {
  edges :: [(GoGraphMod, GoGraphMod)],
  root :: GoGraphMod,
  ....
 . ...
}
```

meghfossa (commented):
Can you include the repeatable testing plan in pr comments (example project) and commands? I can run through these steps for review/testing.

src/Strategy/Go/GoModGraph.hs
csasarak: I noticed this as well and that might be good. One other option I thought of was trying to make wrapper types for each of the different kinds of GoGraphMod sets. At least then even with a function with so many arguments the type system would help make sure they're put in the right order. I will hack on that or try to make an argument record and push those changes.

src/Strategy/Go/GoModGraph.hs
csasarak: This is actually a holdover comment I forgot to add. I do this step in `toGraphModWithReplacement`. I will remove the comment - is there any type safety you think is missing that should be added in addition?

src/Strategy/Go/GoModGraph.hs
csasarak: I noticed this as well and that might be good. One other option I thought of was trying to make wrapper types for each of the different kinds of GoGraphMod sets. At least then even with a function with so many arguments the type system would help make sure they're put in the right order. I will hack on that or try to make an argument record and push those changes.

src/Strategy/Go/GoModGraph.hs
csasarak: I've done this

kitified (approved):

kitified (approved):
Nice, I like the new config type ✨

meghfossa (approved):
LGTM! I didn't get to test this e/e as part of the review.

PR #842, opened 2022-03-17
Moves API types from FossaAPIV1 to Fossa.API.Types.

periodic:
# Overview

We had various types associated with the API scattered across two files.  This moves those types over to `Fossa.API.Types` where some similar types already exists.  This helps separate the types from the HTTP implementation which is important for future refactoring around the API.

## Acceptance criteria

No types are exported from `FossaAPIV1` that aren't related to HTTP.

## Testing plan

This is just moving types, so as long as it all compiles and tests work we should be fine.

I checked `fossa analyze` and `fossa test` on the local build with no problems.

## Checklist

### DIFF
--- a/src/App/Fossa/API/BuildWait.hs
+++ b/src/App/Fossa/API/BuildWait.hs
@@ -24,7 +24,15 @@ import Control.Effect.StickyLogger (StickyLogger, logSticky')
 import Control.Timeout (Cancel, checkForCancel)
 import Data.Text (Text)
 import Effect.Logger (Logger, pretty, viaShow)
-import Fossa.API.Types
+import Fossa.API.Types (
+  ApiOpts,
+  Build (buildTask),
+  BuildStatus (StatusFailed, StatusSucceeded),
+  BuildTask (buildTaskStatus),
+  Issues (issuesStatus),
+  Organization (organizationId),
+  Project (projectIsMonorepo),
+ )
 
 pollDelaySeconds :: Int
 pollDelaySeconds = 8


liftM:
Re: testing plan - I'm not gonna block this PR, but in general (even in Haskell) we should at least do a really quick manual test that the code paths being changed still behave properly. Just upload an analysis that exercises the code paths being changed. 

I've caught many surprising issues that way (e.g. weirdness with Template Haskell, functions doing different things when their inferred types changed, etc.), and it's a really cheap test that provides surprising value.

periodic:
I'll run a quick analysis/upload with it then.

Are there any places in particular you worry about not being covered by the unit tests?

liftM:
It looks like this PR touches `fossa analyze` and `fossa test`'s API routes. So run a quick `fossa analyze` and `fossa test` and if nothing crashes and the web app link works and the uploaded revision looks reasonable in the web app then it's probably fine. I don't think you even need to spin up a local instance for this - just use a test organization in production.

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -64,7 +55,7 @@ import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Word (Word8)
 import Effect.Logger
-import Fossa.API.Types (ApiOpts, ArchiveComponents, Issues, SignedURL, signedURL, useApiOpts)
+import Fossa.API.Types
```
scruffystuffs, line 58: Don't use implicit imports:

We should also clean up the pre-existing implicit imports here, if you're using VSCode there's a code action for making imports explicit automatically (you'll have to run the formatter afterward).

scruffystuffs: Also, we should probably remove the prelude import, it does nothing.

src/App/Fossa/API/BuildWait.hs
```diff
@@ -24,7 +24,7 @@ import Control.Effect.StickyLogger (StickyLogger, logSticky')
 import Control.Timeout (Cancel, checkForCancel)
 import Data.Text (Text)
 import Effect.Logger (Logger, pretty, viaShow)
-import Fossa.API.Types (ApiOpts, Issues (..))
+import Fossa.API.Types
```
scruffystuffs, line 27: Don't use implicit imports:

We should also clean up the pre-existing implicit imports here, if you're using VSCode there's a code action for making imports explicit automatically (you'll have to run the formatter afterward).

scruffystuffs (approved):
Style guide nits, but otherwise LGTM

periodic:
Just the link failing due to that one service being down.

PR #844, opened 2022-03-21
[ANE-113] Report attribution options

kitified:
# Overview

Adds a new option, `--format`, to `fossa report attribution` with the following values:
- `spdx`
- `markdown`
- `json`

Deprecates `--json`. However, `--json` is still supported _and takes precedence over `--format`._

## Acceptance criteria

- I ran a scan of a project: `cabal run fossa -- analyze`
- I ran `cabal run fossa -- report attribution --json`. This output was the same as before.
- I ran `cabal run fossa -- report attribution --format json`. This output was the same as with `--json`.
- I ran `cabal run fossa -- report attribution --format spdx`. This output was the same as downloading the SPDX report from the FOSSA UI.
- I ran `cabal run fossa -- report attribution --format markdown`. This output was the same as downloading the markdown report from the FOSSA UI.
- I ran `cabal run fossa -- report attribution -h` and observed that the help output shows `--json` deprecated and lists viable formats for `--format`.
- I ran `cabal run fossa -- report attribution --format foo` and observed the CLI reporting `foo` as an invalid format.

## Testing plan

I performed the steps in the AC against the FOSSA staging instance. Note: until https://github.com/fossas/FOSSA/pull/7311 is merged to `master`, the `spdx` and `markdown` options will not work in production.

## Risks

This change has been performed in a backwards compatible manner, so I don't believe it to be risky.

## References

Closes https://fossa.atlassian.net/browse/ANE-113

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # Fossa CLI Changelog
 
+## v3.1.5
+
+- Adds `--format` to `fossa report attribution` and deprecates `--json`. ([#844](https://github.com/fossas/fossa-cli/pull/844))
+
 ## v3.1.4
 
 - Handles symlink loops in directory structure. ([#827](https://github.com/fossas/fossa-cli/pull/827))
--- a/docs/differences-from-v1.md
+++ b/docs/differences-from-v1.md
@@ -116,7 +116,7 @@ Following CLI commands are not supported with 3.x:
 - `fossa analyze --options`
 - `fossa test --suppress-issues`
 - `fossa upload` (this is supplemented by `fossa-deps.{yml,json}` file now, refer to "Migrate archive upload targets" section)
-- `fossa report licences` (supplemented by `fossa report attribution --json`)
+- `fossa report licences` (supplemented by `fossa report attribution --format json`)
 - `fossa report dependencies`
 
 ### Language Specific Changes
@@ -221,7 +221,7 @@ If you are integrating a private project and want to share more details, or if y
 
 In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.
 
-In 3.x, only `fossa report attribution --json` is reported. The information in the previously supported v1 commands is contained in the output from `attribution`.
+In 3.x, only `fossa report attribution --format json` is reported. The information in the previously supported v1 commands is contained in the output from `attribution`.
 
 The format for this report has also changed slightly. Until the end of April 2022, FOSSA supports a compatibilty script to convert the format of the report.
 For more information, including usage information, see [FOSSAv1 report compatibility](references/subcommands/report.md#fossav1-report-compatibility) in the user guide.
--- a/docs/differences-from-v1.md
+++ b/docs/differences-from-v1.md
@@ -221,7 +221,7 @@ If you are integrating a private project and want to share more details, or if y
 
 In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.
 
-In 3.x, only `fossa report attribution --format json` is reported. The information in the previously supported v1 commands is contained in the output from `attribution`.
+In 3.x, only `fossa report attribution` is supported. The information in the previously supported v1 commands is contained in the output from `attribution`.
 
 The format for this report has also changed slightly. Until the end of April 2022, FOSSA supports a compatibilty script to convert the format of the report.
 For more information, including usage information, see [FOSSAv1 report compatibility](references/subcommands/report.md#fossav1-report-compatibility) in the user guide.
--- a/docs/differences-from-v1.md
+++ b/docs/differences-from-v1.md
@@ -222,6 +222,7 @@ If you are integrating a private project and want to share more details, or if y
 In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.
 
 In 3.x, only `fossa report attribution` is supported. The information in the previously supported v1 commands is contained in the output from `attribution`.
+This command also now requires a `--format` argument: see [fossa report attribution](references/subcommands/report.md#specifying-a-report-format) for details.
 
 The format for this report has also changed slightly. Until the end of April 2022, FOSSA supports a compatibilty script to convert the format of the report.
 For more information, including usage information, see [FOSSAv1 report compatibility](references/subcommands/report.md#fossav1-report-compatibility) in the user guide.
--- a/docs/references/subcommands/report.md
+++ b/docs/references/subcommands/report.md
@@ -35,7 +35,7 @@ Available options are:
 For example, to render the report in JSON format, use `fossa report attribution --format json`.
 
 In earlier versions of the FOSSA CLI, the `--json` flag was used to denote rendering the report in JSON format.
-This flag is still supported and takes precedence over the `--format` flag if present, but is considered deprecated.
+For backwards compatibility, this flag is still supported and takes precedence over the `--format` flag if present.  However, it is considered deprecated, and should be replaced with `--format json` to prevent future changes from breaking this behavior.
 
 ### FOSSAv1 report compatibility
 
--- a/docs/references/subcommands/report.md
+++ b/docs/references/subcommands/report.md
@@ -34,11 +34,6 @@ Available options are:
 
 For example, to render the report in JSON format, use `fossa report attribution --format json`.
 
-*NOTE: Currently, text reports are not supported, and the report will be*
-*printed as JSON if no format is provided.*
-*It is recommended to use the `--format json` flag anyway, since*
-*the behavior of the command without the flag will change in the future.*
-
 In earlier versions of the FOSSA CLI, the `--json` flag was used to denote rendering the report in JSON format.
 This flag is still supported and takes precedence over the `--format` flag if present, but is considered deprecated.
 
--- a/docs/references/subcommands/report.md
+++ b/docs/references/subcommands/report.md
@@ -42,10 +42,6 @@ For example, to render the report in JSON format, use `fossa report attribution 
 In earlier versions of the FOSSA CLI, the `--json` flag was used to denote rendering the report in JSON format.
 This flag is still supported and takes precedence over the `--format` flag if present, but is considered deprecated.
 
-### Print report as markdown
-
-### Print report as SPDX
-
 ### FOSSAv1 report compatibility
 
 FOSSA supports a compatibility script that converts FOSSAv2 attribution report output to the same format as that in FOSSAv1.
--- a/docs/references/subcommands/report.md
+++ b/docs/references/subcommands/report.md
@@ -35,7 +35,8 @@ Available options are:
 For example, to render the report in JSON format, use `fossa report attribution --format json`.
 
 In earlier versions of the FOSSA CLI, the `--json` flag was used to denote rendering the report in JSON format.
-For backwards compatibility, this flag is still supported and takes precedence over the `--format` flag if present.  However, it is considered deprecated, and should be replaced with `--format json` to prevent future changes from breaking this behavior.
+For backwards compatibility, this flag is still supported and takes precedence over the `--format` flag if present.
+However, it is deprecated, and should be replaced with `--format json` to prevent future changes from breaking this behavior.
 
 ### FOSSAv1 report compatibility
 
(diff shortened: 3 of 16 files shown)


src/App/Fossa/Config/Report.hs
```diff
@@ -52,8 +55,24 @@ instance Show ReportType where
 -- TODO: Add support for text-format reports
 data ReportOutputFormat
   = ReportJson
+  | ReportMarkdown
+  | ReportSPDX
   -- ReportPretty
-  deriving (Eq, Ord, Show)
+  deriving (Eq, Ord, Enum, Bounded)
```
scruffystuffs, line 61: Remove `TODO` and `ReportPretty`.  We had originally planned to reproduce the text reports from v1, but with markdown and SPDX, it seems unlikely that we'll go down that route.

src/App/Fossa/Config/Report.hs
```diff
@@ -136,11 +157,13 @@ mergeOpts cfgfile envvars ReportCliOptions{..} = do
...
-  if doJson
-    then pure ReportJson
-    else fatalText "Plaintext reports are not available for this report."
+validateOutputFormat :: Has Diagnostics sig m => Bool -> Maybe String -> m ReportOutputFormat
+validateOutputFormat True _ = pure ReportJson
+validateOutputFormat False Nothing = fatalText "Plaintext reports are not available for this report."
+validateOutputFormat False (Just format) | format == show ReportJson = pure ReportJson
+validateOutputFormat False (Just format) | format == show ReportMarkdown = pure ReportMarkdown
+validateOutputFormat False (Just format) | format == show ReportSPDX = pure ReportSPDX
+validateOutputFormat False (Just format) = fatalText $ "Report format " <> toText format <> " is not supported"
```
scruffystuffs, line 164: Three notes:
- We can simplify this by deriving `Read` on the `ReportOutputFormat` type, and using the `readMaybe` helper:
- We should use concrete Error types.
- The error for running without options should be updated to say something "You have to use either `--format foo` or `--json`".  Or we should find a good default, since not providing options should generally not cause failure.  This was a bad decision from when I first implemented reports in v2, I'm glad we're cleaning this up as a side-effect of SPDX reports.

```suggestion
validateOutputFormat :: Has Diagnostics sig m => Bool -> Maybe String -> m ReportOutputFormat
validateOutputFormat True _ = pure ReportJson
validateOutputFormat False Nothing = fatal NoFormatProvided
validateOutputFormat False (Just format) = fromMaybe (InvalidReportFormat format) $ readMaybe format
```

You'll have to add 
```haskell
data NoFormatProvided = NoFormatProvided
instance ToDiagnostic NoFormatProvided where
  something something = something $ "Plaintext reports are not available for this report."
```

Or whatever error message you think is best.

kitified: ooh, thanks!

scruffystuffs: Normally, using `Read` is verboten.  But in this case, I think it makes a lot of sense, and we're using the safe `readMaybe` as opposed to the unsafe `read`.

kitified: > Normally, using Read is verboten. But in this case, I think it makes a lot of sense, and we're using the safe readMaybe as opposed to the unsafe read.

Does this mean that in reality, using `read` is verboten, not implementing `Read`?

(Also, deriving `Read` doesn't work for us, because it looks for e.g. `ReportJson`, not `json`, when derived- but I'll manually implement it.)

scruffystuffs: Ah, I mixed up how `Show` and `Read` work together.  For some reason, I thought that deriving `Read` would change based on the impl of `Show`, but that's obviously not true.

kitified: Updated!

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # Fossa CLI Changelog
 
+## v3.1.5
```
periodic, line 3: Aren't we supposed to create an "Unreleased" section?  This saves a step of changing it when doing a release, but doesn't follow the process in our docs.  This feels simpler, but gives the false impression that this release exists.

docs/differences-from-v1.md
```diff
@@ -221,7 +221,7 @@ If you are integrating a private project and want to share more details, or if y
 
 In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.
 
-In 3.x, only `fossa report attribution --json` is reported. The information in the previously supported v1 commands is contained in the output from `attribution`.
+In 3.x, only `fossa report attribution --format json` is reported. The information in the previously supported v1 commands is contained in the output from `attribution`.
```
periodic, line 224: maybe just remove the format flag here now that we support more than one?

docs/references/subcommands/report.md
```diff
@@ -24,17 +24,19 @@ fossa report attribution --timeout 60
...
+- `markdown`
+- `spdx`
 
-```sh
-fossa report attribution --json
-```
+For example, to render the report in JSON format, use `fossa report attribution --format json`.
+
+In earlier versions of the FOSSA CLI, the `--json` flag was used to denote rendering the report in JSON format.
+This flag is still supported and takes precedence over the `--format` flag if present, but is considered deprecated.
```
periodic: Minor nit: "considered" weakens this statement.  We are deprecating it, right?  We should just say that.

src/App/Fossa/Config/Report.hs
```diff
@@ -40,20 +43,45 @@ import Options.Applicative (
...
-  deriving (Eq, Ord, Show)
+  | ReportMarkdown
+  | ReportSpdx
+  deriving (Eq, Ord, Enum, Bounded)
+
+instance Read ReportOutputFormat where
+  readsPrec _ s | s == show ReportJson = [(ReportJson, safeStripPrefix (show ReportJson) s)]
+  readsPrec _ s | s == show ReportSpdx = [(ReportSpdx, safeStripPrefix (show ReportSpdx) s)]
+  readsPrec _ s | s == show ReportMarkdown = [(ReportMarkdown, safeStripPrefix (show ReportMarkdown) s)]
+  readsPrec _ _ = []
```
scruffystuffs, line 67: This is a little embarrassing, but the ONLY reason I suggested using `Read` at all was because I mistakenly thought that somehow it would derive from the `Show` instance (which, frankly, was silly).  Since that was incorrect, we should probably avoid using `Read` altogether (I'm so sorry for making you add this just to remove it).  It's just a bad typeclass overall, I really don't know why I thought this would be a good idea.

We should probably just make this a simple matching function `String -> Maybe ReportOutputFormat`.

src/App/Fossa/Config/Report.hs
```diff
@@ -40,20 +43,45 @@ import Options.Applicative (
...
+
+safeStripPrefix :: String -> String -> String
```
periodic: I think this is the same as `Data.List.Extra.dropPrefix`.  I also like the `drop` name more because it evokes that behavior.  `safe` is often used for functions that throw exceptions and converts them into functions returning `Maybe` or `Either`.  This feels like the opposite.

src/App/Fossa/Config/Report.hs
```diff
@@ -40,20 +43,45 @@ import Options.Applicative (
...
+
+instance Show ReportOutputFormat where
+  show ReportJson = "json"
+  show ReportMarkdown = "markdown"
+  show ReportSpdx = "spdx"
+
+reportOutputFormatList :: String
+reportOutputFormatList = intercalate ", " $ map show allFormats
+  where
+    allFormats :: [ReportOutputFormat]
```
periodic, line 78: This is `Data.List.Extra.enumerate`.

src/App/Fossa/Config/Report.hs
```diff
@@ -136,11 +166,23 @@ mergeOpts cfgfile envvars ReportCliOptions{..} = do
...
     <*> revision
 
-validateOutputFormat :: Has Diagnostics sig m => Bool -> m ReportOutputFormat
-validateOutputFormat doJson =
-  if doJson
-    then pure ReportJson
-    else fatalText "Plaintext reports are not available for this report."
+data NoFormatProvided = NoFormatProvided
+instance ToDiagnostic NoFormatProvided where
+  renderDiagnostic NoFormatProvided = "Provide a format option via '--format' to render this report."
```
periodic, line 166: Should we list the possible formats here?

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -457,15 +458,17 @@ getIssues apiOpts ProjectRevision{..} = fossaReq $ do
 
 ---------------
 
-attributionEndpoint :: Url 'Https -> Int -> Locator -> Url 'Https
-attributionEndpoint baseurl orgId locator = baseurl /: "api" /: "revisions" /: renderLocatorUrl orgId locator /: "attribution" /: "json"
+attributionEndpoint :: Url 'Https -> Int -> Locator -> ReportOutputFormat -> Url 'Https
+attributionEndpoint baseurl orgId locator ReportJson = baseurl /: "api" /: "revisions" /: renderLocatorUrl orgId locator /: "attribution" /: "json"
```
periodic, line 462: This feels really verbose because it repeats the whole URL for all the options.  Could we just make a `where` expression that does a `case` on the format to generate that path segment?

src/App/Fossa/FossaAPIV1.hs
```diff
@@ -457,15 +458,17 @@ getIssues apiOpts ProjectRevision{..} = fossaReq $ do
...
+attributionEndpoint baseurl orgId locator ReportSpdx = baseurl /: "api" /: "revisions" /: renderLocatorUrl orgId locator /: "attribution" /: "full" /: "spdx"
+attributionEndpoint baseurl orgId locator ReportMarkdown = baseurl /: "api" /: "revisions" /: renderLocatorUrl orgId locator /: "attribution" /: "full" /: "MD"
 
-getAttribution ::
+getAttributionJson ::
```
periodic, line 466: Why do we need to handle JSON in a special way?  It looks like this function is only used within this module (I only see four uses of this word in the diff).  Maybe we should just get rid of it and just use bytestring output.

periodic (commented):
Review on the previous version, maybe you fixed this stuff...

Changelog.md
scruffystuffs: We do this when we're planning to release immediately.

src/App/Fossa/FossaAPIV1.hs
scruffystuffs: The querystring options passed to the server are different.  I think we're not sure if we can use them everywhere else yet.

docs/differences-from-v1.md
scruffystuffs: Can you clarify "remove the format flag"?

src/App/Fossa/Config/Report.hs
scruffystuffs: That function doesn't exist.

scruffystuffs: https://github.com/fossas/fossa-cli/blob/master/src/Data/List/Extra.hs

scruffystuffs: I assume this is from a custom prelude package.  We've had trouble using those, it can kill IDE inference, and has made it harder for newcomers to get involved in the codebase.

scruffystuffs: Though it's been a while since we've tried using one, and the one we used was hand-rolled.

src/App/Fossa/FossaAPIV1.hs
kitified: Yep, the query string is unused for the backend route used by the other formats. While technically it wouldn't break anything to pass them anyway today, I don't consider it correct.

docs/differences-from-v1.md
kitified: Updated to:
> In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.
> In 3.x, only `fossa report attribution` is supported. The information in the previously supported v1 commands is contained in the output from `attribution`.

I think we only originally had `--json` in there because it was required.

scruffystuffs: We should leave a note about `--format` being required as well.

kitified: Done.
```markdown
In 1.x, `fossa report` supported `fossa report dependencies`, `fossa report licenses` and `fossa report attribution`.

In 3.x, only `fossa report attribution` is supported. The information in the previously supported v1 commands is contained in the output from `attribution`.
This command also now requires a `--format` argument: see [fossa report attribution](references/subcommands/report.md#specifying-a-report-format) for details.
```

PR #846, opened 2022-03-21
Moves git access into an effect for easier testing

periodic:
# Overview

I was about to write some tests for `App.Fossa.Analyze.Upload` and saw that we still had one IO operation in there: git contributor scanning.  Seemed simple enough to just make it an effect so that it can be mocked in tests.

## Acceptance criteria

- Git access moved to an effect and access to Git made explicit (much like `readFS`)
- Analysis upload still works and reports contributors.

## Testing plan

1. Created an empty commit in a project I was using for testing that had a different author than me.
    ```sh
    git -c "user.name=Not Drew" -c "user.email=drew@fossa.com" \
        commit --allow-empty -m "Test with different author."
    ```
2. Ran `fossa analyze`
3. Checked in the web app that the number of contributors had increased by one, indicating the new one was picked up and uploaded.

## Risks

Very little of the code actually moved around, so that helps reduce the chance of bugs.

## Checklist


csasarak (approved):
LGTM. I do have a few questions if you have the time to answer them though:

Would a valid alternative here have been to use the `Exec` effect? I guess the weakness there is it becomes a bit more difficult to test code that executes multiple different commands. 

It seems like if we applied the reasoning behind this PR across the project we'd have an effect for every kind of command we might run (gradle, bundler, npm, etc.). I can see the advantages of that, though it does mean a proliferation of different handlers. Do you have any thoughts about that or what is a good criterion for writing an effect vs. just using Exec?

BTW, this was a nice introduction for me to writing a simple effect, I hadn't used `fused-effects` before. I think it's probably worth recording that this code exists somewhere because I think it would be good code to learn from for anyone who needs it.

periodic:
We talked about this a bit during our stand-up, but I'll record some of it here for posterity.

- Having an effect makes the testing simpler because we can just think about the inputs and outputs instead of worrying about all the stuff related to running a process.
- Lots of effects can get a little complicated to manage.
- Having many effects at different levels means that code units don't have to cross as many layers of abstraction.
- It splits up the code so that the `Upload` module is no longer dependent on the `VCS.Git` module or `Exec`.  This improves decoupling so that things can be further refactored later and is easier to reason about.

PR #848, opened 2022-03-22
Refactors App.Fossa.Analyze.Upload to use effects

periodic:
# Overview

This is the first place where I refactored the whole file to really use effects for API access.  This is part of a larger project to refactor our API access and make the code more testable.

## Acceptance criteria

- `App.Fossa.Analyze.Upload` should have all it's API and Git access behind an effect.
- There should be tests for the major code paths.

## Testing plan

I chose this one because it's very easy to test locally because uploading analysis is one of the core things we do!

Wrote some unit tests for it.

The part that is tricky to unit-test is the JSON output because that is not behind an effect yet.

## Risks

This is a pretty core flow.  Most of the refactoring is just reorganizing code, so there isn't much danger in the logic changing, but there is still some.

## References

[ANE-130](https://fossa.atlassian.net/browse/ANE-130)

## Checklist


spectrometer.cabal
```diff
@@ -507,6 +509,7 @@ test-suite unit-tests
     , hspec-expectations-pretty-diff  ^>=0.7.2.5
     , hspec-hedgehog                  ^>=0.0.1.2
     , hspec-megaparsec                ^>=2.2
+    , HUnit                           ^>=1.6.0
```
periodic, line 512: I need to get at `assertFailure :: String -> IO a` to put in the API mocks.  The `expectationFailure :: String -> IO ()` is locked to returning unit and so it makes for nasty lint warnings if I want it to fit in something like `m a`

periodic:
Looks like there is a `Locator` type that just wasn't getting used.  I'll update this to use that instead.

kitified (approved):

PR #850, opened 2022-03-25
Don't use NE.fromlist

scruffystuffs:
# Overview

Merge skew and early PR checks caused an `hlint` issue to slip through.  You can see the error here: https://github.com/fossas/fossa-cli/runs/5686530735?check_suite_focus=true

### DIFF
--- a/test/Test/Fixtures.hs
+++ b/test/Test/Fixtures.hs
@@ -74,9 +74,10 @@ projectRevision =
     }
 
 sourceUnits :: NE.NonEmpty SourceUnit
-sourceUnits =
-  NE.fromList
-    [ SourceUnit
+sourceUnits = unit NE.:| []
+  where
+    unit = 
+       SourceUnit
         { sourceUnitName = "testSourceUnitName"
         , sourceUnitType = "testSourceUnitType"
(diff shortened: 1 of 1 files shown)


test/Test/Fixtures.hs
```diff
@@ -74,9 +74,10 @@ projectRevision =
     }
 
 sourceUnits :: NE.NonEmpty SourceUnit
-sourceUnits =
-  NE.fromList
-    [ SourceUnit
+sourceUnits = unit NE.:| []
```
liftM, line 77: IMO, it's fine to import `:|` unqualified into the module namespace. I think it reads nicer that way, and I'm not aware of any operators that conflict with it.

scruffystuffs: Yeah but that changes the import.  I didn't really care, and I don't think this is any less readable.  I'm planning a cleanup soon to take care of inconsistent imports.

periodic (approved):

test/Test/Fixtures.hs
periodic: I only did this because there's no `NE.singleton`... yet.

PR #851, opened 2022-03-25
fix: Accept `yaml` as a config file extension

liftM:
# Overview

This PR adds `.fossa.yaml` as an accepted default configuration name.

## Acceptance criteria

Directories with a `.fossa.yaml` but no `.fossa.yml` now read their configuration from `.fossa.yaml`. A directory with both `.fossa.yml` and `.fossa.yaml` still reads configuration from `.fossa.yml`.

## Testing plan

1. Using a test project, create a `.fossa.yaml` and set the `server` field to something nonsensical.
2. Run `fossa analyze` on the test project and see that the analysis fails because it could not connect to the host.

I also checked that the added test fails without the source changes.

## Checklist


src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -90,18 +94,23 @@ resolveConfigFile base path = do
...
+          defaultConfigFileNames
+      case asum possibleConfigFilePaths of
+        Just actualConfigFilePath -> do
+          configFile <- readContentsYaml actualConfigFilePath
           let version = configVersion configFile
           if version >= 3
             then pure $ Just configFile
             else -- Invalid config found without --config flag: warn and ignore file.
               logWarn (warnMsgForOlderConfig @AnsiStyle version) $> Nothing
+        Nothing -> pure Nothing
```
scruffystuffs, line 113: When I see this in a case statement, I'm usually pretty sure we can just use the monad instead (if there's no `pure`, we can use the functor). something like `asum possibleConfigFilePaths >>= do ...`.  I don't know if that's more readable, but I generally prefer that to case statements.

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -90,18 +94,23 @@ resolveConfigFile base path = do
...
     DefaultConfigLocation dir -> do
-      let realpath = dir </> defaultConfigFileName
-      exists <- doesFileExist realpath
-      if not exists
-        then -- File missing and not requested
-          pure Nothing
-        else do
-          configFile <- readContentsYaml realpath
+      possibleConfigFilePaths <-
+        mapM
```
scruffystuffs, line 98: `mapM = traverse`, prefer `traverse`.  This should probably become an `hlint` rule.

scruffystuffs (commented):
Quick review looks OK, left two nits

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -90,18 +94,23 @@ resolveConfigFile base path = do
...
+          ( \configFileName -> do
+              let configFilePath = dir </> configFileName
+              exists <- doesFileExist configFilePath
+              pure $ if exists then Just configFilePath else Nothing
```
periodic, line 102: This monad doesn't have `Maybe` as it's failure monad case does it?  If so you could just write `guard exists $> configFilePath`

periodic (approved):
LGTM

src/App/Fossa/Config/ConfigFile.hs
liftM: I don't quite want to use `guard` here because I want the result to be `[Maybe a]` not `[a]` because `asum` already conveniently gives me `Nothing` in the empty case.

src/App/Fossa/Config/ConfigFile.hs
liftM: I can add this in a follow-up PR.

src/App/Fossa/Config/ConfigFile.hs
liftM: I've rewritten this using `whenJust`. Does this feel better? Should I refactor `whenJust` into a helper somewhere?

We can't exactly use the `Maybe` monad here because the inner action is itself effectful.

scruffystuffs: Damn, you're right.  I did say that it was a very quick look.  I prefer it the previous way for now.

liftM: Yeah, I agree - it's annoying but clearer. Will revert and then merge.

PR #853, opened 2022-03-25
refactor: New hlint rule: replace mapM with traverse

liftM:
# Overview

Adds "replace `mapM` with `traverse`" as an `hlint` rule. See also https://github.com/fossas/fossa-cli/pull/851#discussion_r835590880.


liftM:
Whoops - reverting this to draft so I can see what hlint breakages I need to fix before requesting review.

kitified (approved):

PR #854, opened 2022-03-25
fix: Config file path should be relative to analyze basedir

liftM:
# Overview

This PR fixes configuration file resolution. Originally, running `fossa analyze /foo/bar` would look at the CWD for a `.fossa.yml` instead of in `/foo/bar`. This PR fixes resolution to instead correctly search for `/foo/bar/.fossa.yml`.

## Acceptance criteria

When a user runs `fossa analyze /foo/bar`, the configuration file at `/foo/bar/.fossa.yml` should be used instead of the configuration file at `$(pwd)/.fossa.yml`. This should work for both relative and absolute basedir paths.

## Testing plan

1. Create a test project somewhere (I have one at `/tmp/fossa-test-project`).
2. Set a configuration file that generates an obvious error when used (I set `server` to `localhost`, so the CLI throws a `HostCannotConnect`).
3. Run `fossa analyze PATH_TO_TEST_PROJECT` (`fossa analyze /tmp/fossa-test-project`) with both absolute and relative paths, and check that the expected configuration error occurs.

## Risks

1. Are there any other filepath forms I should test?
2. Does anyone have advice on how to integration test this?
5. Do we need an integration test for this? I'm ambivalent about writing one since it seems relatively high-effort and low-value.

## Checklist

### DIFF
--- a/src/App/Fossa/Config/Analyze.hs
+++ b/src/App/Fossa/Config/Analyze.hs
@@ -77,7 +77,7 @@ import Effect.Exec (
   Exec,
  )
 import Effect.Logger (Logger, Severity (SevDebug, SevInfo), logWarn)
-import Effect.ReadFS (ReadFS)
+import Effect.ReadFS (ReadFS, resolveDir)
 import Fossa.API.Types (ApiOpts)
 import Options.Applicative (
   Alternative (many),
@@ -252,10 +252,10 @@ loadConfig ::
   ) =>
   AnalyzeCliOpts ->
   m (Maybe ConfigFile)
-loadConfig AnalyzeCliOpts{commons = CommonOpts{optConfig}} = do
-  -- FIXME: We eventually want to use the basedir to inform the config file root
-  configRelBase <- sendIO getCurrentDir
-  resolveConfigFile configRelBase optConfig
+loadConfig AnalyzeCliOpts{analyzeBaseDir, commons = CommonOpts{optConfig}} = do
+  cwd <- sendIO getCurrentDir
+  configBaseDir <- resolveDir cwd (toText analyzeBaseDir)
+  resolveConfigFile configBaseDir optConfig
 
 mergeOpts ::
   ( Has Diagnostics sig m
--- a/src/App/Fossa/Config/Analyze.hs
+++ b/src/App/Fossa/Config/Analyze.hs
@@ -76,7 +76,7 @@ import Discovery.Filters (AllFilters (AllFilters), comboExclude, comboInclude)
 import Effect.Exec (
   Exec,
  )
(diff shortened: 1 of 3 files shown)


src/App/Fossa/Config/Analyze.hs
```diff
@@ -245,10 +245,10 @@ loadConfig ::
...
   AnalyzeCliOpts ->
   m (Maybe ConfigFile)
-loadConfig AnalyzeCliOpts{commons = CommonOpts{optConfig}} = do
-  -- FIXME: We eventually want to use the basedir to inform the config file root
-  configRelBase <- sendIO getCurrentDir
-  resolveConfigFile configRelBase optConfig
+loadConfig AnalyzeCliOpts{analyzeBaseDir, commons = CommonOpts{optConfig}} = do
+  cwd <- sendIO getCurrentDir
+  configBaseDir <- resolveDir cwd (toText analyzeBaseDir)
+  resolveConfigFile configBaseDir optConfig
```
meghfossa, line 251: [Optional] I think the behavior you have in this PR is correct and what ought to be - but this could be a breaking change for some customers. Let's add debug log saying looking for config file xyz path, so diagnostics is easier - if this turns out to be a breaking change.

scruffystuffs: This is definitely a breaking change, I remember seeing a bug about this in the early days of v2 config files.  I think we should consider falling back to the current directory for a config file if none is found at the basedir.

spatten: I think I agree with Wes — we should try to fall back to looking in the current directory if we don't find a config file in the working directory.

I'm not sure if we'd build it that way if were were just implementing this, but I think it's the right change to make in order to avoid breaking customers

spatten: I'm fine with that if @scruffystuffs  and @meghfossa are. Wes and Megh, what do you think?

scruffystuffs: Breaking is breaking.  We're almost certainly going to see bugs filed about this.  I know for a fact that a small number of users are using the wrong behavior, there were bug reports about the config file when we first launched it.

I agree with the change, and I'm willing to make it, AND I'm willing to drop the fallback idea.  My focus here is being proactive about reducing the bug report burden on both on-duty and the post-sales team.

I think the bug report burden here will be small, so I'm OK with this one.

spatten (approved):
:+1:

This may cause some support burden, but this PR makes the behaviour more correct and less confusing than it was before, so let's go with this and deal with the hopefully small number of problems that it causes as they arise.

liftM:
I've given post-sales a heads up at https://teamfossa.slack.com/archives/CBVH5QZ9C/p1648577473749529. Will automerge once tests pass.

PR #856, opened 2022-03-26
API Validation - empty string

saramaebee:
Currently when an API token is provided, it's accepted without validation. This introduces a check for a non-empty string and errors if string is empty

# Overview

This PR adds a check on the length of the ApiKey to make sure it isn't being instantiated with an empty string.

## Acceptance criteria

Running `FOSSA_API_KEY="" fossa analyze` should result in a warning that an empty string was provided.

## Testing plan

Run `fossa analyze` with an empty string as `FOSSA_API_KEY`

## Risks

Is there a better way to do this sort of check?

## References

- _[ANE-159](https://fossa.atlassian.net/browse/ANE-159): Warn when users provide non-sensible looking API Key_

## Checklist

- [ ] ~~If this PR introduced a user-visible change, I added documentation into `docs/`.~~
- [ ] ~~If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).~~

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Configuration: Users can now use `.fossa.yaml` as a configuration file name. Previously, only `.fossa.yml` was supported. ([#851](https://github.com/fossas/fossa-cli/pull/851))
+- Warns user when provided API key is an empty string ([#856](https://github.com/fossas/fossa-cli/pull/856))
 
 ## v3.1.7
 - fossa-deps: Fixes an archive uploading bug for vendor dependency by queuing archive builds individually. ([#826](https://github.com/fossas/fossa-cli/pull/826))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,10 +1,14 @@
-# Fossa CLI Changelog
+# FOSSA CLI Changelog
+
+## Unreleased
+
+- Configuration: Users can now use `.fossa.yaml` as a configuration file name. Previously, only `.fossa.yml` was supported. ([#851](https://github.com/fossas/fossa-cli/pull/851))
 
 ## v3.1.7
-- Fossa-deps: Fixes an archive uploading bug for vendor dependency by queuing archive builds individually. ([#826](https://github.com/fossas/fossa-cli/pull/826)) 
+- fossa-deps: Fixes an archive uploading bug for vendor dependency by queuing archive builds individually. ([#826](https://github.com/fossas/fossa-cli/pull/826))
 
 ## v3.1.6
-- Respects Go module replacement directives in the Go Mod Graph strategy. ([#841](https://github.com/fossas/fossa-cli/pull/841)) 
+- Respects Go module replacement directives in the Go Mod Graph strategy. ([#841](https://github.com/fossas/fossa-cli/pull/841))
 
 ## v3.1.5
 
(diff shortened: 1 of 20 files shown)


Changelog.md
```diff
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Configuration: Users can now use `.fossa.yaml` as a configuration file name. Previously, only `.fossa.yml` was supported. ([#851](https://github.com/fossas/fossa-cli/pull/851))
+- Warns user when provided API key is an empty string ([#856](https://github.com/fossas/fossa-cli/pull/856))
```
scruffystuffs, line 6:
```suggestion
- Configuration: Reports an error when provided API key is an empty string ([#856](https://github.com/fossas/fossa-cli/pull/856))
```

src/App/Fossa/Config/Common.hs
```diff
@@ -62,10 +62,10 @@ import Control.Effect.Diagnostics (
   (<||>),
  )
 import Control.Effect.Lift (Lift, sendIO)
-import Control.Timeout (Duration (Minutes))
+import Control.Timeout.Internal ( Duration(Minutes) )
```
scruffystuffs, line 65: Don't import this from `Internal`.  I really want to prevent VSCode from suggesting that, but for now we have to live with it.

scruffystuffs (approved):
Left some notes, but this is good.  Is this your first ever haskell PR?

src/App/Fossa/Config/Common.hs
```diff
@@ -184,7 +184,11 @@ validateApiKey maybeConfigFile EnvVars{envApiKey} CommonOpts{optAPIKey} = do
       optAPIKey
         <|> (maybeConfigFile >>= configApiKey)
         <|> envApiKey
-  pure $ ApiKey textkey
+  if Data.Text.null textkey
```
meghfossa, line 187: This ignores the case of `textKey` where the only character is space. 

Let's strip the text for spaces on the left-hand side, and right-hand side as well. 

```haskell
if Data.Text.null . Data.Text.strip $ textkey 
then fatalText "An empty FOSSA API key was specified. Please provide a non-empty FOSSA API key."
else pure $ ApiKey textkey
```

src/App/Fossa/Config/Common.hs
liftM: Can we `hlint` against `*.Internal` modules?

scruffystuffs: No, because we have to sometimes import internal modules.  There are certain ways of fully re-exporting symbols that don't get refined like this, (e.g. `Has` gets imported from effect modules, but is really from `Control.Algebra`), we just need to use those for the `Control.Timeout` module.

saramaebee:
> Left some notes, but this is good. Is this your first ever haskell PR?

Yes 🥺

src/App/Fossa/Config/Common.hs
scruffystuffs: Also, you imported `null` on line 68, so instead of `Data.Text.null`, you can just use `null`.  That may give you a problem if some other module has imported `null` somewhere else (it's a common name, I wouldn't be surprised), but if it compiles with no issue then it's fine.

src/App/Fossa/Config/Common.hs
saramaebee: Corrected on d68974e

src/App/Fossa/Config/Common.hs
saramaebee: It was yelling at me in VS Code so I put it, I don't think I would have had I been using vim, I typically opt for shortest possible naming

PR #859, opened 2022-03-28
If shrinking would orphan packages with predecessors, keep them orphans.

csasarak:
# Overview

This PR makes it so that dev dependencies and their transitive deps aren't reported as dependencies in fossa-cli output. It does this by changing one of our shrink functions to not promote dev dependencies to direct deps.

## Acceptance criteria

devDependencies won't be reported as dependencies.

## Testing plan

Currently with a manual test, we are still working out whether this is the right approach to solving this problem or not. 

On `master`, if you make an empty directory with this `package.json`:

```json
{
    "name": "foo",
    "version": "0.0.0",
    "private": true,
    "license": "UNLICENSED",
    "type": "module",
    "dependencies": {
        "underscore": "^1.13.2"
    },
    "devDependencies": {
        "chalk-pipe": "^5.1.1"
    }
}
```

Then run `npm install` and analyze you'll see that `chalk-pipe` as well as its dependencies are reported. The version on this branch will only report `underscore` as a dep.

## Risks

We need to be careful to make sure this isn't inadvertently breaking behavior for users of other tools, since the fix here happens after the individual analyzers build a graph in code that prunes all graphs.

## References

TBD

## Checklist


Changelog.md
```diff
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## Unreleased
```
meghfossa, line 3: Let's do a release!

test/GraphingSpec.hs
```diff
@@ -196,3 +264,117 @@ spec = do
...
+
+  getRootsOfSpec
+
+  shrinkRootsSpec
+
+  splitGraphSpec
+
+  stripRootSpec
+
+  describe "promoteToDirect" $ do
```
meghfossa, line 363: Why not move these to its own spec items as you have done for other ones?

csasarak: I thought they were maybe too small, but I've moved them now.

meghfossa (approved):

src/Graphing.hs
```diff
@@ -279,19 +294,7 @@ deeps = Graphing . AM.vertices . map Node
...
-
-    builtEdges :: [(res, res)]
-    builtEdges = map (bimap toDependency toDependency) (edgesFrom seed)
-
-    edgesFrom :: [dep] -> [(dep, dep)]
-    edgesFrom nodes = do
-      node <- nodes
-      let children = getDeps node
-      map (node,) children ++ edgesFrom children
+unfold = unfold' directs
```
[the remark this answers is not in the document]
csasarak: I'm actually not sure I'm seeing what is so bad here. If someone passes in an infinitely large seed list then it would never terminate. Likewise if you do an unfold with a function that has a code-path which might never produce a stopping condition. The main difference is that a graph cannot be generated lazily like a list can. I guess for me these scenarios feel expected rather than being a surprising foot-gun.

One thing that might help is to specify that the stopping condition for the unfold is producing an empty list, and that every output of it should get closer to producing `[]` when it is called again with that output as an input.

PR #860, opened 2022-03-28
Prevent VSCode from refining duration imports

scruffystuffs:
# Overview

VSCode really wants us to import `Duration` from `Control.Timeout.Internal` instead of `Control.Timeout`, where it's re-exported for API cleanliness.

This is from `App.Fossa.Config.Common`, but it occurs in multiple places.

image

We can fix this issue by moving `Duration` into `Control.Timeout`, since it didn't need to be in the internal module anyway.

## Acceptance criteria

- VSCode stops suggesting that you refine your imports when importing `Duration` from `Control.Timeout`.

### DIFF
--- a/src/Control/Timeout.hs
+++ b/src/Control/Timeout.hs
@@ -28,12 +28,37 @@ import Control.Effect.Lift (Has, Lift, sendIO)
 import Control.Monad (when)
 import Control.Timeout.Internal (
   Cancel (..),
-  Duration (..),
-  durationToMicro,
  )
 import Data.Functor (($>))
 
--- Whether a cancellable should cancel right now, or poll again.
+-- | Several time-based functions accept 'Int', which mmakes it difficult to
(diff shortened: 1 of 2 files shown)


src/Control/Timeout.hs
```diff
@@ -28,12 +28,37 @@ import Control.Effect.Lift (Has, Lift, sendIO)
 import Control.Monad (when)
 import Control.Timeout.Internal (
   Cancel (..),
-  Duration (..),
-  durationToMicro,
  )
 import Data.Functor (($>))
 
--- Whether a cancellable should cancel right now, or poll again.
+-- | Several time-based functions accept 'Int', which mmakes it difficult to
```
kitified, line 34:
```suggestion
-- | Several time-based functions accept 'Int', which makes it difficult to
```
Butterfly keyboard? 🥲

kitified (approved):

PR #861, opened 2022-03-29
Fixes odd verbiage in gradle docs

meghfossa:
Fixes verbiage in Gradle docs regarding test and dev configurations.

Reference: https://teamfossa.slack.com/archives/C0155DTGWB1/p1648512089147009


zlav (approved):
🙌

PR #862, opened 2022-03-29
add `resolvePath` util to `ReadFS`

scruffystuffs:
# Overview

This PR adds a function called `resolvePath`.  If you pass `resolvePath` an arbitrary `Filepath` it will:

- Confirm its existence in the filesystem
- Detect whether it's a file or directory (or neither)
- Detect whether the path is absolute or relative.

## Acceptance criteria

- Effect message handler is written well
- String conversion cleanup is acceptable
- Errors are clear and understandable

## Testing plan

No tests, we don't yet have mocking capabilities for `ReadFS`, and if we did, there's not too much we can do with `FileStatus` anyway.  This is also not yet in any code path.

## References

Needed to implement ANE-24

## Checklist

### DIFF
--- a/src/Effect/ReadFS.hs
+++ b/src/Effect/ReadFS.hs
@@ -89,7 +89,6 @@ import Path (
   File,
   Path,
   SomeBase (Abs),
-  fromAbsFile,
   fromSomeDir,
   fromSomeFile,
   parseSomeDir,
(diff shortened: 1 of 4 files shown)


src/Effect/ReadFS.hs
```diff
@@ -215,81 +275,91 @@ type Parser = Parsec Void Text
...
+    (extractIdentifier <$> Posix.getFileStatus (toString dir))
+      `catchingIO` FileReadError (toString dir)
     where
       extractIdentifier :: Posix.FileStatus -> DirID
       extractIdentifier status =
         let (CIno fileID) = Posix.fileID status
             (CDev devID) = Posix.deviceID status
          in DirID{dirFileID = toInteger fileID, dirDeviceID = toInteger devID}
-
+  ResolvePath path -> join <$> ((identifyPath path <$> Posix.getFileStatus path) `catchingIO` FileReadError path)
```
kitified, line 346: Does `Posix.getFileStatus` work on Windows? [Their docs](https://hackage.haskell.org/package/unix-2.7.2.2) imply the answer is no:
> The package is not supported under Windows.

If not, I think it's not urgent to fix since this isn't used anywhere, but we should have a plan for how we can support windows (or at least throw a nice error message)

scruffystuffs: This is actually from `unix-compat`.

https://github.com/fossas/fossa-cli/pull/862/files#diff-c84a9339a8d0b78e16abd42d8d1933c4138aa85b36da9102625415e726df075dR103

[docs](https://hackage.haskell.org/package/unix-compat-0.5.4/docs/System-PosixCompat-Files.html)

kitified: whoops 😅 carry on!

PR #863, opened 2022-03-29
Converts BuildWait to use API effects.

periodic:
# Overview

This rewrites `BuildWait` to use API effects.  This code is used by the `report`, `test` and `container test` commands.

As part of this I made a few changes.  Of these, the last deserves a longer explanation, for that, see the module comment on `Test.MockApi`.

- `ScotlandYard` was moved to be an internal module of the `FossaApiClient` carrier.  It wasn't used anywhere else and this keeps the details of the proxy hidden from the callers.
- I made the time out configuration into a reader value that is derived from the surrounding config.  This helps with testing and will make it easier to make these into config options in the future.
- Code for the ScotlandYard requests was using `Text` for custom locators so I converted it to use `SrcLib.Locator`
- I reworked API-effect mocking so that we can express changes in effects across calls.

## Acceptance criteria

- All API calls in `BuildWait` go through the API Effects
- Call-sites have been updated as necessary.
- Tests are added where reasonable.

## Testing plan

I added a bunch of unit tests, mostly to test the mocking framework.  The mocking framework itself was tested mostly by letting the tests fail and then checking that they behaved appropriately.  It may be reaching the complexity that we should have unit tests for it as well.

I also checked by running uploads and generating reports from the command line.  They should produce the same output as the release command.

## Risks

Most of this change was moving code around and changing call sites and signatures.  There should not be any real changes in business logic, so if it looks like anything may change behavior, please flag it.

## References

- _[ANE-131](https://fossa.atlassian.net/browse/ANE-131?atlOrigin=eyJpIjoiNzRhNmEyNzEwNTE2NGE1NmJjMGYxMjFmMzlhZTYyODEiLCJwIjoiaiJ9): Convert App.Fossa.API.BuildWait to use API Effects_

## Checklist


periodic:
Was assigned to Scott randomly, but I probably need someone more experienced with Haskell like @scruffystuffs  or @meghfossa for something this complicated.

Feel free to make your first comment to split this up.  Most of the complexity is in `Test.MockApi` so I figured I could get away with just leaving in the other refactoring pieces.

test/Test/MockApi.hs
```diff
@@ -0,0 +1,227 @@
...
+  AssertAllSatisfied :: MockApi m ()
+
+-- | Specifies how to handle repetition of a request
+data ExpectationRepetition = Once | Always deriving (Eq, Ord, Show)
+
+-- | The result of a call can be either a value or a diagnostic failure
+data ApiResult a = Return a | Die Text deriving (Eq, Ord, Show)
+
+-- | An expectation of an API call made up of the request and response.
+data ApiExpectation where
```
scruffystuffs, line 77: Can we write a short document explaining the relationship between these mocking structures?

The doc I think we need is this: if you had only laid out the types and the document, someone else could have independently written all of the utilities and the effect handler basically the same as you've done here.

scruffystuffs: Also, I'd like to understand how the mocking works by only reading docs and comments.  We have a precedent for documenting complex solutions like this [in our contributing docs](https://github.com/fossas/fossa-cli/tree/master/docs/contributing).

scruffystuffs (commented):
I took a look at this, I left some comments.  Mostly, I want to sit with this solution in my head for a day or two.  But at first glance, I think I'm pretty happy with this direction for mocking.

test/App/Fossa/API/BuildWaitSpec.hs
```diff
@@ -0,0 +1,181 @@
...
+runWithConfigAndTimeout ::
+  Has (Lift IO) sig m =>
+  (Cancel -> ReaderC WaitConfig m a) ->
+  m ()
+runWithConfigAndTimeout =
+  void
+    . runReader waitConfig
+    . timeout' (Seconds 1)
+
+spec :: Spec
```
scruffystuffs, line 34: Can we put the helper functions _above_ the spec?  I find it's easier to understand tests when they're written in this order:
1. Simple-data-building functions (e.g.: `mkFixture a b = Foo (Bar b) . Just $ Baz a`)
2. Simple data (expectations and inputs)
3. Collections of simple data
4. Complex expectations and inputs (rare in haskell, but it does happen)
5. Trivial test helpers
6. Non-trivial test helpers
7. The actual tests

test/Test/MockApi.hs
```diff
@@ -0,0 +1,227 @@
...
+        Nothing ->
+          assertUnexpectedCall req
+
+runMockApi ::
+  ( Has (Lift IO) sig m
+  ) =>
+  MockApiC m a ->
+  m a
+runMockApi =
+  (fmap snd) . runState [] . runMockApiC
```
scruffystuffs, line 227: [nit]: 
```suggestion
  evalState [] . runMockApiC
```

scruffystuffs (commented):
Left some comments, we talked offline about hiding `WaitConfig` from the developer in prod code paths.  I'll wait on that and the doc before approving.  Structurally, this is fantastic.

test/App/Fossa/API/BuildWaitSpec.hs
```diff
@@ -0,0 +1,186 @@
...
+import Test.Hspec (Spec, describe)
+import Test.MockApi (MockApi, alwaysReturns, returnsOnce)
+
+runWithTimeout ::
+  Has (Lift IO) sig m =>
+  (Cancel -> m a) ->
+  m ()
+runWithTimeout =
+  void
+    . timeout' (Seconds 1)
```
scruffystuffs, line 30: This is small enough to one-line
```suggestion
runWithTimeout = void . timeout' (Seconds 1)
```

test/App/Fossa/API/BuildWaitSpec.hs
periodic: Hmmm, I felt like maybe they weren't the interesting point and the names were descriptive enough.  I sometimes hate having to scroll through lots of data set up just to find the tests themselves.  It's sort of like having `where` blocks on a function.  Easy enough to flip, but it feels like burying the lede.

test/Test/MockApi.hs
```diff
@@ -0,0 +1,227 @@
...
+data MockApi (m :: Type -> Type) a where
+  MockApiOnce :: FossaApiClientF a -> a -> MockApi m ()
+  MockApiAlways :: FossaApiClientF a -> a -> MockApi m ()
+  MockApiFails :: FossaApiClientF a -> Text -> MockApi m ()
+  MockApiRunExpectations :: FossaApiClientF a -> MockApi m (Maybe (ApiResult a))
+  AssertUnexpectedCall :: FossaApiClientF a -> MockApi m a
+  AssertAllSatisfied :: MockApi m ()
+
+-- | Specifies how to handle repetition of a request
+data ExpectationRepetition = Once | Always deriving (Eq, Ord, Show)
```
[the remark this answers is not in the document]
periodic: More lines almost always make things more readable, imo.

test/Test/MockApi.hs
```diff
@@ -0,0 +1,227 @@
...
+
+-- | The result of a call can be either a value or a diagnostic failure
+data ApiResult a = Return a | Die Text deriving (Eq, Ord, Show)
```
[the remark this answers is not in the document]
periodic: I did this because I was originally thinking there might be more cases, but these were the only two I had so far.  I could see some more specific ones like timeouts, but it seems most of our errors are communicated via diagnostic messages and not by using another exception handling effect.

I sort of like your second example there.  Let me try that.  Also, fortunately, `Either Text` is a pretty common way to express errors, so that's easy enough to read.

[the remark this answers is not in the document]
periodic: Did this.  I'm not sure it reads better because pattern matching becomes wordy like `ApiResult (Left (ApiFail msg))` instead of just `Die msg`. I'm not too particular though.

It would be cool if `Either` were a typeclass that could be derived instead.

test/App/Fossa/API/BuildWaitSpec.hs
periodic: Interesting.  I view the most important part of tests to be the description of behavior and the high-level steps taken to test it.  A user can then dig into the data if they want more information, but they probably don't care too much about the details of `expectGetApiOpts`.  Maybe I'm thinking that by naming these expectations they are clear enough that the details of setting the expectation doesn't have to be spelled out and the noise of selecting fixtures can be pushed down.

test/Test/MockApi.hs
scruffystuffs: There's no need to duplicate the info, the missing info is my bigger concern.

test/App/Fossa/API/BuildWaitSpec.hs
periodic: Huh, I come at it from a totally different perspective.  I'm more behavior-driven-design oriented.

I first want to check that the right things are being tested, then I have the context to look at the data and make sure it's the right data.  Without context the data doesn't mean much to me.  Let's say one of the data set-up functions has an empty list.  Is that correct or not?  I don't feel like I have the context to evaluate the data until I've read the tests.  The tests show me intention, the data doesn't.

Is this just the case that I read tests in the opposite order you do?

This may also come from spending most of my time on tests reviewing them.  When I review tests the most important thing to me is that the author has good coverage of tests over the behavior of the code.  If something is missing or the tests are messy then I send it back without looking too much at the details.  Once we have the right tests then I can start to look at the details to see how they are setting things up and make sure it matches what they say they are doing.

Also, sometimes to understand what code does I'll go browse the test cases because they should describe the behavior.

periodic:
Changes should all be made other than the docs one and the code order one.

periodic:
Go for it, @meghfossa.  @scruffystuffs brought up that it could use some more docs, so I'd appreciate if you flag anything that takes you a while to understand and could use more docs so I can get a better idea of what's missing.

test/Test/MockApi.hs
scruffystuffs: Got it, I can be more specific.  I bet I was having the inverse problem trying to express what was missing.

test/App/Fossa/API/BuildWaitSpec.hs
scruffystuffs: Frankly, these are both valid ways to read tests, so I'm not going to keep blocking you on this.  I also don't think this is something to standardize on, since it's clearly a very individual-perspective topic.

scruffystuffs: It's interesting that we read these in polar opposite ways though.  I never would have expected that.

test/Test/MockApi.hs
scruffystuffs: Few things I'd like to see added:

- How does `handleRequest` work, in the abstract?
- How are we hijacking the `FossaApiClient` effect?

The ideal audience for effects shenanigans is almost always "haskell dev that can use effects in basic forms, but may not understand how to write them or how they work in depth, and doesn't necessarily need to learn yet".  So, for the effect rewriting, it's worth pointing out that effects only have meaning because of their handlers, and `runApiWithMock` is using a completely different handler that _requires_ a `MockApi` effect, stuff like that.  I only understand that stuff because I've been diving into the code and have experience with the effect system, but we can make it easier for others to understand this, even if the code is the only way to get ALL of the details.

periodic: Okay, I think maybe I got that all in a doc.

test/Test/MockApi.hs
```diff
@@ -0,0 +1,233 @@
...
+  , Has MockApi sig m
+  ) =>
+  FossaApiClientMockC m a ->
+  m a
+runApiWithMock f = do
+  result <- interpret runRequest f
+  assertAllSatisfied
+  pure result
+  where
+    runRequest req = do
```
meghfossa, line 220: [Optional] [Minor] Let's add types for runRequest function in this where bock, so it's easier to read

test/Test/MockApi.hs
```diff
@@ -0,0 +1,233 @@
...
+handleRequest req = do
+  expectations <- get
+  case testExpectations expectations of
+    Just (resp, expectations') -> do
+      put expectations'
+      pure (Just resp)
+    Nothing ->
+      pure Nothing
+  where
+    testExpectations [] = Nothing
```
meghfossa, line 196: [Optional] [Minor] prefer types annot. for `testExpectations`

docs/contributing/api-mocking.md
```diff
@@ -0,0 +1,30 @@
...
+
+Expectations are checked and matched in the order in which they are added.
+
+Any unsatisfied expectations at the end of the test will trigger an assertion error.
+
+## MockApi Effect
+
+The `MockApi` effect provides the mechanisms to access the expectations.  It can add expectations, run a request against the expectations or trigger assertions.  These should not be used directly.  The exported helpers should be used instead.
+
+Internally the `MockApi` keeps track of the expectations and provides an interface to them that `FossaApiClientMockC` can use to implement `FossaApiClient` effects against the expectations.
```
meghfossa, line 30: Would like a tiny example of how to test hypothetical ApiCall.

meghfossa (commented):
did quick glance - will defer to wes for approval, but looks ~good~ great to me  🚀

docs/contributing/api-mocking.md
periodic: Maybe this is resolved.  I added some examples and more explanation since you left this comment.  I must have forgotten to push it.

scruffystuffs (approved):
Looks good, we'll do more docs work later.

periodic:
:tada:

PR #866, opened 2022-03-31
Wraps organizationId in a new type

periodic:
# Overview

I noticed that `Organization.organizationId` was just a bare `Int` and that made me uneasy, so I wrapped it.  This turned out to be particularly easy because the only thing we ever do with organization IDs seems to be to `show` them and it's easy to make that a pass-through.

## Acceptance criteria

Organization IDs are never accessed as a raw type.

## Testing plan

1. Ran all the unit tests.
2. Did an analysis upload from a local project
3. Pulled a report from that project.

## Risks

There's always a risk that something was using the organization IDs in a weird way, but the type system should catch that.

## Checklist

### DIFF
--- a/src/Fossa/API/Types.hs
+++ b/src/Fossa/API/Types.hs
@@ -258,7 +258,7 @@ data Organization = Organization
 
 instance FromJSON Organization where
   parseJSON = withObject "Organization" $ \obj ->
-    Organization <$> (OrgId <$> obj .: "organizationId")
+    Organization <$> obj .: "organizationId"
       <*> obj .:? "usesSAML" .!= False
       <*> obj .:? "supportsCliLicenseScanning" .!= False
 
--- a/src/Fossa/API/Types.hs
+++ b/src/Fossa/API/Types.hs
@@ -238,16 +238,10 @@ instance Pretty Issues where
   pretty = renderedIssues
 
 newtype OrgId = OrgId Int
-  deriving (Eq, Ord)
+  deriving (Eq, Ord, FromJSON, ToJSON)
 
 instance Show OrgId where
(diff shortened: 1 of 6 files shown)


src/Fossa/API/Types.hs
```diff
@@ -236,8 +237,20 @@ instance ToJSON IssueRule where
...
+  deriving (Eq, Ord)
+
+instance Show OrgId where
+  show (OrgId orgId) = show orgId
+
+instance FromJSON OrgId where
+  parseJSON = fmap OrgId . parseJSON
+
+instance ToJSON OrgId where
+  toJSON (OrgId orgId) = toJSON orgId
```
scruffystuffs, line 250: You can probably use `deriving (FromJSON, ToJSON)`  I would test it out, though, to make sure it derives how you expect it.  I _think_ HLS will tell you which extension you need to make this work, but we have other examples in the codebase.  Maybe grep for `deriving.*FromJSON`

scruffystuffs: I think it's `GeneralizedNewtypeDeriving`, docs here: https://github.com/fossas/fossa-cli/blob/6c6fda19616b8141a1bd2a0fe39aeb48abcff068/src/Discovery/Archive.hs#L59-L67

periodic: Yep, perfect.

```
Prelude Fossa.API.Types Data.Aeson> encode (OrgId 42)
"42"
Prelude Fossa.API.Types Data.Aeson> decode "42" :: Maybe OrgId
Just 42
```

scruffystuffs: Nice.

PR #868, opened 2022-03-31
Unarchive vendored deps + Handle empty directories

scruffystuffs:
# Overview

This PR deals with 2 tickets, since their logic is very closely intertwined.

Delivers ANE-24

Scenario: a user runs with `--experimental-native-license-scan` and the following `fossa-deps` file:

```yaml
vendored-dependencies:
- name: foo
  path: /path/to/foo.zip
```

This PR accepts this form by unarchiving the `zip` file (or other archive that we support) to a temp directory, and running the scan from there.  We also try to report errors related to the archive using the path to the unzipped archive, rather than the temp directory (which is why we have `ScannableArchive`).

Delivers ANE-158

Scenario:  same as above, with this file instead:

```yaml
vendored-dependencies:
- name: foo
  path: /path/to/foo
```

If `/path/to/foo` is an empty directory (i.e. walking the tree downwards produces 0 files), we need to report sensible errors to the user.

## Acceptance criteria

- [ ] Happy path works, pointing to non-empty directory
- [ ] Happy path works, pointing to well-formed archive with the correct file extension.
- [ ] Good errors during known failure cases 

## Testing plan

- [ ] Both happy paths above
- [ ] Empty directory
- [ ] Archive with no file extension
- [ ] Archive with incorrect file extension (i.e. rename `foo.zip` to `foo.tar.gz`)
  - Note that this also works for ANY file.  If we don't recognize the extension, or the format doesn't match the extension we see, we should handle that cleanly.
- [ ] Archive with no files
- [ ] Unsupported archive format with valid extension
  - The formats we support are listed [here](https://github.com/fossas/fossa-cli/blob/6c6fda19616b8141a1bd2a0fe39aeb48abcff068/src/Discovery/Archive.hs#L59-L67).

Automated tests are not possible since we can't mock `Exec` and `ReadFS` in tests (YET!).

## Risks

None, this feature is about to be beta, and touches no non-beta features or code.

## Checklist

### DIFF
--- a/src/App/Fossa/LicenseScanner.hs
+++ b/src/App/Fossa/LicenseScanner.hs
@@ -43,15 +43,21 @@ import Srclib.Types (
   Locator (..),
  )
 
-data NoSuccessfulScans = NoSuccessfulScans
+data LicenseScanErr
+  = NoSuccessfulScans
+  | NoLicenseResults (Path Abs Dir)
+  | EmptyDirectory (Path Abs Dir)
+  | EmptyArchive (Path Abs File)
+  | UnsupportedArchive (Path Abs File)
 
-instance ToDiagnostic NoSuccessfulScans where
-  renderDiagnostic _ = "No native license scans were successful"
-
-newtype NoLicenseResults = NoLicenseResults (Path Abs Dir)
-
-instance ToDiagnostic NoLicenseResults where
+instance ToDiagnostic LicenseScanErr where
+  renderDiagnostic NoSuccessfulScans = "No native license scans were successful"
   renderDiagnostic (NoLicenseResults path) = "No license results found after scanning directory: " <> pretty (toText path)
+  renderDiagnostic (EmptyDirectory path) = "vendored-dependencies path has no files and cannot be scanned: " <> pretty (toString path)
+  renderDiagnostic (EmptyArchive path) = "vendored-dependencies archive contains no files and cannot be scanned: " <> pretty (toString path)
+  renderDiagnostic (UnsupportedArchive path) = case fileExtension path of
+    Just ext -> "fossa-cli does not support archives of type " <> squotes (pretty ext) <> ": " <> pretty (toString path)
+    Nothing -> "fossa-cli does not support archives without file extensions: " <> pretty (toString path)
 
 runLicenseScanOnDir ::
   ( Has Diagnostics sig m
--- a/src/App/Fossa/LicenseScanner.hs
+++ b/src/App/Fossa/LicenseScanner.hs
@@ -127,7 +127,7 @@ scanVendoredDep ::
   m (NonEmpty LicenseUnit)
 scanVendoredDep baseDir VendoredDependency{..} = context "Scanning vendored deps for license data" $ do
   logSticky $ "License Scanning '" <> vendoredName <> "' at '" <> vendoredPath <> "'"
-  scanPath <- resolvePath' $ toString vendoredPath
+  scanPath <- resolvePath' baseDir $ toString vendoredPath
   case scanPath of
     SomeFile (Abs path) -> scanArchive $ ScannableArchive path
     SomeFile (Rel path) -> scanArchive . ScannableArchive $ baseDir </> path
--- a/src/App/Fossa/LicenseScanner.hs
+++ b/src/App/Fossa/LicenseScanner.hs
@@ -70,6 +70,8 @@ instance ToDiagnostic LicenseScanErr where
     Just ext -> "fossa-cli does not support archives of type " <> squotes (pretty ext) <> ": " <> pretty (toString path)
     Nothing -> "fossa-cli does not support archives without file extensions: " <> pretty (toString path)
 
+newtype ScannableArchive = ScannableArchive {scanFile :: Path Abs File} deriving (Eq, Ord, Show)
+
 runLicenseScanOnDir ::
   ( Has Diagnostics sig m
   , Has (Lift IO) sig m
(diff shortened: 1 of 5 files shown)


src/App/Fossa/LicenseScanner.hs
```diff
@@ -99,22 +119,80 @@ scanVendoredDep ::
...
 scanVendoredDep baseDir VendoredDependency{..} = context "Scanning vendored deps for license data" $ do
   logSticky $ "License Scanning '" <> vendoredName <> "' at '" <> vendoredPath <> "'"
-  vendoredDepDir <- case parseRelDir $ toString vendoredPath of
-    Left err -> fatalText ("Error constructing scan dir for vendored scan: " <> showT err)
-    Right val -> pure val
-  let cliScanDir = baseDir </> vendoredDepDir
+  scanPath <- resolvePath' $ toString vendoredPath
+  case scanPath of
+    SomeFile (Abs path) -> scanArchive $ ScannableArchive path
+    SomeFile (Rel path) -> scanArchive . ScannableArchive $ baseDir </> path
```
spatten, line 132: This doesn't work when you run `fossa analyze` from another directory.

E.g. if you run it like this:

```
cabal run fossa -- analyze --experimental-native-license-scan --debug -e http://localhost:9578 ~/fossa/license-scan-dirs/krb5-1.18-compressed
```

You get error messages like this:

```
Error reading file ./krb5-1.18.2.tar.gz:
              ./krb5-1.18.2.tar.gz: getFileStatus: does not exist (No such file or directory)
```

Happy to demo if this is hard to recreate

scruffystuffs: Relative file paths should be based on the directory containing `fossa-deps.yml`.

scruffystuffs: Are we picking up that file from the current directory, or the base directory?

spatten: It looks like we're finding the `fossa-deps.yml` in the scan directory, but looking for the `.tar.gz` file in the current working directory.

In ~/fossa/license-scan-dirs/krb5-1.18-compressed I have two files: `fossa-deps.yml` and `krb5-1.18.2.tar.gz`

`fossa-deps.yml` looks like this:

```yaml
vendored-dependencies:
  - name: krb5.tar.gz
    path: ./krb5-1.18.2.tar.gz
    version: 1.18.2.1
```

I've tried with and without the `./` in the `path`, but we get the problem either way.

```
# cabal run fossa -- analyze --experimental-native-license-scan --debug -e http://localhost:9578 ~/fossa/license-scan-dirs/krb5-1.18-compressed
Up to date                                                                  
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/krb5-1.18-compressed/"
[DEBUG] License Scanning 'krb5.tar.gz' at 'krb5-1.18.2.tar.gz'                                                                                          
[ERROR] ----------                                                          
  An issue occurred                                                                                                                                     
                                                                            
  >>> Relevant errors                                                       
                                                                            
    Error                                                                   
                                                                            
      No native license scans were successful                               
                                                                                                                                                        
      Traceback:                                                            
        - Converting fossa-deps to partial API payload                                                                                                  
        - fossa-deps
        - fossa-analyze

  >>> Possibly-related warnings

    Warning

      >>> Relevant errors

        Error

          Error reading file krb5-1.18.2.tar.gz:
              krb5-1.18.2.tar.gz: getFileStatus: does not exist (No such file or directory)

          Traceback:
            - Scanning vendored deps for license data
            - Processing vendored dependency
            - Converting fossa-deps to partial API payload
            - fossa-deps
            - fossa-analyze
```

spatten: Oh, and I manually added the change from https://github.com/fossas/fossa-cli/pull/865 and set `    enable_cli_license_scan: true,` in my local version of Core

spatten: I think the problem is occuring on line 129. That's just using a relative path, right?

```
scanPath <- resolvePath' $ toString vendoredPath
```

scruffystuffs: Yep, that's it.

scruffystuffs: Can you try running from the current dir for the rest of the tests?  That's an easy fix, but it'll take a while before I can push it.

scruffystuffs: Fixed in 9e02916

spatten (approved):
This looks good.

The only nit I have is that the error you get when unarchiving fails is a bit inscrutable.

E.g. I changed the extension of a file from `.tar.gz` to `.zip`, and then got this output:

```
# cabal run fossa -- analyze --experimental-native-license-scan --debug -e http://localhost:9578 ~/fossa/license-scan-dirs/krb5-1.18-compressed
Up to date                                                                                     
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/krb5-1.18-compressed/"
[DEBUG] License Scanning 'krb5.tar.gz' at 'krb5-1.18.zip'                                      
[ERROR] ----------                                                                             
  An issue occurred                                                                            
                                                                                               
  >>> Relevant errors                                                                          
                                                                                               
    Error                                                                                      
                                               
      No native license scans were successful  
                                                                                               
      Traceback:                                                                               
        - Converting fossa-deps to partial API payload                                         
        - fossa-deps                                                                           
        - fossa-analyze                                                                                                                                                                       
                                               
  >>> Possibly-related warnings                                                                
                                                                                               
    Warning                                                                                                                                                                                   
                                                                                               
      >>> Relevant errors                                                                      
                                                                                               
        Error                                  
                                                                                               
          caught exception: Parsing of archive structure failed:                               
          Cannot locate end of central directory              
          in "/Users/scott/fossa/license-scan-dirs/krb5-1.18-compressed/krb5-1.18.zip"         
                                                                                               
          Traceback:                                                                                                                                                                          
            - withArchive                                                                      
            - Scanning vendored deps for license data                                          
            - Processing vendored dependency                                                   
            - Converting fossa-deps to partial API payload                                     
            - fossa-deps                                                                       
            - fossa-analyze                                      
```

If it's easy to make it a bit more obvious that unzipping failed, then that would be a nice add. If not, then let's ship this as is

scruffystuffs:
That's definitely out-of-scope, it's in the pre-existing archive work, but I'll see if it's an easy add.

scruffystuffs:
@spatten I added better errors, but it's late, so I'm not going to test it yet.  Would you mind testing out the new error for failed archive unpacking?

spatten:
> @spatten I added better errors, but it's late, so I'm not going to test it yet. Would you mind testing out the new error for failed archive unpacking?

Nice! I've tested and I can confirm that it works and that it's going to be much easier for the user to figure out why things failed.

PR #870, opened 2022-04-01
Wip/grapher update

csasarak:
# Overview

In some package-lock.json peerDependencies can be used to form cycles with other dependencies. With no way to stop processing when this happens we can go into an infinite loop. 

## Acceptance criteria

NPM package-lock.json analysis should stop even if there's a cycle because of peerDependency processing. 

## Testing plan

I've added a unit test that replicates the situation. 

I've also manually tested on the project that we first noticed the problem on. 

## Risks

Please pay special attention to what conditions the graphing function will stop under. 

I think the risk is fairly low since, the code that checks if we've handled a dependency and stops is localized to just `addPeerDeps`. I didn't guard the full `buildGraph`function because it depends on being able to process dependencies multiple times. 

Consider the case where we first encounter a dep in a `requires` block. `buildGraph` will add a vertex for the dep, but location information isn't added because it doesn't have it at that time. If `buildGraph` queries the graph to see if it's handled that dependency later when processing its full dependency object it looks like it's been handled already. Location/environment info never makes it into the graph. I tried a few approaches to handling this but couldn't get all the tests to work, limiting the stop condition checking to just `addPeerDeps` seemed more reasonable for now.

## Checklist

### DIFF
--- a/src/Effect/Grapher.hs
+++ b/src/Effect/Grapher.hs
@@ -29,7 +29,8 @@ module Effect.Grapher (
 
   -- * Re-exports
   module X,
-hasDep) where
+  hasVertex,
+) where
 
 import Control.Algebra as X
 import Control.Carrier.Diagnostics (ToDiagnostic (..))
@@ -49,7 +50,7 @@ data SGrapher ty k where
   Direct :: ty -> SGrapher ty ()
   Edge :: ty -> ty -> SGrapher ty ()
   Deep :: ty -> SGrapher ty ()
-  HasDep :: ty -> SGrapher ty Bool
+  HasVertex :: ty -> SGrapher ty Bool
 
 type Grapher ty = Simple (SGrapher ty)
 
@@ -62,8 +63,8 @@ edge parent child = sendSimple (Edge parent child)
 deep :: Has (Grapher ty) sig m => ty -> m ()
 deep = sendSimple . Deep
 
-hasDep :: Has (Grapher ty) sig m => ty -> m Bool
-hasDep = sendSimple . HasDep
+hasVertex :: Has (Grapher ty) sig m => ty -> m Bool
+hasVertex = sendSimple . HasVertex
 
 evalGrapher :: (Ord ty, Algebra sig m) => GrapherC ty m a -> m (G.Graphing ty)
 evalGrapher = fmap fst . runGrapher
@@ -75,7 +76,7 @@ runGrapher = interpretState G.empty $ \case
   Direct ty -> modify (G.direct ty <>)
   Edge parent child -> modify (G.edge parent child <>)
   Deep n -> modify (G.deep n <>)
-  HasDep ty -> gets (G.hasVertex ty)
+  HasVertex ty -> gets (G.hasVertex ty)
 
 ----- Labeling
 
--- a/src/Effect/Grapher.hs
+++ b/src/Effect/Grapher.hs
@@ -12,8 +12,6 @@ module Effect.Grapher (
   edge,
   deep,
   evalGrapher,
-  hasEdge,
-  hasVertex,
   runGrapher,
 
   -- * Labeling
@@ -51,8 +49,6 @@ data SGrapher ty k where
   Direct :: ty -> SGrapher ty ()
   Edge :: ty -> ty -> SGrapher ty ()
   Deep :: ty -> SGrapher ty ()
-  HasVertex :: ty -> SGrapher ty Bool
-  HasEdge :: ty -> ty -> SGrapher ty Bool
 
 type Grapher ty = Simple (SGrapher ty)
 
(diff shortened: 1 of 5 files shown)


test/Node/PackageLockSpec.hs
```diff
@@ -329,6 +373,26 @@ buildGraphSpec testDir =
         ]
         graph
 
+    -- What we're really trying to test here is that buildGraph finishes its
+    -- work in a reasonable amount of time. If buildGraph starts to recurse
+    -- infinitely, eventually that thread will be forced to stop and we'll
+    -- signal a failure. This style of test is prone to be flakey, so
+    -- it may need to be removed/turned off in the future.
+    it "Should avoid recursing into peerDep cycles" $
```
meghfossa, line 381: I'm presuming that without this PR change - this case would be stuck in an infinite loop, or _just_ extremely poor performance?

If it's the former then I'm not in favor of adding threading/timing into this test case. I know this is the pragmatic approach, but I would prefer if we have more of a pure unit test. Since your test case is not changing (i.e. `peerDepCycle`), I rather that we just assert the graph's correctness.

meghfossa: sigh! Letting the unit test run infinitely is an a-ok signal of failure (I'm personally a-ok with this) but, it is not ideal (hoard up CI). Another approach would be to introduce some form of a `MaybeCircuitBreaker` into the graphing effect - but this seems out of scope for this PR. Can you add this as a tech debt (e.g. Identify a better way to test/break infinite loops in graphing effect) into our backlog?

csasarak: ok, I will definitely add that and reference this conversation. Thanks!

src/Effect/Grapher.hs
```diff
@@ -29,6 +29,7 @@ module Effect.Grapher (
 
   -- * Re-exports
   module X,
+  hasVertex,
```
scruffystuffs, line 32: [nit]: We should probably move this import up to a more relevant section.

src/Effect/Grapher.hs
```diff
@@ -49,6 +50,7 @@ data SGrapher ty k where
   Direct :: ty -> SGrapher ty ()
   Edge :: ty -> ty -> SGrapher ty ()
   Deep :: ty -> SGrapher ty ()
+  HasVertex :: ty -> SGrapher ty Bool
```
scruffystuffs, line 53: Can we also add a `HasEdge` helper here?  I think it make sense to do both, and it may help with other strategies down the line.

test/Node/PackageLockSpec.hs
scruffystuffs: The long term solution is to form the graph early, in a way that does not infinitely recurse (the `dependencies` field is recursive, but only because the data is physically nested, so the recursion there is not infinite).  Then, if needed, we can check for cycles in the graph using existing `AdjacencyMap` utils.

scruffystuffs: Data-traversal recursion without jumping is basically the only long term solution here.  This prevents us from even having to test for cycles.

scruffystuffs (approved):
Looks good.  If the test is flaky, we'll remove it.  If the test is flaky AND we miss some cycle due to the test missing, we'll come back to this tactic later.

test/Node/PackageLockSpec.hs
csasarak: I'm wondering now if it was a mistake for `addPeerDeps` to recurse at all.

If we can assume that all peer dep versions are also represented in the `dependencies` object, couldn't `addPeerDeps` just make edges and the rest of the information to be added when the top-level traversal gets to it? I'm already making the above assumption in `addPeerDeps` because `addPeerDeps` looks up the corresponding dependency object in order to recurse into it.

If the above works, we don't need to look up whether a vertex has been added and the peer deps support becomes less invasive.

scruffystuffs: If you can think of a solution that doesn't use jumping recursion for peer deps, that's worth doing now.

src/Effect/Grapher.hs
csasarak: I've gone ahead and added this.

src/Graphing.hs
```diff
@@ -209,6 +211,12 @@ shrinkRoots (Graphing gr) = Graphing $ foldr AME.shrinkSingle gr currentDirect
     currentDirect :: [Node ty]
     currentDirect = Set.toList $ AM.postSet Root gr
 
+hasVertex :: Ord ty => ty -> Graphing ty -> Bool
```
meghfossa, line 214: [Optional] Avoid adding dead code, unless required by acceptance criteria or subsequent PR.

src/Strategy/Node/Npm/PackageLock.hs
```diff
@@ -6,6 +6,7 @@ module Strategy.Node.Npm.PackageLock (
   PkgLockJson (..),
   PkgLockDependency (..),
   NpmResolved (..),
+  PkgLockPackage (..),
```
meghfossa, line 9: [Minor] Let's add a unit test to signify that peerCycleDep is a known and covered case.

meghfossa (approved):
🚀

src/Strategy/Node/Npm/PackageLock.hs
csasarak: I think this could be good, but if I wrote it it would probably be similar to the `race` thing I had in an earlier revision. I don't really like the idea of infinite recursion being the failure case because of its effect on CI. 

Do you think it's worth bringing back a test case like this? It seemed a bit less necessary now since we've removed the possibility by not recursing at all for peer dependencies.

src/Graphing.hs
csasarak: Wes and I talked about this, and we decided to leave it in for future use. You're definitely right to flag this though.

PR #874, opened 2022-04-05
Periodic/api effects report

periodic:
# Overview

Converts the `report` command to use API effects.  This is really just the conversion of one endpoint, `getAttribution`.

Two other small things that were included

1. Made a `fetchReport` function that is exported from `Report` so it can be tested without IO.
2. I noticed we had two sets of logic for rendering locators, so I merged them.

## Acceptance criteria

- All requests made by the `report` command are done through the `FossaApiClient` effect.
- Unit tests added where appropriate.

## Testing plan

I added unit tests for the basic reporting logic.

Otherwise I tested the reports by uploading analysis and then downloading a report using my local Core instance a test project.  I verified that this worked as it did with release.

## Risks

The riskiest thing here was probably changing the locator rendering.  It seems to work fine locally, so I think I got it correct.  There was a tiny difference where one of them would render a trailing `$` when there was no version, but the other wouldn't.  Core seems to treat these the same.

## References

- [ANE-133](https://fossa.atlassian.net/browse/ANE-133): Convert App.Fossa.Report to use API Effects

## Checklist


spatten (approved):
Sorry this took so long. I took the opportunity to dig into how the Reader effect works so I did a lot of reading.

This looks good to me!

PR #875, opened 2022-04-05
feat: Resolve vendored podspecs to git locators

liftM:
# Overview

This PR adds the ability to resolve vendored local podspecs into their source Git repositories. This is desirable because the reason why the podspecs are vendored in the first place is because they're usually unavailable in the Cocoapods registry. It does this by shelling out to `pod ipc spec`.

It also refactors the existing Podfile.lock tactic, and fixes a random typo I came across.

## Remaining work

- [x] This is really slow. I need to make it faster.
  - Maybe parallelize the traversal?
  - It looks like many nodes are being re-traversed due to `gtraverse` - cache or dedupe?
- [x] Add documentation.

## Acceptance criteria

When `pod` is locally available, Cocoapods dependencies listed under `EXTERNAL SOURCES` whose vendored podspecs have a `git` source should now be uploaded as correctly constructed `git` locators, not `pod` locators.

## Testing plan

1. Clone https://github.com/react-native-maps/react-native-maps.
2. Change into the `example` directory.
3. Run `yarn`.
4. Run `fossa analyze --output` in the cloned repository.
5. Examining the output should show that dependencies with external sources (such as `DoubleConversion`) now upload git locators.

## Risks

1. Should I write an integration test for this? If so, can I get a pointer towards an example integration test setup?
2. This PR relies on undocumented semantics for the meaning of `:path` and `:podspec` in `EXTERNAL SOURCES`. In particular, it assumes that `:path` always points to a folder with a `.podspec` file named the same as its pod, and that `:podspec` always points to a `.podspec` file. A [quick GitHub search](https://github.com/search?l=&p=1&q=filename%3APodfile.lock&ref=advsearch&type=Code) seems to support this (examples: [Expo](https://github.com/expo/expo/blob/67e936ed86ce2c3248273617eeba9020e2e83a87/ios/Podfile.lock), [WeatherApp](https://github.com/luizrmm/WeatherApp-2.0/blob/e01c2d124b507219ea2365fda98bb81e6c3a114e/ios/Podfile.lock), [schedule](https://github.com/ibrahimmohamed2411/schedule), [async-dataflow](https://github.com/bancolombia/async-dataflow/blob/c0cb9053589bd25bb989eb5de1c23e3b336a0530/examples/app_async_flutter/ios/Podfile.lock))

## References

- https://fossa.atlassian.net/browse/ANE-155
- https://teamfossa.slack.com/archives/C0155DTGWB1/p1647873192769469

## Checklist

### DIFF
--- a/src/Strategy/Cocoapods/PodfileLock.hs
+++ b/src/Strategy/Cocoapods/PodfileLock.hs
@@ -115,6 +115,9 @@ buildGraphStatic PodLock{lockPods, lockDeps, lockExternalSources} = staticGraph
       where
         pkg = PodfilePkg podName
 
+-- We use this cache to avoid repeated invocations since gtraverse may revisit
+-- nodes. In practice, this saves quite a bit of time (e.g. on example projects,
+-- it has historically turned ~2 minute analyses into ~15 seconds).
 type PodSpecCache = Map Text PodSpecJSON
 
 buildGraph ::
--- a/src/Strategy/Cocoapods/PodfileLock.hs
+++ b/src/Strategy/Cocoapods/PodfileLock.hs
@@ -35,7 +35,7 @@ import Data.SemVer qualified as SemVer
 import Data.SemVer.Internal (Version (..))
 import Data.Set (Set)
 import Data.String.Conversion (decodeUtf8, toString, toText)
-import Data.Text (Text)
+import Data.Text (Text, strip)
 import Data.Text.Extra (showT)
 import Data.Void (Void)
 import Data.Yaml ((.:), (.:?))
@@ -154,7 +154,7 @@ buildGraph lockFilePath lockFile@PodLock{lockExternalSources} = do
   case podVersionOutput of
     Left _ -> fatalText "could not analyze vendored pods: `pod --version` was not at least `0.29.0`"
     Right podVersionStdoutRaw -> do
-      let podVersionStdout = decodeUtf8 podVersionStdoutRaw
+      let podVersionStdout = strip $ decodeUtf8 podVersionStdoutRaw
 
       -- `pod ipc spec` is supported for versions 0.29.0 and above. See also:
       -- https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md#0290-2013-12-25
--- a/src/Strategy/Cocoapods/PodfileLock.hs
+++ b/src/Strategy/Cocoapods/PodfileLock.hs
@@ -9,6 +9,7 @@ module Strategy.Cocoapods.PodfileLock (
   ExternalGitSource (..),
 ) where
 
+import Control.Carrier.State.Strict (evalState)
 import Control.Effect.Diagnostics (
   Diagnostics,
   Has,
@@ -16,6 +17,7 @@ import Control.Effect.Diagnostics (
   run,
   (<||>),
  )
+import Control.Effect.State (State, get, put)
 import Data.Aeson (FromJSON (parseJSON))
 import Data.Aeson qualified as JSON
 import Data.Aeson.Types (FromJSONKey)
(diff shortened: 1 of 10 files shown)


src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -23,89 +26,149 @@ import Data.HashMap.Lazy qualified as HashMap (toList)
...
 import Text.Megaparsec.Char (char)
 import Text.Megaparsec.Char.Lexer qualified as L
+import Data.Text.Extra (showT)
 
-analyze' :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m (Graphing Dependency)
+analyze' :: (Has ReadFS sig m, Has Diagnostics sig m, Has Exec sig m) => Path Abs File -> m (Graphing Dependency)
 analyze' file = do
   podfileLock <- readContentsYaml file
-  context "Building dependency graph" $ pure (buildGraph podfileLock)
+  context "Building dependency graph" $ buildGraph file podfileLock
```
meghfossa, line 50: I know this is still wip, but this caught my eye.

Prefer the use of alternatives, so it's easier to reason on limitations, and warnings can be embedded outside of graphing logic.
 
```haskell
-- Function and variable names, etc. for example. 
-- Semmatically there will likely be better names than what I've 

podLockDynamicAnalysis podLockFile = do
 someGraph <- someBuildGraph podLockFile
 graph <- warnOnErr MissingVendorReplacements (enrichGraph someGraph)  <||> pure someGraph
 pure graph

data PodlockGraphEnrichmentFailed = PodlockGraphEnrichmentFailed (Path Abs File)

instance ToDiagnostic MissingVendorReplacements where
  renderDiagnostic (MissingVendorReplacements podLockFile) = "..."
```

liftM:
I investigated adding integration tests for this, but decided to defer that after conversation in the team channel. The fallback tactic (static graphs) is reliably unit tested.

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -23,89 +28,168 @@ import Data.HashMap.Lazy qualified as HashMap (toList)
...
+-- nodes. In practice, this saves quite a bit of time (e.g. on example projects,
+-- it has historically turned ~2 minute analyses into ~15 seconds).
+type PodSpecCache = Map Text PodSpecJSON
+
+buildGraph ::
+  (Has ReadFS sig m, Has Exec sig m, Has Diagnostics sig m) =>
+  Path Abs File ->
+  PodLock ->
+  m (Graphing Dependency)
+buildGraph lockFilePath lockFile@PodLock{lockExternalSources} =
```
meghfossa, line 128: [Nit] Prefer if we fail early here if the user cannot execute `pod --version`, or the version is below `0.29.0`(https://github.com/CocoaPods/CocoaPods/blob/master/CHANGELOG.md#0290-2013-12-25).

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -23,89 +28,168 @@ import Data.HashMap.Lazy qualified as HashMap (toList)
...
+    readPodSpecAt podSpecPath = context ("Resolving vendored podspec at " <> showT podSpecPath) $ do
+      podSpecCache :: PodSpecCache <- get
+      (PodSpecJSON ExternalGitSource{urlOf, tagOf, commitOf, branchOf}) <-
+        case Map.lookup podSpecPath podSpecCache of
+          Just cached -> pure cached
+          Nothing -> do
+            podSpec <-
+              execJson
+                lockFileDir
+                Command
```
meghfossa, line 177: [Optional] Although this command is only used here, I would move it outside of `buildGraph` scope, so it follows the stylistic convention of other analyzers. 

[Optional] Add ref URL in comment - https://guides.cocoapods.org/terminal/commands.html#pod_ipc_spec

meghfossa (approved):
LGTM left 1 nit, and 2 optional. 

I ran the provided test case - it worked. Although since lots of external sources have the same git-locator, there is squashing of lots of edges, I checked a couple of node dependencies - before and after, and they seemed correct.

PR #876, opened 2022-04-05
Converts ArchiveUploader to use API effects

periodic:
# Overview

Converts the archive upload functions to use the new `FossaApiClient` effects.

As part of this, I renamed a few of the methods so they would be clearer in the `verbNoun` format, e.g. `queueArchiveBuild` instead of `archiveBuildUpload`.

## Acceptance criteria

- All API requests in the `ArchiveUploader` use API effects

## Testing plan

The `ArchiveUploader` uses a lot of IO, so it's very hard to write unit tests for.  Instead I just did a bunch of manual testing by setting up a `fossa-deps.yml` file that includes a vendored dep then verified that this got picked up by my local core.  Specifically it seems to want to compress archives into a temporary file before uploading.

## Risks

There's a possibility that my testing missed something important, since it was all manual.

## References

- [ANE-132](https://fossa.atlassian.net/browse/ANE-132): Convert App.Fossa.ArchiveUploader to use API Effects

## Checklist


src/Control/Carrier/FossaApiClient/Internal/Core.hs
```diff
@@ -102,3 +119,35 @@ getIssues ::
...
+getSignedUploadUrl ::
+  ( Has (Lift IO) sig m
+  , Has Diagnostics sig m
+  , Has (Reader ApiOpts) sig m
+  ) =>
+  ProjectRevision ->
+  m SignedURL
+getSignedUploadUrl ProjectRevision{..} = do
+  apiOpts <- ask
+  API.getSignedURL apiOpts projectRevision projectName
```
kitified, line 132: Can we please give this its own type? Archive uploads never have a `projectBranch` associated, which means that any time we want to use this the value provided to the `ProjectRevision` instance is silently ignored.

While I recognize it's convenient to use the type since it mostly fits, I think it's important to make sure APIs use types that reflect what they actually _need_.

kitified (changes requested):

src/Control/Carrier/FossaApiClient/Internal/Core.hs
periodic: Okay, I wasn't sure, and the previous code used `Text` for both, which was awkward.  I'll just `newtype` them or something.

kitified: Yeah! I think newtype wrappers or a new type that is just name and revision would work great!

periodic: Done.

kitified (approved):
🎉

PR #877, opened 2022-04-05
Naive discovery exclusion

scruffystuffs:
# Overview

Delivers ANE-185

I made a few orthogonal changes to aid in development here:

- `simpleDiscover` - Most of our discovery can be proven to take a simplified form, and `simpleDiscover` helps prove that.  The module includes a comment explaining the difference between simple and non-simple.
- `$` as a type-level operator - Remove parens from big monad stacks.
- Remove `MonadFail` from fused-effect carriers, we don't use them, and trying to use them can cause ambiguous usage.  This is actually one of the bigger reasons for effect systems in general, a `fused-effects` dev did a great talk about this problem [here](https://youtu.be/vfDazZfxlNs). 

## A Caveat of Path Filtering
Note that path filtering is not available for walkers that attempt to collect files from multiple directories, such as maven and nodejs (which both have static-analysis workspace support).  Another way to think of it is that directories emitted by the walk must have a 1:1 relationship with projects emitted by the discovery, and must be independent of each other.  In cases of tool-based workspace support (like cargo), we CAN safely apply path filtering, because the walker does not emit subprojects, it skips them via `WalkSkipSome/WalkSkipAll`.

This is similar to the simple vs non-simple comparison for discovery, except that simple discovery is only determined by the ability to define `mkProject` as `a -> DiscoveredProject a`.

## The Legacy of NodeJS

After reading this PR, you might ask: why isn't the `DiscoveredProjectType` that we pass to `simpleDiscover` inferred from the type of the `a` in `DiscoveredProject a`?  The short answer is that there is not always a single value for all cases of `a`.

During the recent `yarn`/`npm` revamp, nodejs-based discovery became a single process.  However, users' filters still reflected the old model, which used `yarn` and `npm`.  We didn't have a good reason for changing them at the time, so there was no real argument for the breaking change of calling all of these `Node` projects.  However, that has presented us with the problem of not being able make `DiscoveredProjectType` a type-guaranteed match based on the `a` in `DiscoveredProject a`.  Doing this requires a 1:1 relationship between the `DiscoveredProjectType` and the `a`, which is only not true for nodejs, and only because of this exact legacy problem.

While that definitely limits the solution space for this feature, and is fixable, I think it should be in a follow-up discussion/PR.  The addition of `simpleDiscovery` makes this easier to refactor later, as well.  Also, we could probably add this for `simpleDiscover`-eligible projects only, and ship it here, but this PR is already large due to modifying every strategy and affecting tests, so it's a good idea to table it anyway.

## Acceptance criteria

- When a user runs `fossa analyze --exclude-target cabal`, we should not run the `cabal` discovery function.
- When a user runs `fossa analyze --only-target cabal`, we should only run the `cabal` discovery function.
- When a user runs `fossa analyze --only-path hello`, we should not enter any other directory (see caveats above).
- When a user runs `fossa analyze --exclude-path hello`, we should not descend into that directory (see caveats above).

## Testing plan

- unit tests and integration tests must pass
- AC above must be manually tested.

## Checklist

### DIFF
--- a/src/App/Fossa/ListTargets.hs
+++ b/src/App/Fossa/ListTargets.hs
@@ -49,7 +49,8 @@ import Effect.Logger (
   annotate,
   color,
   logDebug,
-  logInfo, logWarn
+  logInfo,
+  logWarn,
  )
 import Effect.ReadFS (ReadFS)
 import Path (Abs, Dir, Path, toFilePath)
--- a/src/App/Fossa/ListTargets.hs
+++ b/src/App/Fossa/ListTargets.hs
@@ -49,7 +49,7 @@ import Effect.Logger (
   annotate,
   color,
   logDebug,
-  logInfo,
+  logInfo, logWarn
  )
 import Effect.ReadFS (ReadFS)
 import Path (Abs, Dir, Path, toFilePath)
@@ -71,7 +71,7 @@ listTargetsMain ::
   m ()
 listTargetsMain ListTargetsConfig{..} = do
   capabilities <- sendIO getNumCapabilities
-
+  logWarn "fossa list-targets does not apply any filtering, you may see projects which are not present in the final analysis."
   ignoreDebug
     . runStickyLogger SevInfo
     . runFinally
--- a/src/App/Fossa/ListTargets.hs
+++ b/src/App/Fossa/ListTargets.hs
@@ -38,6 +38,7 @@ import Data.Aeson.Extra (encodeJSONToText)
 import Data.Foldable (for_, traverse_)
 import Data.Set qualified as Set
 import Data.Set.NonEmpty (toSet)
+import Discovery.Filters (AllFilters)
 import Discovery.Projects (withDiscoveredProjects)
 import Effect.Exec (Exec)
 import Effect.Logger (
@@ -77,6 +78,9 @@ listTargetsMain ListTargetsConfig{..} = do
     . withTaskPool capabilities updateProgress
     . runAtomicCounter
     . runReader experimental
+    -- The current version of `fossa list-targets` does not support filters.
+    -- TODO: support both discovery and post-discovery filtering.
+    . runReader (mempty :: AllFilters)
     $ runAll (unBaseDir baseDir)
 
 runAll ::
@@ -89,6 +93,7 @@ runAll ::
   , Has Debug sig m
   , Has Stack sig m
   , Has (Reader ExperimentalAnalyzeConfig) sig m
+  , Has (Reader AllFilters) sig m
   , Has Telemetry sig m
   ) =>
   Path Abs Dir ->
--- a/src/Discovery/Filters.hs
+++ b/src/Discovery/Filters.hs
@@ -121,11 +121,13 @@ toolMatches ty = \case
 pathAllowed :: AllFilters -> Path Rel Dir -> Bool
 pathAllowed AllFilters{..} path = allowedInclude && allowedExclude
   where
-    allowedExclude = notElem path $ combinedPaths excludeFilters
+    allowedExclude = not . any (matchPath path) $ combinedPaths excludeFilters
     allowedInclude
       -- If the included path array is empty, allow everything
-      | combinedPaths includeFilters == (mempty :: [Path Rel Dir]) = True
-      | otherwise = elem path $ combinedPaths includeFilters
+      | combinedPaths includeFilters == mempty = True
+      | otherwise = any (matchPath path) $ combinedPaths includeFilters
+    -- child "hello/world/" matches "hello/" and "hello/world/"
+    matchPath child parent = parent == child || parent `isProperPrefixOf` child
 
 comboInclude :: [TargetFilter] -> [Path Rel Dir] -> FilterCombination Include
 comboInclude = FilterCombination
--- a/src/Discovery/Filters.hs
+++ b/src/Discovery/Filters.hs
@@ -16,8 +16,13 @@ module Discovery.Filters (
   combinedTargets,
   Include,
   Exclude,
+  pathAllowed,
+  toolAllowed,
+  withMultiToolFilter,
+  withToolFilter,
 ) where
 
+import Control.Effect.Reader (Has, Reader, ask)
 import Data.Aeson (ToJSON (toEncoding), defaultOptions, genericToEncoding)
 import Data.List ((\\))
 import Data.List.NonEmpty qualified as NE
@@ -38,7 +43,7 @@ import Text.Megaparsec (
   (<|>),
  )
 import Text.Megaparsec.Char (alphaNumChar, char)
-import Types (BuildTarget (..), FoundTargets (FoundTargets, ProjectWithoutTargets), TargetFilter (..))
+import Types (BuildTarget (..), DiscoveredProjectType, FoundTargets (FoundTargets, ProjectWithoutTargets), TargetFilter (..))
 
 -- | filterIsVSIOnly is a very naive check for whether the user provided a filter for the VSI target *only*, at the root of the project with no path specified.
 -- This decision was made because predicating VSI only scans is a stopgap until we have more broad support for *preemptive* analysis filtering.
@@ -86,6 +91,40 @@ instance Semigroup (FilterCombination a) where
 
 instance Monoid (FilterCombination a) where
   mempty = FilterCombination mempty mempty
+
+withMultiToolFilter :: Has (Reader AllFilters) sig m => [DiscoveredProjectType] -> m [a] -> m [a]
+withMultiToolFilter tools act = do
+  filters <- ask
+  if any (toolAllowed filters) tools
+    then act
+    else pure mempty
+
+withToolFilter :: Has (Reader AllFilters) sig m => DiscoveredProjectType -> m [a] -> m [a]
+withToolFilter tool = withMultiToolFilter [tool]
+
+toolAllowed :: AllFilters -> DiscoveredProjectType -> Bool
+toolAllowed AllFilters{..} ty = allowedInclude && allowedExclude
+  where
+    allowedExclude = not . any (toolMatches ty) $ combinedTargets excludeFilters
+    allowedInclude
+      -- If the included target array is empty, allow everything
+      | combinedTargets includeFilters == (mempty :: [TargetFilter]) = True
+      | otherwise = any (toolMatches ty) $ combinedTargets includeFilters
+
+toolMatches :: DiscoveredProjectType -> TargetFilter -> Bool
+toolMatches ty = \case
+  TypeTarget txt -> txt == toText ty
+  TypeDirTarget txt _ -> txt == toText ty
+  TypeDirTargetTarget txt _ _ -> txt == toText ty
+
+pathAllowed :: AllFilters -> Path Rel Dir -> Bool
+pathAllowed AllFilters{..} path = allowedInclude && allowedExclude
+  where
+    allowedExclude = notElem path $ combinedPaths excludeFilters
+    allowedInclude
+      -- If the included path array is empty, allow everything
+      | combinedPaths includeFilters == (mempty :: [Path Rel Dir]) = True
+      | otherwise = elem path $ combinedPaths includeFilters
 
 comboInclude :: [TargetFilter] -> [Path Rel Dir] -> FilterCombination Include
 comboInclude = FilterCombination
--- a/src/Discovery/Filters.hs
+++ b/src/Discovery/Filters.hs
@@ -105,6 +105,7 @@ withToolFilter tool = withMultiToolFilter [tool]
 toolAllowed :: AllFilters -> DiscoveredProjectType -> Bool
 toolAllowed AllFilters{..} ty = allowedInclude && allowedExclude
   where
+    -- TODO: Use FilterCombination's phantom type to confirm that we aren't mixing these up.
     allowedExclude = not . any (toolMatches ty) $ combinedTargets excludeFilters
     allowedInclude
       -- If the included target array is empty, allow everything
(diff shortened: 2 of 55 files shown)


scruffystuffs:
@meghfossa Can you review this approach?  I've done basic testing, the idea definitely works, but I'd like to check-in before going too far on this.

src/Strategy/Python/Setuptools.hs
```diff
@@ -7,29 +7,41 @@ module Strategy.Python.Setuptools (
...
+  DependencyResults (..),
+  DiscoveredProject (..),
+  DiscoveredProjectType (SetuptoolsProjectType),
+  GraphBreadth (Partial),
+ )
 
-discover :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> m [DiscoveredProject SetuptoolsProject]
-discover dir = context "Setuptools" $ do
+discover :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters) sig m) => Path Abs Dir -> m [DiscoveredProject SetuptoolsProject]
+discover dir = withToolFilter SetuptoolsProjectType $ context "Setuptools" $ do
```
meghfossa, line 39: I'm a bit hesitant about this - It couples concerns of filtering within individual discovery and findProject function. Although, I can't pin a better alternative at the moment (maybe some form of factory?). Let me have some time with this. 

These are some of the thoughts that came at the moment, although I need some time to refine my thoughts. 

1) why not move this (withToolFilter a $ ..) decorator to `App.Fossa.Analyze.Discover`?
2) since this ProjectType lives as a value (declared) in DiscoveredProject as opposed to type makes this bit awkward - otherwise, maybe with `-XTypeApplications` this can be easier to filter, and have more strict guarantees?

scruffystuffs: So, the `DiscoveredProjectType` is not deterministic for a few `discover` functions, so we can't make a filter that's generic for all of them that's automatically applied.  Even if we could, and did: type-enforcing those constant values is _really hard_ without fully dependent types, which haskell does not really have.

scruffystuffs: While this setup is not error-proof, it's incredibly unobtrusive, and easily human-verifiable.  It's also very unlikely to change.

scruffystuffs: Additionally, I put these into their associated strategies modules _intentionally_.  Each strategy has to decide _for itself_ whether to use the filters or not.  In some cases, it doesn't make sense to do so, particularly where we cross-pollinate strategies (see node/yarn, and scala/maven).

scruffystuffs: This PR should be seen as a band-aid that allows us to deliver the general discovery exclusion feature.  The _real_ problem is that we have a 2-phase system (discovery/analysis), when we need a 3-phase system:
1. Walk the tree and note _potential_ project locations
2. Iterate through valid locations from step 1 to identify valid project roots (currently we do this in discovery as part of `findProjects`)
3. analyze (no change here).

scruffystuffs: If you can find a way to make this less error-prone without sacrificing conciseness, introducing a massive refactor or adding complex type-level trickery, I'm 100% on board.  But I already went down those rabbit holes and I found nothing that didn't make things worse.

meghfossa: > If you can find a way to make this less error-prone without sacrificing conciseness, introducing a massive refactor or adding complex type-level trickery, I'm 100% on board. But I already went down those rabbit holes and I found nothing that didn't make things worse.

I'm slowly coming to the same conclusion without having to redo/refactor the entirety of discovery plumbing.

scruffystuffs: Posterity: We discussed this offline and agreed that this solution is good enough for now, but it would definitely be nice to have a less error-prone version.

Also, the addition of `simpleDiscover` can hopefully help with this.

src/Discovery/Filters.hs
```diff
@@ -87,6 +92,40 @@ instance Semigroup (FilterCombination a) where
...
 
+withMultiToolFilter :: Has (Reader AllFilters) sig m => [DiscoveredProjectType] -> m [a] -> m [a]
+withMultiToolFilter tools act = do
+  filters <- ask
+  if any (toolAllowed filters) tools
+    then act
+    else pure mempty
+
+withToolFilter :: Has (Reader AllFilters) sig m => DiscoveredProjectType -> m [a] -> m [a]
+withToolFilter tool = withMultiToolFilter [tool]
```
scruffystuffs, line 103: - [ ] Unit tests needed

src/Discovery/Filters.hs
```diff
@@ -87,6 +92,40 @@ instance Semigroup (FilterCombination a) where
...
+      | otherwise = any (toolMatches ty) $ combinedTargets includeFilters
+
+toolMatches :: DiscoveredProjectType -> TargetFilter -> Bool
+toolMatches ty = \case
+  TypeTarget txt -> txt == toText ty
+  TypeDirTarget txt _ -> txt == toText ty
+  TypeDirTargetTarget txt _ _ -> txt == toText ty
+
+pathAllowed :: AllFilters -> Path Rel Dir -> Bool
+pathAllowed AllFilters{..} path = allowedInclude && allowedExclude
```
scruffystuffs, line 121: - [ ] Unit tests needed

src/Discovery/Filters.hs
```diff
@@ -87,6 +92,40 @@ instance Semigroup (FilterCombination a) where
 instance Monoid (FilterCombination a) where
   mempty = FilterCombination mempty mempty
 
+withMultiToolFilter :: Has (Reader AllFilters) sig m => [DiscoveredProjectType] -> m [a] -> m [a]
+withMultiToolFilter tools act = do
```
scruffystuffs, line 96: - [ ] Unit tests needed

src/Discovery/Filters.hs
```diff
@@ -87,6 +92,40 @@ instance Semigroup (FilterCombination a) where
...
+  filters <- ask
+  if any (toolAllowed filters) tools
+    then act
+    else pure mempty
+
+withToolFilter :: Has (Reader AllFilters) sig m => DiscoveredProjectType -> m [a] -> m [a]
+withToolFilter tool = withMultiToolFilter [tool]
+
+toolAllowed :: AllFilters -> DiscoveredProjectType -> Bool
+toolAllowed AllFilters{..} ty = allowedInclude && allowedExclude
```
scruffystuffs, line 106: - [ ] Unit tests needed

src/Discovery/Simple.hs
```diff
@@ -0,0 +1,28 @@
...
+
+import Control.Effect.Diagnostics (Diagnostics, context)
+import Control.Effect.Reader (Reader)
+import Data.String.Conversion (toText)
+import Discovery.Filters (AllFilters, withToolFilter)
+import Effect.ReadFS (Has, ReadFS)
+import Path (Abs, Dir, Path)
+import Types (DiscoveredProject, DiscoveredProjectType)
+
+simpleDiscover ::
```
scruffystuffs, line 11: - [ ] Add comment explaining difference between simple and non-simple discovery

src/Strategy/Node.hs
```diff
@@ -24,10 +24,11 @@ import Control.Effect.Diagnostics (
   recover,
   warnOnErr,
  )
+import Control.Effect.Reader (Reader)
 import Control.Monad (void, (<=<))
 import Data.Glob (Glob)
 import Data.Glob qualified as Glob
-import Data.List.Extra (singleton) -- singleton is in Data.List in base 4.16
```
scruffystuffs, line 30: Had to remove this comment, it was causing failed idempotence checks in the formatter.  It's also not a major concern, since it will be a name conflict once we upgrade to GHC 9.2 and `base-4.16`.

test/Test/Effect.hs
```diff
@@ -49,26 +35,30 @@ import Path.IO (createDirIfMissing, removeDirRecur)
...
-                    )
-                )
-            )
-        )
-    )
-    a
+-- The FossaApiClient must be outside of Diagnostics because it
+-- depends on it.
+-- MockApiC must be inside Diagnostics so its state is not
+-- lost when diagnostics short-circuits on an error.
```
scruffystuffs, line 60: I moved these comments out because the formatter was making this completely unreadable.

src/Discovery/Filters.hs
scruffystuffs: Actually, this isn't needed, since it's just a convenience alias.

csasarak (approved):

csasarak:
LGTM!

src/App/Fossa/ListTargets.hs
```diff
@@ -77,6 +78,9 @@ listTargetsMain ListTargetsConfig{..} = do
     . withTaskPool capabilities updateProgress
     . runAtomicCounter
     . runReader experimental
+    -- The current version of `fossa list-targets` does not support filters.
```
meghfossa, line 81: Can we include this comment as a warning or info message, so the user is aware that we have not applied any discovery exclusion from the configuration file.

src/Strategy/Carthage.hs
```diff
@@ -22,26 +31,59 @@ import Data.String.Conversion (toString, toText)
...
-findProjects = walk' $ \dir _ files -> do
+import Types (
+  DependencyResults (..),
+  DiscoveredProject (..),
+  DiscoveredProjectType (BundlerProjectType, CarthageProjectType),
+  GraphBreadth (Complete),
+ )
+
+discover :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters) sig m) => Path Abs Dir -> m [DiscoveredProject CarthageProject]
+discover = simpleDiscover findProjects mkProject BundlerProjectType
```
meghfossa, line 83: ought to be `CarthageProject`

test/Discovery/FiltersSpec.hs
```diff
@@ -161,7 +186,86 @@ spec = do
...
+        let filts = excludeTool CargoProjectType <> excludeTool GomodProjectType
+        toolAllowed filts CargoProjectType `shouldBe` False
+        toolAllowed filts GomodProjectType `shouldBe` False
+        toolAllowed filts SetuptoolsProjectType `shouldBe` True
+      it "should reject conflicted tools" $ do
+        -- Conflicting filters, members of exclude are NEVER allowed
+        toolAllowed (includeTool CargoProjectType <> excludeTool CargoProjectType) CargoProjectType `shouldBe` False
+
+    describe "Path-based matching" $ do
+      it "should include paths correctly" $ do
```
meghfossa, line 213: [nit] prefer line breaks between consecutive `it` for readability

meghfossa (changes requested):
CarthageProject filtering is broken, left two nits.

meghfossa (approved):
LGTM - before you merge, can you update the changelog, and scan the summary (as our standup discussion)

scruffystuffs:
Updated and set to auto-merge

thernstig:
@scruffystuffs as part of this change, I think the output from `fossa analyze` becomes a bit weird:

```sh
[ INFO] [ 0 Waiting / 2 Running / 33 Completed ]
[ INFO] [ 0 Waiting / 1 Running / 34 Completed ]
[ INFO] [ 1 Waiting / 1 Running / 34 Completed ]
[ INFO] [ 1 Waiting / 2 Running / 34 Completed ]
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/apps/
[ INFO] Skipping npm project at "/home/userA/code/projectFoo/": no filters matched
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/client/
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/server/
[ INFO] [ 0 Waiting / 3 Running / 34 Completed ]
[ INFO] [ 0 Waiting / 3 Running / 36 Completed ]
[ INFO] [ 0 Waiting / 2 Running / 37 Completed ]
[ INFO] [ 0 Waiting / 1 Running / 38 Completed ]
```

Notice the 34-38. This does not make a lot of sense to users when only scanning with a `fossa.yaml` that looks like:

```yaml
version: 3
server: https://someserver.com/
project:
  id: projectFoo
targets:
  only:
    - type: npm
      path: apps/
    - type: npm
      path: client/
    - type: npm
      path: server/
```

scruffystuffs:
> I think the output from fossa analyze becomes a bit weird

The output wasn't changed here.

> Notice the 34-38

You didn't include line numbers, so I'm assuming you mean this chunk

```
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/apps/
[ INFO] Skipping npm project at "/home/userA/code/projectFoo/": no filters matched
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/client/
[ INFO] Analyzing npm project at /home/userA/code/projectFoo/server/
```

> People would not able to grok with 34-38 means.

I'm not sure how we could be more clear on meaning.  We tell you that there's a project of a specific type, at a specific location, and what we're going to do with it.  What information could we add that makes this clearer?

> In addition, why is the root scanned when it is not in the targets?

Great question.  First, a clarification:  we _discover_ project, but do not _analyze_ them, even with discovery exclusion.

We scan the root because our project discovery is a top-down traversal of the filesystem.  If we stopped at the root because it didn't match the files, then we would never get to the subfolders, and we wouldn't ever scan anything.

thernstig:
I wrote a lengthy reply why the output of the CLI is not great in e.g. CI systems where terminals do not support `\r` carriage return to re-write lines, giving output that is contextually hard to grok. But realized that is a battle I wish not to take here 😆

But regarding this:

```yaml
targets:
  only:
    - type: npm
      path: apps/
    - type: npm
      path: client/
    - type: npm
      path: server/
```

I don't follow the explanation about why you discover the root. Even if it is a top-down system and I specify the above, then seeing the output for `[ INFO] Skipping npm project at "/home/userA/code/projectFoo/": no filters matched` is confusing, since I never wanted to scan the root. I didn't specify it.

Aside: How would I then specify to _only_ scan the root and nothing else in the config? That is a very valid scenario for npm workspaces where you only have a top-level `package-lock.json`.

PR #879, opened 2022-04-06
Converts assertRevisionBinaries to FossaApiClient

periodic:
# Overview

Converts `assertRevisionBinaries` API call to go through `FossaApiClient`

## Testing plan

Due to the extra IO of fingerprinting I tested this manually using the [C/C++ VSI Demo](https://github.com/fossas/cpp-vsi-demo) project and a local Core.

First, I implemented the call with a `fatalText`:

```
[ INFO] Fingerprinting assertion directory contents
[ INFO] Uploading assertion to FOSSA
[ERROR] ----------
  An issue occurred

  >>> Relevant errors

    Error

      assertRevisionBinaries is not implemented

      Traceback:
        - upload-results
        - fossa-analyze
```

Then I implemented it as a pass-through to the old implementation:

```
[ INFO] Fingerprinting assertion directory contents
[ INFO] Uploading assertion to FOSSA
```

## References

_Example:_

- [ANE-135](https://fossa.atlassian.net/browse/ANE-135): Convert App.Fossa.VSI.IAT.AssertRevisionBinaries to use API Effects

## Checklist

### DIFF
--- a/src/Control/Effect/FossaApiClient.hs
+++ b/src/Control/Effect/FossaApiClient.hs
@@ -4,7 +4,10 @@ module Control.Effect.FossaApiClient (
   ArchiveRevision (..),
   FossaApiClientF (..),
   FossaApiClient,
+  assertRevisionBinaries,
+  assertUserDefinedBinaries,
   getApiOpts,
+  getAttribution,
   getIssues,
   getLatestBuild,
   getLatestScan,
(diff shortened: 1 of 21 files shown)


src/Control/Effect/FossaApiClient.hs
```diff
@@ -137,3 +139,6 @@ queueArchiveBuild = sendSimple . QueueArchiveBuild
 
 assertUserDefinedBinaries :: Has FossaApiClient sig m => IAT.UserDefinedAssertionMeta -> [Fingerprint Raw] -> m ()
 assertUserDefinedBinaries meta fprints = sendSimple (AssertUserDefinedBinaries meta fprints)
+
+assertRevisionBinaries :: Has FossaApiClient sig m => Locator -> [Fingerprint Raw] -> m ()
```
meghfossa, line 143: [No action] too many methods (about ~4 now) in different modules called the `assertRevisionBinaries` 😄  This adds a bit of friction in readability, but this is convention anyways, so no action but wanted to denote this.

meghfossa (approved):
lgtm. I did not perform the test plan as the changes were small and they look correct.

src/Control/Effect/FossaApiClient.hs
periodic: I'm hoping we can mess with names later, and we'll get rig of the ones in `FossaV1` soon.

PR #882, opened 2022-04-07
[ANE-211] Missing peer deps of namespaced packages

scruffystuffs:
# Overview

I wrote `Graphing.Trace` to debug this problem, and solved the problem to prove that the tracing module worked.

`package-lock.json` take the following rough form (simplified for explanation):
```json
{
  ...,
  "packages": {
    "node_modules/foo": {
      "peerDeps": [ ... ]
    },
    "node_modules/@org/bar": {
      "peerDeps": [ ... ]
    },
    "node_modules/baz/node_modules/quux": {
      "peerDeps": [ ... ]
    }
  },
  "dependencies": {
    "foo": { ... },
    "@org/bar": { ... },
    "baz": {
      "dependencies": {
        "quux": { ... }
      }
    }
  }
}
```

To understand the problem, you must know two things about the file:

- Most relevant dependency info is in the top-level `dependencies` field, or in one of the sub-`dependencies` fields, (see `baz -> quux`).  Peer dependencies info is only available in `packages`.
- The link between keys in `packages`/`dependencies` is not explicitly specified, and must be manually constructed.

## The Actual Problem

When traversing the `dependencies` field, we query the associated item on the `packages` field for peer dep info.  To make the names match, we mangle the key from the `dependencies` field.  However, the mangling process was doing a string split on `/` characters, which would screw with namespaced packages (like `@org/bar` in the above example).  This caused us to miss peer deps for ALL namespaced packages.  If a dependency was only brought in via the peer deps of a namespaced package, we would miss it.

JIRA: [ANE-211](https://fossa.atlassian.net/browse/ANE-211)

## Checklist

### DIFF
--- a/src/Graphing/Debug.hs
+++ b/src/Graphing/Debug.hs
@@ -6,8 +6,7 @@ module Graphing.Debug (
   renderGvFile,
 ) where
 
-import Algebra.Graph.Acyclic.AdjacencyMap qualified as Acyclic
-import Algebra.Graph.AdjacencyMap qualified as Cyclic
+import Algebra.Graph.AdjacencyMap qualified as AM
 import Control.Effect.Lift (Has, Lift, sendIO)
 import Data.Bifunctor (Bifunctor (bimap))
 import Data.Set (Set)
@@ -52,9 +51,9 @@ versionTxt = \case
   CGreaterOrEq txt -> txt
   CNot txt -> txt
 
-data EmittedGraph
-  = Cyclic [Text]
-  | Acyclic [Text]
+newtype EdgeText = EdgeText
+  { unEdgeText :: Text
+  }
 
 debugGraph :: Has (Lift IO) sig m => GraphDebug a => Graphing a -> m ()
 debugGraph gr = sendIO $ do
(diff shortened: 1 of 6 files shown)


src/Graphing/Debug.hs
```diff
@@ -0,0 +1,103 @@
...
+
+envsToText :: Set DepEnvironment -> Text
+envsToText = Text.intercalate ", " . map showT . Set.toList
+
+versionTxt :: VerConstraint -> Text
+versionTxt = \case
+  CEq txt -> txt
+  CURI txt -> txt
+  CCompatible txt -> txt
+  CAnd vc vc' -> versionTxt vc <> "AND" <> versionTxt vc'
```
csasarak, line 47: [nit] This might be overkill for now, but with nested combinations of `AND` and `OR` it could get confusing to see how things associate. Maybe it would be worth adding explicit parens around the operands?

src/Graphing/Debug.hs
```diff
@@ -0,0 +1,103 @@
...
+  let graphTxt = renderGvFile gr
+  TIO.writeFile gvFileName graphTxt
+{-# WARNING debugGraph "This function is for debug purposes only and should not be used in production." #-}
+
+renderGvFile :: GraphDebug a => Graphing a -> Text
+renderGvFile gr = emit graphEdges
+  where
+    adjMap = unGraphing gr
+    graphEdges = case Acyclic.toAcyclic adjMap of
+      Nothing -> emitGraphEdges adjMap
```
csasarak, line 70: Aren't even the cyclic graphs technically directed? I'm thinking it might be useful to know which direction a cycle goes in. Of course, I would probably be looking at the `package-lock.json` file manually anyways so I don't think I really need any changes here - I'm just curious.

csasarak (approved):
LGTM, I left a few questions but don't have any requested changes.

src/Graphing/Debug.hs
scruffystuffs: Good point, we can actually merge these two.  I was thinking of DAG vs. non-DAG, but that's not quite right.

scruffystuffs: I'll merge these quickly, shouldn't take long.

scruffystuffs: Done.

src/Graphing/Debug.hs
scruffystuffs: For now, I don't think this is going to be a problem.  I haven't actually seen these used in quite some time, and never in JS.

meghfossa (approved):
LGTM

thernstig:
@scruffystuffs @csasarak are you aware that the top-level `packages` property is a npm lockfile version 2/3 thing only, where `dependencies` is only kept for backwards compatibility?

I.e. in npm lockfile version 1  you use `dependencies`, in npm lockfile version 2 `dependencies` is **only** kept for backwards compatability and in npm lockfile version 3 the `dependencies` is removed completely.

I.e. the new in version 2 and 3 is under `packages`.

Meaning that if this fix has created an inter-dependency between `dependencies` and `packages` then this entire implementation has been done on incorrect and false assumptions.

PR #881, opened 2022-04-06
Converts VSI.IAT.Resolve to FossaApiClient

periodic:
# Overview

Converts both `resolveUserDefinedBinary` and `resolveProjectDependencies` to use `FossaApiClient`.

## Acceptance criteria

- Both functions should be exclusively accessed through effects.
- Unit tests where possible.

## Testing plan

In addition to any unit tests, I did manual testing.

1. Set up both functions to trigger a fatal diagnostic error in the carrier implementation.
2. Linked a project binary
    ```sh
    fossa analyze internal-json-parser \    
      -e "$ENDPOINT" --fossa-api-key "$API_KEY" \
      --project internal-json-parser --revision $(date +%s) \
      --experimental-link-project-binary bin/libjson_internal \
      --enable-vsi --debug
    ```
3. Observed the diagnostic error for `resolveUserDefinedBinary` in the output.
4. Implemented the function as a pass-through to the old API function.
5. Re-ran and observed the error disappeared and the command was successful.
6. Ran analysis on the project
    ```sh
fossa analyze example-internal-project \
  -e "$ENDPOINT" --fossa-api-key "$API_KEY" \
  --project internal-project --revision $(date +%s) \
  --enable-vsi --debug
    ````
7. Observed the error for `resolveProjectDependencies` in the output.
8. Implemented the function as a pass-through to the old API function
9. Re-ran and observed the error disappeared and the command was successful

## References

_Add links to any referenced GitHub issues, Zendesk tickets, Jira tickets, Slack threads, etc._

_Example:_

- [ANE-136](https://fossa.atlassian.net/browse/ANE-136): Convert App.Fossa.VSI.IAT.Resolve to use API Effects

## Checklist

- [ ] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

PR #883, opened 2022-04-07
Converts VSI.Analyze to use FossaApiClient

periodic:
# Overview

Converts `App.Fossa.VSI.Analyze` to use `FossaApiClient`.

## Acceptance criteria

All API access from `App.Fossa.VSI.Analyze` goes through `FossaApiClient`.

Tests written where appropriate.

## Testing plan

I checked that this worked on the examples given in the [C/C++ VSI Demo](https://github.com/fossas/cpp-vsi-demo).  Automated tests are difficult because this code it tightly intertwined with other IO that has not been made into effects.

Here's what I did during development:

1. Created empty implementations for each function using `fatalText`.
2. Ran the analyze command on the demo project.
    ```sh
    fossa analyze example-internal-project \
      -e "$ENDPOINT" --fossa-api-key "$API_KEY" \
      --project internal-project --revision $(date +%s) \
      --enable-vsi --debug
    ```
3. Observed failures and then implemented the failing function as a pass through to the old implementation
4. Proceed until the analysis succeeds without error.

## Risks

This code is pretty complicated and does a lot.  There's a risk that I missed something when I was converting things.  Thankfully, the changes are simply refactoring.  I tested manually to try to mitigate the risks, but I'm not an expert in VSI behavior so it's possible I missed something.

## References

- [ANE-137](https://fossa.atlassian.net/browse/ANE-137): Convert App.Fossa.VSI.IAT.Analyze to use API Effects

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## v3.2.4
+- nodejs: Fixed a bug where dev deps that only appear in requires were considered production dependencies. ([#884](https://github.com/fossas/fossa-cli/pull/884))
+
 ## v3.2.3
 - nodejs: Fixed a bug where some dev dependencies weren't removed during shrinking. ([#859](https://github.com/fossas/fossa-cli/pull/859))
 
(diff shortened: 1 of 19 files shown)


periodic:
@kitified, would you please review this change?  You are most familiar with VSI code so I'd trust you to notice if I did something dangerous.

periodic:
Oops, I had the wrong base.  This was including changes from #881.

Also, I tried changing some names to make them clearer or more consistent, but I don't know if they are good ones.

kitified (approved):

PR #885, opened 2022-04-08
Convert LicenseScanner to FossaApiClient

periodic:
# Overview

Converts `LicenceScanning` to use `FossaApiClient`.

Also, I had a structured called `ArchiveRevision` which I renamed to `PackageRevision`.  It captures the notion of an external dependency that will never have a branch and so `ProjectRevision` is not appropriate.  I'm not sure if this is the best naming, but it makes the API clearer to bundle these instead of passing around pairs of `Text` values.  I welcome discussion on this.

I also noticed there were two places where we were ignoring the value of the call, so I converted those to just return unit.

## Acceptance criteria

- All of `LicenceScanning` uses `FossaApiClient`.
- Tests added where feasible

## Testing plan

The function exposed by `LicenseScanner` actually executes Themis.  Our test harnesses don't seem to have mock execution set up.  This makes testing hard because I'd need to create and run against a fake dependency.  That feels out of scope for this PR, so I didn't create any unit tests.

I tested by incrementally implementing the endpoints to ensure everything was working.

1. First, set up my local core to do local license scanning.
2. Run the CLI with local scanning against a vendored dep and observe the behavior, verifying it invoked local scanning.
3. Refactor the code up to the point that it would invoke the legacy API functions. 
 Those are replaced with dummy implementations using `fatalText`.
4. Run the CLI and observe the dummy errors, updating the triggered dummy implementation with a pass-through to the API V1 implementation.
5. Repeat until license scanning succeeds.
6. Double-check that all expected end-points were invoked to ensure I have full coverage.

Future testing is just to run a local licence scan and observe that it works.

## References

- [ANE-217](https://fossa.atlassian.net/browse/ANE-217): Convert License Scanning to use FossaApiClient

## Checklist

### DIFF
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -15,7 +15,7 @@ import Codec.Compression.GZip qualified as GZip
 import Control.Carrier.Diagnostics qualified as Diag
 import Control.Carrier.StickyLogger (StickyLogger, logSticky)
 import Control.Effect.Diagnostics (context)
-import Control.Effect.FossaApiClient (ArchiveRevision (ArchiveRevision), FossaApiClient, getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
+import Control.Effect.FossaApiClient (PackageRevision (PackageRevision), FossaApiClient, getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
 import Control.Effect.Lift
 import Control.Effect.Path (withSystemTempDir)
 import Control.Monad (unless)
@@ -90,7 +90,7 @@ compressAndUpload arcDir tmpDir VendoredDependency{..} = context "compressing an
     Nothing -> sendIO $ hashFile compressedFile
     Just version -> pure version
 
-  signedURL <- getSignedUploadUrl $ ArchiveRevision vendoredName depVersion
+  signedURL <- getSignedUploadUrl $ PackageRevision vendoredName depVersion
 
   logSticky $ "Uploading '" <> vendoredName <> "' to secure S3 bucket"
   res <- uploadArchive signedURL compressedFile
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -15,7 +15,7 @@ import Codec.Compression.GZip qualified as GZip
 import Control.Carrier.Diagnostics qualified as Diag
 import Control.Carrier.StickyLogger (StickyLogger, logSticky)
 import Control.Effect.Diagnostics (context)
-import Control.Effect.FossaApiClient (PackageRevision (PackageRevision), FossaApiClient, getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
+import Control.Effect.FossaApiClient (FossaApiClient, PackageRevision (PackageRevision), getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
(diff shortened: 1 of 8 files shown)


src/App/Fossa/LicenseScanner.hs
```diff
@@ -12,16 +12,15 @@ import App.Fossa.ArchiveUploader (
...
  )
 import App.Fossa.EmbeddedBinary (ThemisBins, withThemisAndIndex)
-import App.Fossa.FossaAPIV1 qualified as Fossa
 import App.Fossa.RunThemis (
   execThemis,
  )
 import Control.Carrier.Finally (runFinally)
 import Control.Effect.Diagnostics (Diagnostics, ToDiagnostic (renderDiagnostic), context, fatal, fatalText, fromMaybe, recover)
-import Control.Effect.FossaApiClient (FossaApiClient, getApiOpts)
+import Control.Effect.FossaApiClient (FossaApiClient, PackageRevision (PackageRevision, packageName, packageVersion), finalizeLicenseScan, getOrganization, getSignedLicenseScanUrl, uploadLicenseScanResult)
```
meghfossa, line 20: [optional] [nit] line break in the listing for better formatting

src/App/Fossa/ArchiveUploader.hs
```diff
@@ -15,7 +15,7 @@ import Codec.Compression.GZip qualified as GZip
 import Control.Carrier.Diagnostics qualified as Diag
 import Control.Carrier.StickyLogger (StickyLogger, logSticky)
 import Control.Effect.Diagnostics (context)
-import Control.Effect.FossaApiClient (ArchiveRevision (ArchiveRevision), FossaApiClient, getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
+import Control.Effect.FossaApiClient (FossaApiClient, PackageRevision (PackageRevision), getOrganization, getSignedUploadUrl, queueArchiveBuild, uploadArchive)
```
meghfossa, line 18: [optional] [nit] line break in the listing for better formatting

meghfossa, line 18: [optional] [nit] line break in the listing for better formatting

src/App/Fossa/LicenseScanner.hs
```diff
@@ -216,10 +213,10 @@ uploadVendoredDep apiOpts VendoredDependency{..} themisScanResult = do
     Nothing -> pure . showT . md5 . encodeUtf8 . show . NE.sort $ themisScanResult
     Just version -> pure version
 
-  signedURL <- Fossa.getSignedLicenseScanURL apiOpts depVersion vendoredName
+  signedURL <- getSignedLicenseScanUrl $ PackageRevision{packageVersion = depVersion, packageName = vendoredName}
```
meghfossa, line 216: [optional] [nit] why not `PackageRevision vendoredName depVersion` like you have in the Archive uploader?

meghfossa, line 216: [optional] [nit] why not `PackageRevision vendoredName depVersion` like you have in the Archive uploader?

meghfossa (approved):
left some options. Did not do e/e test but the changes look straightforward.

src/App/Fossa/LicenseScanner.hs
periodic: I actually think they should both be this way.  The point is to name these `Text` parameters that are being passed in so it's hard to accidentally get them switched around.

periodic: Looks like GitHub posted some duplicates.

PR #886, opened 2022-04-08
[CLI License Scan] Calculate the version from a hash of the vendored deps contents

spatten:
# Overview

This PR changes how we calculate the version of a vendored dependency when the version is not supplied in `fossa-deps.yml`.

This is the first requirement for implementing [avoiding rescanning vendored deps](https://fossa.atlassian.net/browse/ANE-190).

Previously we got the version of a vendored dependecy by calculating the hash of the results of the license scan. This was bad for two reasons. First, it meant we always had to license scan.

Second, it meant that we got different versions for cli-side license scans and archive uploads.

This PR fixes both of those problems by calculating the version of a cli-side license scanned vendored dependency in the same way that we calculate it for an archive-uploaded vendored dependency: by archiving the directory and calculating a hash of its contents.

## Acceptance criteria

- The version should be calculated using the directory contents
- The version calculated should be exactly the same as the version calculated by an archive upload for both archived vendored deps and unarchived vendored deps.

## Testing plan

clone the https://github.com/fossas/example-projects repo.

We are going to run `fossa analyze` on the `fossa-deps/vendored-dependencies` and `fossa-deps/unarchived-vendored-dependencies` projects, with and without the `--experimental-native-license-scan` flag.

Remove the version field from `fossa-deps/vendored-dependencies/fossa-deps.yml` and `fossa-deps/unarchived-vendored-dependencies/fossa-deps.yml`

```
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/example-projects/fossa-deps/vendored-dependencies
```

Note the version of the `Hello-Vendored` dependency.

Now re-scan with the `--experimental-native-license-scan` flag:

```
cabal run fossa -- analyze -e http://localhost:9578 --experimental-native-license-scan ~/fossa/example-projects/fossa-deps/vendored-dependencies
```

The version of the two `Hello-Vendored` revisions should be the same. For me it was `aeee3b54f8f3917e504b566cfc2a910e`.

Screenshot 2022-04-08 at 4 22 40 PM

Screenshot 2022-04-08 at 4 22 54 PM

Do the same thing for an unarchived vendored dependency:

```
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/example-projects/fossa-deps/unarchived-vendored-dependencies
cabal run fossa -- analyze -e http://localhost:9578 --experimental-native-license-scan ~/fossa/example-projects/fossa-deps/unarchived-vendored-dependencies
```

Those analyses should both result in the same revision for the `Hello-Vendored-Unarchived` dependency.

Screenshot 2022-04-08 at 4 32 19 PM

Screenshot 2022-04-08 at 4 32 32 PM

Change the contents of the `vendored/hello-1.0.0.2` directory and scan again:

```
echo hello > unarchived-vendored-dependencies/vendor/hello-1.0.0.2/hello.txt   
```

Then re-run the analyses:

```
cabal run fossa -- analyze -e http://localhost:9578 ~/fossa/example-projects/fossa-deps/unarchived-vendored-dependencies
cabal run fossa -- analyze -e http://localhost:9578 --experimental-native-license-scan ~/fossa/example-projects/fossa-deps/unarchived-vendored-dependencies
```

The revision for the `Hello-Vendored-Unarchived` dependency should change from your previous scans, but it should be the same for the archive-upload and the cli-side license scan.

Screenshot 2022-04-08 at 4 40 13 PM

Screenshot 2022-04-08 at 4 40 29 PM

## Risks

## Checklist

### DIFF
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -3,6 +3,7 @@
 module App.Fossa.ArchiveUploader (
   archiveUploadSourceUnit,
   arcToLocator,
+  compressFile,
   forceVendoredToArchive,
   duplicateFailureBundle,
   duplicateNames,
--- a/src/App/Fossa/LicenseScanner.hs
+++ b/src/App/Fossa/LicenseScanner.hs
@@ -8,8 +8,10 @@ module App.Fossa.LicenseScanner (
 import App.Fossa.ArchiveUploader (
   VendoredDependency (..),
   arcToLocator,
+  compressFile,
   duplicateFailureBundle,
   duplicateNames,
+  hashFile,
  )
 import App.Fossa.EmbeddedBinary (ThemisBins, withThemisAndIndex)
 import App.Fossa.FossaAPIV1 qualified as Fossa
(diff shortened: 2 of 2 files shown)


src/App/Fossa/ArchiveUploader.hs
```diff
@@ -3,6 +3,7 @@
 module App.Fossa.ArchiveUploader (
   archiveUploadSourceUnit,
   arcToLocator,
+  compressFile,
```
periodic, line 6: At some point soon we ought to move utilities like this to a common location so that the scanner doesn't depend on the uploader.  I'll let you make the call on whether that is now or not.

periodic (approved):
Looks good.  It solves the issue quickly and succinctly.

Seeing all this IO without effects makes me uncomfortable.  If you want to try to build more effects for this stuff I'd be more than happy. :smile:

PR #891, opened 2022-04-12
Rewrites the MockApi docs

periodic:
# Overview

Rewrites the `MockApi` docs to be a bit clearer about usage and implementation.

@scruffystuffs and I had a conversation about making the docs both a good source of usage and of design for this module.  I've attempted to make the docs useful for consumers of the interface as well as anyone who might want to contribute to the interface or understand it more deeply.

## Acceptance criteria

- Drew and Wes have a common understanding of the purposes and audiences of this doc.
- The doc satisfies that purposes for those audiences.

## References

_Example:_

- [ANE-188](https://fossa.atlassian.net/browse/ANE-188): Create contributor docs for API effects/mocking

## Checklist

### DIFF
--- a/docs/contributing/api-mocking.md
+++ b/docs/contributing/api-mocking.md
@@ -34,7 +34,7 @@ An expectation is considered _satisfied_ when it has been used to fulful the
 minimum number it expects.  An expectation is considered _exhausted_ if has
 fulfulled the maximum number of requests it expects.  At the end of a test all
 expectations should be satisfied, though they need not be exhausted.  Reusing an
-exhaused expectation is an error.
+exhausted expectation is an error.
 
 The helpers are:
 
--- a/docs/contributing/api-mocking.md
+++ b/docs/contributing/api-mocking.md
@@ -1,28 +1,60 @@
 # API Mocking
 
-Testing functions that interact with the API can be complicated because the API can have many different responses or errors, but the behavior is outside the scope of the test.  APIs can have errors, they may be stateful, there may be network instability and a host of other issues that conspire to cause test failures.  Unit tests shouldn't make any API calls because it makes the test slow, brittle and non-determistic which undermines many of the advantages of unit tests such as using them for CI on every change.
+Testing functions that interact with the API can be complicated because the API
+can have many different responses or errors, but the behavior is outside the
+scope of the test.  APIs can have errors, they may be stateful, there may be
+network instability and a host of other issues that conspire to cause test
+failures.  Unit tests shouldn't make any API calls because it makes the test
+slow, brittle and non-determistic which undermines many of the advantages of
+unit tests such as using them for CI on every change.
 
-The code that calls the API still needs to be tested in a predictable and reproducible way.  We need a way to remove all that uncertainty without introducing burdensome complexity in our testing environment such as running an entire application and networking stack.
+The code that calls the API still needs to be tested in a predictable and
+reproducible way.  We need a way to remove all that uncertainty without
+introducing burdensome complexity in our testing environment such as running an
+entire application and networking stack.
 
-The solution is mock the API.  This means that we will tell the test harness how to interpret calls to the API so that the code under test can experience different scenarios.  This is not a fake API, which would seek to mimic the API behavior, but rather a collection of requests and responses.
+The solution is mock the API.  This means that we will tell the test harness how
+to interpret calls to the API so that the code under test can experience
+different scenarios.  This is not a fake API, which would seek to mimic the API
+behavior, but rather a collection of requests and responses.
 
 ## Usage
 
-Using the Mock API is a matter of setting expectations of calls that will be made and at the same time stating what those calls should return.  This lets us tell the test how to handle the expected calls but it also tells the test that anything else is unexpected.  This is important because it guards against accidental calls so you never have to say, "I didn't know that function had stateful behavior!"
+Using the Mock API is a matter of setting expectations of calls that will be
+made and at the same time stating what those calls should return.  This lets us
+tell the test how to handle the expected calls but it also tells the test that
+anything else is unexpected.  This is important because it guards against
+accidental calls so you never have to say, "I didn't know that function had
+stateful behavior!"
 
-Expectations are added through helpers, such as `returnsOnce`.  Each of these take in a request to match against (from the effect) and then a behavior to execute when the request is made.
+Expectations are added through helpers, such as `returnsOnce`.  Each of these
+take in a request to match against (from the effect) and then a behavior to
+execute when the request is made.
 
 The helpers are:
 
 - ``req `returnsOnce` a`` - Expects a request of `req` for which it will return `a`.
-- ``req `alwaysReturns` a`` - Expects zero or requests of `req` for which it will always return `a`.
-- ``req `fails` "Mock Error"`` - Expects exactly one request matching `req` for which it will raise a `Diagnostics` error with the text `"Mock Error"`.
+- ``req `alwaysReturns` a`` - Expects zero or requests of `req` for which it
+  will always return `a`.
+- ``req `fails` "Mock Error"`` - Expects exactly one request matching `req` for
+  which it will raise a `Diagnostics` error with the text `"Mock Error"`.
 
-Expectations are added to the end of a list of expectations that are checked whenever a new request comes in.  That means you can specify a series of values to be returned on subsequent calls, but if you specify that a request should always return the same value it will never reach any expectations you add for the quest after that.
+Expectations are added to the end of a list of expectations that are checked
+whenever a new request comes in.  That means you can specify a series of values
+to be returned on subsequent calls, but if you specify that a request should
+always return the same value it will never reach any expectations you add for
+the quest after that.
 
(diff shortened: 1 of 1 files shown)


scruffystuffs:
Can we reflow the line lengths to ~80 chars?

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,175 @@
...
+Any unsatisfied expectations at the end of the test will trigger an assertion
+error.  Remember that the requests must match exactly (via `Eq`) in order to be
+satisfied..
+
+### Examples
 
-The solution is mock the API.  This means that we will tell the test harness how to interpret calls to the API so that the code under test can experience different scenarios.  This is not a fake API, which would seek to mimic the API behavior, but rather a collection of requests and responses.
+```haskell
+-- return the same value every time a call is made.
+GetOrganization `alwaysReturns` Fixtures.organization
```
meghfossa, line 63: I would like to see an end/end example of how to test a given function.  

```haskell
someFoo :: Has FossaAPIClient sig m => m Int
someFoo = do
     val <- API.someFunction
     case val of
          True -> 100
          False -> 200


someFoo_return0_when_APIFails... :: Spec
someFoo_return100_when_... :: Spec
someFoo_return200_when_... :: Spec
someFoo_always_return200_when_called_second_time_... :: Spec
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
 # API Mocking
 
-Testing functions that interact with the API can be complicated because the API can have many different responses or errors, but the behavior is defined outside this app.  The code that calls the API needs to be tested in a predictable and reproducible way.
+API testing can be complicated because the API can have many different responses
+or errors. Additionally the API behavior is generally outside the scope of the
+test.  APIs can have errors, they may be stateful, there may be network
+instability and a host of other issues that conspire to cause test failures.
+Unit tests shouldn't make any API calls because it makes the test slow, brittle
+and non-determistic which undermines many of the advantages of unit tests such
```
scruffystuffs, line 8:
```suggestion
and non-determistic, which undermines many of the advantages of unit tests such
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+as using them for CI on every change.
+
+The code that calls the API still needs to be tested in a predictable and
+reproducible way.  We need a way to remove all that uncertainty without
+introducing burdensome complexity in our testing environment such as running an
+entire application and networking stack.
+
+The solution is mock the API.  This means that we will tell the test harness how
+to interpret calls to the API so that the code under test can experience
+different scenarios.  This is not a fake API, which would seek to mimic the API
```
scruffystuffs, line 18: > This is not a fake API

Great line!  Stamp out potential mis-assumptions early.

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+Expectations are added through helpers, such as `returnsOnce`.  Each of these
+take in a request to match against (from the effect) and then a behavior to
+execute when the request is made.
+
+An expectation is considered _satisfied_ when it has been used to fulful the
+minimum number it expects.  An expectation is considered _exhausted_ if has
+fulfulled the maximum number of requests it expects.
+`returnsOnce` and `fails` are both satisfied and exhausted after one matching
+request.  `alwaysReturns` expects zero or more requests so it is always
+satisfied but never exhausted.
```
scruffystuffs, line 38: Is there a justification for some "one or more times" helper?  Or is that easily definable using `oneOrMore exp = returnsOnce exp >> alwaysReturns exp`?

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+
+The helpers are:
+
+- ``req `returnsOnce` a`` - Expects a request of `req` for which it will return
+  `a`.  This is both fulfilled and exhausted after one request.
```
scruffystuffs, line 43: Consistent naming
```suggestion
  `a`.  This is both satisfied and exhausted after one request.
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+guard against accidental calls so you never have to say, "I didn't know that
+function had stateful behavior!"
+
+Expectations are added through helpers, such as `returnsOnce`.  Each of these
+take in a request to match against (from the effect) and then a behavior to
+execute when the request is made.
+
+An expectation is considered _satisfied_ when it has been used to fulful the
+minimum number it expects.  An expectation is considered _exhausted_ if has
+fulfulled the maximum number of requests it expects.
```
scruffystuffs, line 35: Add two points here to prime the user for the details later:
- At the end of a test, all expectations must be satisfied (no expectation may go unused)
- Trying to reuse an exhausted expectation will raise an error

scruffystuffs: Maybe a third point:
- expectation do NOT have to be exhausted at the end of a test

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+satisfied but never exhausted.
+
+The helpers are:
+
+- ``req `returnsOnce` a`` - Expects a request of `req` for which it will return
+  `a`.  This is both fulfilled and exhausted after one request.
+- ``req `alwaysReturns` a`` - Expects zero or more requests of `req` for which it
+  will always return `a`.  This is always satisfied and never exhausted.
+- ``req `fails` "Mock Error"`` - Expects exactly one request matching `req` for
+  which it will raise a `Diagnostics` error with the text `"Mock Error"`.  This is both fulfilled and exhausted after one request.
```
scruffystuffs, line 47: Consistent naming
```suggestion
  which it will raise a `Diagnostics` error with the text `"Mock Error"`.  This is both satisfied and exhausted after one request.
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+
+Expectations are added to the end of a list of expectations that are checked
```
scruffystuffs, line 49: Shorten and force the user into thinking about list order semantically.
```suggestion
Expectations are added to the end of a list of expectations that are checked
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
+whenever a new request comes in.  That means you can specify a series of values
+to be returned on subsequent calls by specifying multiple expectations. Note
+that if you specify that a request should always return the same value it will
+never reach any expectations you add for the request after that.
```
scruffystuffs, line 53:
```suggestion
never reach any expectations you add for the request after that, and those
expectations will never be satisfied, causing a test error.
```

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,176 @@
...
-```
+The repetition specifies how to exhaust the the expectation.  It's currently
+only two options, but there is room for expansion.
+
+- `Once` means the expectation should be exhausted after one match.
+- `Always` means the expectation should never be exhausted.
+
+The request is used to match incoming requests with `Eq`.  It would be more
+flexible as a predicate, but this would make error messages harder to write.  A
+copy of the request is very useful in debugging and writing errors.
```
scruffystuffs, line 143: It may be worth coming up with some error-friendly comparison typeclass based on `Eq`, but that's a later improvement if we need it..

scruffystuffs (approved):
Left a bunch of comments, but this is great.  I didn't have to check back on the code once while reading this.

docs/contributing/api-mocking.md
periodic: Whoops, I kept wanting to use the term "fulfilled" but I've seen "satisfied" for expectations more often.

docs/contributing/api-mocking.md
periodic: It could exist, I just didn't need it yet and we could add it later.

Another thing we might want to add at some point is ordering of requests.

docs/contributing/api-mocking.md
periodic: Yeah, I'd really like it to figure this out automatically, which might mean we need a new typeclass (hopefully not TH) or to use Typeable.

scruffystuffs: Yeah this is non-trivial, but if we need better equality comparison in general, then a `StructuralDiff` typeclass might be necessary.

docs/contributing/api-mocking.md
```diff
@@ -1,53 +1,215 @@
...
+
+Expectations are added through helpers, such as `returnsOnce`.  Each of these
+take in a request to match against (from the effect) and then a behavior to
+execute when the request is made.
+
+An expectation is considered _satisfied_ when it has been used to fulful the
+minimum number it expects.  An expectation is considered _exhausted_ if has
+fulfulled the maximum number of requests it expects.  At the end of a test all
+expectations should be satisfied, though they need not be exhausted.  Reusing an
+exhaused expectation is an error.
```
scruffystuffs, line 37:
```suggestion
exhausted expectation is an error.
```

PR #896, opened 2022-04-14
Update to GHC 9.0

scruffystuffs:
# Overview

All code and CI changes needed to update to using GHC 9.0.

## Acceptance Criteria

- [x] HLS works when installed by `ghcup` or with `vscode` extension. i.e. no manual install _required_ (It's always optionally available)
  - `ghcup compile hls ...` may be required, but it should be a single command and no shenanigans.
- [ ] Code compiles on all 4 platforms:
  - [x] Linux
  - [x] Windows
  - [x] x86_64 Mac OSX
  - [ ] M1 Mac OSX
- [ ] CI works
  - [x] linux container image for static builds has a tag for GHC 9.0.2
  - [x] haskell-dev-tools has a tag for GHC 9.0.2
  - [ ] Binary packaging and release still works (testable with pre-release)
- [x] Automated tests pass
  - [x] Unit tests
  - [x] Integration tests
- [ ] Manual tests pass
  - [x] Dogfooding
  - [ ] [example-projects](https://github.com/fossas/example-projects) repo
  - [ ] VSI demo
  - [ ] CLI license scan
  - [ ] Archive upload
- [ ] TH-specific tests pass
  - [ ] `--version`
  - [ ] Debug bundle


scruffystuffs:
I'll be working on this over the weekend, but it'll wait until next week to merge, at minimum.

meghfossa:
@scruffystuffs let's include `arm64/linux` on the code compilation list and updates to `HACKING.md`.

scruffystuffs:
@meghfossa what do you mean by "code compilation list"?  Do you mean adding M1 as a platform for release binaries?  If so, that can be follow-up work, I'd really like to limit this to ONLY what's strictly necessary for GHC 9 and the relevant tooling.

cabal.project
```diff
@@ -54,8 +54,9 @@ source-repository-package
...
   subdir: x509-system
 
-index-state: hackage.haskell.org 2022-02-02T15:17:53Z
+index-state: hackage.haskell.org 2022-04-13T23:32:58Z
 
 package *
   extra-include-dirs: /opt/homebrew/include
-  extra-lib-dirs: /opt/homebrew/lib
\ No newline at end of file
+  extra-lib-dirs: /opt/homebrew/lib
```
periodic, line 61: Seems weird to have absolute paths in our `cabal.project`.  They won't exist on all systems.  Shouldn't these be in a `macos` specific version?

periodic (approved):
One comment.  Otherwise it looks like we just need to wrap up the testing and we should be good to go.

cabal.project
scruffystuffs: They're ignored if they aren't present.  This is also something we can fix later, it just happened to be relevant to get @kitified to be able to build GHC 9 on her machine.  I expect someone else to run into this, but I think we can fix it reactively instead of proactively without too much trouble.

PR #897, opened 2022-04-14
Improve debug bundles.

scruffystuffs:
# Overview

We currently leak API keys in debug bundles if the key is in the config file or the CLI args.  We also do not capture the CLI version in the debug bundle.  This resolves both of those problems with the following changes

- We no longer capture the config file text, only the resolved configuration settings.
- We redact the api key from the command line using the same util as telemetry
- We add the version text to the debug bundle.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.2.7
+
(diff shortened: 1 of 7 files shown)


meghfossa (approved):
lgtm - let's add this to changelog?

PR #898, opened 2022-04-14
Makes GetCurrendDir into a ReadFS effect

periodic:
# Overview

This makes an effect for `GetCurrentDir` as part of `ReadFS`.  This was one of the simplest changes identified in an inventory of our `sendIO` usage.

The main target here is config-file loading, which I standardized a bit so that each file wasn't reading in the directory independently.  For that I introduced a function to load the config file with the current dir as the default, which most commands use.

Other than that there was some usage for finding the directory to run Syft.

## Acceptance criteria

- There are fewer usages of `getCurrentDir` in the codebase.
- Other places we get the current dir are updated to use `ReadFS`

## Testing plan

One change here is that we now always print the config file as debug information for all the commands, instead of just a few.  This made it easy to run various comnmands and make sure that it was working in the new common code path.

I also added unit tests for `analyze` and `test` config loading.

## Checklist

### DIFF
--- a/src/App/Fossa/Config/ConfigFile.hs
+++ b/src/App/Fossa/Config/ConfigFile.hs
@@ -15,6 +15,7 @@ module App.Fossa.Config.ConfigFile (
   ExperimentalGradleConfigs (..),
   mergeFileCmdMetadata,
   empty,
+  resolveLocalConfigFile,
 ) where
 
 import App.Docs (fossaYmlDocUrl)
@@ -48,10 +49,11 @@ import Effect.Logger (
   Doc,
   Logger,
   Pretty (pretty),
+  logDebug,
   logWarn,
   vsep,
  )
-import Effect.ReadFS (ReadFS, doesFileExist, readContentsYaml)
+import Effect.ReadFS (ReadFS, doesFileExist, getCurrentDir, readContentsYaml)
 import Path (
   Abs,
   Dir,
@@ -75,6 +77,18 @@ data ConfigLocation
   = DefaultConfigLocation (Path Abs Dir)
   | SpecifiedConfigLocation (Path Abs File)
 
+resolveLocalConfigFile ::
+  ( Has Diagnostics sig m
+  , Has (Lift IO) sig m
+  , Has Logger sig m
+  , Has ReadFS sig m
+  ) =>
+  Maybe FilePath ->
+  m (Maybe ConfigFile)
+resolveLocalConfigFile path = do
+  cwd <- getCurrentDir
+  resolveConfigFile cwd path
+
 resolveConfigFile ::
   ( Has Diagnostics sig m
   , Has (Lift IO) sig m
@@ -87,6 +101,7 @@ resolveConfigFile ::
   Maybe FilePath ->
   m (Maybe ConfigFile)
 resolveConfigFile base path = do
+  logDebug $ "Loading configuration file from " <> pretty (show base)
   loc <- resolveLocation base path
   case loc of
     DefaultConfigLocation dir -> do
--- a/src/App/Fossa/Config/ConfigFile.hs
+++ b/src/App/Fossa/Config/ConfigFile.hs
@@ -51,6 +51,7 @@ import Effect.Logger (
   Pretty (pretty),
   logDebug,
   logWarn,
+  viaShow,
   vsep,
  )
 import Effect.ReadFS (ReadFS, doesFileExist, getCurrentDir, readContentsYaml)
@@ -101,7 +102,7 @@ resolveConfigFile ::
   Maybe FilePath ->
   m (Maybe ConfigFile)
 resolveConfigFile base path = do
-  logDebug $ "Loading configuration file from " <> pretty (show base)
+  logDebug $ "Loading configuration file from " <> viaShow base
   loc <- resolveLocation base path
   case loc of
     DefaultConfigLocation dir -> do
--- a/src/App/Fossa/Config/Container.hs
+++ b/src/App/Fossa/Config/Container.hs
@@ -76,8 +76,7 @@ loadConfig = \case
   ContainerParseFile _ -> pure Nothing
   ContainerDumpScan _ -> pure Nothing
   -- Only parse config file if we're running analyze or test
-  cmd -> do
-    resolveLocalConfigFile $ getCfgFilePath cmd
+  cmd -> resolveLocalConfigFile $ getCfgFilePath cmd
 
 getCfgFilePath :: ContainerCommand -> Maybe FilePath
 getCfgFilePath = \case
--- a/src/App/Fossa/Config/Container.hs
+++ b/src/App/Fossa/Config/Container.hs
@@ -15,7 +15,7 @@ import App.Fossa.Config.Common (
  )
 import App.Fossa.Config.ConfigFile (
   ConfigFile,
-  resolveConfigFile,
+  resolveLocalConfigFile,
  )
 import App.Fossa.Config.Container.Analyze (ContainerAnalyzeConfig, ContainerAnalyzeOptions (..))
 import App.Fossa.Config.Container.Analyze qualified as Analyze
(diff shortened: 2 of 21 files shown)


src/Control/Effect/Replay/TH.hs
```diff
@@ -73,15 +73,24 @@ decodeSingle keyVal valVal con = do
...
-      noBindS [e|pure (EffectResult $(conAppliedE) $(varE resultNm))|]
-    ]
+    ( argBindingsE
+        <> [
+             -- decodedResultValue <- fromRecordedValue valVal
+             bindS (sigP (varP resultNm) (getGadtTy con)) [e|fromRecordedValue $(varE valVal)|]
+           , -- pure (EffectResult (Bar a b) decodedResultValue)
+             noBindS [e|pure (EffectResult $(conAppliedE) $(varE resultNm))|]
+           ]
+    )
```
scruffystuffs, line 93: Is this change related?  I'd prefer to leave it out if not, TH code is really hard to review, I don't want to block the new effect constructor for this.

src/Effect/Exec.hs
```diff
@@ -218,6 +220,15 @@ execThrow dir cmd = context ("Running command '" <> cmdName cmd <> "'") $ do
...
     Right stdout -> pure stdout
 
+-- | A variant of 'execThrow' that runs the command in the currend directory
+execThrow' :: (Has Exec sig m, Has ReadFS sig m, Has Diagnostics sig m) => Command -> m BL.ByteString
+execThrow' cmd = context ("Running command '" <> cmdName cmd <> "'") $ do
+  dir <- getCurrentDir
+  result <- exec dir cmd
+  case result of
+    Left failure -> fatal (CommandFailed failure)
+    Right stdout -> pure stdout
```
scruffystuffs, line 230: DRY:
```suggestion
  execThrow dir cmd
```

src/Effect/ReadFS.hs
```diff
@@ -351,6 +367,9 @@ runReadFSIO = interpret $ \case
   -- NB: these never throw
   DoesFileExist file -> sendIO (Directory.doesFileExist (fromSomeFile file))
   DoesDirExist dir -> sendIO (Directory.doesDirectoryExist (fromSomeDir dir))
+  GetCurrentDir -> do
+    PIO.getCurrentDir
+      `catchingIO` CurrentDirError
```
scruffystuffs, line 372: If this throws on some occasions, put it above the NB comment.

src/App/Fossa/Config/LinkUserBinaries.hs
```diff
@@ -94,9 +93,8 @@ loadConfig ::
   ) =>
   LinkUserBinsOpts ->
   m (Maybe ConfigFile)
-loadConfig LinkUserBinsOpts{..} = do
-  curdir <- sendIO getCurrentDir
-  resolveConfigFile curdir $ optConfig commons
+loadConfig =
+  resolveLocalConfigFile . optConfig . commons
```
scruffystuffs, line 97:
```suggestion
loadConfig = resolveLocalConfigFile . optConfig . commons
```

src/App/Fossa/Config/Container.hs
```diff
@@ -78,8 +77,7 @@ loadConfig = \case
   ContainerDumpScan _ -> pure Nothing
   -- Only parse config file if we're running analyze or test
   cmd -> do
-    curdir <- sendIO getCurrentDir
-    resolveConfigFile curdir $ getCfgFilePath cmd
+    resolveLocalConfigFile $ getCfgFilePath cmd
```
scruffystuffs, line 80: Remove `do` and one-line this

scruffystuffs (changes requested):
Mostly nits, but I'd like to undo the TH changes if possible.  If not, can you leave a comment in the PR explaining why we had to make a change there?

src/Control/Effect/Replay/TH.hs
periodic: Oh, I forgot to explain that.  The TH we have doesn't seem to support no-arg constructors.  It ended up with missing the ReplayableData instance on `Unit(String)`.  I'm not sure what that means though, and it didn't seem necessary to match anyway so I did this instead.

I'll take another look tomorrow.

periodic:
Did all the little changes.  Still have to figure out if the TH can be removed.

src/Control/Effect/Replay/TH.hs
periodic: I removed the `deriveReplayable` and reverted this, as you suggested.  I don't think it was correct anyway.

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -87,6 +101,7 @@ resolveConfigFile ::
   Maybe FilePath ->
   m (Maybe ConfigFile)
 resolveConfigFile base path = do
+  logDebug $ "Loading configuration file from " <> pretty (show base)
```
scruffystuffs, line 104: [nit]: use helper
```suggestion
  logDebug $ "Loading configuration file from " <> viaShow base
```

src/Effect/Exec.hs
```diff
@@ -39,9 +40,10 @@ import Data.String.Conversion (decodeUtf8, toString, toText)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Void (Void)
+import Effect.ReadFS (ReadFS, getCurrentDir)
 import GHC.Generics (Generic)
 import Path
-import Path.IO
+import Path.IO (AnyPath (makeAbsolute))
```
scruffystuffs, line 46: Let's make all imports explicit here

test/App/Fossa/Config/Utils.hs
```diff
@@ -0,0 +1,44 @@
...
+
+-- | Tests that the config loader uses the directory set in the arguments
+itShouldLoadFromTheConfiguredBaseDir ::
+  Parser opts -> (opts -> EffectStack (Maybe ConfigFile)) -> Spec
+itShouldLoadFromTheConfiguredBaseDir parser loadConfig =
+  it' "should load from the configured base dir"
+    . withTempDir "AnalyzeSpec"
+    $ \tempDir ->
+      do
+        sendIO $ BS.writeFile (toFilePath (tempDir </> $(mkRelFile ".fossa.yml"))) "version: 42\n"
```
scruffystuffs, line 29: [suggestion]: if you use `version: 3`, you'll be able to use `mempty` for the expected config file value.

scruffystuffs (approved):
Few nits, but this looks good.  another effect in the bag.

src/Effect/Exec.hs
periodic: You sure? It's a lot. :D

test/App/Fossa/Config/Utils.hs
periodic: I want to be sure that we don't also default the version because we can't find the config.  I want to be sure that the version came from the file I created so I set it to an unusual value.

scruffystuffs: We don't have a default config file, we handle a `Maybe ConfigFile` everywhere

scruffystuffs: This is not a big deal, just making clear that this suggestion would be semantically safe.

periodic: That's true.  I think this makes it abundantly clear that there's no way it's falling back to defaults so it's less likely the test could give a false positive.

periodic: In a way what I want to test is that the config file is read, not that the config file matches a default, so I think this better describes the test.

We should have some tests about the default value though. :D

scruffystuffs: If you just want to test that the file is read, you should use `isJust` instead.  That way the value doesn't matter at all.  As I said, we NEVER fallback to a default config file.

PR #901, opened 2022-04-14
No discovery filters

scruffystuffs:
Turn off discovery filters for debug and unblocking purposes.

The flag has a `debug` prefix, and the help text warns against regular use of this flag.


src/App/Fossa/Config/Analyze.hs
```diff
@@ -243,6 +250,7 @@ cliParser =
     <*> flagOpt UnpackArchives (long "unpack-archives" <> help "Recursively unpack and analyze discovered archives")
     <*> flagOpt JsonOutput (long "json" <> help "Output project metadata as json to the console. Useful for communicating with the FOSSA API")
     <*> flagOpt IncludeAll (long "include-unused-deps" <> help "Include all deps found, instead of filtering non-production deps.  Ignored by VSI.")
+    <*> flagOpt NoDiscoveryExclusion (long "debug-no-discovery-exclusion" <> help "Ignore filters during discovery phase.  This is for debugging only and may be removed without warning.")
```
zlav, line 253: Can we make this hidden as well?
```suggestion
    <*> flagOpt NoDiscoveryExclusion (long "debug-no-discovery-exclusion" <> help "Ignore filters during discovery phase.  This is for debugging only and may be removed without warning" <> hidden)
```

Also very small nit, we're inconsistent about using periods.

zlav (approved):
LGTM!

PR #902, opened 2022-04-14
Discovery exclusion reads target filters wrong

scruffystuffs:
# Overview

When a user has 2 project of the same type, and explicitly excludes only one of them with `tool@path/`, discovery exclusion should not treat that tool as forbidden.  

See new unit test for a clearer example


csasarak (approved):
LGTM !

PR #903, opened 2022-04-15
Skip embedding files if FOSSA_SKIP_EMBED_FILE_IN_HLS env var is set

spatten:
# Overview

HLS's memory usage blows up when it has to open `EmbeddedBinary.hs` or when you compile with a change to that file.

This PR gets rid of that problem by adding an environment variable that, if set, turns off embedding.

You use it by setting the environment variable when running HLS. This is documented in a change to `HACKING.md` in the PR.

## Acceptance criteria

* HLS's memory footprint should stay small, even when you open or change `EmbeddedBinary.hs`
* Nothing else should change

## Testing plan

I tested by making changes to `EmbeddedBinary.hs` with this change and with the environment variable set properly in VS Code. I think it works. HLS's memory usage always stayed below 2 GB, and without this change it can balloon up to > 15 GB.

## Checklist

### DIFF
--- a/src/Data/FileEmbed/Extra.hs
+++ b/src/Data/FileEmbed/Extra.hs
@@ -11,7 +11,7 @@ import Prelude
 
 embedFileIfExists :: FilePath -> Q Exp
 embedFileIfExists inputPath = do
-  skipEmbedEnvVar <- runIO $ lookupEnv "FOSSA_SKIP_EMBEDDED_FILES_IN_HLS"
+  skipEmbedEnvVar <- runIO $ lookupEnv "FOSSA_SKIP_EMBED_FILE_IN_HLS"
   case (skipEmbedEnvVar, parseRelFile inputPath) of
     (Just _, _) -> do
       pure (LitE $ StringL "")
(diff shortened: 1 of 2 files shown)


src/Data/FileEmbed/Extra.hs
```diff
@@ -6,16 +6,20 @@ import Data.FileEmbed (embedFile)
...
 
 embedFileIfExists :: FilePath -> Q Exp
 embedFileIfExists inputPath = do
-  case parseRelFile inputPath of
-    Just path -> do
+  skipEmbedEnvVar <- runIO $ lookupEnv "FOSSA_SKIP_EMBED_FILE_IN_HLS"
+  case (skipEmbedEnvVar, parseRelFile inputPath) of
+    (Just _, _) -> do
+      pure (LitE $ StringL "")
+    (_, Just path) -> do
```
scruffystuffs, line 18: This can wait till later, but I think we should refactor a few things here:

- `Q Exp` is an untyped haskell expression, but we should make it a typed haskell expression: `Q TExp`.
- We should split the logical checks of env var and file existence.
- We shouldn't be relying on `OverloadedStrings`, this should work even if that extension is disabled.  (`pure (LitE $ StringL "")` only works because `Bytestring` implements `IsString`)

src/Data/FileEmbed/Extra.hs
```diff
@@ -6,16 +6,20 @@ import Data.FileEmbed (embedFile)
 import Language.Haskell.TH
 import Path
 import Path.IO
+import System.Environment (lookupEnv)
 import Prelude
```
scruffystuffs, line 10: [nit]: Please remove the prelude import and make all other imports explicit.

scruffystuffs (approved):
One nit and one future comment, but we can merge this as-is

PR #907, opened 2022-04-18
Don't skip non-included paths

scruffystuffs:
# Overview

Given filters of the shape `include a/b && include a/c`, we would try to skip BOTH directories, because the parent directories wouldn't match the only existing filters.

To fix this, we no longer use `include` paths to do discovery exclusion.  Later, we may be able to limit discovery to paths that are proper prefixes of at least one include filter (if and only if any include filters exist), but for now we just need to fix the bug.

## Acceptance criteria

`--only-path` or `paths.only` does not exclude paths from discovery.

## Testing plan

Given the following folder structure, run `fossa analyze -o /path/to --only-path test-arena/testpy1 --only-path test-arena/testpy2 > /dev/null`.  The scan summary should show that both setuptools projects were found and successfully scanned (there will be warnings, but that's normal for setuptools).

```
/path/to
└── test-arena
    ├── testpy1
    │   └── requirements.txt
    └── testpy2
        └── requirements.txt
```

## Risks

None, this postpones the risky change for later.

## References

https://fossa.zendesk.com/agent/tickets/4194

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## v3.2.8
 
+- Filtering: Don't use included paths for discovery exclusion. ([#907](https://github.com/fossas/fossa-cli/pull/907))
 - Filtering: add `--debug-no-discovery-exclusion` for client-side filter debugging. (#[901](https://github.com/fossas/fossa-cli/pull/901))
 
 ## v3.2.7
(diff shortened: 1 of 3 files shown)


kitified (approved):
this lgtm based on automated tests & code, if you want a double check via running the code lmk!

PR #911, opened 2022-04-21
Require integration tests before release

zlav:
# Overview

When we initially added integration tests we did not mandate them to pass before releases due to the amount of time we believed they would take to run. We have seen that the last few successful tests took <10 minutes which is far less than the 24 minutes I have seen the macOS build take. I don't think that requiring integration tests before a build will slow down development and we should take advantage of the tests that we've already written.

## Acceptance criteria

- Integration tests are required to pass before merging to master.
- Integration tests are passing again.

## Testing plan

- Merging this PR with a passing "Required" integration tests check.

## Risks

- Our integration tests are flaky - Then we need to fix out integration tests.
- Integration tests take too long to run - We can look into making integration tests as a requirement only for releases, although I fear this would allow merging code that breaks integration tests.

## Checklist


zlav:
Closing in favor of https://github.com/fossas/fossa-cli/pull/912

PR #913, opened 2022-04-21
[ANE-82]: Native license scan output

scruffystuffs:
# Overview

Provide an analog to `fossa analyze --output`, but for the results that we upload during native license scanning.

We did not add this as part of the existing `fossa analyze --output` mechanism because the existing pipeline of `fossa analyze` assumes that in output mode, we don't run the license scan at all.  Also, the existing license scan pipeline is heavily tied to the upload process.  Instead, we make a new pipeline entirely out of the existing license scan primitives (with a tiny refactor to give us access to the primitives).

## Acceptance criteria

The output of `fossa experimental-license-scan fossa-deps /path/to/dir/with/fossa-deps/file/` should be a collection of all of the uploads from running `fossa analyze` against that same directory.

Note that because the actual uploads are entirely separate, we cannot match the format of one JSON blob to many, but the output format should be a thin wrapper around the data, and the data within should match perfectly.

## Testing plan

Run `fossa experimental-license-scan fossa-deps` on a directory with a `fossa-deps` file that has the following `vendored-dependencies` (these will probably have to be separate runs):

- One local directory
- One archive
- 2+ local directories
- 2+ archive
- One or more of each

## Risks

Please check that the refactor does not affect the upload of the actual `fossa analyze` process.  You can check this locally with minio.

## Checklist

### DIFF
--- a/src/App/Fossa/Config/LicenseScan.hs
+++ b/src/App/Fossa/Config/LicenseScan.hs
@@ -74,5 +74,5 @@ cliParser = public <|> private
       command
         "direct"
         ( info (DirectScan <$> baseDirArg) $
-            progDesc "Run a license scan directly only the provided path."
+            progDesc "Run a license scan directly on the provided path."
         )
--- a/src/App/Fossa/Config/LicenseScan.hs
+++ b/src/App/Fossa/Config/LicenseScan.hs
@@ -1,34 +1,48 @@
 module App.Fossa.Config.LicenseScan (
   mkSubCommand,
   LicenseScanConfig (..),
-  LicenseScanOpts,
+  LicenseScanCommand,
 ) where
 
-import App.Fossa.Config.Common (baseDirArg, validateDir)
+import App.Fossa.Config.Common (baseDirArg, collectBaseDir)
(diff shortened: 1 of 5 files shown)


src/App/Fossa/Config/LicenseScan.hs
```diff
@@ -40,6 +54,25 @@ mergeOpts ::
...
+        "fossa-deps"
+        ( info (FossaDeps <$> baseDirArg) $
+            progDesc "Like `fossa analyze --output`, but only for native scanning of vendored-dependencies."
+        )
+    directScanCommand =
+      command
+        "direct"
+        ( info (DirectScan <$> baseDirArg) $
+            progDesc "Run a license scan directly on the provided path."
+        )
```
zlav, line 78: I think removing this `direct` argument and making direct license scanning the default behavior would be a more straightforward UX.

zlav (approved):
One nit about setting the default behavior to license scanning, otherwise LGTM!

src/App/Fossa/Config/LicenseScan.hs
zlav: Oh, I don't think I recognized that one of the commands was hidden and that one of them wasn't. Either way thats fine with me.

PR #918, opened 2022-04-28
Go: Skip reporting stdlib packages

kitified:
# Overview

Updates the Go strategy to filter standard library packages from the reported graph.
We use `go list std` to power the filtration; this means that we'll always filter the correct set of standard library dependencies depending on the Go version the user has installed.

`go list std` has been in Go since at least `go1.5`: https://pkg.go.dev/cmd/go@go1.5#hdr-Description_of_package_lists
I can't see documentation older than that but I think it's safe to assume users won't run into issues with this command.

## Acceptance criteria

Go standard library packages are no longer reported for projects.

## Testing plan

I've attached an archive with a reproduction case: [repro.zip](https://github.com/fossas/fossa-cli/files/8578517/repro.zip)

1. Place this project in your `go env GOPATH`.
For me, that's `/Users/kit/go`, so I put it at `/Users/kit/go/src/github.com/kitified/gotest/gomod`.
You may need to rewrite the module name to match your user; sorry, that's Go pre-modules 😬 
2. Set the environment variable `GO111MODULES=off`.
Steps 1 and 2 are required to get the primary tactic, `go list -m -json all`, to fail.
3. Run `fossa analyze -o | jq` on this project. You should see the below (note the `log` and `github.com/kitified/gotest/gomod` deps):
```json
{
  "projects": [
    {
      "type": "gomod",
      "graph": {
        "assocs": [
          [
            1,
            [
              0,
              2
            ]
          ]
        ],
        "deps": [
          {
            "tags": {},
            "name": "github.com/google/uuid",
            "type": "GoType",
            "locations": [],
            "version": {
              "type": "EQUAL",
              "value": "v1.3.0"
            }
          },
          {
            "tags": {},
            "name": "github.com/kitified/gotest/gomod",
            "type": "GoType",
            "locations": [],
            "version": null
          },
          {
            "tags": {},
            "name": "log",
            "type": "GoType",
            "locations": [],
            "version": null
          }
        ],
        "direct": [
          0
        ]
      },
      "path": "/Users/kit/go/src/github.com/kitified/gotest/gomod/"
    }
  ],
  "sourceUnits": [
    {
      "Build": {
        "Dependencies": [
          {
            "imports": [],
            "locator": "go+github.com/google/uuid$v1.3.0"
          },
          {
            "imports": [
              "go+github.com/google/uuid$v1.3.0",
              "go+log$"
            ],
            "locator": "go+github.com/kitified/gotest/gomod$"
          },
          {
            "imports": [],
            "locator": "go+log$"
          }
        ],
        "Imports": [
          "go+github.com/google/uuid$v1.3.0"
        ],
        "Succeeded": true,
        "Artifact": "default"
      },
      "Name": "/Users/kit/go/src/github.com/kitified/gotest/gomod/",
      "Type": "gomod",
      "GraphBreadth": "partial",
      "OriginPaths": [
        "go.mod"
      ],
      "Manifest": "/Users/kit/go/src/github.com/kitified/gotest/gomod/",
      "AdditionalDependencyData": null
    }
  ]
}
```
4. Check out and compile this branch, then run `cabal build && cabal run fossa -- analyze -o | tail -n 1 | jq`.
5. You should see the below (note the `log` and `github.com/kitified/gotest/gomod` deps are no longer showing):
```json
{
  "projects": [
    {
      "type": "gomod",
      "graph": {
        "assocs": [],
        "deps": [
          {
            "tags": {},
            "name": "github.com/google/uuid",
            "type": "GoType",
            "locations": [],
            "version": {
              "type": "EQUAL",
              "value": "v1.3.0"
            }
          }
        ],
        "direct": [
          0
        ]
      },
      "path": "/Users/kit/go/src/github.com/kitified/gotest/gomod/"
    }
  ],
  "sourceUnits": [
    {
      "Build": {
        "Dependencies": [
          {
            "imports": [],
            "locator": "go+github.com/google/uuid$v1.3.0"
          }
        ],
        "Imports": [
          "go+github.com/google/uuid$v1.3.0"
        ],
        "Succeeded": true,
        "Artifact": "default"
      },
      "Name": "/Users/kit/go/src/github.com/kitified/gotest/gomod/",
      "Type": "gomod",
      "GraphBreadth": "complete",
      "OriginPaths": [
        "go.mod"
      ],
      "Manifest": "/Users/kit/go/src/github.com/kitified/gotest/gomod/",
      "AdditionalDependencyData": null
    }
  ]
}
```

## Risks

This makes it to where users will no longer have the Go standard library reported as dependencies in their projects.
For any org that cares about this, this would be a regression. However, as far as we can tell no org on cloud has any issue related to the Go standard library in Core, so we do not believe this would affect anyone.

If it turns out this _does_ affect an org, and they're unwilling to view this in a positive light, we can always add a flag to disable this filtering behavior.

## References

This ticket has been a long time coming but was brought to a head when a user had local deps flagged as missing in [ZD 4098](https://fossa.zendesk.com/agent/tickets/4098), and then I learned that updating the default FOSSA Go version is tricky to do correctly and potentially risky ([Core #7562](https://github.com/fossas/FOSSA/pull/7562)).

Rather than go down that road I thought it best to fix it once and for all.

## Checklist


src/Strategy/Go/Gostd.hs
```diff
@@ -0,0 +1,87 @@
...
+    { cmdName = "go"
+    , cmdArgs = ["list", "std"]
+    , cmdAllowErr = Never
+    }
+
+listGoStdlibPackages :: (Has Exec sig m, Has Diagnostics sig m) => Path Abs Dir -> m [GoStdlibDep]
+listGoStdlibPackages root = execParser parseGoStdlibPackages root goListStdCommand
+
+parseGoStdlibPackages :: Parser [GoStdlibDep]
+parseGoStdlibPackages = many parseGoStdlibPackage <* eof
```
periodic, line 48: Do you not need a separator here?  Maybe I don't know `Megaparsec` well enough, but I don't see any place that you are allowing any separator between them.  Just lots of `ident` back to back.

src/Strategy/Go/Gostd.hs
```diff
@@ -0,0 +1,87 @@
...
+parseGoStdlibPackage = mkDep <$> ident
+
+filterGoStdlibPackages :: [GoStdlibDep] -> Graphing Dependency -> Graphing Dependency
+filterGoStdlibPackages stdlib = Graphing.filter $ not . shouldFilter
+  where
+    stdlib' :: Set Text
+    stdlib' = Set.fromList $ fmap goStdlibDepName stdlib
+
+    isStdLib :: Dependency -> Bool
+    isStdLib = flip Set.member stdlib' . dependencyName
```
periodic, line 60: I've had hlint in the past recommend ``(`Set.member` stdlib')`` instead of `flip Set.member stdlib'`.  Do we not do that?  I personally like the way it reads a bit better.

src/Strategy/Go/Gostd.hs
```diff
@@ -0,0 +1,87 @@
...
+
+    isExtra :: Dependency -> Bool
+    isExtra = flip Set.member extraPackageFilters . dependencyName
```
periodic, line 63: Same as above.

src/Strategy/Go/Gostd.hs
```diff
@@ -0,0 +1,87 @@
...
+
+    shouldFilter :: Dependency -> Bool
+    shouldFilter dep = isStdLib dep || isExtra dep
+
+-- | These aren't part of the standard library, but are extra packages we also want to ignore.
+-- Each package should have a comment explaining what it is.
+extraPackageFilters :: Set Text
+extraPackageFilters =
+  Set.fromList
+    [ "C" -- Not a real package - this is a magic import used by Go to declare C interop: https://go.dev/blog/cgo
```
periodic, line 73: Bonus functionality!  Is this not included in the changelog for some reason?

periodic (approved):
LGTM with minor comments.

src/Strategy/Go/Gostd.hs
kitified: We don't have a hint for it, but I can do it this way. I don't love infix operators in general but meh.

src/Strategy/Go/Gostd.hs
kitified: It is included- despite this not being reported in `go list std` the `C` magic import is definitely part of the standard library.

PR #924, opened 2022-05-05
Fix/osx vendor downloader

meghfossa:
Fixes the `vendor_download.sh` for osx, I wasn't sure why this was failing. 

Previous actions (failing in osx):
- https://github.com/fossas/fossa-cli/actions/runs/2272922869
- https://github.com/fossas/fossa-cli/actions/runs/2272209359
- https://github.com/fossas/fossa-cli/actions/runs/2273447666

This action seems to be working: https://github.com/fossas/fossa-cli/runs/6300351000?check_suite_focus=true

### DIFF
--- a/vendor_download.sh
+++ b/vendor_download.sh
@@ -60,7 +60,7 @@ WIGGINS_TAG="2022-01-19-a647d17"
 echo "Downloading wiggins binary"
 echo "Using wiggins release: $WIGGINS_TAG"
 WIGGINS_RELEASE_JSON=vendor-bins/wiggins-release.json
-curl -sL \
+curl -sSL \
     -H "Authorization: token $GITHUB_TOKEN" \
     -H "Accept: application/vnd.github.v3.raw" \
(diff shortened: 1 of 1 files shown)


zlav (commented):
Do you know why this was failing previously?

Was `-S` showing the failing error and breaking? Or was it related to the `https` addition (which seems unlikely)?
``` 
-S, --show-error

When used with [-s, --silent](https://curl.se/docs/manpage.html#-s), it makes curl show an error message if it fails.
```

meghfossa:
@zlav Tbh I don't know the root cause, 

1) I modified the basis token, but this did not solve the issue (which I initially thought was the issue)
2) Correlating with build failures, I noticed Github had released an update on macos-latest image (3days ago): https://github.com/actions/virtual-environments/commit/3189502fc6eaf85ad34d50457d5610496634a4b2 (which updates curl version 7.83.0)

I couldn't reproduce this locally (osx binary for curl are usually a few versions behind), so I tried to replicate what GH's API docs do - and it seemed to work, so I stopped here.

zlav (approved):
Interesting, well thats enough to approve and unblock but hopefully this doesn't show up again as an issue.

PR #927, opened 2022-05-06
add a random suffix to the path we write .tar.gz files to

spatten:
# Overview

I was in a customer call, and they were running into a problem with cli-side license scans. The bug was that the scans were failing and they were seeing this in the error logs:

```
      An exception occurred: /private/var/folders/45/k76d2pgj6mlb1vpx_2gyl7z80000gn/T/fossa-temp-53acdfdf15c4156e/.: openBinaryFile: inappropriate type (Is a directory)
```

I was able to recreate this locally by creating a `fossa-deps.yml` file like this:

```yaml
vendored-dependencies:
  - name: imgui
    path: .
```

and then a directory structure like this:

```
mkdir imgui
touch imgui/bar.txt
touch imgui/nice.txt
```

```
fossa analyze --experimental-native-license-scan
[ERROR] ----------
  An issue occurred

  >>> Relevant errors

    Error

      An exception occurred: /private/var/folders/45/k76d2pgj6mlb1vpx_2gyl7z80000gn/T/fossa-temp-90eb3f7ccffb27db/.: openBinaryFile: inappropriate type (Is a directory)

      Traceback:
        - fossa-analyze
```

The problem is that in `compressFile` we're attempting to create a tarball at a path that already exists and is a directory.

The fix implemented by this PR is to add a random suffix to the name of the tarball file we create.

Note: I was unable to recreate the exact scenario the customer ran into. They were not using `.` as the path in their `fossa-deps.yml`. I think this should fix their problem as well, and this is a problem that needs to be fixed either way.

This bug also happens without the `--experimental-native-license-scan` flag. I.e. it affects archive uploads as well.

## Acceptance criteria

* You should be able to scan with the scenario outlined above, with and without the `--experimental-native-license-scan` flag.
* The name of the tar file uploaded to S3 should be unchanged. This should ensure that we don't break anything else
* The rest of the archive upload and cli-side license scan workflows in the CLI and on Core should work properly

## Testing plan

```yaml
vendored-dependencies:
  - name: imgui
    path: .
```

and then a directory structure like this:

```
mkdir imgui
touch imgui/bar.txt
touch imgui/nice.txt
cp ~/fossa/themis-license-data/licenses/mit.LICENSE imgui/LICENSE
uuidgen > imgui/VERSION
```

Do the following with and without the `--experimental-native-license-scan` flag:

```
cabal run fossa -- analyze --experimental-native-license-scan -e http://localhost:9578 ~/fossa/license-scan-dirs/927-repro
```

* Verify that the path of the file on minio is not changed by this PR
* Verify that the rest of the workflow works. You should be able to see the MIT license uploaded by the scan

cli-side license scan:

Screenshot 2022-05-11 at 4 28 22 PM

archive upload:
Screenshot 2022-05-11 at 4 29 23 PM

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+
+- Archive upload: Fix a bug when trying to tar to a filename that already exists
 ## v3.2.16
 - Go: When statically analyzing a project, apply reported replacements. ([#926](https://github.com/fossas/fossa-cli/pull/926))
 
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -435,6 +435,7 @@ test-suite unit-tests
     App.Fossa.AnalyzeSpec
     App.Fossa.API.BuildLinkSpec
     App.Fossa.API.BuildWaitSpec
+    App.Fossa.ArchiveUploaderSpec
     App.Fossa.BinaryDeps.JarSpec
     App.Fossa.Config.AnalyzeSpec
     App.Fossa.Config.TestSpec
--- a/src/App/Fossa/ArchiveUploader.hs
+++ b/src/App/Fossa/ArchiveUploader.hs
@@ -38,6 +38,7 @@ import Data.Maybe (fromMaybe)
 import Data.String.Conversion
 import Data.Text (Text)
 import Data.Text qualified as Text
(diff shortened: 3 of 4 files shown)


meghfossa (approved):
LGTM! Make sure to update Changelog.md, I think this is a release worthy!

PR #929, opened 2022-05-10
let wiggins binary control stdin/stdout

enricozb:
# Overview
Running a Monorepo scan execs the embedded wiggins binary. The stdout & stderr is currently captured. This PR makes it so that the stdout & stderr are inherited from the process running the fossa-cli.

## Acceptance criteria
- Running a Monorepo scan should show wiggin's output *as the scan runs*.

## Testing plan

Run the following:
```
fossa analyze --experimental-enable-monorepo aosp <some project>
```

Logs should appear as the CLI runs.

## Risks
None since this only affects Monorepo scans.

## Checklist


scruffystuffs:
A few things are needed to get this to build in CI:

- Emitting `IOException` from the `RawExec` constructor and the `RawException` variant of `ExecErr` is going to cause a few problems, notably with deriving `Ord` and `Recordable` (We don't care about `Replayable`).  To resolve this, we'll just emit the text of the exception, wrapped in a newtype.
- We need to add the error reporting variants for the new error types.

Once that's done, we'll need a few more things before this is merge-ready:

- Documentation for the new functions and constructors.
- A new command type with no `AllowErr`
- Some slight refactoring to make the code a bit more DRY.
- Code comments to explain what we're doing and why.

scruffystuffs:
@periodic I designed most of the changes here, can you just review the overall approach and let me know if you agree?  I've reviewed the code for quality, but I'd like a second opinion on the design.

scruffystuffs:
I'm tentatively approving the code once @periodic does.
@enricozb you'll need to make sure the linter and formatter checks pass.  If you're having trouble with your setup, I can patch those bits into this PR for you.

src/App/Fossa/Monorepo.hs
```diff
@@ -44,9 +44,8 @@ monorepoScan MonorepoAnalyzeConfig{..} = withWigginsBinary $ \binaryPaths -> do
           monorepoMetadata
 
   logInfo "Running monorepo scan"
-  stdout <- context "Monorepo" $ runExecIO $ runWiggins binaryPaths wigginsOpts
```
periodic, line 47: Interesting.  It looks like after this change this anonymous function no longer depends on `Lift IO`, but the whole function still does because `withWigginsBinary` still goes through `IO`.  In the long run it would be cool if we pushed the exec commands up the stack to both load and run the binary.

No action here.  Just glad to see extra `runExecIO` commands getting pulled out of the codebase. :D

periodic (changes requested):
This feels like the quick solution to the problem, but not the best long-term solution.  As a pattern, it gives up a lot of control and may have unexpected consequences.  We use the capturing of the output in a few ways:

- Capturing error descriptions for telemetry
- Ensuring ordering and transactional nature of logging so that messages don't get mixed together.
- Connecting errors to potential causes.

We lose these features if we pipe the output through.  Maybe this is fine for Monorepo/Wiggins, but it doesn't feel like a good pattern to enshrine in our effects.

I think we should write Wiggins output streaming as a one-off.  It would be more wordy, but it would clearly limit this behavior to Monorepo scanning.  We still get a quick solution, but it's more isolated.  You should be able to copy a lot of the code you wrote into that file.  Also, then it's easy to add a comment about why this command is being run differently so that others will be more informed on whether they should follow the same pattern or not.

**Would you try isolating the behavior mostly to the Monorepo file?** Give that a shot.  If there is something that makes that hard of the diff looks like it would be _larger_ than this, let me know because I don't want you having to do something even more complicated.

That might let us address some of Wes's comments too.

scruffystuffs:
I agree with your comment, and I had concerns originally about doing it this way, but I let them go so that we could get this done.  It should be noted that for monorepo, the limitations you mentioned are completely fine, because we essentially treat the entire process as an `exec`-style process replacement (reference: https://man7.org/linux/man-pages/man3/exec.3.html).  I think the comment about polluting the `Exec` effect is valid.

I think we could tackle this in one of three ways:
1. Merge it now, as-is.  We can refactor it to not pollute the `Exec` effect in a followup PR.  This has the advantage of freeing up @enricozb, and delivering the feature.  We can also schedule this ourselves and prevent this from leaking out anywhere with PR  review and team understanding.
2. Take this PR off @enricozb's hands and refactor it now, since he's spent a decent amount of time on something we're asking him to throw away.
3. Have @enricozb redo this himself.  If @enricozb wants to drive this PR home and this isn't time-sensitive, then this is fine with me, otherwise we should probably do one of the first two.

periodic:
Yeah, I get the expediency of getting this merged in to unblock something for a customer.  If we can get a follow up ticket scheduled I'm fine with that.

enricozb:
This PR was initially meant to help me debug an issue that has since been resolved, as I needed streaming logs to figure out what was causing the CLI to be killed. This is no longer an important customer need, and I definitely don't want to merge anything in that you guys aren't comfortable with. No need to work on this any further, thank you all for the reviews!

PR #936, opened 2022-05-16
Makes telemetry enabled by default

meghfossa:
# Overview

This PR enables telemetry by default.

## Acceptance criteria

- When `telemetryScope` is not provided via command line, the environment variable, or configuration file, `FullTelemetry` scope is used.

## Testing plan
  
1) Run an analysis
```
make install-dev
fossa-dev analyze ./docs --fossa-api-key <MY_API_KEY>
```

Since core silently ignores telemetry submitted by the non-released cli version, we won't see the records in the data store. Still, we can assess whether CLI successfully made API call via: `go/datadog`. To do so, go to `api/cli/telemetry` endpoint - and notice an inbound request.

## Risks

N/A

## References

Closes https://fossa.atlassian.net/browse/ANE-261

## Checklist


install-latest.sh
```diff
@@ -382,18 +382,17 @@ print_telemetry_disclaimer() {
   log_info "------"
   log_info ""
   log_info "FOSSA collects warnings, errors, and usage data to improve"
-	log_info "the FOSSA CLI and your experience. The CLI DOES NOT collect"  
-	log_info "telemetry by default, but this will change in future!"
+	log_info "the FOSSA CLI and your experience."
```
csasarak, line 385:
```suggestion
      log_info "the FOSSA CLI and your experience."
```

csasarak (approved):
LGTM!

PR #938, opened 2022-05-18
Ignore request in specs

spatten:
# Overview

_Provide an overview of this change. Describe the intent of this change, and how it implements that intent._

_Example: This PR accomplishes X by doing Y._

## Acceptance criteria

_If this PR is successful, what impact does it have on the user experience?_

_Example: When users do X, Y should now happen._

## Testing plan

_Example:_

1. _Set up test case X._

## Risks

## References

_Example:_

## Checklist

### DIFF
--- a/test/Test/MockApi.hs
+++ b/test/Test/MockApi.hs
@@ -200,7 +200,7 @@ matchExpectation a@(GetOrganization{}) (ApiExpectation _ _ b@(GetOrganization{})
 matchExpectation a@(GetProject{}) (ApiExpectation _ _ b@(GetProject{}) resp) = resp <$ guard (a == b)
 matchExpectation a@(GetScan{}) (ApiExpectation _ _ b@(GetScan{}) resp) = resp <$ guard (a == b)
 matchExpectation a@(GetRevisionInfo{}) (ApiExpectation _ Yes b@(GetRevisionInfo{}) resp) = resp <$ guard (a == b)
-matchExpectation GetRevisionInfo{} (ApiExpectation _ No GetRevisionInfo{} resp) = resp <$ guard (True)
+matchExpectation GetRevisionInfo{} (ApiExpectation _ No GetRevisionInfo{} resp) = Just resp
 matchExpectation a@(GetSignedLicenseScanUrl{}) (ApiExpectation _ _ b@(GetSignedLicenseScanUrl{}) resp) = resp <$ guard (a == b)
 matchExpectation a@(ResolveProjectDependencies{}) (ApiExpectation _ _ b@(ResolveProjectDependencies{}) resp) = resp <$ guard (a == b)
 matchExpectation a@(ResolveUserDefinedBinary{}) (ApiExpectation _ _ b@(ResolveUserDefinedBinary{}) resp) = resp <$ guard (a == b)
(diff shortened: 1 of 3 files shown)


test/Test/MockApi.hs
```diff
@@ -75,21 +77,31 @@ data ExpectationRepetition
...
+
 -- | The result of a call can be either a value or a diagnostic failure
 newtype ApiResult a = ApiResult (Either ApiFail a)
 
 newtype ApiFail = ApiFail {unApiFail :: Text}
 
 -- | An expectation of an API call made up of the request and response.
 data ApiExpectation where
-  ApiExpectation :: ExpectationRepetition -> FossaApiClientF a -> ApiResult a -> ApiExpectation
+  ApiExpectation :: ExpectationRepetition -> ExpectationCheckRequestValue -> FossaApiClientF a -> ApiResult a -> ApiExpectation
```
periodic, line 92: I might actually make this a different type here.  Maybe we have `ExactRequestExpectation` and `AnyRequestExpectation`?  I think the fact that we have to include a `FossaApiClientF a` is sort of accidental.

The other option would be to define some sort of "matcher".  Then instead of a `FossaApiClientF a` it would include a `RequestMatcher` which could have `Exact` and `Any` constructors.  Then we could even pass those into the first argument of `expectOnce` so you get something like 

```haskell
Exact (GetOrganization) `alwaysReturns` resp1
Any (GetProject ...) `returnsOnce` resp2
```

You could even do something cool with a typeclasses so you don't have to write `Exact`.

```haskell
class IsMatchableRequest f where
  toMatcher :: f a -> RequestMatcher a
  
instance IsMatchableRequest RequestMatcher where
  toMatcher = id

instance IsMachableRequest (FossaApiClientF a) where
  toMatcher = Exact

-- Not the real signature, just building the expectation for simplicity
returnsOnce :: IsMatchableRequest f => f a -> a -> ApiExpectation
returnsOnce matchable resp =
  ApiExpectation Once (toMatcher machable) resp
```

test/Test/MockApi.hs
```diff
@@ -156,39 +172,41 @@ instance (Algebra sig m, Has (Lift IO) sig m) => Algebra (MockApi :+: sig) (Mock
...
+matchExpectation GetRevisionInfo{} (ApiExpectation _ No GetRevisionInfo{} resp) = Just resp
+matchExpectation a@(GetSignedLicenseScanUrl{}) (ApiExpectation _ _ b@(GetSignedLicenseScanUrl{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(ResolveProjectDependencies{}) (ApiExpectation _ _ b@(ResolveProjectDependencies{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(ResolveUserDefinedBinary{}) (ApiExpectation _ _ b@(ResolveUserDefinedBinary{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(UploadAnalysis{}) (ApiExpectation _ _ b@(UploadAnalysis{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(UploadArchive{}) (ApiExpectation _ _ b@(UploadArchive{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(UploadContainerScan{}) (ApiExpectation _ _ b@(UploadContainerScan{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(UploadContributors{}) (ApiExpectation _ _ b@(UploadContributors{}) resp) = resp <$ guard (a == b)
+matchExpectation a@(UploadLicenseScanResult{}) (ApiExpectation _ Yes b@(UploadLicenseScanResult{}) resp) = resp <$ guard (a == b)
+matchExpectation UploadLicenseScanResult{} (ApiExpectation _ No UploadLicenseScanResult{} resp) = Just resp
```
periodic, line 209: I think I see how this `Yes`/`No` is useful here because we don't have to drill in too much into various constructors to see the type.

You might just want to write a helper function to do the Yes/No so you don't have to add a case for it.

```haskell
checkResult :: ExpectationCheckRequestValue -> FossaApiClientF a  -> FossaApiClientF a -> ApiResult a -> Maybe (ApiResult a)
checkResult Yes a b resp = resp <$ guard (a == b)
checkResult No _ _ resp = pure resp
```

periodic (commented):
I left a few comments.  I think ultimately this is very close to the simplest solution.

There's possibly a solution involving the `singletons` library which uses template-Haskell to create types and values for each constructor that we can then match on. However, getting it to work and set up the first time is not trivial.  [Wes tried a while ago and had some compilation issues.](https://teamfossa.slack.com/archives/CU3EPRS11/p1649702020815859)  I don't want us to have to maintain such a list manually.

spatten:
Closing, as this was just a spike that is fully implemented in https://github.com/fossas/fossa-cli/pull/946

PR #946, opened 2022-05-26
Api expectations that succeed for any request body

spatten:
# Overview

When testing the API, we sometimes want to test API requests where the request body may change between requests or the body may be complicated to construct (and we do not actually care what the body is).

For these cases, we want a new expectation that does not check that the request body.

This PR adds two new expecations that do this: `returnsOnceForAnyRequest` and `alwaysReturnsForAnyRequest`.

As part of the PR, I thought it would be nice to add some tests that use the new expecations. So I added some tests for the `ArchiveUploader` workflow. To make these easier to test, I changed the `uploadArchive` API call to return a lazy ByteString instead of a `LbsResponse`.

## Acceptance criteria

- The new tests should pass
- All existing tests should pass

## Testing plan

Because I made a small change to the `uploadArchive` endpoint, I ran an archive upload to test that it worked.

```
```

The directory I'm pointing at is this one from the `example projects` repo: https://github.com/fossas/example-projects/tree/main/fossa-deps/two-vendored-dependencies

## Risks

None

## References

This was sketched out in a test PR here: https://github.com/fossas/fossa-cli/pull/938

## Checklist


test/Test/MockApi.hs
```diff
@@ -75,27 +79,43 @@ data ExpectationRepetition
...
+  ApiExpectation :: ExpectationRepetition -> ExpectationRequestType -> FossaApiClientF a -> ApiResult a -> ApiExpectation
 
 -- | Create an expectation that will only be satisfied once.
 returnsOnce :: Has MockApi sig m => FossaApiClientF a -> a -> m ()
 returnsOnce req resp =
   -- ApiExpectation Once req (Return resp)
   send $ MockApiOnce req resp
 
+-- | Create an expectation that will only be satisfied once where we do not check the value of the req
+returnsOnceForAnyRequest :: Has MockApi sig m => FossaApiClientF a -> a -> m ()
```
spatten, line 103: Would `returnsOnceForAnyRequestBody` and `alwaysReturnsForAnyRequestBody` be better names, or is that just too wordy?

test/Test/Fixtures.hs
```diff
@@ -206,3 +217,56 @@ issuesPending =
...
+
+secondVendoredDep :: VendoredDependency
+secondVendoredDep =
+  VendoredDependency
+    "second-archive-test"
+    "vendor/bar"
+    (Just "0.0.1")
+
+vendoredDeps :: NonEmpty VendoredDependency
+vendoredDeps = firstVendoredDep :| [secondVendoredDep]
```
spatten, line 240: I'll get rid of these `:|` in a future PR by changing out hlint config to allow `NE.fromList` in this file and in specs.

spatten: Because `:|` makes me feel :|

src/Control/Carrier/FossaApiClient/Internal/Core.hs
```diff
@@ -164,6 +164,6 @@ uploadArchive ::
   ) =>
   SignedURL ->
   FilePath ->
-  m LbsResponse
+  m ByteString
```
periodic, line 167: I like this change.  It really should have been updated not to leak the network libraries.

test/Test/Fixtures.hs
periodic: It really doesn't seem necessary for fixtures when you passing a list that is defined in-line.  I'd support that!

test/Test/MockApi.hs
```diff
@@ -126,15 +146,23 @@ newtype MockApiC m a = MockApiC
 instance (Algebra sig m, Has (Lift IO) sig m) => Algebra (MockApi :+: sig) (MockApiC m) where
   alg hdl sig ctx = MockApiC $ case sig of
     L (MockApiOnce req resp) -> do
-      let expectation = ApiExpectation Once req (ApiResult (Right resp))
+      let expectation = ApiExpectation Once ExpectingExactRequest req (ApiResult (Right resp))
```
periodic, line 149: This is starting to look repetitive and we could probably pull it out into a helper since 2/3 lines are the same in every case.

test/Test/MockApi.hs
```diff
@@ -156,38 +184,45 @@ instance (Algebra sig m, Has (Lift IO) sig m) => Algebra (MockApi :+: sig) (Mock
...
+matchExpectation a@(ResolveProjectDependencies{}) (ApiExpectation _ requestExpectation b@(ResolveProjectDependencies{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(ResolveUserDefinedBinary{}) (ApiExpectation _ requestExpectation b@(ResolveUserDefinedBinary{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(UploadAnalysis{}) (ApiExpectation _ requestExpectation b@(UploadAnalysis{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(UploadArchive{}) (ApiExpectation _ requestExpectation b@(UploadArchive{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(UploadContainerScan{}) (ApiExpectation _ requestExpectation b@(UploadContainerScan{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(UploadContributors{}) (ApiExpectation _ requestExpectation b@(UploadContributors{}) resp) = checkResult requestExpectation a b resp
+matchExpectation a@(UploadLicenseScanResult{}) (ApiExpectation _ requestExpectation b@(UploadLicenseScanResult{}) resp) = checkResult requestExpectation a b resp
```
periodic, line 225: Some day we'll figure out how to make this more automatic.

periodic (approved):
Nice work.

What are the empty `.gitkeep` files for?  I don't see how those two directories were used.  Maybe we should put a sentence in the file explaining it.

test/Test/MockApi.hs
periodic: I think it might be getting too wordy.  Maybe we can think back to what we would want to read and something will jump out at us.  "request returnsOnceForSimilarRequests response"? "request returnsOnceIgnoringContents responce"?

I'll leave it up to you.  I think the ergonomics are minor and very easy to change.

spatten:
> What are the empty `.gitkeep` files for? I don't see how those two directories were used. Maybe we should put a sentence in the file explaining it.

That will become more clear when I fill them in for use in the same tests for CLI-side license scanning. I think I'll leave it for now since it will be fixed very shortly

test/Test/MockApi.hs
spatten: I'm going to leave it as is for now, since it will be an easy search-and-replace if we come up with a better name in the future

PR #948, opened 2022-05-30
[No Ticket] Provides more example of installation (of specific version), and links walkthoughs in fossa-yml 

meghfossa:
# Overview

This PR improves doc by,
- adding installation example for specific version
- linking analysis target configuration walkthrough in fossa.yml.doc


scruffystuffs (approved):

scruffystuffs:
The link-check failures are legit, those should be cleaned up before merging

PR #950, opened 2022-05-31
Force vendored dependency rescans

spatten:
# Overview

The last part of [ANE-190](https://fossa.atlassian.net/browse/ANE-190) is to add a flag that forces CLI-side license scans to run, even if they have already been analyzed on Core.

This is useful if the customer has provided a version for the vendored-dependency in `fossa-deps.yml`, and wants to force a re-scan without changing the version.

This requires two changes. First, we have to ignore the fact that the vendored dependencies have already been analyzed on Core and always do the CLI-side license scan. Second, we have to force a rebuild on Core. If we don't force the rebuild, then we would end up scanning them and uploading the results, but the build would never get queued and so the changed code would never be analyzed by Core and shown to the user.

We do this by adding a new flag, `--force-vendored-dependency-rescans`. If it exists, then we always rescan all vendored dependencies, and we set `forceRebuild` to true in the payload that we send to the `/api/license_scans/finalize` endpoint on Core.

The core side of this PR is here: https://github.com/fossas/FOSSA/pull/7780

We will make the same change for archive uploads when we implement [ANE-361](https://fossa.atlassian.net/browse/ANE-361). For now, the `--force-vendored-dependency-rescans` flag does nothing for archive uploads, and this is noted in the help text.

## Acceptance criteria

1. If the `--force-vendored-dependency-rescans` flag is present and we are doing a cli-side license scan, then we should re-scan all vendored dependencies and force a rebuild on Core.
2. If we are doing an archive upload, then it should work exactly the same with or without the `--force-vendored-dependency-rescans` flag.
3. If we are pointing at a core instance that does not support forcing rebuilds on the `/api/license_scans/finalize` endpoint, then we will silently ignore the `forceRebuild` flag but otherwise keep going.

### Setup

We're going to scan the `two-vendored-dependencies` project from the example-projects repo: https://github.com/fossas/example-projects/tree/main/fossa-deps/two-vendored-dependencies

I've copied that into ~/fossa/license-scan-dirs/two-vendored-dependencies

### Run with current Core

Set your current core to the `develop` branch, then scan the `two-vendored-dependencies` project. Current develop supports skipping dependencies, but does not support forcing rebuilds. So we're testing the third AC when running this with the `--force-vendored-dependency-rescans` flag.

Run this a few times:

```
cabal run fossa -- analyze --debug --experimental-native-license-scan -e http://localhost:9578 ~/fossa/license-scan-dirs/two-vendored-dependencies
```

The first time you run it you will probably see a build being done. After that, it should skip re-scanning.

Now run it with the new flag:

```
cabal run fossa -- analyze --debug --experimental-native-license-scan --force-vendored-dependency-rescans -e http://localhost:9578 ~/fossa/license-scan-dirs/two-vendored-dependencies
```

You should see that it's running the license scan every time, and you will see a build queued for the project. But, no rebuilds for the vendored dependencies will be queued by Core.

### Run with updated Core

Switch to a version of Core that supports the `forceRebuild` flag:

```
git checkout add-force-rebuild-to-finalize-license-scan-endpoint
```

Run again with the flag:

```
cabal run fossa -- analyze --debug --experimental-native-license-scan --force-vendored-dependency-rescans -e http://localhost:9578 ~/fossa/license-scan-dirs/two-vendored-dependencies
```

You should see builds being queued

Run one more time without the flag. You should see the skipping logic working again:

```
cabal run fossa -- analyze --debug --experimental-native-license-scan -e http://localhost:9578 ~/fossa/license-scan-dirs/two-vendored-dependencies
```

### Sanity check an archive upload

Run that again without the `--experimental-native-license-scan` flag. Do this with and without the `--force-vendored-dependency-rescans` flag.

The first time you do this you will see a build on Core whether or not you have the `--force-vendored-dependency-rescans` flag. This is because we always force a rebuild when you switch between CLI-side license scans and archive uploads.

```
```

```
cabal run fossa -- analyze --debug --force-vendored-dependency-rescans -e http://localhost:9578 ~/fossa/license-scan-dirs/two-vendored-dependencies
```
## Checklist

- [X] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.


src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -434,7 +434,7 @@ archiveBuildUpload apiOpts archive = runEmpty $
     let opts = "dependency" =: True <> "rawLicenseScan" =: True
     -- The API route expects an array of archives, but doesn't properly handle multiple archives so we upload
     -- an array of a single archive.
-    let archiveProjects = ArchiveComponents [archive]
+    let archiveProjects = ArchiveComponents [archive] False
```
spatten, line 437: We'll make this respect the `--force-vendored-dependency-rescans` flag when we do [ANE-361](https://fossa.atlassian.net/browse/ANE-361)

src/App/Fossa/VendoredDependency.hs
```diff
@@ -9,6 +9,10 @@ module App.Fossa.VendoredDependency (
   compressFile,
   hashFile,
   dedupVendoredDeps,
+  skippedDepsDebugLog,
```
spatten, line 12: I moved `SkippableDeps` and `NeedScanningDeps` here as we're going to need them here when we implement skipping for archive uploads. It's also a good spot for them even before we do that

src/App/Fossa/LicenseScanner.hs
```diff
@@ -349,7 +347,7 @@ licenseScanSourceUnit vendoredDependencyScanMode baseDir vendoredDeps = do
 
   -- finalizeLicenseScan takes archives without Organization information. This orgID is appended when creating the build on the backend.
   -- We don't care about the response here because if the build has already been queued, we get a 401 response.
-  finalizeLicenseScan $ ArchiveComponents $ NE.toList archives
+  finalizeLicenseScan $ ArchiveComponents{archives = NE.toList archives, forceRebuild = vendoredDependencyScanMode == SkippingDisabledViaFlag}
```
scruffystuffs, line 350: We should avoid using the `{a = foo, b = bar}` constructor style, especially when the types of the fields are different, and the type system can enforce the correct usage.
```suggestion
  finalizeLicenseScan $ ArchiveComponents (NE.toList archives) (vendoredDependencyScanMode == SkippingDisabledViaFlag)
```

src/App/Fossa/VendoredDependency.hs
```diff
@@ -118,3 +157,14 @@ hashFile fileToHash = do
...
+skippedDepsDebugLog :: NeedScanningDeps -> SkippableDeps -> VendoredDependencyScanMode -> SkippedDepsLogMsg
+skippedDepsDebugLog needScanningDeps skippedDeps scanMode =
+  case (needScanningDeps, scanMode) of
+    (_, SkippingNotSupported) -> SkippingUnsupportedMsg
+    (_, SkippingDisabledViaFlag) -> SkippingDisabledViaFlagMsg
+    (NeedScanningDeps [], SkipPreviouslyScanned) -> AllDepsPreviouslyScannedMsg
+    (_, SkipPreviouslyScanned) -> do
+      case skippedDeps of
+        SkippableDeps [] -> AllDepsNeedScanningMsg
+        _ -> SomeDepsNeedScanningMsg skippedDeps
```
scruffystuffs, line 170: 🔥

src/App/Fossa/VendoredDependency.hs
```diff
@@ -54,8 +59,42 @@ instance FromJSON VendoredDependency where
...
+    "The --force-vendored-dependency-rescans flag was used, so performing a full scan of all vendored dependencies even if they have been scanned previously."
+  pretty AllDepsPreviouslyScannedMsg =
+    "All of the current vendored dependencies have been previously scanned, reusing previous results."
+  pretty AllDepsNeedScanningMsg =
+    "None of the current vendored dependencies have been previously scanned. License scanning all vendored dependencies"
+  pretty (SomeDepsNeedScanningMsg skippedDeps) =
+    vsep
+      [ "Some of the current vendored dependencies have already been scanned by FOSSA."
+      , "Reusing previous results for the following vendored dependencies: " <> (pretty . show $ skippedDeps)
+      ]
```
scruffystuffs, line 96: 🔥

scruffystuffs (approved):
Love the approach we took for the skipped deps logging.

Left a few nits, but this LGTM

PR #954, opened 2022-06-02
Default to cli side license scanning

spatten:
# Overview

~This should not be merged until https://github.com/fossas/FOSSA/pull/7805 is merged and deployed to Ford, and the feature flag is turned on for all of the orgs on the Ford instance (I think there are three of them)~ This has been done on Ford staging and prod

[Delivers ANE-219](https://fossa.atlassian.net/browse/ANE-219)
[Delivers ANE-359](https://fossa.atlassian.net/browse/ANE-359)

This PR makes CLI-side license scans the default for most orgs. We will still do Archive uploads by default if the org has the "do archive uploads by default" feature flag enabled (see https://github.com/fossas/FOSSA/pull/7805) or if the FOSSA server does not support CLI-side license scans.

You can force an archive upload by using either the `--force-vendored-dependency-archive-upload` flag or setting the `vendoredDependencies.scanMode` field in `.fossa.yml` to "ArchiveUpload". 

If the org has the "do archive uploads by default" feature flag and the FOSSA server support CLI-side license scans, then you can still switch to CLI-side license scans by using the `--force-vendored-dependency-license-scan` flag or setting the `vendoredDependencies.scanMode` field in `.fossa.yml` to "CLILicenseScan". 

## Acceptance criteria

* We should do CLI-side license scanning if the backend supports it and the “force archive upload” flag or config-file setting is not present
* We should do an archive upload if the flag or config requests it or if Core reports that it does not support CLI-side license scanning
* We should be able to force a CLI-side license scan if the org defaults to archive upload and the FOSSA server supports it
* If you try to do a CLI-side license scan but the FOSSA server does not support CLI-side license scanning, then we should throw an error
* The new flags should be documented
* The new `vendoredDependencies` section in `.fossa.yml` should be documented and added to the YAML schema
* Flags should win if both a flag and a `.fossa.yml` setting exist

## Testing plan

I'm testing against the "two vendored dependencies" example repo found here: https://github.com/fossas/example-projects/tree/main/fossa-deps/two-vendored-dependencies

Core is running current develop, which supports the "default to archive upload" feature flag.

With no `.fossa.yml`, run the following command twice:

```
```

It should succeed both times, and the second time it should say

```
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/two-vendored-dependencies/"
[DEBUG] All of the current vendored dependencies have been previously scanned, reusing previous results.
```

Force it to be an archive upload via the flag:

```
cabal run fossa -- analyze --debug -e http://localhost:9578 --force-vendored-dependency-archive-upload --force-vendored-dependency-rescans ~/fossa/license-scan-dirs/two-vendored-dependencies
```

The debug logs should show that it is an archive upload:

```
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/two-vendored-dependencies/"
[DEBUG] Compressing 'first-dir-aaa11' at 'first-dir'
[DEBUG] Uploading 'first-dir-aaa11' to secure S3 bucket
[DEBUG] ""
[DEBUG] Compressing 'second-dir-aaa11' at 'second-dir'
[DEBUG] Uploading 'second-dir-aaa11' to secure S3 bucket
[DEBUG] ""
```

Force rescans in CLI-side license scanning mode. It should say that it's license scanning and that a rescan has been forced:

```
cabal run fossa -- analyze --debug -e http://localhost:9578 --force-vendored-dependency-rescans ~/fossa/license-scan-dirs/two-vendored-dependencies
```

```
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/two-vendored-dependencies/"
[DEBUG] Vendored dependency rescans forced on via either the --force-vendored-dependency-rescans flag or the vendoredDependencies.forceRescans entry in .fossa.yml, so performing a full scan of all vendored dependencies even if they have been scanned previously.
[DEBUG] License Scanning 'first-dir-aaa11' at 'first-dir'
[DEBUG] Uploading 'first-dir-aaa11' to secure S3 bucket
[DEBUG] License Scanning 'second-dir-aaa11' at 'second-dir'
[DEBUG] Uploading 'second-dir-aaa11' to secure S3 bucket
```

Now do the same with a config file.

```yaml
version: 3
vendoredDependencies:
  forceRescans: true
  scanMethod: CLILicenseScan
```

You'll want to try it with `forceRescans` set to `true` and `false`, and `scanMethod` set to `CLILicenseScan` and `ArchiveUpload`.

Set `scanMethod` to something invalid and see that it fails.

Test that the flag beats the config file by making `.fossa.yml` look like this:

```yaml
version: 3
vendoredDependencies:
  forceRescans: false
  scanMethod: ArchiveUpload
```

and running this command:

```
cabal run fossa -- analyze --debug -e http://localhost:9578 --force-vendored-dependency-license-scan --force-vendored-dependency-rescans ~/fossa/license-scan-dirs/two-vendored-dependencies
```

You should see a CLI-side license scan with rescans on:

```
[DEBUG] Loading configuration file from "/Users/scott/fossa/license-scan-dirs/two-vendored-dependencies/"
[DEBUG] Vendored dependency rescans forced on via either the --force-vendored-dependency-rescans flag or the vendoredDependencies.forceRescans entry in .fossa.yml, so performing a full scan of all vendored dependencies even if they have been scanned previously.
[DEBUG] License Scanning 'first-dir-aaa11' at 'first-dir'
[DEBUG] Uploading 'first-dir-aaa11' to secure S3 bucket
[DEBUG] License Scanning 'second-dir-aaa11' at 'second-dir'
[DEBUG] Uploading 'second-dir-aaa11' to secure S3 bucket
```

Use both the `--force-vendored-dependency-archive-upload` and `--force-vendored-dependency-license-scan` flags and see that it fails.

## Risks

None

## References

https://fossa.atlassian.net/browse/ANE-219

## Checklist

- [X] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [X] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [X] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).


docs/features/vendored-dependencies.md
```diff
@@ -71,3 +73,27 @@ We also support json-formatted dependencies:
...
+
+There are two methods of vendored dependency scanning: "CLI license scan" and "archive upload".
+
+The default is typically "CLI license scan", but your organization may have opted to default to "archive upload".
+
+Both methods use the same license scanning technology and will give you the same results. The difference is where the license scan is done.
+
+A "CLI license scan" runs a license scan directly on the server where you are running `fossa analyze`, and only uploads the matched license data to FOSSA's servers.
+
+"Archive upload" uploads the files at the specified path to a secure S3 bucket. We license scan the uploaded files on our servers. All files that do not contain licenses are then removed after 30 days.
```
spatten, line 86: This used to say "after 2 weeks", but I was just looking at our S3 lifecycle configuration for something else and I'm pretty sure that this should be 30 days.

src/App/Fossa/Config/Analyze.hs
```diff
@@ -8,6 +8,8 @@ module App.Fossa.Config.Analyze (
   BinaryDiscovery (..),
   ExperimentalAnalyzeConfig (..),
   ForceVendoredDependencyRescans (..),
+  ForceVendoredDependencyCLILicenseScan (..),
+  ForceVendoredDependencyArchiveUpload (..),
```
kitified, line 12: Since only one state is valid, please consider parsing this as an enum instead (ref: [ReportOutputFormat](https://github.com/fossas/fossa-cli/blob/102a09e1919379868739872e4939aff13529bf95/src/App/Fossa/Config/Report.hs#L60)), which lets us avoid an error case in general and is cleaner code-wise.

That'd be most easily done with a slight change in the CLI flag (maybe `--force-vendored-dependency-mode license-scan|archive-upload`), which leaves more room for future modes as well (I know we don't have any plans, but the flexibility would cost us roughly nothing right now, I think, since these are new flags).

kitified (commented):
This looks good to me, super well documented and tested, very nice!

I'd like to have a discussion on these flags before approving - I won't block if you feel strongly that this is the right way to go but I think it'd be more ergonomic to go enum-style with the flags.

src/App/Fossa/Config/Analyze.hs
spatten: Yeah, you're right. I switched to this and the code is much cleaner

Thank you!

src/App/Fossa/Config/Analyze.hs
```diff
@@ -273,6 +290,15 @@ cliParser =
...
     <*> baseDirArg
 
+vendoredDependencyModeOpt :: Parser ArchiveUploadType
+vendoredDependencyModeOpt = option (maybeReader parseType) (long "force-vendored-dependency-scan-method" <> metavar "METHOD" <> help "Force the vendored dependency scan method. The options are 'CLILicenseScan' or 'ArchiveUpload'. 'CLILicenseScan' is usually the default unless your organization has overridden this.")
+  where
+    parseType :: String -> Maybe ArchiveUploadType
+    parseType = \case
+      "ArchiveUpload" -> Just ArchiveUpload
+      "CLILicenseScan" -> Just CLILicenseScan
+      _ -> Nothing
```
kitified, line 300: I think this means that if I pass in an arbitrary string to `force-vendored-dependency-scan-method`, it'll just use the default. Is that correct? (I don't see anything that causes the `Nothing` case to error if they provided a value)

Basically I'm looking for something like this in this changeset: https://github.com/fossas/fossa-cli/blob/102a09e1919379868739872e4939aff13529bf95/src/App/Fossa/Config/Report.hs#L192-L195

src/App/Fossa/Config/Analyze.hs
```diff
@@ -103,13 +107,16 @@ import Options.Applicative (
...
 import Path (Abs, Dir, File, Path, Rel)
 import System.Info qualified as SysInfo
-import Types (TargetFilter)
+import Types (ArchiveUploadType (..), TargetFilter)
 
 -- CLI flags, for use with 'Data.Flag'
 data AllowNativeLicenseScan = AllowNativeLicenseScan deriving (Generic)
 data ForceVendoredDependencyRescans = ForceVendoredDependencyRescans deriving (Generic)
+data ForceVendoredDependencyCLILicenseScan = ForceVendoredDependencyCLILicenseScan deriving (Generic)
+data ForceVendoredDependencyArchiveUpload = ForceVendoredDependencyArchiveUpload deriving (Generic)
```
kitified, line 119: I think these can be removed now, right?

kitified (approved):
lgtm after the comments are addressed!

src/App/Fossa/Config/Analyze.hs
spatten: Oh right! Thank you

src/App/Fossa/Config/Analyze.hs
spatten: It does actually fail, but with a not-so-great error message:

```
option --force-vendored-dependency-scan-method: cannot parse value `CLILicenseScaaaan'
```

I've fixed it by using `eitherReader` instead of `maybeReader`. This makes it more obvious that there's an error message, and gives you the possibility of a better error message too. The error message now looks like this:

```
option --force-vendored-dependency-scan-method: must be either 'CLILicenseScan' or 'ArchiveUpload'. Found CLILicenseScaaaan
```

kitified: Brilliant, thanks! ✨

PR #958, opened 2022-06-04
[ANE-325] Adds initial pnpm support

meghfossa:
# Overview

This PR adds pnpm support. 

## Acceptance criteria

- `pnpm-lock.yaml` file can be analyzed with pnpm analyzer.
- pnpm analyzer reports direct, deep dependencies, and edges amongst them.
- pnpm analyzer by default exclude dev dependencies.
- pnpm analyzer reports git dependencies.
- pnpm analyzer reports URL dependencies.
- pnpm analyzer reports npm registry dependencies.
- pnpm analyzer does not report directory sourced dependencies.
- pnpm analyzer promotes deep dependencies of directory sourced dependencies.
- pnpm analyzer works for lockfile which has no workspaces.
- pnpm analyzer works for lockfile which has workspaces.

## Testing plan

1) Get an example pnpm project from: https://github.com/fossas/example-projects (javascript/pnpm directory) or from open source community
2) Analyze project with:
```
make install-dev
fossa-dev analyze -o | jq
```

Compare the graph to confirm fossa-cli is reporting the correct result. You can also refer to included unit tests.

## Risks
None - It has no impact on existing analyzers, or functionalities.

## References
https://fossa.atlassian.net/browse/ANE-325

## Post Approval Checklist
- [x] Update Changelog
- [ ] Update Product Documentation (in docs.fossa.com)

## Checklist

### DIFF
--- a/src/Strategy/Node/Pnpm/PnpmLock.hs
+++ b/src/Strategy/Node/Pnpm/PnpmLock.hs
@@ -124,7 +124,6 @@ instance FromJSON PnpmProjectSnapshot where
 
 data PnpmPackageSnapshot = PnpmPackageSnapshot
   { isDev :: Bool
-  , isOptional :: Bool
   , name :: Maybe Text -- only provided when non-registry resolver is used
   , resolution :: Resolution
   , dependencies :: Map Text Text
@@ -135,7 +134,6 @@ data PnpmPackageSnapshot = PnpmPackageSnapshot
 instance FromJSON PnpmPackageSnapshot where
   parseJSON = Yaml.withObject "PnpmPackageSnapshot" $ \obj ->
     PnpmPackageSnapshot <$> (obj .:? "dev" .!= False)
-      <*> (obj .:? "optional" .!= False)
       <*> obj .:? "name"
       <*> obj .: "resolution"
       <*> (obj .:? "dependencies" .!= mempty)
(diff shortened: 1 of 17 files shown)


scruffystuffs:
We may want to include optional dependencies by default, but not dev deps.  We almost certainly want to let the user remove them in the UI rather than stifle them here, because it's impossible for the user to opt-in to optional prod dependencies without also including dev dependencies.

meghfossa:
> We may want to include optional dependencies by default, but not dev deps. We almost certainly want to let the user remove them in the UI rather than stifle them here, because it's impossible for the user to opt-in to optional prod dependencies without also including dev dependencies.

I presume this is what happens with npm/yarn today? I looked at pnpm source code, it does seem they _try_ to include optional by default, I can include this in the coming commit.

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -0,0 +1,298 @@
...
+      let parentDep = toDependency depName depVersion pkgMeta
+
+      -- It is ok, if this dependency was already graphed as direct
+      -- @direct 1 <> deep 1 = direct 1@
+      deep parentDep
+
+      for_ deepDependencies $ \(deepName, deepVersion) -> do
+        case (toResolvedDependency deepName deepVersion) of
+          Just childDep -> edge parentDep childDep
+          Nothing -> pure ()
```
scruffystuffs, line 219: [nit]: this can be written more functionally
```suggestion
        maybe (pure ()) (edge parentDep) toResolvedDependency deepName deepVersion
```

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -0,0 +1,298 @@
...
+    -- >> getPkgNameVersion "github.com/something" = Nothing
+    -- >> getPkgNameVersion "/pkg-a/1.0.0" = Just ("pkg-a", "1.0.0")
+    getPkgNameVersion :: Text -> Maybe (Text, Text)
+    getPkgNameVersion pkgKey = case (Text.stripPrefix "/" pkgKey) of
+      Nothing -> Nothing
+      Just txt -> do
+        let (nameWithSlash, version) = Text.breakOnEnd "/" txt
+        case (Text.stripSuffix "/" nameWithSlash, version) of
+          (Just name, v) -> Just (name, v)
+          _ -> Nothing
```
scruffystuffs, line 233: [nit]: I think a lot of this pattern matching can be done with `>>=` instead.  Might be worth refactoring.

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -0,0 +1,298 @@
...
+    toDependency name _ (PnpmPackageSnapshot isDev Nothing (DirectoryResolution _) _ _) =
+      toDep UserType name Nothing isDev
+
+    -- Sometimes package versions include symlinked paths
+    -- of sibling dependencies used for resolution.
+    --
+    -- >> withoutSymLinks "1.2.0" = "1.2.0"
+    -- >> withoutSymLinks "1.2.0_vue@3.0" = "1.2.0"
+    withoutSymConstraint :: Text -> Text
+    withoutSymConstraint version = fst $ Text.breakOn "_" version
```
scruffystuffs, line 289: naming mismatch

test/Pnpm/PnpmLockSpec.hs
```diff
@@ -0,0 +1,256 @@
+{-# LANGUAGE TemplateHaskell #-}
+
+module Pnpm.PnpmLockSpec (
```
scruffystuffs, line 3: This set of tests is pretty big.  Can we reduce the lockfile down to a smaller set artificially, so that the test file is smaller and doesn't retest the same features several times each?

scruffystuffs (approved):
Lots of nits, but nothing blocking.
Loved the clean layout for the `buildGraph` function

PR #971, opened 2022-06-13
Update instances of `Fossa`

kitified:
Updates text references to `Fossa` to be either `FOSSA` or `fossa-`, loosely based on the guidance below:

> in copy: the FOSSA CLI
> referring to the repo or binary: fossa-cli
> absolutely never, if you find this in any docs you should change it because this is a typo: Fossa

I sent a link to https://github.com/fossas/fossa-cli/blob/70a87954ad88e55caf6b089cd8bdd4133abdecf3/docs/walkthroughs/ssl-cert.md earlier and noticed `Fossa`, and it bugged me, so this PR materialized.


src/Strategy/Node/YarnV2/Resolvers.hs
```diff
@@ -39,12 +39,12 @@ import Strategy.Node.YarnV2.Lockfile
...
-  , resolverSupportsLocator :: Locator -> Bool
-  -- ^ Does this resolver support the locator?
-  , resolverLocatorToPackage :: Locator -> Either Text Package
-  -- ^ Convert this locator to a yarn package
+  { -- | Used for error messages
+    resolverName :: Text
+  , -- | Does this resolver support the locator?
+    resolverSupportsLocator :: Locator -> Bool
+  , -- | Convert this locator to a yarn package
+    resolverLocatorToPackage :: Locator -> Either Text Package
```
kitified, line 47: I think this was from auto running cabal fmt, I didn't do this.

docs/contributing/STYLE-GUIDE.md
```diff
@@ -1,4 +1,4 @@
-# Fossa CLI style guide
+# FOSSA CLI style guide
```
spatten, line 1: Should it be `fossa-cli` even if it's in a title?

spatten (approved):
:+1:

I have a slight preference to using `fossa-cli` rather than `FOSSA CLI`, even if it's in a title. I.e. 

# `fossa-cli` something something

rather than 

# FOSSA CLI something something

but it's only a slight preference

docs/contributing/STYLE-GUIDE.md
periodic: It _could_, but I think this reads better.  Maybe "Style Guide for the FOSSA CLI" so that it's not a proper noun.

periodic (approved):

kitified:
I went back and forth on the instances inside a heading, but it does turn out github renders them nicely so I agree I like it this way (and it more closely aligns with the guidelines):
Screen Shot 2022-06-13 at 4 16 41 PM

PR #972, opened 2022-06-14
Effect docs

scruffystuffs:


docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+talk by the end:
+
+- How `mtl`-style monad transformers might work in production code.
+- The benefits of monad transformers, and the guiding principles behind its
+  development.
+- Where `mtl`-style monad transformers break down in some cases.
+  - The _functional dependencies_ point in particular is important.
+- The goals/requirements of the `fused-effects` library.
+
+[Here is the link to the fused-effects talk.][fused-talk]  Pay close attention
```
periodic, line 35: Might what to move up the link to where you first mention the talk.  I was afraid you didn't include it at first.

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+filesystem.  The function to read the value from the file contents comes from a
+third-party library, and we have no idea how it works, other than that it does
+not always contain a valid value, and can therefore return a `Maybe`.
+
+Every time we read a valid value from the file, we'll store that value into a
+ordered buffer, and attempt to read the value again. Once we read an invalid
+value (i.e. the magic file reader returns `Nothing`), we'll stop the loop, and
+print out all of the values we read to stdout.
+
+Let's add one more requirement: we only want to report unique valid values, in
```
periodic, line 61: As I read this through this requirement seems out of place?  Why include this?  This section could use a little explainer of why this example is interesting.  E.g. mixing IO, state and failure.

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+          liftIO sleepForOneSecond
+          doTheProgram
+        else doTheProgram
+```
+
+`mtl`-style monad transformers are quite fast, compared to other, more generic
+solutions, and are on par with the non-`mtl` version.  The _reason_ for this is
+that the `m` in the return value of `doTheProgram` is generic at the definition
+site, but when we call it with an explicit list of monad runners in our `main`
+function, that generic `m` is monomorphized to
```
periodic, line 192: I worry that this term "monomorphized" might be a surprise to some readers.  Also that "remember, generics are slow" feels like it comes out of nowhere because you haven't talked much about performance yet, I think.

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+          -- This is a new value
+          tell [value]
+          sendIO sleepForOneSecond
+          doTheProgram
+        else doTheProgram
+```
+
+Note that other than some different function names (`modify'` -> `modify` and
+`liftIO` -> `sendIO`), and a very different (but still similar) style of listing
+function constraints, `doTheProgram` didn't change at all.  Our main slightly
```
periodic, line 237: > `doTheProgram` didn't change at all

The magic of Monads!  Swap out implementations without changing the program flow!

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+use the `Reader` concept as an example:
+
+| Name | Purpose |
+| --- | --- |
+| Effects | GADT-style datatypes which serve as message formats for an abstract interface. |
+| Carriers | Concrete monads which provide a implementation of an effectful interface. |
+| `Algebra` typeclass | The interface by which concrete implementations are selected and used. |
+| `Member` typeclass | Expressing and proving the effects required for an operation.
+
+Note that `fused-effects` differs from `mtl` in two very distinct ways:
```
periodic, line 259: Looks like you accidentally left one of the ways out of the list.

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+is transferrable to `fused-effects`-style monads, and vice versa.
+
+## Pieces of the Puzzle
+
+Let's look at the pieces of `fused-effects`, and compare their purposes. We'll
+use the `Reader` concept as an example:
+
+| Name | Purpose |
+| --- | --- |
+| Effects | GADT-style datatypes which serve as message formats for an abstract interface. |
```
periodic, line 254: Should we add GADTs to the list of things a reader should be familiar with?  I suppose the examples don't touch on them. Maybe a short definition of some less-important terms in the introduction?

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+The concrete monad is a "transformer" of our base monad (`fused-effects` calls
+these `Carriers`) , which we compose by stacking them on top of each other, and
+we can use a runner to remove it from the stack of monads.  If our base monad is
+`IO`, we can pass that back up to `main`.  If it's a pure base monad, like
+`Identity`, we can use `runIdentity` to simply remove that base monad and get an
+entirely pure value.
+
+When we apply a runner, like `runReader`, that function specifies a _particular_
+carrier monad and removes it from the stack of carriers. Let's take a look at
+some of the definitions needed to make a reader work:
```
periodic, line 288: Might want to link straight to the reader source too, in case anyone wants to see it in full.

docs/contributing/effects-tutorial.md
```diff
@@ -0,0 +1,634 @@
...
+instance Monad        m => Monad        (ReaderC r m)
+  ...
+
+-- Here is where we define the actual operations of the reader, for the specific
+-- carrier ReaderC.
+instance Algebra sig m => Algebra (Reader r :+: sig) (ReaderC r m) where
+  ...
+```
+
+_(Question for critics of this doc: How much of this example code is_
```
periodic, line 338: Hard for me to say.  I know a lot of these things and it still doesn't read easily to me. I think this code uses a lot of extensions that might confuse a newer reader.  It has GADTs, floating instances, type-level operators.

I wonder if we might be better served by a toy example that's simpler.  Those are _much_ harder to write though.

periodic (commented):
I left a few nits in direct comments.

Overall, I'm not clear if this is meant to be a doc about using effects, understanding how they work, or writing them.  You start with an example of usage, but quickly get into the details of how they are implemented, which has a lot of jargon and technical details.

Would it be possible to structure the doc to be a bit more progressive and give some high level understanding before diving into the details of implementation?  Could you come up with a simplified abstraction that is helpful for understanding and present that first?

PR #973, opened 2022-06-14
Make the CLI `cabal install`-able

liftM:
# Overview

This makes it possible to consume the CLI as a library.

The changes in this PR:

- Update the [`cabal-version`](https://cabal.readthedocs.io/en/latest/file-format-changelog.html). Ideally, we'd update this all the way to the latest version, but [`cabal-fmt` only supports up to Cabal 3.2](https://hackage.haskell.org/package/cabal-fmt-0.1.5.1) and crashes with an error on unrecognized `cabal-version` values.
- Update [`extra-source-files`](https://cabal.readthedocs.io/en/latest/cabal-package.html#pkg-field-extra-source-files) to the correct value. This field specifies extra files bundled when running `cabal sdist`, which determines which files go into the file tree built by `cabal install`. To get a successful build, we need to (1) remove entries which are not present (e.g. `vendor-bins`) and (2) add entries which must be present (e.g. `scripts/`).
- Update the git version embedding logic to not crash when `.git` is unavailable (which occurs during `cabal install`).
- Adds a "lint" check that makes sure `cabal install` still works.

## Acceptance criteria

`cabal install` works.

## Testing plan

Run `cabal install`.

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -79,4 +79,5 @@ jobs:
 
     - name: Run `cabal install`
       run: |
+        cabal update
         cabal install
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -57,3 +57,26 @@ jobs:
     - name: "run cabal-fmt"
       run: |
         cabal-fmt --check spectrometer.cabal
+
+  cabal-install-check:
+    name: cabal-install-check
+    runs-on: ubuntu-latest
+    container: ghcr.io/fossas/haskell-dev-tools:9.0.2
+
+    steps:
+    - uses: actions/checkout@v2
+
+    # Run `cabal install`.
+    - uses: actions/cache@v2
+      name: Cache cabal store
+      with:
+        path: ${{ steps.setup-haskell.outputs.cabal-store || '~/.cabal/store' }}
+        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-${{ hashFiles('**/*.cabal') }}
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-
+          ${{ runner.os }}-${{ matrix.ghc }}-
+          ${{ runner.os }}-
+
+    - name: Run `cabal install`
+      run: |
+        cabal install
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -64,7 +64,7 @@ jobs:
     container: ghcr.io/fossas/haskell-dev-tools:9.0.2
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     # Run `cabal install`.
     - uses: actions/cache@v2
@@ -79,5 +79,6 @@ jobs:
 
     - name: Run `cabal install`
       run: |
+        apk add xz-dev bzip2-dev bzip2-static
         cabal update
         cabal install
--- a/src/App/Version.hs
+++ b/src/App/Version.hs
@@ -13,26 +13,29 @@ import Data.String.Conversion (toText)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Version (showVersion)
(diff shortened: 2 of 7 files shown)


japui:
Torne a CLI cabal install #973 capaz

.github/workflows/lint.yml
```diff
@@ -57,3 +57,28 @@ jobs:
...
+        restore-keys: |
+          ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-
+          ${{ runner.os }}-${{ matrix.ghc }}-
+          ${{ runner.os }}-
+
+    - name: Run `cabal install`
+      run: |
+        apk add xz-dev bzip2-dev bzip2-static
+        cabal update
+        cabal install
```
periodic, line 84: `--dry-run` isn't sufficent here?  It might be faster and it avoids modifying the container.

src/App/Version.hs
```diff
@@ -13,26 +13,29 @@ import Data.String.Conversion (toText)
...
-info = $$(tGitInfoCwd)
+info :: Either String GitInfo
+info = $$(tGitInfoCwdTry)
+
+fromRight' :: v -> (r -> v) -> Either l r -> v
+fromRight' v = either (const v)
 
 currentBranch :: Text
-currentBranch = toText $ giBranch info
+currentBranch = toText $ fromRight' "unknown" giBranch info
```
periodic, line 29: Nit: I wonder if these might read better as `toText . fromRight "unknown" $ fmap gitBranch info` or something, just to avoid a non-standard `fromRight'`.  I don't really have a good solution, but it feels like a pattern that should exist without the need for `fromRight'`.

src/App/Version/TH.hs
```diff
@@ -1,4 +1,10 @@
 {-# LANGUAGE TemplateHaskell #-}
+-- We disable overlapping-patterns in this module to handle the case statement
+-- in `getCurrentTag`. GHC always thinks one of the two cases is redundant
+-- because `tGitInfoCwdTry` is compiled by Template Haskell into a concrete
+-- Right or Left. However, both branches are needed to handle different
+-- compilation environments.
+{-# OPTIONS_GHC -Wno-overlapping-patterns #-}
```
periodic, line 7: Nit: If you just saved the output of `$$(tGitInfoCwdTry)` to a variable you could avoid suppressing this warning, right?  That might at least look less scary.  The fact that you put in a clear comment is enough for me.

periodic (approved):
Thanks for getting this finally fixed!

.github/workflows/lint.yml
liftM: `--dry-run` is insufficient because we need to actually execute the build in order to see things like Template Haskell compilation errors.

src/App/Version/TH.hs
liftM: I tried that, but still got the warning. GHC is just too clever about this optimization.

src/App/Version.hs
liftM: If you're opposed to `fromRight'`, I can just use `either` directly. I figured it was a little clearer with the repeated helper factored out.

scruffystuffs:
If we aren't confirming the git status at compile time, then we need to run a check during the release that the version info is attached properly.  That was the reason we didn't use the `try` variant originally, it was previously a compile-time check.

liftM:
@scruffystuffs how does this work for you? https://github.com/fossas/fossa-cli/pull/973/commits/4f19b6fb13737a4c5285843e18e2dd228325709e

scruffystuffs:
I think this will break when we run with a tag

liftM:
Good catch, @scruffystuffs - it should work with tags now. Are there other edge cases I should check for?

scruffystuffs:
I think the critical check is for tags, which are only required on _releases_.  Can we make a tag-only check on the "Create Release" job as well?

liftM:
Alright, I've got the check working for release version strings as we discussed in Slack. Merging.

PR #978, opened 2022-06-16
refactor: Add concurrent in-memory logger

elldritch:
# Overview

This PR adds `withTVarLogger`, which runs a `Logger` into a `Reader TVar`. This is a lot like `withWriterLogger` (#977) except that it also collects logged messages across all threads in the presence of `withTaskPool`.

## Acceptance criteria

`withTVarLogger` allows consumers to run a `Logger` into a `Reader TVar`, and collects all logged messages even when the logged action forks threads.

## Testing plan

See the accompanying unit tests. Run with `make test ARGS="Effect.Logger"`.

## Checklist

### DIFF
--- a/src/Effect/Logger.hs
+++ b/src/Effect/Logger.hs
@@ -12,7 +12,7 @@ module Effect.Logger (
   withLogger,
   withDefaultLogger,
   withWriterLogger,
-  withStateLogger,
+  withConcurrentLogger,
   runLogger,
   ignoreLogger,
   log,
@@ -27,12 +27,14 @@ module Effect.Logger (
 
 import Control.Algebra as X
 import Control.Carrier.Simple
+import Control.Concurrent.STM (atomically, readTVar)
+import Control.Concurrent.STM.TVar (TVar, writeTVar)
 import Control.Effect.ConsoleRegion
 import Control.Effect.Exception
 import Control.Effect.Lift (sendIO)
+import Control.Effect.Reader (Reader, ask)
 import Control.Effect.Record (RecordableValue)
 import Control.Effect.Record.TH (deriveRecordable)
-import Control.Effect.State (State, modify)
 import Control.Effect.Writer (Writer, tell)
 import Control.Monad (when)
 import Data.Aeson (ToJSON)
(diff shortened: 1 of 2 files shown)


src/Effect/Logger.hs
```diff
@@ -134,6 +141,33 @@ withWriterLogger sev = runLogger ctx
         , logCtxWrite = tell . pure @f
         }
 
+-- | Like `withWriterLogger`, except it works even when the logged action forks
+-- threads.
+withTVarLogger ::
```
scruffystuffs, line 146: [nit]: I would prefer a more generic name.  `withConcurrentLogger` is one idea.

src/Effect/Logger.hs
```diff
@@ -134,6 +141,33 @@ withWriterLogger sev = runLogger ctx
...
+  forall f sig m a.
+  ( Applicative f
+  , Monoid (f (Doc AnsiStyle))
+  , Has (Reader (TVar (f (Doc AnsiStyle)))) sig m
```
scruffystuffs, line 150: Why are we using a mutable variable instead of a queue?  Having a single variable means lock contention.

scruffystuffs (commented):
Drive-by comments

src/Effect/Logger.hs
liftM: That was my original name, but I was concerned that future consumers would see the symbol name and think to themselves that they should be using the concurrent logger to do _any_ concurrent logging, when in fact it's purely a testing utility.

scruffystuffs: If it's purely a testing utility, can we remove it from this module and put it somewhere in `test/`?

liftM: Well... it's purely a _programmatic usage_ utility. I can't put it into `test/` because I need it exported into Hubble.

periodic: As a carrier, I like when the names are pretty explicit.  That way we can have other implementations that may also be concurrent without naming conflict.

The function comment could maybe use some more explanation of how it's using a TVar so that readers understand why they would want to use a TVar logger and what the limitations are.  For example, I'm not clear what the interleaving semantics are just from the naming.

src/Effect/Logger.hs
periodic: Maybe add some documentation on what you discovered so that future engineers will feel empowered to make the change if and when it makes sense to them.

scruffystuffs (approved):
LGTM, thanks for taking another look at working with a queue.

PR #981, opened 2022-06-17
[untested]: Make walker resilient to dir access errors

scruffystuffs:
# Overview

When the discovery traverses the file tree, it crashes if there are any errors reading the directory, or executing the directory handler.

This PR only addresses the read errors, since handler errors could be valid.

## Acceptance criteria

Walker is resilient to errors during the `listdir` operation.

## Testing plan

TBD

## Risks

We lose information here, since we can't log the directories that failed (too noisy to do it for every discovery).

## References

Internal slack thread: https://teamfossa.slack.com/archives/C0155DTGWB1/p1655439461261949

## Checklist

### DIFF
--- a/src/Discovery/Walk.hs
+++ b/src/Discovery/Walk.hs
@@ -12,22 +12,41 @@ module Discovery.Walk (
   findFilesMatchingGlob,
 ) where
 
-import Control.Carrier.Writer.Church
-import Control.Effect.Diagnostics
+import Control.Carrier.Writer.Church (
+  Has,
+  WriterC,
+  runWriter,
+  tell,
+ )
+import Control.Effect.Diagnostics (
+  Diagnostics,
+  context,
+  recover,
+ )
 import Control.Effect.Reader (Reader, ask)
-import Control.Monad.Trans
-import Control.Monad.Trans.Maybe
+import Control.Monad.Trans (MonadTrans (lift))
+import Control.Monad.Trans.Maybe (MaybeT (MaybeT, runMaybeT))
 import Data.Foldable (find)
 import Data.Functor (void)
 import Data.Glob qualified as Glob
 import Data.List ((\\))
-import Data.Maybe (mapMaybe)
+import Data.Maybe (fromMaybe, mapMaybe)
 import Data.Set qualified as Set
 import Data.String.Conversion (toString)
 import Data.Text (Text)
 import Discovery.Filters (AllFilters, pathAllowed)
-import Effect.ReadFS
-import Path
+import Effect.ReadFS (ReadFS, getIdentifier, listDir)
+import Path (
+  Abs,
+  Dir,
+  File,
+  Path,
+  dirname,
+  filename,
+  parseRelDir,
+  stripProperPrefix,
+  toFilePath,
+ )
 
 data WalkStep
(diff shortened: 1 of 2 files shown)


src/Discovery/Walk.hs
```diff
@@ -148,7 +167,11 @@ walkDir handler topdir =
         Nothing -> pure $ Just ()
         Just traversed' -> walktree traversed' curdir
     walktree traversed curdir = do
-      (subdirs, files) <- listDir curdir
+      -- If we can't list the directory contents, we just skip the directory
+      -- and claim that it has no children.  We don't log it because it would
+      -- happen for every discovery process and would be very noisy.
+      -- TODO: use a (State (Set SkippedDirs)) effect to track and report this
+      (subdirs, files) <- fmap (fromMaybe ([], [])) $ recover $ listDir curdir
```
periodic, line 174: Any reason not to use `fromMaybe ([], []) <$> recover $ listDir curdir`?  It doesn't seem to be a firm choice in our style guide, but I like that it makes the read order a bit more consistent instead of having to jump way over to the `fmap` to see how that affects the `fromMaybe`.

periodic (approved):
LGTM.

As for a testing plan, you could probably write a test that revokes list permission from a directory, runs the test, then gives back any permissions required for clean-up, if any.

src/Discovery/Walk.hs
scruffystuffs: I think it makes it easier to read since there's no operator confusion this way, but with `a <$> b $ c` it can sometimes not be clear where the functor operation reaches to.  This particular case is a bad example, but this pattern chains well.

scruffystuffs: Also, I think that functions like `fmap` aren't really interesting in terms of reviewing.  If the types are strict, and there's only one way a function could work with the right type, I just generally assume that it's laid out right.  If we had to check that kind of thing anyway, what's the point of having the strict types?

periodic: I don't understand what you mean about "aren't really interesting in terms of reviewing".

I'm just pointing out that as a reader this line is a little awkward because it mixes prefix and infix operators. I think leaning more towards parentheses or more towards infix operators would make it easier to read and understand at a glance.

scruffystuffs: Actually, looking closer at this, there's a good argument for making this pattern a diagnostic helper.

periodic: Sorry, I mispoke.  When said we were mixing operators I meant mixing infix and prefix functions.  It's not really important for my point.

I'm trying to focus on what is easiest to read and understand at a glance.  I found this line harder to read than most and I think we could do something to make it easier to read.  That could be using conventions like `<$>` (which I may just be personally comfortable with), it may be using multiple lines, it may be refactoring the pattern into a helper (as you suggested).  All I'm suggesting is that you could probably make this easier to read before merging, but I'm not blocking on it.

Though I'm confused on one point. When you say "interesting to a reviewer", do you mean me as the code reviewer, or do you mean a general reader reviewing the code?

scruffystuffs: reviewer ~ code reviewer

Also, I made it easier by adding helpers to diagnostics.

PR #984, opened 2022-06-23
[No Ticket] Includes fossa endpoint version in the scan summary and debug bundle

meghfossa:
# Overview

This PR includes the endpoint version in the debug bundle, and in the scan summary. I know this format is changing eventually, but wanted to get this in to help triage some issues as they come up.

## Acceptance criteria

- endpoint version is included in scan summary
- endpoint version is included in debug bundle, when `--debug` is passed
- endpoint version is 'n/a' when --output flag is passed

## Testing plan

```
mkdir sandbox && echo 'numpy' > sandbox/reqs.txt
fossa analyze sandbox
```

You should see endpoint's version in the scan summary, likewise if you pass `--debug`, you should see the version in the debug bundle.

## Risks

N/A

## Checklist

### DIFF
--- a/src/Control/Effect/FossaApiClient.hs
+++ b/src/Control/Effect/FossaApiClient.hs
@@ -15,6 +15,7 @@ module Control.Effect.FossaApiClient (
   getIssues,
   getLatestBuild,
   getLatestScan,
+  getEndpointVersion,
   getOrganization,
   getProject,
   getAnalyzedRevisions,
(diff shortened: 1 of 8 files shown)


src/Control/Effect/FossaApiClient.hs
```diff
@@ -84,6 +85,7 @@ data FossaApiClientF a where
   GetApiOpts :: FossaApiClientF ApiOpts
   GetAttribution :: ProjectRevision -> ReportOutputFormat -> FossaApiClientF Text
   GetIssues :: ProjectRevision -> FossaApiClientF Issues
+  GetEndpointVersion :: FossaApiClientF (Maybe Text)
```
spatten, line 88: We should this new endpoint to the `matchExpectation` definition in `Test.MockApi`

spatten (approved):
One minor nit, but other than that looks great. It will be very nice to have this info when debugging

PR #985, opened 2022-06-23
Creates a parser for Alpine Linux package databases.

periodic:
# Overview

Creates a parser for Alpine Linux installed-packages database files.

These files are simple text files.  They consist of:
- A list of packages separated by an empty line.
- A package is a list of properties separated across lines.
- A property is a key and a value separated by a colon.
- A key is a character that denotes a package attribute.

See https://wiki.alpinelinux.org/wiki/Apk_spec for more information on the spec (there isn't much).

This work somewhat depends on extracting the package database from docker images, but that is a separate PR.

## Acceptance criteria

Alpine Linux package databases can be parsed and understood.

## Testing plan

Unit tests were added for the parser.

An example file was generated by getting the package database from the latest. `alpine` docker image.

```bash
docker run -t alpine:latest cat /lib/apk/db/installed > installed.example
```

## Risks

None that I can think of.  There is still end-to-end testing to do in the future and this code is not yet invoked.

## References

- [ANE-271](https://fossa.atlassian.net/browse/ANE-271): User can scan apk database

## Checklist


src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,126 @@
...
+    Nothing -> pure []
+    Just x -> (x :) <$> many (try (sep >> p))
+
+-- | Parses one or more occurrences of a construct with separators between.
+--
+-- See `sepByWithTry` for an explanation and motivation.
+--
+-- This implementation is written to mirror the source of the generic version:
+--   https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/src/Control.Monad.Combinators.html#sepBy1
+sepByWithTry1 :: MonadParsec e s m => m a1 -> m a2 -> m (NonEmpty a1)
```
scruffystuffs, line 75: Using different letters for type variables is a more commonly seen convention.
```suggestion
sepByWithTry1 :: MonadParsec e s m => m a -> m b -> m (NonEmpty a)
```

If you want to show that `b` is discarded, or used as the separator, you could do
```suggestion
sepByWithTry1 :: MonadParsec e s m => m a -> m sep -> m (NonEmpty a)
```

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
```
scruffystuffs, line 75: Using different letters for type variables is a more commonly seen convention.
```suggestion
sepByWithTry1 :: MonadParsec e s m => m a -> m b -> m (NonEmpty a)
```

If you want to show that `b` is discarded, or used as the separator, you could do
```suggestion
sepByWithTry1 :: MonadParsec e s m => m a -> m sep -> m (NonEmpty a)
```

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+-- the separator consumes input after the last entry then the whole parser will
+-- fail because it will fail to parse another entry after the separator.
+--
+-- TODO: If this is more generally useful we should move it to a common location.
+--
+-- The Megaparsec project chose not to include this in their library:
+--    https://github.com/mrkkrp/megaparsec/issues/401
+-- This implementation is written to mirror the source of the generic version:
+--   https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/src/Control.Monad.Combinators.html#sepBy
+sepByWithTry :: MonadParsec e s m => m a1 -> m a2 -> m [a1]
```
scruffystuffs, line 62: Same type var comment
```suggestion
sepByWithTry :: MonadParsec e s m => m a -> m b -> m [a]
```

periodic: Oh, yep.  Whoops.  I left in the defaults that HLS generated for me.

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
+-- | Parsing functions for APK files.
+-- APK is the package manager for Alpine Linux (https://alpinelinux.org/).
+--
+-- The database itself is just a flat file with entries in text.  These are
+-- specified in a key-value format with obscure names.  See
+-- https://wiki.alpinelinux.org/wiki/Apk_spec#Syntax.
```
scruffystuffs, line 6: Can we add a trivial inline example to these docs?
Something like this, but with explanations of what we do with the first set of kv pairs?  Which ones we read, which ones we discard, if there are relevant ones we don't use yet, etc.

The idea is to prevent what you ran into with having to read an obscure spec that just links to code, or making the reader read THIS module's code.
```
C: whatever
P: sompackage
A: someattr
f: foobar
V: 1.2.3

<another package>

```

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+  many,
+  optional,
+  some,
+  try,
+ )
+import Text.Megaparsec.Char (char, eol, letterChar)
+
+-- | Errors that can occur even in a valid file.  Packages are a lose collection
+-- of fields, but it unclear which are required.  These error represent fields
+-- we expect, but which may be missing in unusual circumstances.
```
scruffystuffs, line 28: [nit]: typo and wording
```suggestion
-- | Errors from file with valid syntax but invalid data.  Packages are a loose collection
-- of fields, but it is unclear which are required.  These errors represent fields
-- we expect, but which may be missing in unusual circumstances.
```

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+-- The Megaparsec project chose not to include this in their library:
+--    https://github.com/mrkkrp/megaparsec/issues/401
+-- This implementation is written to mirror the source of the generic version:
+--   https://hackage.haskell.org/package/parser-combinators-1.3.0/docs/src/Control.Monad.Combinators.html#sepBy
+sepByWithTry :: MonadParsec e s m => m a -> m b -> m [a]
+sepByWithTry p sep = do
+  r <- optional (try p)
+  case r of
+    Nothing -> pure []
+    Just x -> (x :) <$> many (try (sep >> p))
```
scruffystuffs, line 67: Does `many . try` create an infinite loop?

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+versionKey = "V"
+
+-- | Parses zero or more occurrences of a construct with separators between.
+--
+-- This differs from `Control.Monad.Combinators.sepBy` because it will backtrack
+-- failed repetitions.  The generic version has no notion of backtracking so if
+-- the separator consumes input after the last entry then the whole parser will
+-- fail because it will fail to parse another entry after the separator.
+--
+-- TODO: If this is more generally useful we should move it to a common location.
```
scruffystuffs, line 56: I think the inner usages of `try` would make me want to avoid using this generally.

src/Strategy/AlpineLinux/Parser.hs
scruffystuffs: I'm still not sure why we need all of the backtracking logic.  Can you explain why we can't just do `try (sepBy p sep)`?

test/AlpineLinux/ParserSpec.hs
```diff
@@ -0,0 +1,70 @@
...
+          val `shouldBe` []
+    withExampleDatabase $
+      it "parses the example database" $ \exampleDatabase -> do
+        let packages = parse installedPackagesDatabaseParser testFileName exampleDatabase
+        case packages of
+          Left err ->
+            expectationFailure $ "Unexpected parse failure: " <> show err
+          Right p -> do
+            length p `shouldBe` 14
+            (take 1 . drop 5 $ p)
```
meghfossa, line 34: [Nit, Optional] We should assert all of the entries in the file, or not include the other entries in the file

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+  x <- try p
+  (x :|) <$> many (try (sep >> p))
+
+-- | Parses an Alpine Linux installed-packages database file.
+-- This returns a list of packages or errors that represent malformed packages.
+--
+-- Packages are separated from one another by one-or-more newlines.  The end of
+-- a file may have any number of trailing newlines.
+installedPackagesDatabaseParser :: Parser [Either PackageError AlpinePackage]
+installedPackagesDatabaseParser =
```
meghfossa, line 86: [Optional, Nit] might be a good idea to have a lexer (or trim them after), right now adding a space after the value in the field will break the parser. Folks may have modified this file manually, or other tools may have made modifications - good to stay on side of caution. 

```haskell
-- note the space after version value
P:package-name\r\nA:x86-64\r\nV:some-version \r\n\r\n
```

meghfossa (commented):
Left some Optionals - mostly related to readability. Will wrap up the review tomorrow (mostly looking at `propertiesParser`, `installedPackagesDatabaseParser`, and `sepByTry`)

src/Strategy/AlpineLinux/Parser.hs
periodic: Sure, I'll try to explain it and it'll help me understand it more clearly along the way.

Let's say that we have a input that describes two packages, like this:

```
prop1
prop2

prop3
prop4
```

You could generalize this to a stream of `prop<eol>prop<eol><eol>prop<eol>prop` where `prop` is parsed by whatever parses properties and `eol` is the end-of-line.

`sepBy` works by consuming the first `prop`, then many `eol >> prop` sequences.  On this input that means it will consume the first `prop`, then `<eol>prop`, but then it goes to do it's second `eol >> prop`.  The `eol` succeeds, but `prop` fails.  That actually causes the whole `sepBy` to fail!  If we put `try` around the `sepBy` it will still fail, it'll just backtrack when it does it.  What we want is to backtrack that failing separator that gets consumed at the end.

This Issue sort of explains it too: https://github.com/mrkkrp/megaparsec/issues/401

src/Strategy/AlpineLinux/Parser.hs
periodic: It makes the semantic similar to what we get in other parsers like attoparsec, which might be more intuitive.  I find it makes more sense to work this way.

One weird thing is that different constructs have different backtracking behavior in Megaparsec.  (From the docs) "As of Megaparsec 4.4.0, string backtracks automatically (see tokens), so it does not need try."  This means that you might refactor something slightly to have a compound parsers instead of a primitive one and suddenly your `sepBy` would break in surprising ways.  I think this version is "safer" and less error prone.

Also, being restricted to Megaparsec with that `try` at least means it's not going to cause changes to other monads.

src/Strategy/AlpineLinux/Parser.hs
scruffystuffs: I think separating the lexer and parser stages is a great way to solve the oddities of this parser code that arise from the oddities of the format.

scruffystuffs: The try-based `sepBy` combinator might not be needed if operating on a `Seq ApkPackageToken`.

periodic: Trimming makes sense.  That's pretty easy to put in.  I'm not sure we need a full-blown lexer though.

test/AlpineLinux/ParserSpec.hs
periodic: I went back and forth on it.  It was a lot to write manually when I was making the test before the code was done.  But now that I have things working it should be easy enough to run it, see the output, double-check it and copy it into the test file.  That's a lot less work than building all the objects manually.

src/Strategy/AlpineLinux/Parser.hs
scruffystuffs: The phrase "full-blown" makes it sound more difficult than I think it is (or less valuable).  I think that separating the two makes each part simpler, and most of the work you've done handles the lexing part anyway.

I'd like to see more separation of lexing and parsing in the codebase in general, specifically because it makes each part of the code much simpler to understand and is primarily useful in desugaring these overloaded tokens: a lexer could make a clear differentiation between a single new line and 2+ newlines, by tokenizing them as `PropertySeparator` and `PackageSeparator`.

It also allows testing individual parts of the parsing scheme easier.  Testing the details of this is a little clearer than E2E testing a lexing parser:
```haskell
type Lexer = Parsec Void Text
type Parser = Parsec PackageError [ApkToken]

lexer :: Lexer [ApkToken]
...

parser :: Parser AlpinePackage
...

doTheWholeThing someInput = do
  -- Won't actually work out-of-box since the Eithers have different Left types
  apkTokens <- Megaparsec.parse lexer someInput
  package <- Megaparsec.parse parser apkTokens
```

I don't think this is necessarily simpler overall, but its parts are simpler, and we should probably be doing this with our parsers in general.

meghfossa (approved):
Approving tentatively (nothing blocking from my side) - but please make sure to address some of the readability comments, and trimming asks prior to merging. Wes also had good feedback.

scruffystuffs:
I don't want to block this PR on the lexer discussion.  If you can get reliable results, or if you think a split lexer has enough merit
to do, then I'm OK either way for now.

If we have trouble with this implementation later, or in testing now, I'll probably be a little more insistent at that point.  I would have been much more insistent now if the parser was sufficiently complicated without separate stages, but that's not really the case here.

src/Strategy/AlpineLinux/Parser.hs
```diff
@@ -0,0 +1,128 @@
...
+
+-- | Parses a list of properties into a map.
+-- Note: Some properties can be repeated.  This map will only have the last value.
+-- However, none of the properties we are interested in are repeatable.
+propertiesParser :: Parser (Map.Map Text Text)
+propertiesParser =
+  Map.fromList . NE.toList <$> propertyParser `sepByWithTry1` eol
+
+-- | Parses a single property.
+propertyParser :: Parser (Text, Text)
```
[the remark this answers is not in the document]
periodic: I tried putting that for each of the properties and it was really wordy.  I like that I can use the built-in REPL in the Haddock blocks, but it's a lot of noise and the smaller pieces don't seem that interesting.  I'll leave it in so you can see it, but I think it's a bit too verbose and a few top-level examples are probably all that's needed.

What do you think?

periodic:
I just uploaded new code which I think addresses all the smaller comments, but I'll double-check.

I agree that lexers are useful on some more complicated things, but it feels like overkill for a file this simple.  This is basically just some grouped key-value pairs.  I think adding a lexer step would make the code much more verbose (and take longer to write).

I'd like to invoke the principle of incremental development.  Right now this code isn't even used anywhere.  We haven't tried it on more than a handful real-world examples.  If it gets much more complicated I'd vote for a lexer too.  At this point I don't see much benefit.

src/Strategy/AlpineLinux/Parser.hs
periodic: I spend so much time trying to remember to write "lose" instead of "loose" that I made the opposite mistake here. :joy:

src/Strategy/AlpineLinux/Parser.hs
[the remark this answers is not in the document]
meghfossa: yep - agreed 👍

PR #986, opened 2022-06-23
Adds `fossa test --diff` command

meghfossa:
# Overview

This PR, adds `fossa test --diff <ARG>` command.

## Acceptance criteria

- `fossa test --revision a --diff b`, only reports issues that are in revision `a`, but not in revision `b`.
- `fossa test --revision a --diff b`, provides an actionable error message when `diff` is not supported by FOSSA instance.
- `fossa test --revision a --diff b`, provides an actionable error message when revision `b` is not part of the same project.
- `fossa test --revision a --diff a`, always succeeds.

### Current `fossa test` Output

CleanShot 2022-06-23 at 20 05 29@2x

### When `fossa test --diff` Passes

CleanShot 2022-06-23 at 20 06 26@2x

### When `fossa test --diff` Fails

CleanShot 2022-06-23 at 20 07 15@2x

### When `fossa test --diff` is not supported

CleanShot 2022-06-23 at 20 08 15@2x

## Testing plan
Environment:

```bash
git checkout feat/fossa-test-diff && make install-dev
```

(You will need latest FOSSA/develop)

Setup:

```bash
mkdir sandbox && cd sandbox

# numpy is BSD
echo 'numpy' > reqs.txt
fossa-dev analyze --project ane-388 --revision 1

# FFPopSim is GPL
echo 'numpy\nFFPopSim' > reqs.txt
fossa-dev analyze --project ane-388 --revision 2

# black is MIT
echo 'numpy\nFFPopSim\nblack' > reqs.txt
fossa-dev analyze --project ane-388 --revision 3

```

1) Test without diff (for regression)

```bash
fossa-dev test --project ane-388 --revision 1 # should pass
fossa-dev test --project ane-388 --revision 2 # should fail
fossa-dev test --project ane-388 --revision 3 # should fail
```

2) Test with diff (when no new issues are discovered)

```bash
fossa-dev test --project ane-388 --revision 3 --diff 2 # should pass
```

3) Test with diff (when new issues are discovered)

```bash
fossa-dev test --project ane-388 --revision 3 --diff 1 # should fail
fossa-dev test --project ane-388 --revision 2 --diff 1  # should fail
```

## Risks
N/A

## References
https://fossa.atlassian.net/browse/ANE-388

## Checklist

### DIFF
--- a/docs/references/subcommands/test.md
+++ b/docs/references/subcommands/test.md
@@ -25,3 +25,33 @@ By default, `fossa test` displays issues in a human-readable format. To instead 
 ```sh
 fossa test --json
 ```
+
+### Test against previous revision
+
+With `--diff <ARG>`, you can check for _only_ new issue comapred to some `<ARG>` revision. 
+
+```sh
+fossa test --diff revisionToCompare
+```
+
+#### Example
+
+```sh
+fossa test --revision 34021e --diff v2.0.0
+```
+
+This will only report issues that are present in `34021e` revision,
+BUT are not present in revision `v2.0.0`.
+
+For instance, 
+
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issue `A`, 
+  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+  
+
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has 0 issues, 
+  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+
+
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issues `A`, `B`, 
+  * `fossa-cli` will report issue `B`, and will exit with status code of 1.
--- a/docs/references/subcommands/test.md
+++ b/docs/references/subcommands/test.md
@@ -26,32 +26,32 @@ By default, `fossa test` displays issues in a human-readable format. To instead 
 fossa test --json
 ```
 
-### Test against previous revision
+### Test for new issues compared to another revision
 
-With `--diff <ARG>`, you can check for _only_ new issue comapred to some `<ARG>` revision. 
+`--diff <REVISION>` configures FOSSA to only report new issues observed with the current revision that weren't already reported on the specified `<REVISION>`.
 
-```sh
+\```sh
 fossa test --diff revisionToCompare
-```
+\```
 
 #### Example
 
-```sh
+\```sh
 fossa test --revision 34021e --diff v2.0.0
-```
+\```
 
-This will only report issues that are present in `34021e` revision,
+This only reports issues that are present in `34021e` revision,
 BUT are not present in revision `v2.0.0`.
 
 For instance, 
 
 * If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issue `A`, 
-  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+  * `fossa-cli` reports no new issues discovered and exits with status code of 0.
   
 
 * If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has 0 issues, 
-  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+  * `fossa-cli` reports no new issues discovered and exits with status code of 0.
 
 
 * If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issues `A`, `B`, 
-  * `fossa-cli` will report issue `B`, and will exit with status code of 1.
+  * `fossa-cli` reports issue `B` and exits with status code of 1.
--- a/src/App/Fossa/API/BuildWait.hs
+++ b/src/App/Fossa/API/BuildWait.hs
@@ -4,6 +4,7 @@ module App.Fossa.API.BuildWait (
   waitForBuild,
 ) where
 
+import App.Fossa.Config.Test (DiffRevision)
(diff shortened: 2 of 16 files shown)


src/App/Fossa/Config/Test.hs
```diff
@@ -96,6 +106,7 @@ parser =
     <*> optional (option auto (long "timeout" <> help "Duration to wait for build completion in seconds (Defaults to 1 hour)"))
     <*> flag TestOutputPretty TestOutputJson (long "json" <> help "Output issues as json")
     <*> baseDirArg
+    <*> optional (strOption (long "diff" <> help "Checks for new issues of the revision, that does not exist in provided diff revision."))
```
meghfossa, line 109: Feels akwd, Open to suggestions for better verbiage here.

docs/references/subcommands/test.md
```diff
@@ -25,3 +25,33 @@ By default, `fossa test` displays issues in a human-readable format. To instead
...
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issue `A`, 
+  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+  
+
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has 0 issues, 
+  * `fossa-cli` will report no new issues discovered, and will exit with status code of 0.
+
+
+* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issues `A`, `B`, 
+  * `fossa-cli` will report issue `B`, and will exit with status code of 1.
```
jssblck, line 57: [optional] Reworded this to flow a little more nicely. In particular:
- Refactors to keep the focus on what FOSSA does, not on the user (e.g., removes "you").
- Refactors "will do" style statements in favor of "does" style statements, which are more clear and direct.

(Escaping the triple-backticks so that they don't break the suggestion block)

```suggestion
### Test for new issues compared to another revision

`--diff <REVISION>` configures FOSSA to only report new issues observed with the current revision that weren't already reported on the specified `<REVISION>`.

\```sh
fossa test --diff revisionToCompare
\```

#### Example

\```sh
fossa test --revision 34021e --diff v2.0.0
\```

This only reports issues that are present in `34021e` revision,
BUT are not present in revision `v2.0.0`.

For instance, 

* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issue `A`, 
  * `fossa-cli` reports no new issues discovered and exits with status code of 0.
  

* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has 0 issues, 
  * `fossa-cli` reports no new issues discovered and exits with status code of 0.


* If the revision `v2.0.0` has issue: `A`, and the revision `34021e` has issues `A`, `B`, 
  * `fossa-cli` reports issue `B` and exits with status code of 1.
```

src/App/Fossa/API/BuildWait.hs
```diff
@@ -84,15 +85,16 @@ waitForIssues ::
   , Has (Lift IO) sig m
   ) =>
   ProjectRevision ->
+  Maybe DiffRevision ->
   Cancel ->
   m Issues
-waitForIssues revision cancelFlag = do
+waitForIssues revision maybeDiffRevision cancelFlag = do
```
jssblck, line 91: [nit, optional] Can we avoid the [Hungarian Notation](https://en.wikipedia.org/wiki/Hungarian_notation) here? We already know it's `Maybe` from the type, and aren't trying to avoid shadowing.
```suggestion
waitForIssues revision diffRevision cancelFlag = do
```

Only commenting here because this is the first time I'm seeing it, but I see lots of `maybe`-prefixed `diffRevisions` that I think would be more simply named without the prefix.

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -578,14 +579,37 @@ getIssues ::
...
   pure (responseBody response)
 
+data EndpointDoesNotSupportIssueDiffing = EndpointDoesNotSupportIssueDiffing
+instance ToDiagnostic EndpointDoesNotSupportIssueDiffing where
+  renderDiagnostic (EndpointDoesNotSupportIssueDiffing) =
+    vsep
+      [ "Provided endpoint does not support issue diffing."
+      , ""
+      , "If you are an on-prem customer, you will likely need to"
+      , "update your FOSSA server instance."
```
jssblck, line 610: [nit, optional] Suggest dropping "will" and "you" here too:
```suggestion
      , "If this instance of FOSSA is on-premise, it likely needs to be updated."
```

test/App/Fossa/API/BuildWaitSpec.hs
```diff
@@ -88,22 +88,29 @@ spec =
         commonExpectations
         expectIssuesAvailable
         runWithTimeout $ \cancel -> do
-          issues <- waitForIssues Fixtures.projectRevision cancel
+          issues <- waitForIssues Fixtures.projectRevision Nothing cancel
           issues `shouldBe'` Fixtures.issuesAvailable
 
+      it' "should return when the issues diffs are avilable" $ do
```
jssblck, line 94:
```suggestion
      it' "should return when the issues diffs are available" $ do
```

jssblck (approved):
LGTM! All my comments are optional.

Didn't test as your test plan looks good- since the actual diffing is done core side, seems straightforward from the CLI, just lots of moving parts. If you want me to test though let me know.

PR #992, opened 2022-07-06
document workaround for version truncating in yaml

spatten:
# Overview

I just ran into a fun YAML mis-feature when doing something with `fossa-deps.yml`. If you have a `fossa-deps.yml` that has a version like `1.0` or `2.0`, YAML helpfully turns the 1.0 into a 1, and we flag it as an unknown dependency because version 1 of that dependency does not exist (1.0 does, though).

```
referenced-dependencies:
- type: maven
  name: javax.websocket:javax.websocket-client-api
  version: 1.0
```

The workaround is to put quotes around the "1.0". I don't think there's a way to actually fix this: this is just how YAML works, right? 

This PR mentions this behaviour in the documentation and puts quotes around the versions in our sample YAML.

## Acceptance criteria

The docs should help users avoid a footgun

### DIFF
--- a/docs/features/manual-dependencies.md
+++ b/docs/features/manual-dependencies.md
@@ -16,10 +16,13 @@ referenced-dependencies:
   name: iron
 - type: pypi
   name: Django
-  version: 2.1.7
+  version: "2.1.7"
 ```
 
 The `name` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.
+
+Note that YAML will convert a version like `1.0` or `2.0` into an integer, which is probably not what you want. The workaround is to
+surround the version with quotes.
 
 Supported dependency types:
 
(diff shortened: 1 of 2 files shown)


docs/features/manual-dependencies.md
```diff
@@ -16,11 +16,14 @@ referenced-dependencies:
   name: iron
 - type: pypi
   name: Django
-  version: 2.1.7
+  version: "2.1.7"
 ```
 
 The `name` and `type` fields are required and specify the name of the dependency and where to find it. The `version` field is optional and specifies the preferred version of dependency.
 
+Note that YAML will convert a version like `1.0` or `2.0` into an integer, which is probably not what you want. The workaround is to
```
scruffystuffs, line 24:
```suggestion
Note that YAML will convert a version like `1.0` or `2.0` into a number (as opposed to a string), which is probably not what you want. The workaround is to
```

docs/features/manual-dependencies.md
```diff
@@ -48,23 +51,26 @@ Supported dependency types:
...
 - name: foo-wrapper
-  version: 1.2.3
+  version: "1.2.3"
   license: MIT
   metadata:
     homepage: https://www.foowrapper.com/about
     description: Provides foo and a helpful interface around foo-like tasks.
 ```
 
+Note that YAML will convert a version like `1.0` or `2.0` into an integer, which is probably not what you want. The workaround is to
```
scruffystuffs, line 72:
```suggestion
Note that YAML will convert a version like `1.0` or `2.0` into a number (as opposed to a string), which is probably not what you want. The workaround is to
```

scruffystuffs (approved):
Unfortunately, this is just how YAML works.

jssblck (approved):

meghfossa (commented):
1 more instance of version example in: https://github.com/fossas/fossa-cli/blob/master/docs/features/vendored-dependencies.md.

PR #994, opened 2022-07-07
[ANE-419] Adds more logging, considers nested subspec, fixes a bug for podspec path

meghfossa:
# Overview

This PR,
- Adds more debug logging for the Podlock file strategy. 
- Updates Podlock strategy to use file suffix to infer external source type instead of attribute
- Considers nested subspecs when using sources of vendored subspecs
- Fixes a bug where exception in `ipc spec` was leading to non-complete results

## Acceptance criteria

- Podlock strategy replaces dependency (of nested subspec) with vendored subspec's git source (if source present)
- Podlock strategy uses alternatives for exception handling
- Podlock strategy has better debug messages, indicating which subspec is in consideration for replacement and which are getting replaced

## Testing plan

```bash
git clone https://github.com/react-native-maps/react-native-maps/ && cd react-native-maps && git checkout 71727a5f && cd example && yarn && cd ios
fossa analyze --project ane419 --revision pr994a
```

You should not see any unknown dependency in FOSSA Web UI.

Note:
- `turbomodule/core` is a nested subspec of ReactCommon

## Risks

N/A

## References

https://fossa.atlassian.net/browse/ANE-419

## Checklist

### DIFF
--- a/src/Strategy/Cocoapods/PodfileLock.hs
+++ b/src/Strategy/Cocoapods/PodfileLock.hs
@@ -17,7 +17,7 @@ import Control.Effect.Diagnostics (
   fatalText,
   fromEither,
   run,
-  (<||>)
+  (<||>),
  )
 import Control.Effect.State (State, get, put)
 import Control.Monad (when)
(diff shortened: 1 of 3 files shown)


src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -209,58 +214,66 @@ buildGraph lockFilePath lockFile@PodLock{lockExternalSources} = do
...
             then pure d
             else do
               revisedDep <- case Map.lookup parentSpec lockExternalSources of
-                Just (ExternalPodSpec podSpecPath) -> fromMaybe d <$> (readPodSubSpecSourceAt podSpecPath parentSpec)
-                Just (ExternalPath podPath) -> fromMaybe d <$> (readPodSubSpecSourceAt (toPodSpecPath podPath parentSpec) candidateSubSpec)
+                Just (ExternalPodSpec podSpecPath) ->
+                  (fromMaybe d <$> (readPodSubSpecSourceAt podSpecPath childSpec)) <||> pure d
+                Just (ExternalPath podPath) ->
+                  (fromMaybe d <$> (readPodSubSpecSourceAt (toPodSpecPath podPath parentSpec) childSpec))
+                    <||> pure d
```
periodic, line 227: It feels a bit odd to have `<||> pure d` on both of these, particularly when they already have `fromMaybe d`.  It looks like the difference between the two is just what path to use.  Maybe this could be refactored into a few smaller steps to create less reuse and make all the alternatives clearer.

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -209,58 +214,66 @@ buildGraph lockFilePath lockFile@PodLock{lockExternalSources} = do
...
-    readPodSubSpecSourceAt podSpecPath candidateSubSpec = context ("Resolving vendored subspec of podspec at " <> showT podSpecPath) $ do
-      podSpecCache :: PodSpecCache <- get
-      (PodSpecJSON ExternalGitSource{urlOf, tagOf, commitOf, branchOf} subSpecs) <-
-        case Map.lookup podSpecPath podSpecCache of
-          Just cached -> pure cached
+    readPodSubSpecSourceAt podSpecPath candidateSubSpec = context
+      ( "Trying to resolve (" <> candidateSubSpec <> "), It is potentially a vendored subspec of podspec at: " <> showT podSpecPath
+      )
+      $ do
+        podSpecCache :: PodSpecCache <- get
```
periodic, line 250: Does this really need a type annotation since there is only one `State` in the signature?

src/Strategy/Cocoapods/PodfileLock.hs
```diff
@@ -209,58 +214,66 @@ buildGraph lockFilePath lockFile@PodLock{lockExternalSources} = do
...
+        (PodSpecJSON ExternalGitSource{urlOf, tagOf, commitOf, branchOf} subSpecs) <-
+          case Map.lookup podSpecPath podSpecCache of
```
periodic, line 252: This caching logic might read better in its own function (top level or `where`-clause)

periodic (approved):
LGTM, though I'd love a few more comments and I wish we had a test, though I know that may be difficult to set up.

PR #997, opened 2022-07-18
Fixes project inference for windows svn

meghfossa:
# Overview

Fixes project inference for SVN in windows environment.

## Acceptance criteria

- We do not include `\r` in revision, branch, and project data.

## Testing plan

1. Setup windows 
2. Setup Svn
3. Run Fossa Analyze

## Risks

N/A

## References

To write up

## Checklist

### DIFF
--- a/src/App/Fossa/ProjectInference.hs
+++ b/src/App/Fossa/ProjectInference.hs
@@ -12,7 +12,7 @@ module App.Fossa.ProjectInference (
   InferredProject (..),
 
   -- * for testing
-  linesWithCR,
+  linesWithoutCR,
 ) where
 
 import App.Types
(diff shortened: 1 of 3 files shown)


src/App/Fossa/ProjectInference.hs
```diff
@@ -103,13 +106,22 @@ inferSVN dir = context "Inferring project from SVN" $ do
...
+
     toProp :: Text -> Maybe (Text, Text)
     toProp propLine =
       case Text.splitOn ": " propLine of
         [key, val] -> Just (key, val)
         _ -> Nothing
 
+-- Removes Windows `\r` from suffix if any
+-- We do this since: Text.lines does not remove \r.
+linesWithCR :: Text -> [Text]
```
csasarak, line 119: [Question] Should this be called `linesWithoutCR`? It is removing the carriage return.

Also, is there a particular reason not to use [replace](https://hackage.haskell.org/package/text-2.0/docs/Data-Text.html#v:replace):
```haskell
linesWithCR = lines . Text.replace "\r\n" "\n"
```
The only thing I can think of is that its time-complexity may be a bit worse, though it's a bit hard to reason about. I don't feel strongly that `replace` is better, besides maybe being shorter. What you have is fine too.

csasarak (approved):
This LGTM. I have a few questions, but they don't require any changes.

PR #1002, opened 2022-07-24
Container Scanning: Adds ManifestJson

meghfossa:
# Overview

This PR adds Os Release parser as part of container scanning scaffolding works.

## Acceptance criteria

- Can parse manifest.json from exported container image tarball

## Testing plan

I have added unit tests. As native container scanning is not configured E/E it cannot be tested manually yet.

## Risks

N/A

## References

## Checklist

### DIFF
--- a/src/Container/Docker/Manifest.hs
+++ b/src/Container/Docker/Manifest.hs
@@ -1,0 +1,70 @@
+{-# LANGUAGE DeriveAnyClass #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+module Container.Docker.Manifest (
+  ManifestJson (..),
+  ManifestJsonImageEntry (..),
+  decodeManifestJson,
+  getLayerPaths,
+  manifestFilename,
+) where
+
+import Control.DeepSeq (NFData)
+import Data.Aeson (
+  FromJSON,
+  ToJSON,
+  defaultOptions,
+  eitherDecode,
+  genericToEncoding,
+  parseJSON,
+  toEncoding,
+  withObject,
+  (.!=),
+  (.:),
+ )
+import Data.ByteString.Lazy qualified as ByteStringLazy
+import Data.List.NonEmpty qualified as NonEmpty
+import Data.Text (Text)
+import GHC.Generics (Generic)
+
+manifestFilename :: Text
+manifestFilename = "manifest.json"
+
+-- | The manifest of container image for the top-level image, and optionally for parent images
+-- that this image was derived from. It consists of an array of metadata entries.
+--
+-- Reference:
+-- * https://github.com/moby/moby/blob/master/image/spec/v1.2.md
+-- * https://github.com/moby/moby/blob/master/image/spec/v1.1.md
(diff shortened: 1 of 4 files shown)


src/Container/Docker/Manifest.hs
```diff
@@ -0,0 +1,70 @@
...
+  deriving (Show, Eq, Ord, Generic, NFData)
+
+instance ToJSON ManifestJsonImageEntry where
+  toEncoding = genericToEncoding defaultOptions
+
+instance FromJSON ManifestJsonImageEntry where
+  parseJSON = withObject "" $ \o ->
+    ManifestJsonImageEntry
+      <$> o .: "Config"
+      <*> o .: "RepoTags" .!= []
```
zlav, line 63: Are there scenarios where we would see no repo tags and pass through an empty `[]`? If so can you also add a test case for this? I don't see it represented in the spec file.

src/Container/Docker/Manifest.hs
```diff
@@ -0,0 +1,70 @@
...
+  (.!=),
+  (.:),
+ )
+import Data.ByteString.Lazy qualified as ByteStringLazy
+import Data.List.NonEmpty qualified as NonEmpty
+import Data.Text (Text)
+import GHC.Generics (Generic)
+
+manifestFilename :: Text
+manifestFilename = "manifest.json"
```
zlav, line 33: Is the file guaranteed to be named `manifest.json`? Even if it can be modified I don't think we should support it until necessary, I'm just curious.

src/Container/Docker/Manifest.hs
```diff
@@ -0,0 +1,70 @@
...
+  deriving anyclass (NFData) -- explicitly use DeriveAnyClass strategy to avoid deriving-defaults warning.
+
+instance FromJSON ManifestJson
+instance ToJSON ManifestJson where
+  toEncoding = genericToEncoding defaultOptions
+
+data ManifestJsonImageEntry = ManifestJsonImageEntry
+  { config :: Text -- references another file in the tar which includes the image JSON for this image.
+  , repoTags :: [Text] -- references pointing to this image.
+  , layers :: NonEmpty.NonEmpty FilePath -- points to the filesystem changeset tars
```
zlav, line 52: Why does `layers` need to be a `NonEmpty FilePath`? Doesn't `FromJSON` already guarantee that it will parse a `FilePath` or it will fail?

src/Container/Docker/Manifest.hs
meghfossa: yes - this is part of Spec for tarball exported docker image.

src/Container/Docker/Manifest.hs
meghfossa: yes - but the container will be of type `[FilePath]`, which does not guarantee at least one layer property; Also, it's easier to perform `head` safely later on biz code.

src/Container/Docker/Manifest.hs
meghfossa: I don't expect to, but I chose to stay on side of caution - I've removed the default assoc now

src/Container/Docker/Manifest.hs
zlav: Makes sense, thanks for the clarification!

zlav (approved):
Looks great! Thanks for answering my questions

PR #1003, opened 2022-07-25
Container Scanning: Adds entry point scaffolding

meghfossa:
# Overview

This PR adds entry point scaffolding for experimental container scanning.

## Acceptance criteria

- When `-x` or `--experimental-scanner` switch is used, a native container scanning path is used.

## Testing plan

- Compile, and `fossa-dev container analyze '' -x`

## Risks

N/A

## References

## Checklist

### DIFF
--- a/src/App/Fossa/Config/Container/Analyze.hs
+++ b/src/App/Fossa/Config/Container/Analyze.hs
@@ -46,6 +46,7 @@ import Options.Applicative (
   progDesc,
   short,
   strOption,
+  switch,
  )
 
 data NoUpload = NoUpload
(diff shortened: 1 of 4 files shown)


src/App/Fossa/Config/Container/Analyze.hs
```diff
@@ -95,6 +98,7 @@ cliParser =
       )
     <*> metadataOpts
     <*> imageTextArg
+    <*> switch (long "experimental-scanner" <> short 'x' <> help "Uses experimental fossa native container scanner")
```
jssblck, line 101:
```suggestion
    <*> switch (long "experimental-scanner" <> help "Uses experimental fossa native container scanner")
```
Explicit goals for the `experimental-` prefix were:
- Make it extremely clear to users that they're doing something outside our normal SLA.
- Avoid colliding with another flag (especially since we maintain flags ~forever).

Would prefer we do not provide a shorthand version.

jssblck (approved):
LGTM but please consider my comment blocking- if you strongly believe we should have the shorthand alias let's talk about it!

src/App/Fossa/Config/Container/Analyze.hs
meghfossa: Agree with your assessment; I will remove the `x` flag.

PR #1004, opened 2022-07-25
Container Scanning: Adds Tarball Parsing Scaffolding

meghfossa:
# Overview

This PR adds Tarball parsing as part of container scanning scaffolding works.

## Acceptance criteria

- Can parse docker exported tarball file (and nested layer file)

## Testing plan

I have added unit tests. As native container scanning is not configured E/E it cannot be tested manually yet.

## Risks

N/A

## References

## Checklist


.gitattributes
```diff
@@ -0,0 +1 @@
+*.tar filter=lfs diff=lfs merge=lfs -text
```
meghfossa, line 1: From retro, it seems we are ok with git lfs

src/Container/Tarball.hs
```diff
@@ -0,0 +1,183 @@
...
+  NLE.NonEmpty FilePath ->
+  Either (NLE.NonEmpty ContainerImgParsingError) ContainerImageRaw
+mkImage entries layerTarballPaths =
+  if null errs
+    then case parsedLayers of
+      [] -> Left $ NLE.fromList [ContainerNoLayersDiscovered]
+      l -> Right $ ContainerImageRaw $ NLE.fromList l
+    else -- This is safe, since we asserted that layers are not null
+    -- earlier in main clause.
+      Left $ NLE.fromList errs
```
csasarak, line 80: It's possible to avoid the partial functions here using pattern matches. I've compressed the two layers of logic here into one using a tuple, but if it's easier to read it can be converted to nested `case`s too.

```suggestion
mkImage ::
  TarEntries ->
  NLE.NonEmpty FilePath ->
  Either (NLE.NonEmpty ContainerImgParsingError) ContainerImageRaw
mkImage entries layerTarballPaths =
  case (errs, parsedLayers) of
    ((e : es), _) -> Left $ e NLE.:| es
    (_, []) -> Left $ NLE.singleton ContainerNoLayersDiscovered
    (_, (l : ls)) -> Right . ContainerImageRaw $ l NLE.:| ls
```

src/Container/Tarball.hs
```diff
@@ -0,0 +1,183 @@
...
+isDoubleWhiteOut :: TarPath -> Bool
+isDoubleWhiteOut = fileNameHasPrefix ".wh..wh."
+
+-- | True if tar path has whiteout marker.
+isWhiteOut :: TarPath -> Bool
+isWhiteOut = fileNameHasPrefix ".wh"
+
+-- | True if tarpath's filename has provided prefix. Otherwise False.
+fileNameHasPrefix :: Text -> TarPath -> Bool
+fileNameHasPrefix prefix path = Text.isPrefixOf prefix (fileNameOf $ toText . fromTarPath $ path)
```
csasarak, line 166: [nit]
```suggestion
fileNameHasPrefix prefix = Text.isPrefixOf prefix . fileNameOf . toText . fromTarPath
```

src/Container/Tarball.hs
```diff
@@ -0,0 +1,183 @@
...
+
+-- | Filename from text path.
+fileNameOf :: Text -> Text
+fileNameOf path = snd $ Text.breakOnEnd "/" path
```
csasarak, line 170: Do we need to support Windows here? Would this fail if the paths there use "\\"?

src/Container/Types.hs
```diff
@@ -0,0 +1,45 @@
...
+newtype ContainerImageRaw = ContainerImageRaw {layers :: NonEmpty.NonEmpty ContainerLayer}
+  deriving (Show, Generic, Eq)
+  deriving anyclass (NFData)
+
+baseLayer :: ContainerImageRaw -> ContainerLayer
+baseLayer img = NonEmpty.head $ layers img
+
+data ContainerLayer = ContainerLayer
+  { layerChangeSets :: !(Seq ContainerFSChangeSet)
+  , lastOffset :: !TarEntryOffset
```
csasarak, line 28: I think `StrictData` makes these `!`s redundant.

src/Container/Types.hs
```diff
@@ -0,0 +1,45 @@
...
+
+instance Monoid ContainerLayer where
+  mempty = ContainerLayer Seq.empty 0
+
+-- | Record indicating a change fs event, with filepath and reference.
+data ContainerFSChangeSet
+  = -- | Record which is inserted or to be modified.
+    InsertOrUpdate !FilePath !TarEntryOffset
+  | -- | Record which is to be removed
+    Whiteout !FilePath
```
csasarak, line 44: I think `StrictData` makes these `!`s redundant.

src/Container/Tarball.hs
```diff
@@ -0,0 +1,183 @@
...
+
+-- | Removes whiteout prefix from the filepath. If no whiteout prefix is detected returns Nothing.
+--
+-- Path with whiteout markers do not have "/" suffix in their filepath, even for directories.
+--
+-- >> removeWhiteOut "etc/hello.txt" = Nothing
+-- >> removeWhiteOut "etc/.wh.hello.txt" = Just "etc/hello.txt"
+-- >> removeWhiteOut "etc/w.h.os" = Just "etc/os"
+removeWhiteOut :: FilePath -> Maybe FilePath
```
csasarak, line 179: [nit] This function is a bit confusing to me because it does two things really, removes white out but also reports whether there was whiteout. The name reflects just one of those things. But I really can't think of a better one myself. 

The one call-site I can find that uses this function converts an existing `TarPath` to a `FilePath` first. It might make sense to have this function just accept the `TarPath` and return another cleaned one. Then you can reuse functions like `isWhiteOut`.

csasarak (changes requested):
I've left some questions and comments.

src/Container/Tarball.hs
```diff
@@ -0,0 +1,183 @@
...
+ )
+import Data.Either (lefts, rights)
+import Data.List.NonEmpty qualified as NLE
+import Data.Sequence (Seq, ViewL (EmptyL, (:<)), viewl, (|>))
+import Data.Sequence qualified as Seq
+import Data.String.Conversion (ToText (toText), toString)
+import Data.Text (Text)
+import Data.Text qualified as Text
+
+data TarEntries = TarEntries
```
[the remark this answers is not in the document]
meghfossa: Current container scanning uses syft application, so it's not FOSSA owned, and we don't have docs such. I can add some comments in this header here to explain the rationale behind the type

src/Container/Errors.hs
```diff
@@ -0,0 +1,37 @@
...
+  show (ManifestJsonParsingFailed err) = "ManifestJsonParsingFailed: " <> show err
+  show (TarParserError err) = "TarParserError: " <> show err
+  show (TarMissingLayerTar path) = "TarMissingLayerTar: " <> show path
+  show (TarLayerNotAFile path) = "TarLayerNotAFile: " <> show path
+  show (ContainerNoLayersDiscovered) = "ContainerNoLayersDiscovered - could not analyze or detect any layer."
+
+instance Exception ContainerImgParsingError
+
+instance NFData ContainerImgParsingError where
+  rnf !_ = () -- fully strict by construction
```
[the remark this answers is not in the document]
csasarak: I also forgot to mention, I learned a lot of this from Apfelmus's [blog](https://apfelmus.nfshost.com/articles/lazy-eval.html). I'll probably re-read myself because this is a complicated topic.

src/Container/Tarball.hs
meghfossa: We only need to support POSIX, will add comment reflecting to this fact;

test/Container/TarballSpec.hs
```diff
@@ -0,0 +1,170 @@
...
+                           ]
+
+spec :: Spec
+spec = do
+  removeWhiteOutSpec
+  mkEntriesSpec
+  mkImageSpec
+
+tarEntriesToPathsOffset :: TarEntries -> [(FilePath, TarEntryOffset)]
+tarEntriesToPathsOffset (TarEntries entries _) = map (first $ fromTarPath . entryTarPath) $ toList entries
```
meghfossa, line 167: use: `fromTarPathToPosixPath`

src/Container/Tarball.hs
```diff
@@ -0,0 +1,180 @@
...
+fileNameHasPrefix :: Text -> FilePath -> Bool
+fileNameHasPrefix prefix = Text.isPrefixOf prefix . fileNameOf . toText
+
+-- | Filename from text path.
+fileNameOf :: Text -> Text
+fileNameOf path = snd $ Text.breakOnEnd "/" path
+
+-- | Retrieves filepath from tar path.
+filePathOf :: TarPath -> FilePath
+filePathOf = fromTarPath
```
meghfossa, line 170: use: `fromTarPathToPosixPath`

src/Container/Types.hs
```diff
@@ -0,0 +1,45 @@
+{-# LANGUAGE DeriveAnyClass #-}
```
csasarak, line 1: [nit] These types seem to only be used in `Tarball`, tests aside. Is there a reason they should not be included in that module?

csasarak (approved):
LGTM.

src/Container/Types.hs
meghfossa: Leaving it open for other sources which may not be tarball centric

PR #1018, opened 2022-08-15
refactor: Update the Hackage index-state

elldritch:
# Overview

This PR updates the Hackage index-state. This allows newer versions of dependencies to be used when solving for the dependency install plan, although it does not necessarily change the versions of dependencies if they still solve to the same version.

## Acceptance criteria

No behavior changes.

## Testing plan

I'm hoping the CI tests will suffice for this.

## Checklist

- [x] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

PR #1019, opened 2022-08-15
refactor: Upgrade Aeson to 2.1.0.0

elldritch:
# Overview

This PR builds on top of #1018.

This PR upgrades Aeson to the latest version, 2.1.0.0. This unblocks a dependency upgrade that Hubble requires.

The main interface change for `aeson-2.1.0.0` is a change to `Object`s to be an ADT instead of literally a `HashMap`. Accordingly, I've fixed some usages of accessors and done some type conversions between `Text` and the new `Key` type where needed.

I've also made import lists explicit where I came across implicit ones.

## Acceptance criteria

No behavior changes.

## Testing plan

I'm banking on CI to be sufficient for this.

## Checklist

### DIFF
--- a/src/Control/Carrier/Debug.hs
+++ b/src/Control/Carrier/Debug.hs
@@ -12,18 +12,49 @@ module Control.Carrier.Debug (
   module X,
 ) where
 
-import Control.Carrier.AtomicState
-import Control.Carrier.Diagnostics
-import Control.Carrier.Output.IO
+import Control.Carrier.AtomicState (AtomicStateC, modify, runAtomicState)
+import Control.Carrier.Diagnostics (
+  Algebra (..),
+  Diagnostics,
+  Handler,
+  Has,
+  SomeDiagnostic (..),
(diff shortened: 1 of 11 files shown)


scruffystuffs:
Did you target the wrong base branch?  Looks like both PR's update the index-state.

elldritch:
I was planning to merge #1018, then rebase and merge this one.

src/Control/Carrier/Debug.hs
```diff
@@ -12,18 +12,49 @@ module Control.Carrier.Debug (
...
+  debugError,
+  debugLog,
+  debugMetadata,
+  debugScope,
+  run,
+  send,
+  thread,
+  (~<~),
+  type (:+:) (..),
+ )
```
scruffystuffs, line 48: Can you leave these two big expansions as implicit for now, and we'll address it separately?  We don't want to limit the re-exports with explicit lists.

scruffystuffs (approved):
LGTM

PR #1025, opened 2022-08-17
Container Scanning: Adds podman support

meghfossa:
# Overview

This PR, 
- Adds podman support for experimental container scanning.

## Acceptance criteria

- `fossa-cli` can use podman for container scanning

#### Setup
```bash
git checkout feat/adds-podman-support
make install-local
./fossa --version
```

- Make sure you have podman installed, and docker engine/client is not accessible.

#### Test

1. Use podman to build the image (so only podman knows about it)
```bash
echo 'FROM alpine:latest\n\nRUN apk add tree' > Dockerfile
podman build . -t testpodman:1.0.0
```

2. Use fossa-container analyzer:
```bash
fossa container analyze testpodman:1.0.0 --experimental-scanner
```

## Risks

N/A

## References

https://fossa.atlassian.net/browse/ANE-279
## Checklist

### DIFF
--- a/src/App/Fossa/Config/Container/Analyze.hs
+++ b/src/App/Fossa/Config/Container/Analyze.hs
@@ -121,7 +121,6 @@ mergeOpts cfgfile envvars cliOpts@ContainerAnalyzeOptions{..} = do
   let scanDest = collectScanDestination cfgfile envvars cliOpts
       severity = getSeverity cliOpts
       imageLoc = containerAnalyzeImage
-      dockerHost = collectDockerHost envvars
       arch = collectArch
       revOverride =
         collectRevisionOverride cfgfile $
@@ -134,7 +133,7 @@ mergeOpts cfgfile envvars cliOpts@ContainerAnalyzeOptions{..} = do
     <*> pure revOverride
     <*> pure imageLoc
     <*> pure containerExperimentalScanner
-    <*> pure dockerHost
+    <*> collectDockerHost envvars
     <*> pure arch
     <*> pure severity
 
--- a/src/App/Fossa/Config/Container/Analyze.hs
+++ b/src/App/Fossa/Config/Container/Analyze.hs
@@ -21,6 +21,8 @@ import App.Fossa.Config.Common (
 import App.Fossa.Config.ConfigFile (ConfigFile)
 import App.Fossa.Config.Container.Common (
   ImageText,
+  collectArch,
+  collectDockerHost,
   imageTextArg,
  )
 import App.Fossa.Config.EnvironmentVars (EnvVars)
@@ -57,7 +59,10 @@ data ContainerAnalyzeConfig = ContainerAnalyzeConfig
   { scanDestination :: ScanDestination
   , revisionOverride :: OverrideProject
   , imageLocator :: ImageText
-  , usesExperimentalScanner :: Bool
+  , -- \* For Experimental Scanner
+    usesExperimentalScanner :: Bool
+  , dockerHost :: Text
+  , arch :: Text
   , severity :: Severity
   }
   deriving (Eq, Ord, Show, Generic)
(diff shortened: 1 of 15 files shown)


src/App/Fossa/Config/Container/Analyze.hs
```diff
@@ -57,7 +59,10 @@ data ContainerAnalyzeConfig = ContainerAnalyzeConfig
   { scanDestination :: ScanDestination
   , revisionOverride :: OverrideProject
   , imageLocator :: ImageText
-  , usesExperimentalScanner :: Bool
+  , -- \* For Experimental Scanner
```
meghfossa, line 62: Migrating some of the config/const values here, so it's easy to diagnose in FOSSA_CONFIG_DEBUG, debug bundle etc, and easier to provide override ux option in future

src/App/Fossa/Container/Sources/Podman.hs
```diff
@@ -0,0 +1,60 @@
...
+podmanInspectImage img =
+  Command
+    { cmdName = "podman"
+    , cmdArgs = ["image", "inspect", img]
+    , cmdAllowErr = Never
+    }
+
+-- | Saves container image to a location in docker archive (tarball) format.
+-- Refer to: https://docs.podman.io/en/latest/markdown/podman-save.1.html
+podmanExtractImage :: Text -> Path Abs File -> Command
```
meghfossa, line 32: We don't need to talk to socket over ssh, we can use the executable itself...

image

src/App/Fossa/Container/AnalyzeNative.hs
```diff
@@ -115,11 +117,23 @@ analyze ::
   m ()
 analyze cfg = do
   logInfo "Running container scanning with fossa experimental-scanner!"
-  parsedSource <- runDockerEngineApi $ parseContainerImageSource (unImageText $ imageLocator cfg)
+
+  let host :: Text
+      host = (dockerHost cfg)
+
+  let defaultArch :: Text
+      defaultArch = arch cfg
```
jssblck, line 125: Why are we pre-declaring the type of these variables? These can be inferred, my preference would be to just let them be inferred.

src/App/Fossa/Container/Sources/DockerEngine.hs
```diff
@@ -26,16 +26,17 @@ analyzeFromDockerEngine ::
...
     logInfo . pretty $
       "Exporting docker image to temp file: "
         <> toString tempTarFile
         <> "! This may take a while!"
-    runDockerEngineApi $ getDockerImage imgTag tempTarFile
+    runDockerEngineApi dockerHost $ getDockerImage imgTag tempTarFile
 
     logInfo . pretty $ "Analyzing exported docker tarball: " <> toString tempTarFile
-    analyzeExportedTarball tempTarFile
+    analyzeFromDockerArchive tempTarFile
```
jssblck, line 42: If we're replacing "tarball" with "archive" in the types, we should probably do the same for logs/errors.

jssblck (approved):

src/App/Fossa/Config/Container/Common.hs
```diff
@@ -18,3 +24,30 @@ instance ToJSON ImageText where
...
+      else arch
+
+collectDockerHost :: EnvVars -> Text
+collectDockerHost EnvVars{envDockerHost} =
+  case envDockerHost of
+    Nothing -> defaultDockerHost
+    Just host ->
+      if Text.isPrefixOf "unix://" host
+        then withoutUnixSocketScheme host
+        else defaultDockerHost
```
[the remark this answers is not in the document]
meghfossa: fixed bu adding warning in [94ebef8](https://github.com/fossas/fossa-cli/pull/1025/commits/94ebef8dd403cdd4aa2723af554e7e695e0b6f84)

PR #1031, opened 2022-08-26
Parse BerkeleyDB container format

jssblck:
# Overview

Adds a parser for BerkeleyDB container format.
This parser is a separate binary and is heavily based on the Go implementation here: https://github.com/knqyf263/go-rpmdb

I'm sorry that this PR is relatively large but this is a very atomic unit of work 🥲

## Acceptance criteria

We can now get blobs from BerkeleyDB. Parsing these actual blobs is done in #1029.

## Testing plan

I relied on automated tests. They're in a separate branch to make review easier/possible.
Tests are all ported from the Go implementation, but testing against the raw blobs instead of the parsed entries.

To run them, run `cargo nextest run` from within the `fossa-cli` root directory after checking out #1032.

## Risks

Not risky.

## References

Between this and #1029, this is the lion's share of [ANE-274](https://fossa.atlassian.net/browse/ANE-274).
All that's left is actually bundling and running this.

## Notes

I'm going to add Rust unit tests to CI later, but I didn't want to bloat this PR larger than it had to be.

## Checklist


scruffystuffs:
Can we pull out the test data into a separate PR so this is easier to review?

jssblck:
> Can we pull out the test data into a separate PR so this is easier to review?

Done @scruffystuffs!

extlib/berkeleydb/src/metadata/generic.rs
```diff
@@ -0,0 +1,123 @@
...
+        if generic.is_big_endian() {
+            r.seek(SeekFrom::Start(0)).context("seek to start")?;
+            let generic = Self::parse_inner::<BigEndian>(r).context("big endian")?;
+            Ok((generic, true))
+        } else {
+            Ok((generic, false))
+        }
+    }
+
+    pub fn parse_inner<E: byteorder::ByteOrder>(r: &mut impl Read) -> Result<Self> {
```
scruffystuffs, line 73: [nit]: Let's add comments explaining why we can't make this (or any other incompatible struct) `ByteParser`

extlib/berkeleydb/src/read/header.rs
```diff
@@ -0,0 +1,73 @@
...
+    /// https://github.com/jssblck/go-rpmdb/blob/956701287363101ee9ade742d6bf1d5c5495f62a/pkg/bdb/constants.go#L16
+    const UNSORTED_PAGE_TYPE: u8 = 2;
+
+    /// https://github.com/jssblck/go-rpmdb/blob/956701287363101ee9ade742d6bf1d5c5495f62a/pkg/bdb/constants.go#L19
+    const SORTED_PAGE_TYPE: u8 = 13;
+
+    /// Read [`Self`] out of a file in plain byte order.
+    ///
+    /// Reference: https://github.com/jssblck/go-rpmdb/blob/160242deff7a9ee82d1b493b62b7e50fd4c3e81c/pkg/bdb/hash_page.go#L24
+    pub fn parse<E: ByteOrder>(r: &mut impl Read) -> Result<Self> {
```
scruffystuffs, line 39: [nit]: Leave a comment explaining why this can't be `ByteParser`

extlib/berkeleydb/src/read/index.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// https://github.com/jssblck/go-rpmdb/blob/956701287363101ee9ade742d6bf1d5c5495f62a/pkg/bdb/constants.go#L10
+    const ENTRY_SIZE: usize = 2;
+
+    /// https://github.com/jssblck/go-rpmdb/blob/160242deff7a9ee82d1b493b62b7e50fd4c3e81c/pkg/bdb/hash_page.go#L102
+    const PAIR_SIZE: usize = Self::ENTRY_SIZE * 2;
+
+    /// Read [`Self`] out of a file in plain byte order.
+    ///
+    /// Reference: https://github.com/jssblck/go-rpmdb/blob/160242deff7a9ee82d1b493b62b7e50fd4c3e81c/pkg/bdb/hash_page.go#L89
+    pub fn parse<E: ByteOrder>(data: &[u8], entries: usize) -> Result<Self> {
```
scruffystuffs, line 23: [nit]: Leave a comment explaining why this can't be `ByteParser`

extlib/berkeleydb/src/read/index.rs
```diff
@@ -0,0 +1,93 @@
...
+
+fn parse_pair_u16<E: ByteOrder>(mut r: impl Read) -> Result<(u16, u16)> {
+    let k = r.read_u16::<E>().context("read pair 0 index")?;
+    let v = r.read_u16::<E>().context("read pair 1 index")?;
+    Ok((k, v))
+}
+
+fn snd<T>((_, t): (T, T)) -> T {
+    t
+}
```
scruffystuffs, line 93: [no action]: I know it's not necessary at all, but this could be more generic:
```suggestion
fn snd<T, X>((_, t): (X, T)) -> T {
    t
}
```

extlib/berkeleydb/src/read/index.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// https://github.com/jssblck/go-rpmdb/blob/160242deff7a9ee82d1b493b62b7e50fd4c3e81c/pkg/bdb/hash_page.go#L102
+    const PAIR_SIZE: usize = Self::ENTRY_SIZE * 2;
+
+    /// Read [`Self`] out of a file in plain byte order.
+    ///
+    /// Reference: https://github.com/jssblck/go-rpmdb/blob/160242deff7a9ee82d1b493b62b7e50fd4c3e81c/pkg/bdb/hash_page.go#L89
+    pub fn parse<E: ByteOrder>(data: &[u8], entries: usize) -> Result<Self> {
+        if entries % 2 != 0 {
+            bail!("entries must only be requested in pairs: {entries}");
+        }
```
scruffystuffs, line 26: [optional]: In general:
```rust
if $negative_condition {
  bail!($err, $args*);
}
```

is identical to `ensure!($positive_condition, $err, $args*);`

For example:
```suggestion
        ensure!(entries % 2 == 0, "entries must only be requested in pairs: {entries}");
```

This is identical, so it's not at all blocking, but I personally prefer reading `ensure!` over `if -> bail!`, so that I can gloss over it while reading data-flow logic, and read it when I care about validation logic.  I think one-liners are easier to ignore for readers who know what they're looking for.

There's quite a few of these in this PR, so I'm comfortable marking that as TODO or just leaving it.

scruffystuffs (approved):
Nothing blocking.  I left quite a few nits and suggestions, but we can leave them.

extlib/berkeleydb/src/metadata/hash.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::io::Read;
+
+use byteorder::{ByteOrder, ReadBytesExt};
+
+use crate::parse::ByteParser;
+
+/// Hash metadata.
+///
+/// This is intended to be parsed immediately _after_ parsing [`crate::metadata::Generic`];
+/// as such the byte offsets reflect the offset after doing so.
```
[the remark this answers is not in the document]
jssblck: It could, but I think it is an extra syscall we don't need- I'd rather handle that in the type system if it becomes a problem.

extlib/berkeleydb/src/read.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn read(&mut self) -> Result<Vec<Value>> {
+        debug!("🔬 Reading values from DB");
+
+        // Not all `u32` fit into `usize` without overflowing, so we have to check this conversion.
+        let page_size = self
+            .metadata
+            .generic()
+            .page_size()
+            .try_into()
+            .context("convert page size from u32")?;
```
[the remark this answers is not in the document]
jssblck: On 32+ bit systems this should just compile to a no-op; I'd rather avoid casting since it's already written. But yes, we could have just casted.

extlib/berkeleydb/src/metadata/hash.rs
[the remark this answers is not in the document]
scruffystuffs: Interesting.  My reasoning was that in the current model, you could parse the structure as a standalone.

Also I don't think you would even need to handle it in the type system, you've already handled it in the API.  Either way, no action needed here.

extlib/berkeleydb/src/read.rs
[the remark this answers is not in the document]
scruffystuffs: TIL that error checks can optimize away, but the playground ASM output confirms it: https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=bd2f9d577a50e6d305ac3fe532095fa6

jssblck:
Created ticket to move BerkeleyDB to FFI: https://fossa.atlassian.net/browse/ANE-550

PR #1032, opened 2022-08-27
Parse BerkeleyDB container format: tests

jssblck:
Adds the tests to #1031.

The only file that needs review is here: https://github.com/fossas/fossa-cli/blob/container/bdb-native-rs-tests/extlib/berkeleydb/tests/it/main.rs


scruffystuffs (approved):
I've read through the test runner file, nothing jumped out at me.  I'm going to rubber-stamp this without checking, unless there's any other review needed.

PR #1033, opened 2022-08-29
Update link

jssblck:
Fixes the link check that's currently failing. It seems the paper "Data Types a la Carte" has been access restricted or removed. I've updated it to a Cambridge University Press link which appears to be live and pointing to the same document.


zlav (approved):
Nice fix, just for reference if the link is flaky there is a way to ignore the link checker

PR #1035, opened 2022-08-30
Automatically filter dist-newstyle from target discovery

zlav:
# Overview

This PR addresses https://fossa.atlassian.net/browse/ANE-538

`dist-newstyle` is a directory used exclusively by Haskell projects to store build artifacts.

> This PR makes me realize that we should probably address or better document how these automatic filters operate. Thankfully we have not seen too many issues pop up related to them.

## Acceptance criteria

Targets from `dist-newstyle` are automatically excluded from the analysis.

## Testing plan

I manually compiled the CLI and validated that targets in dist-newstyle are not picked up. The array of filters does not have an accompanying test suite, and while I think it would be great to build one, I don't think thats necessary for this fix.

## Risks

A user _could_ be relying on targets in `dist-newstyle`, however, this is very unlikely and we have very few customers using Haskell.

## References

[ANE-538](https://fossa.atlassian.net/browse/ANE-538)

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add `dist-newstyle` to the list of automatically filtered directories. ([#1030](https://github.com/fossas/fossa-cli/pull/1035))
 - Fix a bug in fossa-deps.schema.json. It is now valid JSON. ([#1030](https://github.com/fossas/fossa-cli/pull/1030))
 
 ## 3.3.12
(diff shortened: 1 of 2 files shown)


jssblck (approved):
> This PR makes me realize that we should probably address or better document how these automatic filters operate. Thankfully we have not seen too many issues pop up related to them.

I agree, but in a slightly different manner: I think we should implement filters _specific to strategies_. Specifically, `dist-newstyle` is the output dir for Haskell projects, but it's entirely possible that a project of another type actually has a directory named the same but with actual targets inside. This problem becomes very apparent if we consider the end state of having all the output dirs for all languages we support ignored.

Anyway, not super important for now, so this lgtm. But I think we should do something like this before adding more 😄

PR #1038, opened 2022-08-31
Rpm blob doc

csasarak:
# Overview

This is the follow up PR to try to clean up and make the RPM blob parser more readable.

This also includes a small amount of work to expose the `Epoch` field from the RPM headers.

## Acceptance criteria

It should be a bit easier to understand what's going on.

## Testing plan

I just made sure that after my changes each test still ran and added a test for Epoch data.

## Risks

None

## References

[ANE-273](https://fossa.atlassian.net/browse/ANE-273)

## Checklist


jssblck (approved):
LGTM after comments!

PR #1043, opened 2022-09-09
Includes CI environment in telemetry record

meghfossa:
# Overview

This PR, 
- adds CI environment identifier in the telemetry record.

## Acceptance criteria

- When CLI is ran in github actions, telemetry include github_action ci indicator
- When CLI is ran in bamboo (bitbucket), telemetry include bamboo ci indicator
- When CLI is ran in azure pipeline, telemetry include azure ci indicator
- When CLI is ran in generic CI, telemetry includes ci indicator

## Testing plan

```bash
mkdir sandbox

# should be unknown_ci
CI='some-ci' ./fossa analyze sandbox --debug && cat fossa.telemetry.json | jq '.cliCIEnvironment' 

# should be github_action
GITHUB_ACTIONS=true ./fossa analyze sandbox --debug && cat fossa.telemetry.json | jq '.cliCIEnvironment'

# should be azure_pipeline
TF_BUILD=true ./fossa analyze sandbox --debug && cat fossa.telemetry.json | jq '.cliCIEnvironment'

# should be bamboo
bamboo.buildKey=123 ./fossa analyze sandbox --debug && cat fossa.telemetry.json | jq '.cliCIEnvironment'
```

Sibling PR:
- https://github.com/fossas/FOSSA/pull/8362

## Risks

N/A

## Checklist

### DIFF
--- a/src/Control/Carrier/Telemetry/Types.hs
+++ b/src/Control/Carrier/Telemetry/Types.hs
@@ -70,7 +70,7 @@ data TelemetryRecord = TelemetryRecord
   , cliTotalDurationInSec :: Double
   , cliUsageCounter :: Map CountableCliFeature Int
   , cliVersion :: Text
-  , cliCIEnv :: Maybe CIEnvironment
+  , cliCIEnvironment :: Maybe CIEnvironment
   }
   deriving (Show, Eq, Ord, Generic)
 
--- a/src/Control/Carrier/Telemetry/Types.hs
+++ b/src/Control/Carrier/Telemetry/Types.hs
@@ -135,8 +135,6 @@ data CIEnvironment
   | UnknownCI
   deriving (Eq, Ord, Generic)
 
--- Modifying values below, also requires modifying value in
--- FOSSA's server's cli telemetry type for compatibility.
 instance Show CIEnvironment where
   show GithubAction = "github_action"
   show AzurePipeline = "azure_pipeline"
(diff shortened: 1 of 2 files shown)


src/Control/Carrier/Telemetry/Utils.hs
```diff
@@ -91,3 +99,58 @@ getSystemInfo =
     Info.arch
     <$> Conc.getNumCapabilities
     <*> Conc.getNumProcessors
+
+lookupCIEnvironment :: IO (Maybe CIEnvironment)
```
meghfossa, line 103: `IO` is a ramification of the telemetry effect being the top, following fmt of `getSystemInfo`...

src/Control/Carrier/Telemetry/Types.hs
```diff
@@ -125,3 +127,21 @@ data TelemetryTimeSpent = TelemetryTimeSpent
...
+
+data CIEnvironment
+  = GithubAction
+  | AzurePipeline
+  | Bamboo
+  | UnknownCI
+  deriving (Eq, Ord, Generic)
+
+-- Modifying values below, also requires modifying value in
+-- FOSSA's server's cli telemetry type for compatibility.
```
spatten, line 139: Where are these mentioned in the server? I don't see them mentioned in https://github.com/fossas/FOSSA/pull/8362/files or elsewhere in the core codebase

spatten (approved):
This all looks good to me. 

I was unable to test with the `bamboo.buildKey` env variable (`.` is not allowed in env variable names in MacOS and I think Linux), but other than that it all worked.

Also, I'm assuming that it's hard to add tests for this. If not, it would be nice to have something.

src/Control/Carrier/Telemetry/Types.hs
meghfossa: Ah! good catch - I had a fixed implementation approach, so this comment is not needed anymore. I will remove the comment.

PR #1044, opened 2022-09-09
Feat/rpm container sqlite ane 273

csasarak:
# Overview

Should be able to do native container scanning with sqlite based RPM package databases.

## Acceptance criteria

Native scanning will be able to read RPMs from a package database that is sqlite.

## Testing plan

You can verify that this runs with these commands:

```
docker pull fedora:35
fossa container analyze --experimental-scanner fedora:35
```
Then navigate to the page it outputs. You can run the same command without the `--experimental-scanner` option and see that it outputs the same thing.

## Risks

None

## References

[ANE-273](https://fossa.atlassian.net/browse/ANE-273)

## Checklist

### DIFF
--- a/src/Data/Rpm/DbHeaderBlob.hs
+++ b/src/Data/Rpm/DbHeaderBlob.hs
@@ -23,8 +23,6 @@ import DepTypes (
 newtype PkgConversionError = PkgConversionError PkgInfo
   deriving (Show)
 
--- |Turn a list of RPM header blobs into a graph where every package is
--- a direct dependency.
 pkgInfoToDependency :: OsInfo -> PkgInfo -> Either PkgConversionError Dependency
 pkgInfoToDependency
   (OsInfo os osVersion)
--- a/src/Data/Rpm/DbHeaderBlob.hs
+++ b/src/Data/Rpm/DbHeaderBlob.hs
@@ -48,5 +48,5 @@ pkgInfoToDependency
       version = CEq $ pkgArch <> "#" <> epoch <> (pkgVersion <> "-" <> pkgRelease)
 
       epoch :: Text
-      epoch = maybe "" (<> ":") (toText . show <$> pkgEpoch)
+      epoch = maybe "" ((<> ":") . toText . show) pkgEpoch
 pkgInfoToDependency _ pkg = Left $ PkgConversionError pkg
--- a/src/Data/Rpm/DbHeaderBlob.hs
+++ b/src/Data/Rpm/DbHeaderBlob.hs
@@ -2,10 +2,51 @@
 --Description: Functions for reading RPM information out of binary package headers
 module Data.Rpm.DbHeaderBlob (
   PkgInfo (..),
+  PkgConversionError (..),
   readPackageInfo,
+  pkgInfoToDependency,
 ) where
 
+import Container.OsRelease (OsInfo (..))
 import Data.Rpm.DbHeaderBlob.Internal (
   PkgInfo (..),
   readPackageInfo,
  )
+import Data.String.Conversion (ToText (toText))
+import Data.Text (Text)
+import DepTypes (
+  DepType (LinuxRPM),
+  Dependency (..),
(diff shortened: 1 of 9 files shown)


src/Discovery/Walk.hs
```diff
@@ -124,6 +125,10 @@ fileName = toFilePath . filename
 findFileNamed :: String -> [Path a File] -> Maybe (Path a File)
 findFileNamed name = find (\f -> fileName f == name)
 
+-- |Find the first file that matches any of the provided names.
+findFirstMatchingFile :: [String] -> [Path a File] -> Maybe (Path a File)
+findFirstMatchingFile names = find (\f -> fileName f `elem` names)
```
jssblck, line 130: Given `findFirstMatch :: [a] -> [b] -> Maybe b`
Searching all `b`s for one matching any `a` has to rescan all of `[a]` for potentially every `b` (unless GHC is way smarter than I think it is- if you think this is the case please let me know!)

Please convert `names` to a set for faster matches:
```suggestion
-- | Find the first file that matches any of the provided names.
findFirstMatchingFile :: [String] -> [Path a File] -> Maybe (Path a File)
findFirstMatchingFile names = find (\f -> fileName f `Set.member` names')
  where
    names' = Set.fromList names
```

src/Strategy/Sqlite.hs
```diff
@@ -0,0 +1,179 @@
+{-# LANGUAGE TupleSections #-}
+
+-- Description: Discovery and analysis functions for Sqlite3 backed RPM databases
+module Strategy.Sqlite (
+  -- for testing
```
jssblck, line 5:
```suggestion
```

src/Strategy/Sqlite.hs
```diff
@@ -0,0 +1,179 @@
...
+  ) =>
+  OsInfo ->
+  Path Abs Dir ->
+  m [SqliteDB]
+findProjects osInfo = walkWithFilters' $ \dir _ files -> do
+  case findFirstMatchingFile
+    [ "Packages.sqlite"
+    , "rpmdb.sqlite"
+    ]
+    files of
```
jssblck, line 72: [optional] Keep in mind splitting lines increases cognitive load while reading. Prefer leaving stuff like this on one line or perhaps splitting the list into a separate `where` function to make it easier to read instead of splitting.
```suggestion
  case findFirstMatchingFile ["Packages.sqlite", "rpmdb.sqlite"] files of
```

jssblck (approved):
lgtm after comments

src/Data/Rpm/DbHeaderBlob.hs
```diff
@@ -2,10 +2,51 @@
...
+  Dependency (..),
+  VerConstraint (CEq),
+ )
+
+newtype PkgConversionError = PkgConversionError PkgInfo
+  deriving (Show)
+
+-- |Turn a list of RPM header blobs into a graph where every package is
+-- a direct dependency.
+pkgInfoToDependency :: OsInfo -> PkgInfo -> Either PkgConversionError Dependency
```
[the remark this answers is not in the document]
csasarak: Good catch, I had more ambitious plans for this than I ended up implementing and didn't update the comment.

src/Discovery/Walk.hs
csasarak: GHC won't do something smart to my knowlege, my reasoning here was that I wasn't using this function for more than two filenames so the `Set` wouldn't be much benefit. Using a `Set` is much better from an API perspective though since who knows what future folks will do with it.

jssblck: Yeah, I agree with your intuition for sufficiently small `a`s, but since this is a general function better to be safe.
Either that, or just throw a comment in there calling out this as a potential performance footgun.

csasarak: I committed your Set change, after you asked for it I understood that it makes more sense than what I was doing since it's in such a widely used module. Thanks!

PR #1047, opened 2022-09-14
[No Ticket] Container scanning supports hardlinks in image, and increases timeout

meghfossa:
# Overview

This PR:
- adds support for hardlinks
- increases default timeout for docker engine api

## Acceptance criteria

- Can scan very large images using docker engine api

## Testing plan

```bash
➜  fossa-cli git:(fix/experimental-scanner-issues) ✗ docker pull apache/nifi:latest                                                       
➜  fossa-cli git:(fix/experimental-scanner-issues) ✗ ./fossa container analyze apache/nifi:latest --experimental-scanner -o     
```
^ should succeed!

```bash
➜  fossa-cli git:(fix/experimental-scanner-issues) ✗ docker pull busybox:latest                                                      
➜  fossa-cli git:(fix/experimental-scanner-issues) ✗ ./fossa container analyze busybox:latest --experimental-scanner -o (yes it will not find anything, but it should not throw error)
```

^ should succeed!

## Risks

N/A

## References

N/A

## Checklist

- [] I added tests for this PR's change (or explained in the PR description why tests don't make sense).

### DIFF
--- a/src/Container/Tarball.hs
+++ b/src/Container/Tarball.hs
@@ -204,7 +204,7 @@ isSymLink :: Tar.Entry -> Bool
 isSymLink (TarEntry.Entry _ (SymbolicLink _) _ _ _ _) = True
 isSymLink _ = False
 
--- | True if tar entry is for a symbolic link, otherwise False.
+-- | True if tar entry is for a hard link, otherwise False.
 isHardLink :: Tar.Entry -> Bool
 isHardLink (TarEntry.Entry _ (HardLink _) _ _ _ _) = True
 isHardLink _ = False
--- a/src/Container/Tarball.hs
+++ b/src/Container/Tarball.hs
@@ -14,7 +14,7 @@ module Container.Tarball (
 
 import Codec.Archive.Tar (
   Entry (entryContent),
-  EntryContent (NormalFile, SymbolicLink),
+  EntryContent (HardLink, NormalFile, SymbolicLink),
  )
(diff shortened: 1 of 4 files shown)


src/Container/Tarball.hs
```diff
@@ -204,6 +204,11 @@ isSymLink :: Tar.Entry -> Bool
 isSymLink (TarEntry.Entry _ (SymbolicLink _) _ _ _ _) = True
 isSymLink _ = False
 
+-- | True if tar entry is for a symbolic link, otherwise False.
```
spatten, line 207: Should this read "True if tar entry is for a hard link, otherwise False."?

src/Control/Carrier/DockerEngineApi.hs
```diff
@@ -91,8 +95,15 @@ unixSocketClient socketPath = socketHttpManager
     socketHttpManager :: Has (Lift IO) sig m => m HTTP.Manager
     socketHttpManager = sendIO $ HTTP.newManager socketManagerSettings
 
+    fiveMinutes :: ResponseTimeout
+    fiveMinutes = ResponseTimeoutMicro 3_0000_0000
```
spatten, line 99: Nit: can we write this as 300_000_000?

src/Container/Tarball.hs
meghfossa: yes - lazy copy/paste error. fixed.

PR #1050, opened 2022-09-19
[ANE-561] Consistent Project naming for Containers

meghfossa:
# Overview

This PR ensures, 

- Logs of finalized "project" and "revision" value for both `--output` and `uploadScan` targets
- For the project name, for syft scanner, we break on last `:` to ensure port numbers are included in the project name
- For the project name, for the experimental scanner we include the registry host URL in the repository tag to match `syft` scanner's project naming
- incorrect/or/not-supported yet experimental-scanner docs re `fossa-deps` are removed

## Acceptance criteria

- Syft scanner included the host's port in the project name
- Experimental scanner includes registry URL in the project name (just like syft)

## Testing plan

1) Syft scanner included host's port in the project name

```bash
# make local container registry
docker run -d -p 6000:5000 --name registry registry:2.7

# Push an image to this registry
docker pull ubuntu
docker tag ubuntu localhost:6000/ubuntu
docker push localhost:6000/ubuntu

# Run Syft
./fossa container analyze localhost:6000/ubuntu -o # should see [ INFO] Using project name: `localhost:6000/ubuntu`

# Run Experimental Scanner
./fossa container analyze localhost:6000/ubuntu --experimental-scanner -o # should see [ INFO] Using project name: `localhost:6000/ubuntu`
```

2) Experimental scanner includes registry url in the project name (just like syft)

```bash
docker image rm ghcr.io/fossas/haskell-dev-tools:8.10.4
./fossa container analyze ghcr.io/fossas/haskell-dev-tools:8.10.4 --experimental-scanner -o # should see [ INFO] Using project name: `ghcr.io/fossas/haskell-dev-tools:8.10.4`
./fossa container analyze redis --experimental-scanner -o # should see [ INFO] Using project name: `redis`
```

## Risks

1) It's breaking change for any users who were using registry with port `:` - it may create an additional project if they upgrade to CLI with this commit. We expect that not many users (or none at all) are using `fossa-cli` to scan container image from registry with some port value in it's image input

## References

https://fossa.atlassian.net/browse/ANE-561

## Checklist

### DIFF
--- a/src/App/Fossa/Container/Analyze.hs
+++ b/src/App/Fossa/Container/Analyze.hs
@@ -56,6 +56,12 @@ analyze ContainerAnalyzeConfig{..} = do
   logDebug "Running embedded syft binary"
   containerScan <- errCtx (SyftScanFailed imageLocator) (runSyft imageLocator >>= toContainerScan)
   let revision = extractRevision revisionOverride containerScan
+
+  logInfo ("Using project name: `" <> pretty (projectName revision) <> "`")
+  logInfo ("Using project revision: `" <> pretty (projectRevision revision) <> "`")
+  let branchText = fromMaybe "No branch (detached HEAD)" $ projectBranch revision
+  logInfo ("Using branch: `" <> pretty branchText <> "`")
+
   case scanDestination of
     OutputStdout -> logStdout . decodeUtf8 $ encode containerScan
     UploadScan apiOpts projectMeta ->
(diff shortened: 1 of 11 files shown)


src/App/Fossa/Container/Analyze.hs
```diff
@@ -56,6 +56,12 @@ analyze ContainerAnalyzeConfig{..} = do
   logDebug "Running embedded syft binary"
   containerScan <- errCtx (SyftScanFailed imageLocator) (runSyft imageLocator >>= toContainerScan)
   let revision = extractRevision revisionOverride containerScan
+
+  logInfo ("Using project name: `" <> pretty (projectName revision) <> "`")
```
zlav, line 60: I see this as one of the PR acceptance criteria, but why should we be doing this?

`Logs of finalized "project" and "revision" value for both --output and uploadScan targets`

meghfossa: This is so that whichever command we run (regardless of whether the result is to be stdout'd or uploaded); we always print the project name, and revision. Previously, we were only logging project names and revisions if the result was to be uploaded. I included this in AC so, in the future defect/questions related to inferred project names and revisions are easier to diagnose.

```
fossa container analyze
fossa container analyze --output

fossa container analyze --experimental-scanner
fossa container analyze --experimental-scanner --output
```

zlav: Sounds great to me, I love that it will be easier to get at project inference information when debugging.

PR #1052, opened 2022-09-20
fix the v1 installer for windows

spatten:
# Overview

The Windows v1 installers (`install.ps1` and `install-v1.ps1`) were downloading the HTML for a release page and extracting the download URL for the windows version from it.

The HTML no longer contains the download URL, so the installer broke.

The `install-latest.ps1` script does not have this bug. This only needs to be fixed for the v1 installers.

## Acceptance criteria

- `install.ps1` and `install-v1.ps1` should work when tested on a Windows machine
- The `test-installation-scripts` test for Windows should pass again. Here's a link to a run before this fix: https://github.com/fossas/fossa-cli/actions/runs/3086005620/jobs/4989967136

## Testing plan

I tested by creating a windows instance on EC2, connecting using "Microsoft Remote Desktop" and then running both scripts on it.

The original script looks like this when you run it:

```
 .\install-original.ps1


-------------------
Deprecation Warning
-------------------


You are installing FOSSA CLI v1, which is no longer in active
development. FOSSA will not address new defects found in CLI v1.


Please upgrade to the latest FOSSA CLI.


Please upgrade to FOSSA CLI v3 by using install-latest script:
--------------------------------------------------------------
    Set-ExecutionPolicy Bypass -Scope Process -Force; iex  ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/fossas/fossa-cli/master/install-latest.ps1'))


Migration guide for FOSSA CLI v3:
---------------------------------
    https://github.com/fossas/fossa-cli/blob/master/docs/differences-from-v1.md#how-to-upgrade-to-fossa-3x


Did not find latest Windows release at https://github.com/fossas/fossa-cli/releases/v1.1.10
At C:\Users\Administrator\Documents\install-original.ps1:45 char:5
+     throw "Did not find latest Windows release at $latestUri"
+     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : OperationStopped: (Did not find la...eleases/v1.1.10:String) [], RuntimeException
    + FullyQualifiedErrorId : Did not find latest Windows release at https://github.com/fossas/fossa-cli/releases/v1.1.10
```

The new scripts run fine:

```
PS C:\Users\Administrator\Documents> .\install-v1.ps1


-------------------
Deprecation Warning
-------------------


You are installing FOSSA CLI v1, which is no longer in active
development. FOSSA will not address new defects found in CLI v1.


Please upgrade to the latest FOSSA CLI.


Please upgrade to FOSSA CLI v3 by using install-latest script:
--------------------------------------------------------------
    Set-ExecutionPolicy Bypass -Scope Process -Force; iex  ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/fossas/fossa-cli/master/install-latest.ps1'))


Migration guide for FOSSA CLI v3:
---------------------------------
    https://github.com/fossas/fossa-cli/blob/master/docs/differences-from-v1.md#how-to-upgrade-to-fossa-3x


Installed fossa-cli at: C:\ProgramData\fossa-cli\fossa.exe
Get started by running: fossa.exe --help
C:\ProgramData\fossa-cli\fossa.exe
```

I only tested `install-v1.ps1`, as it is exactly the same as `install.ps1`, so there's no need to test both.

## Risks

I did this by hard-coding the download URL, which is safe because this URL is not going to change. 

Another fix would be to copy the approach used by `install-latest.ps1`, which downloads a JSON blob and grabs the download URL from there. I'm happy to go with that approach if anyone feels strongly about it, but I think I like the slightly lazy way I did this fix better.

## References

https://teamfossa.slack.com/archives/C039KE5ERNE/p1663626413126779

## Checklist

### DIFF
--- a/install-v1.ps1
+++ b/install-v1.ps1
@@ -1,4 +1,4 @@
-#Requires -Version 5
+﻿#Requires -Version 5
 
 <#
     .SYNOPSIS
@@ -11,8 +11,8 @@ Param()
 $OldEAP = $ErrorActionPreference #Preserve the original value
 $ErrorActionPreference = "Stop"
 $release_v1_upperbound="v1.1.10"
+$releaseVersionSemver = $release_v1_upperbound.TrimStart("v");
 $github = "https://github.com"
-$latestUri = "$github/fossas/fossa-cli/releases/$release_v1_upperbound"
 $extractDir = "$env:ALLUSERSPROFILE\fossa-cli"
 
 Write-host "`n"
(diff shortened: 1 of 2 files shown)


meghfossa (approved):
lgtm - cI install tests are passing as well!

PR #1045, opened 2022-09-09
[ANE-492] - Documents subproject filtering capabilities (for gradle, etc.)

meghfossa:
# Overview

Documents subproject filtering via target.

## Testing plan

1) Get https://docs.gradle.org/current/samples/zips/sample_building_java_applications_multi_project-kotlin-dsl.zip
2) Update `utilities/build.gradle.kts` to following:

```kts
plugins {
    id("demo.java-library-conventions")
}

dependencies {
    api(project(":list"))

    // Adding dependency for testing
    // https://mvnrepository.com/artifact/org.flywaydb/flyway-core
    implementation("org.flywaydb:flyway-core:9.3.0")
}
```
3) Run `fossa analyze -o | jq`, and observe that `mvn+org.apache.commons:commons-text$1.9` is included from `:app`
4) Create `.fossa.yml`
```yml
version: 3

targets:
  only:
    - type: gradle
      path: ./
      target: ':utilities'
```
5) Run `fossa analyze -o | jq`, and observe that `mvn+org.apache.commons:commons-text$1.9` is not included, only `mvn+org.flywaydb:flyway-core$9.3.0`

## Risks

N/A

## References

https://teamfossa.slack.com/archives/C0155DTGWB1/p1660151930784179?thread_ts=1660142390.251299&cid=C0155DTGWB1
https://fossa.atlassian.net/browse/ANE-492

## Checklist

- [x] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [ ] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [ ] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

PR #1059, opened 2022-09-26
[Docs Only] Fixes minor ambiguity in SSL docs, adds docs on submodule filtering

meghfossa:
# Overview

This PR, 
- Adds docs on submodule filtering
- Fixes minor SSL doc on windows which was ambiguous for windows users

## Testing plan

1) Get https://docs.gradle.org/current/samples/zips/sample_building_java_applications_multi_project-kotlin-dsl.zip
2) Update `utilities/build.gradle.kts` to following:

```kts
plugins {
    id("demo.java-library-conventions")
}

dependencies {
    api(project(":list"))

    // Adding dependency for testing
    // https://mvnrepository.com/artifact/org.flywaydb/flyway-core
    implementation("org.flywaydb:flyway-core:9.3.0")
}
```
3) Run `fossa analyze -o | jq`, and observe that `mvn+org.apache.commons:commons-text$1.9` is included from `:app`
4) Create `.fossa.yml`
```yml
version: 3

targets:
  only:
    - type: gradle
      path: ./
      target: ':utilities'
```
5) Run `fossa analyze -o | jq`, and observe that `mvn+org.apache.commons:commons-text$1.9` is not included, only `mvn+org.flywaydb:flyway-core$9.3.0`

## Risks

N/A

### DIFF
--- a/docs/references/strategies/languages/gradle/gradle.md
+++ b/docs/references/strategies/languages/gradle/gradle.md
@@ -309,7 +309,7 @@ Type   Path         Path   Target      Target
        separator           separator
 ```
 
-1) Now to analyze only `utilities`, use a `.fossa.yml` file in the project root.
+2) Now to analyze only `utilities`, use a `.fossa.yml` file in the project root.
 
 ```yaml
 # filename: .fossa.yml
--- a/docs/references/strategies/languages/gradle/gradle.md
+++ b/docs/references/strategies/languages/gradle/gradle.md
@@ -31,6 +31,7 @@ Gradle users generally specify their builds using a `build.gradle`/`settings.gra
   - [Manually specifying Gradle dependencies](#manually-specifying-gradle-dependencies)
   - [Configurations For Development and Testing](#configurations-for-development-and-testing)
   - [Android Gradle Configurations For Development and Testing](#android-gradle-configurations-for-development-and-testing)
+  - [Only analyzing specific sub project](#only-analyzing-specific-sub-project)
   - [(Experimental) Only Selecting Set of Configurations For Analysis](#experimental-only-selecting-set-of-configurations-for-analysis)
 
 ## Concepts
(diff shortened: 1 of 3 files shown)


meghfossa:
@csasarak this is duplicate of https://github.com/fossas/fossa-cli/pull/1045 (which i had to close due to some rebasing issues)

docs/references/strategies/languages/gradle/gradle.md
```diff
@@ -282,6 +283,58 @@ We classify following configurations, and dependencies originating from it as a
...
+```yaml
+# filename: .fossa.yml
+#
+# analyze only gradle@./::utilities
+version: 3
+targets:
+  only:
+    - type: gradle
+      path: ./
+      target: ':utilities'
```
jssblck, line 313: Two colons in comment and in `list-targets` output, but one in the `target` setting- is this intentional? If so please explicitly call it out in the comment. Same with the other example below.

meghfossa: yes, it is intentional; added: https://github.com/fossas/fossa-cli/pull/1059/commits/405cc0cee0dd5940663c9baa228dd86db1ab783e

docs/references/strategies/languages/gradle/gradle.md
```diff
@@ -282,6 +283,68 @@ We classify following configurations, and dependencies originating from it as a
...
+[ INFO] Found target: gradle@./::app
+[ INFO] Found target: gradle@./::list
+[ INFO] Found target: gradle@./::utilities
+```
+
+Note that, targets are denoted in following format `type@path:target`. For 
+example `gradle@./::utilities`:
+
+```
+gradle  @           ./      :         :utilities
```
scruffystuffs, line 306: We should explicitly call out the extra `:` here as coming from gradle.  I think with that example and the comment, we are fully communicating what we need to.  Something like `gradle attaches leading colons to submodules, so the utilities submodule here is referenced by ":utilities"`.  Probablybest placed above this visual breakdown.

scruffystuffs (approved):
One change needed for clarification of the gradle example

PR #1066, opened 2022-10-04
Package Linux binaries as both tar.gz and zip

rolodato:
# Overview

This PR changes our GitHub Actions workflow to release Linux binaries packaged as both tar.gz and zip. This improves compatibility for installing fossa-cli under Linux distributions that do not come with the `unzip` command installed by default.

**Note**: This PR does not yet change `install-latest.sh` to use tar.gz files, as this would immediately break when trying to download the current latest release. An additional change should be merged after this PR is released to actually start using the tar.gz files when installing:

```diff
diff --git a/install-latest.sh b/install-latest.sh
index 1ec9c3de..10dcdbbf 100755
--- a/install-latest.sh
+++ b/install-latest.sh
@@ -102,6 +102,9 @@ adjust_format() {
 }
 adjust_os() {
   # adjust archive name based on OS
+  case $(uname_os) in
+    linux) FORMAT=tar.gz ;;
+  esac
   true
 }
 adjust_arch() {
```

## Acceptance criteria

After this PR and the above diff are merged, using `install-latest.sh` should work under Linux distributions that do not come with `unzip` installed.

## Testing plan

This depends on GitHub Actions so I haven't been able to test it directly yet.

## Risks

It's technically possible that some customers might be running the install script from a Linux environment that has `unzip` but not `tar` installed, but this seems extremely unlikely. CLI 1.x Linux releases were originally released as tar.gz.

We might want to add a fallback in the install script to try the zip release if the tar.gz release fails for some reason, but IMO this is unnecessary complexity for an extremely unlikely scenario.

## References

- [ANE-79](https://fossa.atlassian.net/browse/ANE-79)

## Checklist

- [ ] ~I added tests for this PR's change (or explained in the PR description why tests don't make sense).~
- [ ] ~If this PR introduced a user-visible change, I added documentation into `docs/`.~

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -218,7 +218,10 @@ jobs:
 
         cat *.sha256 | sha256sum --check --status
 
-    - name: Upload release artifacts
+    # Uploads the generated archives (tar.gz/zip) as build artifacts to allow
+    # verifying them without needing to do an actual release. This step does not
+    # need to run for tagged release versions.
+    - name: Upload release archives
       if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
       uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -171,6 +171,12 @@ jobs:
         [ "$(Linux-binaries/fossa --version)" = "fossa-cli version ${{ steps.get-version.outputs.VERSION }} (revision ${GITHUB_SHA:0:12} compiled with ghc-9.0)" ] && echo "CLI version OK"
 
     # This uses names compatible with our install script.
+    #
+    # Originally, CLI >=2.x Linux releases were only packaged as zip files, but
+    # we added tar.gz to improve compatibility. Our install script depends on
+    # the unzip command, which is not installed in most Linux distributions by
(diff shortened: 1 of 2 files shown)


rolodato:
Confirmed that the `release-archives` artifact is being created: https://github.com/fossas/fossa-cli/suites/8628755062/artifacts/387533340

Seems to be working as expected in an unmodified Debian installation which doesn't come bundled with `unzip`:

```
rolodato@MSI:~$ ls
fossa_9fa715256b13c18866be8199a3aaff474561ac15_linux_amd64.tar.gz
rolodato@MSI:~$ unzip
-bash: unzip: command not found
rolodato@MSI:~$ tar xvf fossa_9fa715256b13c18866be8199a3aaff474561ac15_linux_amd64.tar.gz
fossa
rolodato@MSI:~$ ./fossa --version
fossa-cli branch feat/tar.gz-linux-releases (revision 9fa715256b13 compiled with ghc-9.0)
```

We can keep this PR as it is now, or I can remove commits https://github.com/fossas/fossa-cli/pull/1066/commits/844284ad9048385423e843c5bcdeb5de0918511b and https://github.com/fossas/fossa-cli/pull/1066/commits/2f8d91a4828afe0abebaf75c7dac4f648bc082d1 to keep the behavior similar to the original one.

meghfossa (approved):

PR #1069, opened 2022-10-07
[No Ticket] Adds install script test for specific version

meghfossa:
# Overview

This PR adds an installation script test for installing a specific CLI version


spatten (approved):
Looks good to me!

PR #1074, opened 2022-10-13
[ANE-551] Scala multi-project support

meghfossa:
# Overview

This PR,

- Adds support for scala multi-project build with `sbt-dependency-graph` plugin

## Acceptance criteria

- Multi-project sbt project analysis includes deep dependencies

## Testing plan

1. Prereq
```bash
git checkout feat/sbt-multi-target-support
make install-dev
```

2. Test against a multi-build project (you should see deep dependencies, and edges)
```bash
# get a sample project
git clone https://github.com/fossas/example-projects.git

# add plugin
cd example-projects/scala/multi/
touch project/plugin.sbt
echo 'addDependencyTreePlugin' > project/plugin.sbt

# analyze
fossa-dev analyze --only-target scala -o | jq
```

2. Test against single build project (you should see deep dependencies and edges)
```bash
# get a sample project
git clone https://github.com/fossas/example-projects.git
cd example-projects/scala/single/

# analyze
fossa-dev analyze --only-target scala -o | jq
```

## Error View

1. When `sbt-dependency-graph` plugin is not accessible for multi-project-build

CleanShot 2022-10-12 at 19 26 37@2x

## Notes

I chose the `dependencyBrowseTreeHTML` task from the `sbt-dependency-graph` plugin as it generates JSON, which is much easier to parse. This `dependencyBrowseTreeHTML` task is _not explicitly documented in the plugin's GitHub docs_, but it is present from `v0.10.0-RC1` onwards of the plugin (for any `sbt 1.4.0+` - when `addDependencyTreePlugin` is used, it includes `dependencyBrowseTreeHTML` task)

## Risks

N/A

## References

https://fossa.atlassian.net/browse/ANE-551

## Checklist

### DIFF
--- a/integration-test/Analysis/ScalaSpec.hs
+++ b/integration-test/Analysis/ScalaSpec.hs
@@ -34,4 +34,4 @@ scalaExampleProject =
 
 spec :: Spec
 spec = do
-  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 2 1 1 1 Complete)
+  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 3 2 1 1 Complete)
--- a/integration-test/Analysis/ScalaSpec.hs
+++ b/integration-test/Analysis/ScalaSpec.hs
@@ -34,4 +34,4 @@ scalaExampleProject =
 
 spec :: Spec
 spec = do
-  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 3 2 1 1 Complete)
+  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 2 1 1 1 Complete)
--- a/src/Strategy/Scala/Common.hs
+++ b/src/Strategy/Scala/Common.hs
@@ -1,11 +1,14 @@
 module Strategy.Scala.Common (
   removeLogPrefixes,
+  withoutStdLibs,
   SbtArtifact (..),
 ) where
 
 import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import Data.Text qualified as Text
+import DepTypes (DepType (MavenType), Dependency (dependencyName, dependencyType))
+import Graphing (Graphing, filter)
 
 data SbtArtifact = SbtArtifact
   { groupId :: Text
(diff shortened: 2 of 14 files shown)


integration-test/Analysis/ScalaSpec.hs
```diff
@@ -34,4 +34,4 @@ scalaExampleProject =
 
 spec :: Spec
 spec = do
-  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 3 2 1 1 Complete)
+  testSuiteDepResultSummary scalaExampleProject ScalaProjectType (DependencyResultsSummary 2 1 1 1 Complete)
```
meghfossa, line 37: this is because, we ignore stdlib explicitly now

src/Strategy/Scala/Common.hs
```diff
@@ -31,3 +34,19 @@ removeLogPrefixes content = Text.unlines . map removePrefix $ Text.lines content
 
     maybeRemovePrefix :: Text -> Text -> Text
     maybeRemovePrefix prefix candidate = fromMaybe candidate $ Text.stripPrefix (prefix <> " ") candidate
+
+-- | Removes scala standard library from sbt
+-- Refer: https://github.com/sbt/sbt/blob/master/main/src/main/scala/sbt/internal/graph/GraphTransformations.scala#L39
+--
+-- Note that, as general case FOSSA CLI prefers not to report standard library,
+-- to avoid dependency noise, issue noise arising from inclusion of standard library.
+withoutStdLibs :: Graphing Dependency -> Graphing Dependency
```
meghfossa, line 43: From latest retro, it seems we want to include stdlib for scala. 

todo:
- [ ] include stdlib in final analysis

src/Strategy/Scala/SbtDependencyTreeJson.hs
```diff
@@ -0,0 +1,148 @@
...
+    sc = Lexer.space (void $ some $ char ' ' <|> char '\t') empty empty
+
+    lexeme :: ParserT a -> ParserT a
+    lexeme = Lexer.lexeme sc
+
+    symbol :: Text -> ParserT Text
+    symbol = Lexer.symbol sc
+
+    parseEviction :: ParserT (Text)
+    parseEviction = lexeme ("(evicted by" <|> "(evicted by") *> parseValidProjectIdentifier <* symbol ")"
```
spatten, line 111: Maybe I just don't understand what's going on here, but having `"(evicted by"` twice in a row within parens looks wrong to me. Is that correct?

spatten: Also, from the specs it looks like you expect a line like "a:b.c:1.0 (evicted by 2.0)" to get parsed as having a version of 2.0. Can you add a comment explaining why that is?

spatten (approved):
I'm not sure I can give the parsing part a full review, but other than the few comments I made, everything looks good to me

PR #1084, opened 2022-10-26
Generate conda locators with channel/platform where possible.

csasarak:
# Overview

Basis PR [1768](https://github.com/fossas/basis/pull/1768) adds support to the fetchers API service to accept channel and platform in order to return more accurate results. This PR adds support to the CLI to produce those new locators, where possible. The format is: `conda+{org/}?'{channel}':{platform}:{name}${version}`

## Risks

I've put the channel value in quotes in the package spec because if we were to support an optional `org/` at the beginning of the locator it makes it impossible to know whether the package spec has no org and a channel with a slash or an org with a channel. For example `org/conda-forge` could be interpreted as an `org = 'org'` and `channel = 'conda-forge'` or as having no org and `channel = 'org/conda-forge'`. Slashes must be allowed in channels because the `conda` tool reports channels such as `pkgs/main`. 

This format with quotes satisfies the [regex](https://github.com/fossas/FOSSA/blob//shared/Locator.ts#L63) in fossa core, but is there any reason not to do it this way?

## References

- [ANE-571](https://fossa.atlassian.net/browse/ANE-571)

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+- Conda: Output more detailed locators for channel/platform specific analysis. ([#1084])(https://github.com/fossas/fossa-cli/pull/1084)
+
 ## v3.5.1
 
 - Contributor counting: update the contributor count range from 90 days to 365 days. ([#1083](https://github.com/fossas/fossa-cli/pull/1083))
--- a/docs/references/strategies/languages/python/conda.md
+++ b/docs/references/strategies/languages/python/conda.md
@@ -9,7 +9,11 @@ Find the first file named `environment.yml`
 ## Analysis: conda list
 
 We run `conda list --json` in order to get a complete list of all installed conda packages in the conda environment.
+This tactic tries to determine which channel/platform a package was installed for to get the most accurate results.
+If it can't determine a channel/environment it will use just the name to get best-effort results for that package.
 
 ## Analysis: environment.yml
 
 If the `conda list --json` operation does not succeed, we fall back to parsing the `environment.yml`, and getting the list of dependencies from the `dependencies` section of this file.
+`environment.yml` does not include data about which channel a package was installed from.
+Instead, FOSSA will try to make a best effort to retrieve package data based on the name.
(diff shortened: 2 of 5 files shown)


docs/references/strategies/languages/python/conda.md
```diff
@@ -9,7 +9,11 @@ Find the first file named `environment.yml`
 ## Analysis: conda list
 
 We run `conda list --json` in order to get a complete list of all installed conda packages in the conda environment.
+This tactic tries to determine which channel/platform a package was installed for to get the most accurate results.
```
csasarak, line 12: I made these additions to explain how this works, but I was on the fence on whether to include it. The behavior of fossa cli isn't very different from a user perspective, they'll mainly see changes on the back-end. On the other hand, this might be where someone ends up first looking for an explanation if they do see those changes.

zlav: I'm a fan of including this, I think it makes this all clear and shows people that we're taking into account the location of each package. If users question how we know, we can point them here.

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## Unreleased
```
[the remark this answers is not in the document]
zlav: Would this block someone else from releasing? Or are you planning to wait to merge this until 584 is also ready to merge?

src/Strategy/Conda/CondaList.hs
```diff
@@ -28,11 +29,14 @@ condaListCmd =
 buildGraph :: [CondaListDep] -> Graphing Dependency
 buildGraph deps = Graphing.fromList (map toDependency deps)
   where
+    makeV2Dep :: Text -> Text -> Text -> Text
```
zlav, line 32: Where does the name `V2Dep` come from? I don't love this, especially since we plan to support the other dependency format indefinitely. I think something like `depWithlocation`would be more clear.

zlav (approved):
Looks good! Nice simple PR

Changelog.md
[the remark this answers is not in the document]
csasarak: The plan was to wait, but I can revisit that if you'd like. I made this PR before our discussions re: better supporting conda environments.

src/Strategy/Conda/CondaList.hs
csasarak: I called it that because I felt it would make it easy to disambiguate fields if we were to add another format in the future. I can see how this would make it seem like we planned to deprecate the old one. I will rethink this, `depWithLocation` is probably more descriptive.

csasarak:
Now that I think about it, most of this PR will be superseded by ANE-584 now. I'm going to keep this around until that merges and then probably close it.

PR #1091, opened 2022-11-05
Use ISO timestamp for revision fallback

jssblck:
# Overview

We generate the revision for a project automatically, eventually falling back to the current timestamp.
Currently that timestamp is just a unix string. @rolodato suggested that this should be an ISO timestamp instead to make it more clear to users.

I thought this was super simple and a good idea so I just threw it in, but if anyone has a major issue with it happy to just close the PR.

This change isn't tested because I think we don't really have tests for this and adding it into tests would be more time investment than I think this change warrants. If you disagree though, let me know!

## Acceptance criteria

The final fallback for `--revision` is an ISO8601 timestamp instead of a unix epoch timestamp.

## Testing plan

I created a project with no `.git`:
```shell
; pwd
/Users/jessica/projects/scratch/tiny
; ls -al
total 24
drwxr-xr-x   7 jessica  staff   224 Nov  4 16:56 .
drwxr-xr-x  48 jessica  staff  1536 Oct 31 15:47 ..
-rw-r--r--   1 jessica  staff     8 Mar 16  2022 .gitignore
-rw-r--r--   1 jessica  staff   378 Mar 16  2022 Cargo.lock
-rw-r--r--   1 jessica  staff   195 Mar 16  2022 Cargo.toml
drwxr-xr-x   3 jessica  staff    96 Mar 16  2022 src
drwxr-xr-x@  5 jessica  staff   160 Aug 22 13:29 target
```

I then ran the scan:
```shell
; cabal run fossa -- analyze ~/projects/scratch/tiny
# ...
[ INFO] ------------
[ INFO]
[ INFO] Using project name: `tiny`
[ INFO] Using revision: `2022-11-04T23:57:26.932386Z`
[ INFO] Using branch: `No branch (detached HEAD)`
# ...
```

And checked the UI:
Screenshot 2022-11-04 at 5 02 58 PM

## Risks

Does not seem risky.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+
+- Uses an ISO timestamp for the revision if no better revision can be inferred [#1091](https://github.com/fossas/fossa-cli/pull/1091).
+
 ## v3.6.0
 
 - Promote C/C++ features to general availability [#1087](https://github.com/fossas/fossa-cli/pull/1087).
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,21 +2,21 @@
 
 ## Unreleased
 
-- Uses an ISO timestamp for the revision if no better revision can be inferred [#1091](https://github.com/fossas/fossa-cli/pull/1091).
+- Uses an ISO timestamp for the revision if no better revision can be inferred ([#1091](https://github.com/fossas/fossa-cli/pull/1091)).
 
 ## v3.6.0
 
-- Promote C/C++ features to general availability [#1087](https://github.com/fossas/fossa-cli/pull/1087).
+- Promote C/C++ features to general availability ([#1087](https://github.com/fossas/fossa-cli/pull/1087)).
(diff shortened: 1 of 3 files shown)


zlav (approved):
Agreed that this is much clearer! Nice quick fix

rolodato:
Thanks for this ✨ 

One nitpick - removing the milliseconds might look a lot cleaner, but either way this is a lot better than the current behavior.

jssblck:
@rolodato Milliseconds removed, now it looks like this:

```
[ INFO] Using project name: `tiny`
[ INFO] Using revision: `2022-11-05T00:36:46`
```

Screenshot 2022-11-04 at 5 37 43 PM

Agreed, this looks even nicer! Thanks!

rolodato (approved):
The screenshot in this message caught me off guard as it doesn't have the UTC `Z` indicator: https://github.com/fossas/fossa-cli/pull/1091#issuecomment-1304358352

But I can see that there was a change added afterwards to make this compliant with the spec. Ship it! ✨

PR #1092, opened 2022-11-07
Don't automatically promote deep deps to being direct

zlav:
# Overview

This PR is similar to this #8578 on FOSSA core. Currently, the following dependency graph is impossible. `itsdangerous` would be automatically promoted to being a direct dependency.
```
      "Imports": ["pip+pipenv$2018.11.26"],
      "Dependencies": [
        { "locator": "pip+itsdangerous$1.1.0" },
        { "locator": "pip+enum34$1.1.6" },
        { "locator": "pip+pipenv$2018.11.26", "imports": ["pip+enum34$1.1.6"] }
```

This can happen when a direct dependency is pruned from the graph for being out of scope (i.e. a test dep) and its immediate transitive dependencies are promoted to direct dependencies themselves. This PR removes that behavior and keeps the dependencies as transitive dependencies.

## Acceptance criteria

- Transitive dependencies are no longer promoted to direct if their parent is pruned.
- All existing strategies behave how we previously expected.

## Testing plan

1. Manually test this using an example pom file that demonstrates the behavior. Running `fossa analyze` in a directory with only the pom file below will test this logic. We should see that `httpclient` is **not** listed in the `Imports` array.
```
<?xml version="1.0" encoding="UTF-8"?>
<project
	xmlns="http://maven.apache.org/POM/4.0.0"
	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<groupId>com.example</groupId>
	<artifactId>example</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<dependencies>
		<dependency>
			<groupId>com.amazonaws</groupId>
			<artifactId>aws-java-sdk-kms</artifactId>
			<version>1.11.415</version>
		</dependency>
		<dependency>
			<groupId>com.jayway.restassured</groupId>
			<artifactId>rest-assured</artifactId>
			<version>2.9.0</version>
			<scope>test</scope>
		</dependency>
	</dependencies>
</project>
```
2. Ensuring all of the integration tests pass, and the ones that do fail are failing for a good reason.

## Risks

We could break existing strategies that rely on this behavior for unknown and untested reasons. I think this is a fair trade-off if they are unknown and untested. I also don't believe that any analyzer's results should be reliant on generic pre-upload pruning to deliver correct results, this logic should live with that analyzer's graph building logic.

## References

[ANE-659](https://fossa.atlassian.net/browse/ANE-659)

## Checklist

- [ ] If this PR introduced a user-visible change, I added documentation into `docs/`. - I need to add docs for Maven specifically

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.6.2
+
+- Don't promote transitive dependencies  [#1092](https://github.com/fossas/fossa-cli/pull/1092).
+
 ## v3.6.1
 
 - Container Scanning: Fixes a bug where image source parser ignored '-' in host. Also fixes an issue regarding to redirect headers when communicating with registry. [#1089](https://github.com/fossas/fossa-cli/pull/1089) 
--- a/docs/references/strategies/languages/maven/maven.md
+++ b/docs/references/strategies/languages/maven/maven.md
@@ -14,6 +14,17 @@ Maven analysis attempts these analysis methods in order:
 3. Run the maven plugin command version 3.3.0.
 4. Scan `pom.xml` files located in the file tree.
 
+## FAQ
+### One of my transitive dependencies does not have path information
+The mavenplugin and treecmd tactic can result in transitive dependencies which do not display paths to parents. This example graph shows how that can happen:
+```
++- com.amazonaws:aws-java-sdk-kms:1.11.415:compile
+|  +- com.amazonaws:aws-java-sdk-core:1.11.415:compile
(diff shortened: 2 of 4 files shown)


spatten (approved):
This looks good to me

meghfossa (commented):
Ran through following regression test from:
- https://github.com/fossas/team-analysis/issues/811
- https://github.com/fossas/team-analysis/issues/809

They both passed 🚀

PR #1096, opened 2022-11-10
Fix `--detect-vendored` on Windows

jssblck:
# Overview

Fixes VSI scans for windows

## Acceptance criteria

VSI scans work when running on windows.
We don't currently have a great story for integration testing this - I plan to test this with my personal windows PC and list the results before and after here to prove this works.

Longer term integration testing is handled more cleanly with the `foundation-libs` powered VSI client, but this is intended to be a quicker fix than integrating that. Since integration testing is handled more cleanly on that repo I plan to focus my efforts on integrating that instead of implementing integration testing in this repo.

## Testing plan

First, I validated that the results still work on non-windows:
```
; cabal run fossa -- analyze --detect-vendored -p pre-windows-test ~/projects/cpp-vsi-demo
```
Screenshot 2022-11-09 at 4 52 25 PM
(Note: the `libssh2` is because of my local working state; that's not in the standard `cpp-vsi-demo`)

I tested the artifact built for this PR in Windows, which led to a successful scan:
Screenshot 2022-11-10 at 3 30 40 PM
([Scan link in org jessoteric](https://app.fossa.com/projects/custom%2B24357%2Ffrom-windows-vsi-1096/refs/branch/master/1668122842/browse/dependencies))

I validated with the currently published version of the CLI (v3.6.3) that the scan failed on my Windows machine.

## Risks

Not really risky.

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## v3.6.4
 
-- Fixes `--detect-vendored` on Windows. ([#1096](https://github.com/fossas/fossa-cli/pull/1096))
+- C/C++: Fixes `--detect-vendored` on Windows. ([#1096](https://github.com/fossas/fossa-cli/pull/1096))
 - Uses an ISO timestamp for the revision if no better revision can be inferred ([#1091](https://github.com/fossas/fossa-cli/pull/1091)).
 
 ## v3.6.3
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## v3.6.4
 
+- Fixes `--detect-vendored` on Windows. ([#1096](https://github.com/fossas/fossa-cli/pull/1096))
 - Uses an ISO timestamp for the revision if no better revision can be inferred ([#1091](https://github.com/fossas/fossa-cli/pull/1091)).
 
 ## v3.6.3
--- a/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
+++ b/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
@@ -88,8 +88,9 @@ import Data.ByteString.Lazy (ByteString)
 import Data.List.NonEmpty (NonEmpty)
 import Data.List.NonEmpty qualified as NE
 import Data.Map (Map)
+import Data.Map qualified as Map
 import Data.Maybe (catMaybes, fromMaybe)
-import Data.String.Conversion (decodeUtf8, toStrict, toText)
+import Data.String.Conversion (decodeUtf8, toStrict, toString, toText)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Word (Word64, Word8)
@@ -151,6 +152,7 @@ import Network.HTTP.Req.Extra (httpConfigRetryTimeouts)
 import Network.HTTP.Types qualified as HTTP
 import Parse.XML (FromXML (..), child, parseXML, xmlErrorPretty)
 import Path (File, Path, Rel)
+import Path qualified
 import Srclib.Types (
   LicenseSourceUnit,
   Locator (..),
@@ -158,6 +160,7 @@ import Srclib.Types (
   parseLocator,
   renderLocator,
  )
+import System.FilePath (pathSeparator, splitDirectories)
 import Text.URI qualified as URI
 
 -- | Represents error emitted via FOSSA instance.
@@ -1136,7 +1139,7 @@ vsiCreateScan apiOpts ProjectRevision{..} = fossaReq $ do
   body <- responseBody <$> req POST (vsiCreateScanEndpoint baseUrl) (ReqBodyJsonCompat reqBody) jsonResponse baseOpts
   pure $ unVSICreateScanResponseBody body
 
-newtype VSIAddFilesToScanRequestBody = VSIAddFilesToScanRequestBody {vsiAddFilesToScanRequestBodyFiles :: Map (Path Rel File) Fingerprint.Combined}
+newtype VSIAddFilesToScanRequestBody = VSIAddFilesToScanRequestBody {vsiAddFilesToScanRequestBodyFiles :: Map FilePath Fingerprint.Combined}
 
 instance ToJSON VSIAddFilesToScanRequestBody where
   toJSON VSIAddFilesToScanRequestBody{..} = object ["ScanData" .= vsiAddFilesToScanRequestBodyFiles]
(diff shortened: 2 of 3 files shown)


Changelog.md
```diff
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## v3.6.4
 
+- Fixes `--detect-vendored` on Windows. ([#1096](https://github.com/fossas/fossa-cli/pull/1096))
```
meghfossa, line 5: [minor] let's add c/c++ prefix in front i.e. C/C++: Fixes `--detect-vendored` on Windows?

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -151,13 +152,15 @@ import Network.HTTP.Req.Extra (httpConfigRetryTimeouts)
 import Network.HTTP.Types qualified as HTTP
 import Parse.XML (FromXML (..), child, parseXML, xmlErrorPretty)
 import Path (File, Path, Rel)
+import Path qualified
```
meghfossa, line 155: [minor] we should do the explicit import for toFilePath

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1148,10 +1151,23 @@ vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts ->
...
 
   pure ()
+  where
+    normalizeFiles :: [(Path Rel File, Fingerprint.Combined)] -> [(FilePath, Fingerprint.Combined)]
+    normalizeFiles = fmap normalizeFile
+
+    normalizeFile :: (Path Rel File, Fingerprint.Combined) -> (FilePath, Fingerprint.Combined)
+    normalizeFile (path, fp) = (normalizePath path, fp)
+
+    normalizePath :: Path Rel File -> FilePath
```
meghfossa, line 1168: [optional] unit test

meghfossa (approved):
some options/minors 🚀

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
jssblck: I don't think this is really feasible for us to unit test since it is inside the API effect (and I don't think we should have it outside). Plus it's extremely simple. I'd rather solve testing by setting up an integration test with VSI scans E2E.

Going to leave it as is since this is marked optional, but if you disagree happy to talk more!

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
jssblck: As in, instead of `Path.toFilePath` you think we should `import Path (File, Path, Rel, toFilePath)`?

jssblck: Went ahead and made this change but if you meant something else let me know

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1148,10 +1151,23 @@ vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts ->
...
+  -- This is in the API layer because it's an implementation detail of the API.
+  -- The API expects all file paths to be normalized to forward slashes.
+  let normalized = Map.fromList . normalizeFiles $ Map.toList files
+  let body = VSIAddFilesToScanRequestBody normalized
   _ <- req POST (vsiAddFilesToScanEndpoint baseUrl scanID) (ReqBodyJsonCompat body) ignoreResponse baseOpts
 
   pure ()
+  where
+    normalizeFiles :: [(Path Rel File, Fingerprint.Combined)] -> [(FilePath, Fingerprint.Combined)]
+    normalizeFiles = fmap normalizeFile
```
spatten, line 1162: nit (maybe?): should this just be a `map` seeing as we're mapping over a list?

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1148,10 +1151,23 @@ vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts ->
 vsiAddFilesToScan apiOpts scanID files = fossaReq $ do
   (baseUrl, baseOpts) <- useApiOpts apiOpts
 
-  let body = VSIAddFilesToScanRequestBody files
+  -- This is in the API layer because it's an implementation detail of the API.
+  -- The API expects all file paths to be normalized to forward slashes.
+  let normalized = Map.fromList . normalizeFiles $ Map.toList files
+  let body = VSIAddFilesToScanRequestBody normalized
```
spatten, line 1156: [Not blocking] It would be good to have some tests for `vsiAddFilesToScan`

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
meghfossa: yep yea - feel free to ignore anything with [optional] in my pr comments

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1148,10 +1151,23 @@ vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts ->
...
+  where
+    normalizeFiles :: [(Path Rel File, Fingerprint.Combined)] -> [(FilePath, Fingerprint.Combined)]
+    normalizeFiles = fmap normalizeFile
+
+    normalizeFile :: (Path Rel File, Fingerprint.Combined) -> (FilePath, Fingerprint.Combined)
+    normalizeFile (path, fp) = (normalizePath path, fp)
+
+    normalizePath :: Path Rel File -> FilePath
+    normalizePath input | pathSeparator == '/' = Path.toFilePath input
+    normalizePath input = toString . Text.intercalate "/" $ toText <$> splitDirectories (Path.toFilePath input)
```
spatten: According to this, you can still have `\` in the list entries after running `splitDirectories`: 
https://hackage.haskell.org/package/filepath-1.4.100.0/docs/System-FilePath-Posix.html#v:splitDirectories

```
Windows: splitDirectories "C:\\test\\file" == ["C:\\", "test", "file"]
Windows: splitDirectories "C:\\test\\\\\\file" == ["C:\\", "test", "file"]
```

spatten (approved):
Two minor nits, but I think there might be a problem with still having back-slashes in your path after running `splitDirectories`, unless we're certain that we're always using relative paths and this isn't a problem for relative paths

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
spatten: I'm not sure if this is a problem that we can encounter, though. It looks like it doesn't happen for relative paths and we might only be encountering relative paths here?

jssblck: > unless we're certain that we're always using relative paths and this isn't a problem for relative paths

We are; the parent function is this:
```haskell
vsiAddFilesToScan :: (Has (Lift IO) sig m, Has Diagnostics sig m) => ApiOpts -> VSI.ScanID -> Map (Path Rel File) Fingerprint.Combined -> m ()
```

So we are enforcing that we get relative files via types.

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
jssblck: Longer term integration testing is handled more cleanly with the `foundation-libs` powered VSI client, but this is intended to be a quicker fix than integrating that. 

Since integration testing is handled more cleanly on that repo I plan to focus my efforts on integrating that instead of implementing integration testing in this repo.

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
jssblck: Oh, yeah, good catch lol

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
spatten: Right, you mentioned that in the PR description. Sounds reasonable to me!

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
spatten: Perfect. Thanks!

PR #1098, opened 2022-11-14
Expands instances of Simple and Finally

periodic:
# Overview

We are moving Hubble over to use Fused Effects.  We needed to update `SimpleC` and `FinallyC` a little to accommodate this.  Those are the two general carriers we plan to utilize.

First, we need a few more instances.  Servant, our HTTP server framework, requires instances of `MonadError`.  Blammo/aeson-logger, our logging framework, needs instances of `MonadUnliftIO`.  These are simple enough to derive as they can both percolate up from a base monad.

Second, I added some small helpers to `Simple` to make writing `sendX` functions easier.  `sendSimple2` takes a constructor of two args. `sendSimple3` takes a constructor of three arguments.  They convert the constructor into a function within the context of the algebra instance monad.

I bumped the version of `fused-effects` because `MonadUnliftIO` instances were added to the stock effects in 1.1.2.0

## Acceptance criteria

`Simple` and `Finally` derive the needed instances without impacting the CLI in any negative way.

## Testing plan

These changes _should_ be simply additive.  I already have changes in Hubble that depend on these by pointing cabal at the specific commit.  All other spectrometer tests should continue to pass.

The instances themselves have very small and mechanical implementations.

## Risks

This is low risk as it should not change any existing code-paths.

## References

- [PLAT-438](https://fossa.atlassian.net/browse/PLAT-438): Make updates to the CLI carriers 

## Checklist

### DIFF
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -85,7 +85,7 @@ common deps
     , file-embed                   ^>=0.0.11
     , filepath                     ^>=1.4.2.1
     , filepattern                  ^>=0.1.2
-    , fused-effects                ^>=1.1.0.0
+    , fused-effects                ^>=1.1.2.0
     , fused-effects-exceptions     ^>=1.1.0.0
     , git-config                   ^>=0.1.2
     , githash                      ^>=0.1.4.0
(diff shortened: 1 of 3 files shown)


periodic:
Okay, now I'm blocked on this. :D

jssblck (approved):

PR #1099, opened 2022-11-14
Ane 584 new dynamic strategy

csasarak:
# Overview

* Uses a new command in the dynamic strategy for conda. This command will simulate creating an environment from an `environment.yml` file rather than reading package lists from the currently active environment.
* Produces enhanced locators with the channel and platform in the dynamic strategy.

## Acceptance criteria

When users scan a conda project, it should return a list of dependencies that match those in `environment.yml`. Before it would list packages from the currently enabled conda environment and not read `environment.yml` at all.

## Testing plan

1. I tested the v2 locators against our staging server, which has the new server support in `fetchers-api` that we need.
2. I tested that current locators with limited information are still produced and work. These locators are needed for static analysis as well as users of older versions of `fossa-cli`.

## Risks

The new conda command can take a while to run because it does more work/might need to download package information.

## References

[ANE-584](https://fossa.atlassian.net/browse/ANE-584)
[ANE-620](https://fossa.atlassian.net/browse/ANE-620)

## Checklist


src/Strategy/Conda/CondaEnvCreate.hs
```diff
@@ -0,0 +1,122 @@
...
+  ) =>
+  Path Abs Dir ->
+  Path Abs File ->
+  m (Graphing Dependency)
+analyze dir file = do
+  buildGraph <$> (parseEnvCreateDeps =<< execJson @CondaEnvCreateOut dir (condaEnvCmd file))
+
+parseEnvCreateDeps :: (Monad m, Has Diagnostics sig m) => CondaEnvCreateOut -> m [CondaEnvDep]
+parseEnvCreateDeps CondaEnvCreateOut{dependencies = deps} = do
+  let (lefts, rights) = partitionEithers . map (parse parseCondaEnvDep "") $ deps
```
csasarak, line 79: In the event it can't parse a package string this code will emit a warning but move on rather than failing fatally. 

Do we think this is ok? I like the idea of finding whatever information we can and doing the right thing with those, but I wonder if warnings would be visible enough in CI environments.

src/Strategy/Conda/CondaEnvCreate.hs
```diff
@@ -0,0 +1,122 @@
...
+instance FromJSON CondaEnvCreateOut where
+  parseJSON = withObject "CondaEnvOutput" $ \obj ->
+    CondaEnvCreateOut
+      <$> obj
+        .: "dependencies"
+
+type Parser = Parsec Void Text
+
+-- Parses a dep specification from a string like the following:
+--   "conda-forge/osx-64::bzip2==1.0.8=h0d85af4_4"
```
zlav, line 103: What does this portion of the dependency mean "h0d85af4_4"?

src/Strategy/Conda/CondaEnvCreate.hs
zlav: I think emitting a warning instead of failing is preferable here. I think you're right that the warnings won't be readily visible in CI environments, but I'm not sure what other option we have. I think this is an issue with how visible our warnings are in general as opposed to this specific strategy.

I think it would be ideal if we could upload this as part of the CLI telemetry so that we can get alerted if this happens frequently, but I don't think that should be in scope for this PR.

src/Strategy/Conda/CondaEnvCreate.hs
```diff
@@ -0,0 +1,122 @@
...
+
+instance FromJSON CondaEnvCreateOut where
+  parseJSON = withObject "CondaEnvOutput" $ \obj ->
+    CondaEnvCreateOut
+      <$> obj
+        .: "dependencies"
+
+type Parser = Parsec Void Text
+
+-- Parses a dep specification from a string like the following:
```
zlav, line 102: You should include a link to where this specification is defined if you have one.

zlav (approved):
Looks good to me! Left a few small comments, but overall really clean!

src/Strategy/Conda/CondaEnvCreate.hs
csasarak: ok, sounds good. I just wasn't sure if we had made some decision about that and now I know. Thanks!

src/Strategy/Conda/CondaEnvCreate.hs
csasarak: That is the build string, we don't/haven't done anything with it. I can add some explanation though.

src/Strategy/Conda/CondaEnvCreate.hs
csasarak: I don't, it was found just by reading the output.

src/Strategy/Conda/CondaEnvCreate.hs
csasarak: I've added more description of the format.

PR #1100, opened 2022-11-15
Adds auto bot response on new issue

meghfossa:
# Overview

Reference: https://teamfossa.slack.com/archives/C039KE5ERNE/p1653092147543659

## Checklist


meghfossa:
I think this requires a merge with master before it can be tested, if this fails - I will create another PR to remove this WF

spatten (approved):
Looks good to me!

PR #1105, opened 2022-11-16
doc: Document internals of Finally

elldritch:
This PR adds notes I made while debugging through the internals of `Finally`. Hopefully they're useful. Reviewers, it would be great if you could double-check them for correctness.


jssblck (approved):
> Reviewers, it would be great if you could double-check them for correctness.

I'm not super familiar with the internals of the effect system, but I believe these are correct  after reading through it.

PR #1107, opened 2022-11-16
Add documentation about how manual deps are cached

jssblck:
# Overview

Include documentation about how manual dependencies are cached and on which config keys.

## Acceptance criteria

This is a docs only change. I believe these docs are accurate but would love a double check.

## Risks

If these docs are inaccurate users may be confused by the actual behavior we exhibit. Overall low risk.

## References

In response to ZD 4559.

## Checklist

### DIFF
--- a/docs/features/manual-dependencies.md
+++ b/docs/features/manual-dependencies.md
@@ -156,15 +156,15 @@ remote-dependencies:
 
 ## Performance
 
-The FOSSA service caches the results of dependency analysis by different keys depending on the type of remote dependency specified (explained below).
-Due to this caching setup, it is normal for the first analysis to take some time, especially for larger projects, but future analysis of dependencies with the cache keys unchanged should be fast.
+The FOSSA service caches the results of dependency analysis depending on the type of remote dependency specified (explained below).
+Due to this caching setup, it is normal for the first analysis to take some time, especially for larger projects, but future analysis of dependencies with the same information should be fast.
 
 ### Referenced dependencies
 
-Most `referenced-dependencies` are cached by their `(type, name, version)` tuple.
+Most `referenced-dependencies` are cached by the combination of their `(type, name, version)` fields.
 If `version` is not provided, the system assumes the version is "latest", and caching is usually not applied.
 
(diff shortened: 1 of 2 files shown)


docs/features/manual-dependencies.md
```diff
@@ -153,3 +153,23 @@ remote-dependencies:
...
+
+For dependency types that require `arch`, `os`, and `osVersion` attributes, these fields additionally form part of the cache tuple.
+
+### Custom dependencies
+
+`custom-dependencies` do not require analysis by the FOSSA backend and are therefore not cached.
+
+### Remote dependencies
+
+`remote-dependencies` are cached by their `(name, url, version)` tuple, which are all required fields.
```
meghfossa, line 175: [optional] We should have a few lines about rebuilding (what to do if you want to reanalyze without cache). This will be a nice place to send customers instead of having to write this: https://fossa.zendesk.com/agent/tickets/3037

meghfossa (approved):

docs/features/manual-dependencies.md
```diff
@@ -153,3 +153,23 @@ remote-dependencies:
     description: Provides foo and a helpful interface around foo-like tasks.
     homepage: https://www.foowrapper-home.com
 ```
+
+## Performance
+
+The FOSSA service caches the results of dependency analysis by different keys depending on the type of remote dependency specified (explained below).
```
zlav, line 159: [nit] "Different keys" sounds a little awkward, I'm not sure the idea of a "key" is immediately obvious. I would maybe remove this part, or say something more generic like "we cache things in different ways depending on the type".

zlav (approved):
Great change! You should add a note for "vendored dependencies" as well and how we cache those dependencies, but I wouldn't block on that change.

docs/features/manual-dependencies.md
jssblck: Added!

docs/features/manual-dependencies.md
jssblck: Good call! Updated!

PR #1118, opened 2022-11-30
Support quotes in go module directives

csasarak:
# Overview

During static analysis of a go mod file the CLI would fail with an error like

```
[ERROR] ----------
  An issue occurred

  >>> Relevant errors

    Error

      Error parsing file: /foo/go.mod.

          /foo/go.mod:1:8:
            |
          1 | module "github.com/<organization>/modulename"
            |        ^
          unexpected '"'
          expecting '-', '.', '/', '_', or alphanumeric character
```

Quotes in a module path are valid, as shown in this real world example: https://github.com/go-yaml/yaml/blob/v3/go.mod

This PR also adds a missing target to our Makefile.

## Acceptance criteria

The CLI should correctly parse a quoted module 
## Testing plan

When using static analysis on a go.mod file with a quoted module path the CLI should succeed.

## Testing

Disable dynamic analysis on [master](https://github.com/fossas/fossa-cli/blob/master/src/Strategy/Gomodules.hs#L69) then analyze the above repo. Change to this branch, disabling dynamic analysis and analyze the same repo. It should succeed. 

## Risks

None.

## References

None.

## Checklist

### DIFF
--- a/src/Strategy/Go/Gomod.hs
+++ b/src/Strategy/Go/Gomod.hs
@@ -287,6 +287,7 @@ gomodParser = do
     packageName :: Parser PackageName
     packageName = toText <$> lexeme (some (alphaNumChar <|> char '.' <|> char '/' <|> char '-' <|> char '_'))
 
+    modulePath :: Parser Text
     modulePath =
       packageName
         <|> between (char '"') (char '"') packageName
--- a/src/Strategy/Go/Gomod.hs
+++ b/src/Strategy/Go/Gomod.hs
@@ -199,7 +199,7 @@ gomodParser :: Parser Gomod
 gomodParser = do
   _ <- scn
   _ <- lexeme (chunk "module")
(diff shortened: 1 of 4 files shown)


src/Strategy/Go/Gomod.hs
```diff
@@ -287,6 +287,10 @@ gomodParser = do
     packageName :: Parser PackageName
     packageName = toText <$> lexeme (some (alphaNumChar <|> char '.' <|> char '/' <|> char '-' <|> char '_'))
 
+    modulePath =
```
meghfossa, line 290: [prefer] type annotations

meghfossa (approved):

src/Strategy/Go/Gomod.hs
csasarak: I've added it.

PR #1122, opened 2022-12-06
refactor: Allow callers to customize Themis flags

elldritch:
This PR refactors `generateThemisArgs` so that library consumers can pass different flags to it. I'm relying on the automated tests for this one - seems like a safe PR.


spatten (approved):
Looks good to me!

PR #1125, opened 2022-12-06
refactor: Match strings only appear in some Themis formats

elldritch:
This PR fixes the typing for `LicenseUnitMatchData` so that `fromJSON` no longer crashes when `match_string` is missing (e.g. when running Themis in `--srclib` mode).

Surprisingly, no other code changes are required for this, since we never actually examine the field (we merely deserialize and emit it).


spatten (approved):
:+1:

PR #1126, opened 2022-12-06
Yarn: See if the tar resolver will work before trying git.

csasarak:
# Overview

The git resolver for yarn will sometimes look at a tarball url and think that it's a git url. This PR swaps the ordering of the git and tar resolvers to give the tar resolver a first chance at resolving a locator.

## Acceptance criteria

Following the testing plan should not result in an error.

## Testing plan

1. Create an empty directory
2. Inside the directory, run `yarn init`
3. Run `yarn add "colorjs-test@https://gitpkg.now.sh/api/pkg.tgz?url=colorjs%2fcolor-name&commit=0f12d6e6ad4ab04e5cbc26360b759b446b0c6a4e"`
4. Run `fossa analyze`

On master, running `fossa analyze` will exit with a GitResolver error saying that the URL was invalid. On this branch `fossa analyze` should exist successfully.

## Risks

This solution relies on the internal ordering of a list. A more robust but time-consuming solution would be to make the GitResolver not match a tarball URL like the one in the test.

## References

ZD-5281

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## v3.6.9
+- Yarn: Fix a bug where tarball URLs were recognized as git urls. ([#1126](https://github.com/fossas/fossa-cli/pull/1126))
+
 ## v3.6.8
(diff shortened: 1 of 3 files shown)


zlav (approved):
LGTM! Nice comment calling out this weird behavior .

PR #1127, opened 2022-12-08
fix: Parser should actually be optional on match strings

elldritch:
This PR is a follow-up to #1125, because I'm an idiot.


spatten (approved):
👍

PR #1131, opened 2022-12-17
Gh actions fix

spatten:
# Overview

Our integration and benchmark tests were failing due to some errors when starting up nix.

This PR fixes that by doing two things:

1) Creating the nixbuild users and group before installing nix
2) vendoring the install-nix.sh from https://github.com/cachix/install-nix-action/blob/master/install-nix.sh and commenting out one line that was causing a problem.

## Acceptance criteria

The CI workflows defined in `bench.yml` and `integration-test.yml` should install nix and run properly.

## Risks

The only risk is that you will look at the commit history when reviewing this and think that I'm a total hack. So don't do that :D.

## Checklist


meghfossa (approved):
🚀

PR #1134, opened 2023-01-12
Support custom lib yarn protocol

zlav:
# Overview

By default, we support Yarn's [default protocols](https://yarnpkg.com/features/protocols). One user brought up an issue to us where they use a custom protocol named `lib` that causes their scans to fail. `lib` is a custom implementation of the `portal` protocol that essentially supports vendored dependencies.

This PR adds support for treating `lib` deps the same way we treat `portal` deps which makes sure we don't fail if we see a `lib` dep.

An example of a lib dep in a `yarn.lock` file can be found in this [thread](https://teamfossa.slack.com/archives/C02A1ERTXDZ/p1673561528991159).

## Acceptance criteria

`lib` type deps do not cause the CLI to fail.

## Testing plan
I tested this using a `yarn.lock` file found in this [thread](https://teamfossa.slack.com/archives/C02A1ERTXDZ/p1673561528991159).

I also added a test for the `libResolver` in the `ResolverSpec`

## Risks

I honestly can't think of 1 risk associated with this change.

## References

https://teamfossa.slack.com/archives/C02A1ERTXDZ/p1673561528991159

## Checklist

### DIFF
--- a/src/Strategy/Node/YarnV2/Resolvers.hs
+++ b/src/Strategy/Node/YarnV2/Resolvers.hs
@@ -56,7 +56,6 @@ data Package
   | TarPackage Text -- url
   | FilePackage Text
   | -- LibPackages are a custom protocol supported by a user.
-    -- TODO: Support all custom protocols out of the box.
     LibPackage Text
   | LinkPackage Text
   | PortalPackage Text
--- a/src/Strategy/Node/YarnV2/Resolvers.hs
+++ b/src/Strategy/Node/YarnV2/Resolvers.hs
@@ -40,12 +40,12 @@ import Strategy.Node.YarnV2.Lockfile
 import Text.Megaparsec
 
 data Resolver = Resolver
-  { resolverName :: Text
-  -- ^ Used for error messages
-  , resolverSupportsLocator :: Locator -> Bool
-  -- ^ Does this resolver support the locator?
-  , resolverLocatorToPackage :: Locator -> Either Text Package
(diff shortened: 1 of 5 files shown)


src/Strategy/Node/YarnV2/Resolvers.hs
```diff
@@ -47,12 +48,16 @@ data Resolver = Resolver
...
 
+-- Default Yarn Protocols can be found at https://yarnpkg.com/features/protocols.
 data Package
   = WorkspacePackage Text -- relative reference to a directory. not quite a Path Rel Dir because it may contain '..'
   | NpmPackage (Maybe Text) Text Text -- scope, package, version
   | GitPackage Text Text -- url, commit
   | TarPackage Text -- url
   | FilePackage Text
+  | -- LibPackages are a custom protocol supported by a user.
+    -- TODO: Support all custom protocols out of the box.
```
jssblck, line 59: [nit] TODO comments on their own are basically never actioned and are largely just noise since we have no TODO triage or audit process. Can we create a ticket for this instead?

jssblck (approved):

src/Strategy/Node/YarnV2/Resolvers.hs
zlav: Will do, I definitely don't disagree with that

zlav: https://fossa.atlassian.net/browse/ANE-751

PR #1138, opened 2023-01-14
Maven: Parse all scopes

jssblck:
# Overview

It is possible for a dependency to be included via multiple sub projects.
Repro case: [maven-demo-project.zip](https://github.com/fossas/fossa-cli/files/10416398/maven-demo-project.zip).

Roughly translated, given this graph:
```
pie
    apple
        commons-io
            scope: ["test"]
        commons-net
    crust
        commons-io
        commons-net
```

Prior to this PR, `commons-io` in this scenario is filtered, because we were only parsing `"test"` scopes: https://github.com/fossas/fossa-cli/blob/219bdc6f38d401df2bdb7991114c54083a75f56b/src/Strategy/Maven/PluginStrategy.hs#L170

In this PR, we instead now parse all scopes.
This causes the dependency to no longer be filtered, because rather than _excluding when test_, we _include when production_: https://github.com/fossas/fossa-cli/blob/41f1fbb99658e09c0b0b176acc637aad0217b04b/src/Srclib/Converter.hs#L94-L96

## Acceptance criteria

Dependencies with both `compile` and `test` scopes are still reported.

## Testing plan

Reproduction case attached above.

Running `fossa analyze` should result in both these dependencies being identified:
```
"mvn+commons-io:commons-io$2.6"
"mvn+commons-net:commons-net$3.3"
```

Screenshot 2023-01-13 at 7 11 44 PM

## Risks

This will result in more dependencies being reported, which may be confusing to users.
I think that this is likely to be relatively easily explained to users as being more accurate.

## References

- [ZD-4290](https://fossa.zendesk.com/agent/tickets/4290)

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## v3.6.12
 
-- Maven: If a package is both `"test"` and `"compile"`, it is no longer filtered.
+- Maven: If a package is both `"test"` and `"compile"`, it is no longer filtered ([#1138](https://github.com/fossas/fossa-cli/pull/1138)).
 
 ## v3.6.11
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.6.12
+
+- Maven: If a package is both `"test"` and `"compile"`, it is no longer filtered.
+
 ## v3.6.11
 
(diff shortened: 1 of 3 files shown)


spatten (approved):
This looks good to me once the tests pass

PR #1139, opened 2023-01-15
[ANE-689] [Hand off] Container can be scanned when tar entry with path of `''` exists! 

meghfossa:
# Overview

Works on ANE-689

This PR, 
- uses modified haskell/tar
- updates `barkelydb` strategy, such that single pkg parsing failure does not cause entire tactic to fail

## Acceptance criteria

- can scan following without fatal failure!

## Testing plan

```bash
git checkout uses-safe-tar-reader
cabal clean
make install-local

# Before
fossa -V # ensure latest
fossa container analyze -o cypress/base:16.13.0  

# After
./fossa container analyze -o cypress/base:16.13.0  
```

Remaining tests,

- [ ] `alpine:edge`
- [ ] `public.ecr.aws/lambda/python:3.9`
- [ ] and many more ... 
- [ ] Please test ticketed container images (refer to ane-689 references)
- [ ] Berkely db output with redhat base image (refer to output in FOSSA web ui to ensure only non-parsable pkgs were not reported)

## Risks

N/A.

## References

ANE-689

## Checklist

### DIFF
--- a/src/Container/Tarball.hs
+++ b/src/Container/Tarball.hs
@@ -54,7 +54,7 @@ data TarEntries = TarEntries
 
 -- | Parses Container Image from Tarball Byte string.
 parse :: ByteStringLazy.ByteString -> Either (NLE.NonEmpty ContainerImgParsingError) ContainerImageRaw
-parse content = case mkEntries $ Tar.read content of
+parse content = case mkEntries $ Tar.read' content of
   Left err -> Left $ NLE.singleton err
   Right te -> do
     -- Exported docker image must have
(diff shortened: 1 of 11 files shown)


src/Container/Tarball.hs
```diff
@@ -54,7 +54,7 @@ data TarEntries = TarEntries
 
 -- | Parses Container Image from Tarball Byte string.
 parse :: ByteStringLazy.ByteString -> Either (NLE.NonEmpty ContainerImgParsingError) ContainerImageRaw
-parse content = case mkEntries $ Tar.read content of
+parse content = case mkEntries $ Tar.read' content of
```
meghfossa, line 57: We also use `Tar.read` on Archive extractor for unarchiver - which I did not change. 

Errors we are seeing when extracting archives in recursive unarchiving, may or may not solved by this (I believe there was a ticket in backlog for this). They may have also been due to filepath limit (since haskell/tar does not support PAX tar file fully).

jssblck:
@meghfossa closing this one since we implemented this in a separate PR!

PR #1141, opened 2023-01-23
Document that we only have a dynamic strategy for Scala.

csasarak:
# Overview

The docs say that we have a static analysis strategy for Scala. We don't, this PR changes the documentation.


spatten (approved):
:+1:

PR #1145, opened 2023-02-02
Support project labels

zlav:
# Overview

The CLI should be able to handle a new “label” option which will allow a string parameter with the label name. Up to 5 label options can be provided. This data will be accepted by the CLI upload route.

The CLI route accepts an array of strings as a query param “labels”.

## Acceptance criteria

- Project labels are accepted through the command line and the configuration file.
- Only 5 are accepted at a time

## Testing plan

The route is already supported on FOSSA cloud, so testing this was as easy as enabling labels and ensuring that the correct behavior was met when setting a `--project-label` flag.

## References

- [ANE-715](https://fossa.atlassian.net/browse/ANE-715): Implement .

## Checklist

[ANE-715]: https://fossa.atlassian.net/browse/ANE-715?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/references/files/fossa-yml.md
+++ b/docs/references/files/fossa-yml.md
@@ -131,9 +131,9 @@ The `name:` and `release:` of the release group's release to add your project to
 If you choose to associate a project with a release group, you **must** supply both name and release.
 
 #### `project.labels:`
-The `labels` field allows you to add labels to projects so that you can classify certain projects how you would like. This adds a more flexible way to query for projects as opposed to assigning a project to a team.
+The `labels` field allows you to add labels to projects so that you can classify certain projects how you would like. This adds a more flexible way to query for projects in the FOSSA UI as opposed to assigning a project to a team.
 
-Up to 5 labels are allowed to be associated with a project at a time.
+Up to 5 labels are allowed to be associated with a project at a time. Labels will be applied in order up to the limit of 5 labels and then ignored.
 
 ### `revision:`
 The revision fields are used to help FOSSA differentiate between one upload for a project and another, just as GitHub uses commit hashes and branch names.
--- a/docs/references/files/fossa-yml.md
+++ b/docs/references/files/fossa-yml.md
@@ -22,6 +22,9 @@ project:
   releaseGroup:
     name: release-group-name
     release: 123-release-candidate
+  labels:
+    - project-label-1
+    - test-project
 
 revision:
   commit: "12345"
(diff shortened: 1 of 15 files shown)


docs/references/files/fossa-yml.md
```diff
@@ -127,6 +130,11 @@ The `name:` and `release:` of the release group's release to add your project to
 
 If you choose to associate a project with a release group, you **must** supply both name and release.
 
+#### `project.labels:`
+The `labels` field allows you to add labels to projects so that you can classify certain projects how you would like. This adds a more flexible way to query for projects as opposed to assigning a project to a team.
```
csasarak, line 134: [nit] Maybe mention "querying in the FOSSA web UI." Someone who is mainly used to using the CLI may not understand where they would be querying.

docs/references/files/fossa-yml.md
```diff
@@ -127,6 +130,11 @@ The `name:` and `release:` of the release group's release to add your project to
...
+
+Up to 5 labels are allowed to be associated with a project at a time.
```
csasarak, line 136: If I scan the project with one label, then scan it later with two different labels do the new project labels get added or replace the old ones? If it's additive, what happens if I add something above 5?

Since this behavior is probably controlled by core, I can see this being more appropriately documented elsewhere. If it is, maybe include a link?

src/App/Fossa/Config/Analyze.hs
```diff
@@ -517,6 +517,7 @@ collectScanDestination maybeCfgFile envvars AnalyzeCliOpts{..} =
     else do
       apiOpts <- collectApiOpts maybeCfgFile envvars commons
       let metaMerged = maybe analyzeMetadata (mergeFileCmdMetadata analyzeMetadata) (maybeCfgFile)
+      when (length (projectLabel metaMerged) > 5) $ fatalText "Projects are only allowed to have 5 associated project labels"
```
csasarak, line 520: Is it possible to add a unit test for the failure on `> 5` behavior?

csasarak (approved):
I added some questions for things to consider, but lgtm.

docs/references/files/fossa-yml.md
zlav: This isn't publicly documented yet, but I added information about what happens if 5 labels already exist. Thats a good thought, thanks.

PR #1146, opened 2023-02-02
Ane 635 add links to cli issues

csasarak:
# Overview

This PR adds cve and next safe version information to vulnerability issues and adds dashboard links for all issues in fossa test.

## Acceptance criteria

 * The output of the fossa test shows how the weblink to ENDPOINT’s web app’s issue page for all issues, not just CVE issues.
* The output of the fossa test, shows CVE ID for vulnerability issues if CVE ID exists.
* The output of the fossa test shows version fixing vulnerability for vulnerability.
* `fossa test` should also work for versions of FOSSA which do not provide CVE or link information.

## Testing plan
I ran manual tests against a local build of FOSSA which outputs the new data and verified that it was present.
1. Populate the local database with vuln information
2. I created an npm project with one of those vulnerabilities in it and analyzed it.
3. I ran `fossa test` and verified that the output was correct and links were valid.

I also ran repeated this against FOSSA in production to ensure that `fossa test` still behaves properly for versions of FOSSA that don't output this information.

## Risks

* I didn't add automated testing since the output is human readable text, which we don't generally test. The responses from the API are also arbitrary text, so I felt there wasn't a lot of value to testing. Reviewers should read this with an eye towards any automated testing they'd like that I haven't thought of.

## References

[ANE-635](https://fossa.atlassian.net/browse/ANE-635)

## Checklist

- [x] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

[ANE-635]: https://fossa.atlassian.net/browse/ANE-635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1148, opened 2023-02-03
Change builds builds to cache dependencies once

csasarak:
# Overview
I noticed that benchmarks, integration, and cabal-install-check all were using the same cache. This is because in a cache directive like:
```
    # Build FOSSA CLI.
    - uses: actions/cache@v2
      name: Cache cabal store
      with:
        path: ${{ steps.setup-haskell.outputs.cabal-store || '~/.cabal/store' }}
        key: ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-${{ hashFiles('**/*.cabal') }}
        restore-keys: |
          ${{ runner.os }}-${{ matrix.ghc }}-cabal-cache-
          ${{ runner.os }}-${{ matrix.ghc }}-
          ${{ runner.os }}-
```
`matrix.ghc` is empty. So each of those workflows/jobs were sharing a cache like `Linux--cabal-cache-41b67edbc852479d95e48e3b00587ab9d8396f1adcbc66b7e5d243bb0b1b16ec`. 

However, on all runs both integration tests and benchmarks were rebuilding a bunch of dependencies when I would have hoped that they would be sharing between them.

I had a look at our logs as well as the workflows and found that in specific circumstances we were doing more work than we have to. The three workflows below were doing builds using different projects. Or benchmark's case using both project files.
| Workflow                         | cabal.project.linux.ci | cabal.project |
| ------------------------- | --------------------- | ------------- |
| Integration tests               | ✅                               |                      |
| Lint (cabal-install-check) |                                    | ✅                 |
| Benchmarks                      | ✅                              | ✅                 | 

Keep in mind that running `cabal test` or `cabal bench` will trigger a rebuild if the profile they're run with is different than what cabal built before.

What I think would happen in some circumstances is this:

1. All three of benchmarks, integration tests, and lint (cabal-install-check) start running. Whichever one gets to the `Cache cabal store` step first reserves a key and gets the right to store its `~/.cabal/store` under that key as a new cache.
2. If Benchmarks makes the cache, then we're in luck because both the default profile and the linux ci profile get built.
3. If either of integration tests or lint makes the cache, then the other two are forced to do a rebuild because build products from one of project files are missing. Benchmarks will always do a full rebuild at least once since at different times it tries to use both project files.

You can see an example in the build logs for this [PR](https://github.com/fossas/fossa-cli/pull/1146).
* [cabal-install-check](https://github.com/fossas/fossa-cli/actions/runs/4107711674/jobs/7087521196) doesn't rebuild anything but `spectrometer`, because it made the cache that is being used. It builds everything using the default profile. 
* [benchmarks](https://github.com/fossas/fossa-cli/actions/runs/4107711676/jobs/7087520418) runs and is forced to rebuild everything using the `cabal.project.linux.ci` profile. It then tries to run the benchmarks using the default profile, so falls back to the cached versions of packages. It also has to build spectrometer a second time.
* [integration tests](https://github.com/fossas/fossa-cli/actions/runs/4107711672/jobs/7087520312) are built using the `cabal.project.linux.ci` profile, whose deps are not in the cache so all dependencies need to be built once again.

# What this PR does
1. As before there are 3 platform specific caches, but integration tests and the `cabal-install-check` now use the Linux release cache instead of sharing with benchmarks.
2. Benchmarks has its own cache now because it was built with `-O0` settings. I preserved that behavior because I remember discussing this when the benchmarks were written and at the time it sounded preferable to test without optimizations. Benchmarks only use the default `cabal.project`.
3. I configured it so that changes to `cabal.project.*` trigger building a new cache. When we update the haskell index state, I think it should make a new cache. It did not before.
4. This PR was initially to add a new Makefile target, so that's there too.

# Testing
You should be able to look at the build logs for the most recent set of checks and see that they build far fewer library dependencies.

# Future Work/Potential Problems
* The wall clock time here may not be affected too much because the build process lasts as long as its slowest component. In the future we may want to think about caching `dist-newstyle/`. This wouldn't be perfect, but besides space utilization it wouldn't be worse than rebuilding everything every time. 
* The current set up relies on lint, integration test, and the Linux release builds all building everything because we don't know which one will get to produce a new cache. This is unfortunate because integration tests don't rely on something like `pathfinder` building successfully.
* `cache/actions` will try to match cache key prefixes. I'm not sure this is desirable when we need to build a new cache, because it will download the old cache, add any new items it needs, and then store that as the new cache. I'm not positive, but I think we'll slowly be accumulating unused deps in our caches.


elldritch:
If you want to take a look at caching `dist-newstyle/` (which we do in Hubble), check out my old attempt in https://github.com/fossas/spectrometer/pull/439 - be careful to make sure you correctly invalidate Template Haskell code.

csasarak:
> If you want to take a look at caching `dist-newstyle/` (which we do in Hubble), check out my old attempt in [fossas/spectrometer#439](https://github.com/fossas/spectrometer/pull/439) - be careful to make sure you correctly invalidate Template Haskell code.

Thanks! I'll have a look at that sometime.

jssblck (approved):
This looks great, thanks for doing this! Definitely a step in the right direction.

PR #1150, opened 2023-02-09
Exports the Themis command

periodic:
# Overview

Hubble needs to be able to cancel Themis execution.  We can't use the current `Exec` effect because it runs the process synchronously and does not provide a process handle to send the signal to. 

This change just exports the `themisCommand` so that Hubble can create and run the same Themis command that the CLI does, but can use some slightly lower-level functions to manage the timeout.

## Acceptance criteria

The Themis command is available to Hubble.

## Testing plan

I have a Hubble PoC using this change already.

## Risks

None, just an export.  There's a chance that we won't end up using this and forget to hide it, but the threat is low.

## References

- [PLAT-555](https://fossa.atlassian.net/browse/PLAT-555): Themis timeouts should be considered partial successes

## Checklist

[PLAT-555]: https://fossa.atlassian.net/browse/PLAT-555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/RunThemis.hs
+++ b/src/App/Fossa/RunThemis.hs
@@ -3,6 +3,7 @@
 module App.Fossa.RunThemis (
   execThemis,
   execRawThemis,
+  -- `themisCommand` is used by Hubble.
   themisCommand,
   themisFlags,
 ) where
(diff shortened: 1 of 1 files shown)


jssblck (approved):

jssblck:
Actually - @periodic will you add a comment to this export explaining that this is used by Hubble?
That way we don't accidentally remove it later because "it's not being used anywhere" or something.

PR #1152, opened 2023-02-10
Handle Leiningens deduping of maven groupID and artifactID

zlav:
# Overview

The Leiningen tactic “dedupes” the groupID and artifcatID if they are the same. This results in locators that cannot be resolved because they don’t match maven dependency formats.

## Acceptance criteria

Scanning a leiningen repository with deps that appear as "unknown license" now resolve in the FOSSA UI. My test repo has been https://github.com/pedestal/pedestal.

## Testing plan

1. Download https://github.com/pedestal/pedestal.
1. Run `fossa analyze` with an existing version and observe all the unknown license dependencies
1. Run `fossa analyze` with the updated version and observe that these deps are found and that there are no regressions with the previously found dependencies.

Pedestal - https://github.com/pedestal/pedestal
Eastwood - https://github.com/jonase/eastwood
I don't know the specific number of deps that are fixed between the two of these, but it seems like 5% of deps are unknown. After I applied the fixes from this PR I found that all of these unknown licensed deps are fixed but the total count remained the same.

## Risks

The biggest risk is that the parser addition I'm writing accidentally breaks a dependency format that we are unaware of and are not currently testing for. The only way I see this happening is if there is a dep in the leiningen manifest file that has `:` in its name, although I'm fairly confident this is not an accepted dependency format because it would lead to oddly formated Maven dependencies.

## References

[ANE-718](https://fossa.atlassian.net/browse/ANE-718)

## Checklist

[ANE-718]: https://fossa.atlassian.net/browse/ANE-718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,10 @@
 # FOSSA CLI Changelog
 
-<!-- title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+<!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+
+## v3.6.17
+
+- Handle Leinging deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
 
 ## v3.6.17
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -4,7 +4,7 @@
 
 ## v3.6.17
 
-- Handle Leinging deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
+- Handle Leiningen deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
 
 ## v3.6.17
 
--- a/src/Strategy/Leiningen.hs
+++ b/src/Strategy/Leiningen.hs
@@ -201,7 +201,12 @@ buildEdges deps = Map.traverseWithKey single (depsTree deps) $> ()
           buildEdges deeper
 
 toClojureNode :: ClojureDep -> ClojureNode
-toClojureNode dep = ClojureNode (Text.replace "/" ":" (depName dep)) (depVersion dep)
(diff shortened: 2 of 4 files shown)


Changelog.md
```diff
@@ -1,6 +1,10 @@
 # FOSSA CLI Changelog
 
-<!-- title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+<!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+
+## v3.6.17
+
+- Handle Leinging deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
```
spatten, line 7:
```suggestion
- Handle Leiningen deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
```

src/Strategy/Leiningen.hs
```diff
@@ -201,7 +201,12 @@ buildEdges deps = Map.traverseWithKey single (depsTree deps) $> ()
           buildEdges deeper
 
 toClojureNode :: ClojureDep -> ClojureNode
-toClojureNode dep = ClojureNode (Text.replace "/" ":" (depName dep)) (depVersion dep)
+toClojureNode dep = ClojureNode (reverseDepNameMerge $ Text.replace "/" ":" (depName dep)) (depVersion dep)
+
+-- When a dependencies groupID and artifactID are identical, leiningen merges them.
+-- In order to satisfy the Maven dependency format, we reverse the merge of these names.
+reverseDepNameMerge :: Text -> Text
+reverseDepNameMerge dep = if Text.any (== ':') dep then dep else dep <> ":" <> dep
```
spatten, line 209: Would `Text.elem` be slightly better here?

```
if ':' `elem` dep then dep else dep <> ":" <> dep
```

spatten (approved):
Two minor nits, but looks good to me!

src/Strategy/Leiningen.hs
zlav: Thanks for the suggestion, this is much better

PR #1153, opened 2023-02-15
Ane 665 warn on unarchiving error

csasarak:
# Overview

When we fail to unpack an archive, we should emit a warning rather than failing. 

## Acceptance criteria

  * When we fail to unarchive a file, we should keep going and still scan the vendored dependency

  * When we fail to unarchive a file, we should emit a warning

## Testing plan

1. Find a project with a bad archive in and and use it as a vendored dep. I made an empty project folder and used [grafana](https://github.com/grafana/grafana) as a vendored dep with:
```
vendored-dependencies:
- name: Grafana
  path: grafana/
  version: master
```
2. Run the `master` branch cli with: `fossa analyze --force-vendored-dependency-rescans <dir>`
This should emit errors for the targets which have bad archives.

3. Using the cli from this branch, run the same thing. The failing targets from (2) should succeed with warnings. Use `--debug` to see the warnings printed out. 

## Risks

This change touches all code that does unarchiving. The only place a bit outside of license scanning where we do this is with Jar [BinaryDeps](https://github.com/fossas/fossa-cli/blob/ANE-665-warn-on-unarchiving-error/src/App/Fossa/BinaryDeps/Jar.hs#L62). I think this is ok because we `warnOnErr` in that code anyways and I have documented the new behavior.

## References

[ANE-665](https://fossa.atlassian.net/browse/ANE-665)

## Checklist

[ANE-665]: https://fossa.atlassian.net/browse/ANE-665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,9 @@
 
 <!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
 
+## v3.6.18
+- License Scanning: Emit a warning if unarchiving fails rather than a fatal error. ([#1153](https://github.com/fossas/fossa-cli/pull/1153))
+
 ## v3.6.17
 
 - Handle Leiningen deduped deps: expand groupID and artifactID in the leiningen tactic to satisfy the Maven fetcher ([#1152]](https://github.com/fossas/fossa-cli/pull/1152))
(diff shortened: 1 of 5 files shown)


spatten (approved):
This looks great!

The test plan worked for me, and, IMO, `withArchive'` is more readable after this change, too :)

PR #1157, opened 2023-02-17
Call out C/C++ strategies as different

jssblck:
# Overview

[Rendered version](https://github.com/fossas/fossa-cli/tree/call-out-c-cpp/docs/references/strategies#strategies-by-type)

In the [strategies reference](https://github.com/fossas/fossa-cli/tree/master/docs/references/strategies#strategies-by-type),
we don't currently mention that C/C++ are _fundamentally different_ than other strategies in that list.

This PR resolves that by adding a note that's hopefully hard to miss without messing with the overall readability too badly.


csasarak (approved):
looks good!

spatten (approved):
:+1:

elldritch:
I would maybe make this even more dramatic, changing the check boxes in the dynamic and static columns to ⚠️ s.

elldritch (approved):

PR #1158, opened 2023-02-21
Enable sha256 verification of downloads with install-latest.sh.

csasarak:
# Overview

This re-enables checking that the archives we download from GH are checked against their checksums by `install_latest.sh`.

## Acceptance criteria

* `install_latest.sh` checks that the sha256 of the downloaded archive matches what is published by GH.

## Testing plan

I ran the script and made sure that it succeeded.

## Risks

* People who use this in CI that don't happen to have the `sha256sum` program installed could have this start to break.

## Checklist

### DIFF
--- a/.github/workflows/install-script-test.yml
+++ b/.github/workflows/install-script-test.yml
@@ -27,8 +27,9 @@ jobs:
       - name: install latest script can install a specific version
         shell: bash
         run: |
-          ./install-latest.sh -b . v3.1.1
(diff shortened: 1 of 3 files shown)


spatten (approved):
I think this is pretty low risk seeing as the `install.sh` does this checking and we I don't think we've gotten any complaints

I do think we should add this to the changelog, though, before merging

csasarak:
I've added it to the changelog under unreleased, I don't think this really warrants a full release by itself.

PR #1159, opened 2023-02-23
Make verify sha_256 work for fossa cl versions before 3.4.7 on linux.

csasarak:
# Overview

An earlier PR broke `install-latest.sh` because it would try to download the wrong checksum file to check the archive with. 

## Acceptance criteria

* When installing versions of fossa before 3.4.7 on linux, we should successfully check the sha_256 rather than failing.

## Testing plan

1. Clone the master branch onto a linux system (or in a linux container with something like `docker run -it`)
2. Run `./install-latest.sh v3.2.3` and notice that it fails when trying to fetch the checksum file.
3. Check out this branch and run `./install-latest.sh v3.2.3` and see that it successfully installs the right version without an error.
4. Try `./install-latest.sh` and see that it installs the latest version correctly still.

## Risks

I think the main risk is if I've missed something in my version comparison function. 

## Checklist

I didn't update the changelog since the broken version was never actually released. We had a customer using the script from our master branch.


zlav (dismissed):
Can you add a test for the old versions as well? You should be able to pull the old test from the previous PR and add that in addition to the new one

install-latest.sh
```diff
@@ -389,7 +397,37 @@ get_binary_name() {
...
+            return 0
+        fi
+
+        if [ "$VERSION_PIECE1" -gt "$VERSION_PIECE2" ]
+        then
+            return 1
+        fi
+    done
 
+    return 1
```
spatten, line 429: Will this return true if you compare `3.4.7` to `3.4.7`? I.e. is this `<=`?

csasarak:
> Can you add a test for the old versions as well? You should be able to pull the old test from the previous PR and add that in addition to the new one

Good idea, I've added it.

install-latest.sh
csasarak: You're right, let me change that to be properly `<`.

install-latest.sh
```diff
@@ -389,7 +397,37 @@ get_binary_name() {
...
+# 2.1.4 < 3.1.4
+# 3.5.0 < 3.6.5
+# 3.5.5 < 3.5.15
+#
+# Returns 0 when the first version argument < the second.
+# Returns 1 otherwise.
+version_less_than() {
+    for i in 1 2 3
+    do
+        VERSION_PIECE1=$(echo "$1" | cut -d '.' -f $i)
```
spatten, line 415: First, I think I was wrong. I wrote a test harness and it worked properly with `return 1` at the end. I think I was thinking "1 is true", but this is bash so `0` is true. Is that right?

But, I think we do have a problem because we are comparing strings so we'll get things like "3.10.1" < "3.4.7".

```bash
#!/bin/sh
set -e
version_less_than() {
    for i in 1 2 3
    do
        VERSION_PIECE1=$(echo "$1" | cut -d '.' -f $i)
        VERSION_PIECE2=$(echo "$2" | cut -d '.' -f $i)

        if [ "$VERSION_PIECE1" -lt "$VERSION_PIECE2" ]
        then
            return 0
        fi

        if [ "$VERSION_PIECE1" -gt "$VERSION_PIECE2" ]
        then
            return 1
        fi
    done

    return 1
}

test() {
  if version_less_than $1 $2
  then
    echo "$1 < $2"
  else
    echo "$1 NOT < $2"
  fi
}


test "3.4.6" "3.4.7"
test "3.4.7" "3.4.7"
test "2.11.1" "3.4.7"
test "3.11.1" "3.4.7"
```

```
3.4.6 < 3.4.7
3.4.7 NOT < 3.4.7
2.11.1 < 3.4.7
3.11.1 NOT < 3.4.7
```

csasarak: 1. You're right, when writing this I initially made that same mistake. Thinking that 1 was true and 0 was false, but 0/1 indicate the presence or absence of an error in sh.
2. I initially thought that `-lt` was numeric, but looking at the manpage I see that's I messed that up. Let me see what the better way to do that is. I got errors with `<` not reliably being part of `sh`, so let me see if there's some other way.

spatten: I'm wrong here, again. Chris had it correct from the beginning, as the test case clearlly shows

csasarak: After discussing this with Scott, it looks like everything is ok.

spatten (approved):
This looks good to me :+1:

PR #1160, opened 2023-02-23
Stack git locator

csasarak:
# Overview

Stack allows the specification of git repos as `extra-deps` in `stack.yml`. We don't report them from the CLI in a way that FOSSA can scan them, however.

## Acceptance criteria

* Stack projects that define `extra-deps` pointing to a git repo shouldn't break fossa-cli and should appear correctly in FOSSA.

## Testing plan

Use the procedure [here](https://gist.github.com/gfarrell/051d96ec26b4d165abe4d958772baf8f) to create a new stack project with the supplied `stack.yaml`. 

Run the version of fossa from this branch against the project directory: `fossa analyze --only-target stack <project dir>`.

Make sure to include the `--only-target` option or the cli will try to analyze it as a cabal project as well as a stack one. 

## References

ZD-5728

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,8 +2,10 @@
 
 <!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
 
-## v3.7.0
+## v3.7.1
+- Stack: Git based dependencies are detected and handled correctly. ([#1160](https://github.com/fossas/fossa-cli/pull/1160))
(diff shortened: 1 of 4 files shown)


gfarrell:
Would be great to see this merge so that we can use FOSSA in our projects.

jssblck (approved):

PR #1168, opened 2023-03-22
Ane 891 dynamic go tactic

csasarak:
# Overview

This is an initial implementation of a new experimental Go strategy based on `go list -json -deps all`. 

This PR does not implement everything that needs to be done for ANE-891. But I wanted to get some review early. ACs that still need to be done in this or a future PR are:

- [ ] Exclude test dependencies
- [ ] Integration test
- [ ] Handle pseudo-versions
- [ ] Improve efficiency - This isn't a problem that I've seen so far, but there is a little low-hanging fruit in my implementation IMO. 

I've added some minimal documentation for this tactic, there is going to be another ticket to write something more extensive. I will update the changelog if reviewers think that this is ok to incorporate into a release.

Also, the flag itself is a bit long and unwieldy (`experimental-package-centric-go`) and also talks about implementation details, which I think is inappropriate. If anyone has other ideas I'd like to hear them.

## Acceptance criteria

   * There should be a new dynamic tactic for Go using `go list -json -deps all` available.
   * The new tactic is under an experimental flag and not be the default.
   * Should handle both local and package replacements.
   * Should Label deep and direct dependencies.
   * Should detect path dependencies and not include them or their child deps in the graph. 
   * Handle the "C" special package. 

## Testing plan

1. Build the cli on this branch.
2. Run `fossa analyze --experimental-package-centric-go <dir>` on a Go project.
3. Look over the results to make sure they make sense. 

Repos that demonstrate the functionality of this strategy:

https://github.com/oss-review-toolkit/ort/tree/7725f9285eb6d6671937384d0f2a7896cfeb3332/analyzer/src/funTest/assets/projects/synthetic/gomod - Shows removing modules with unused packages from graph. 
https://github.com/ipfs/kubo - Shows ignoring the special "C" package. The current strategy will also do this.
[local_path_deps.tar.gz](https://github.com/fossas/fossa-cli/files/11081554/local_path_deps.tar.gz) this demonstrates pruning out local path deps. You can see the difference by running it with the experimental flag and then without it.
https://github.com/kubernetes/kubernetes - (I found a bug with scanning this, I will update when I fix it)
https://github.com/coredns/coredns 

## Risks
Any major risks I can think of are inherent to trying to replace our dynamic Go tactic and not specific to this PR.

## References

- [ANE-891](https://fossa.atlassian.net/browse/ANE-891)

## Checklist

[ANE-891]: https://fossa.atlassian.net/browse/ANE-891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,240 @@
...
+buildGraph :: (Has Diagnostics sig m) => [GoPackage] -> m (Graphing.Graphing Dependency, GraphBreadth)
+buildGraph rawPackages =
+  fmap ((,Complete) . pruneUnreachable) -- Remove deps that are only children of removed path deps.
+    . evalGrapher
+    . traverse_ graphEdges
+    $ pkgsNoStdLibImports
+  where
+    (stdLibImportPaths, pkgsNoStdLibImports) = foldl' go (HashSet.empty, HashMap.empty) rawPackages
+
+    go :: (HashSet.HashSet ImportPath, HashMap.HashMap ImportPath GoPackage) -> GoPackage -> (HashSet.HashSet ImportPath, HashMap.HashMap ImportPath GoPackage)
```
zlav, line 171: Why are you taking `(HashSet.empty, HashMap.empty)` as an argument here? It feels like this would be better set inside the `go` function.

src/Strategy/Gomodules.hs
```diff
@@ -75,6 +77,8 @@ getDeps project = do
       , dependencyManifestFiles = [gomodulesGomod project]
       }
   where
+    -- `go list -json -deps all` marks std lib deps with a boolean, so Strategy.Go.GoListPackages does this filtering itself.
+    -- I think this can be removed when it becomes the default.
```
zlav, line 81:
```suggestion
    -- I think this can be removed when `go list -json -deps all` becomes the default.
```

Be explicit instead of `it`.

Also just out of curiosity, the reason this can be removed is because we are explicity filtering the `stldlib` deps during graph creation in the new method right?

src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,240 @@
...
+    go (stdLibPaths, otherPackages) g@GoPackage{standard, importPath}
+      | standard = (HashSet.insert importPath stdLibPaths, otherPackages)
+      | otherwise = (stdLibPaths, HashMap.insert importPath (removeIgnoredPackages g) otherPackages)
```
zlav, line 174: I may be totally missing something here, I'm guessing this is recursive and I'm missing where that recursion is happening, but how is it possible to use `removeIgnoredPackages` here inside `go` if `removeIgnoredPackages` uses `ignoredPackages` which uses `stdLibImportPaths` which depends on the output of `foldl' go`? Is this not a circular dependency?

src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,240 @@
...
+            addChildEdge :: (Has Diagnostics sig m, Has (Grapher Dependency) sig m) => Dependency -> m ()
+            addChildEdge childDep = do
+              when (childDep /= currDep) $
+                edge currDep childDep
+
+            makeEdge :: (Has (Grapher Dependency) sig m, Has Diagnostics sig m) => ImportPath -> m ()
+            makeEdge = importToModule >=> pure . modToDep >=> addChildEdge
+        traverse_ makeEdge packageDeps
+        if indirect
+          then deep currDep
```
zlav, line 202: Whats the purpose of this `deep` function? Why do we have to explicitly set the dep to be deep? Isn't knowing if it is direct or not enough?

src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,240 @@
...
+          else direct currDep
+
+    -- Look up module info for an import path, performing replacements if necessary.
+    importToModule :: Has Diagnostics sig m => ImportPath -> m GoModule
+    importToModule impPath = do
+      -- Return a stand-in value for missing import paths so it's visible to users? even if unscannable.
```
zlav, line 208: Why the `?`? Is this a leftover TODO wondering what to do if an importPath doesn't have a match in the `HashMap`?

Have you seen this in the example projects you have worked with?

src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,240 @@
...
+instance ToDiagnostic MissingModuleErr where
+  renderDiagnostic (MissingModuleErr (ImportPath i)) = pretty $ "Could not find module for " <> i
+
+-- |A module is a path dep if its import path starts with './' or '../'.
+isPathDep :: GoModule -> Bool
+-- I think this method of detection is crude, but I'm not sure how else to distinguish path deps from regular ones.
+-- If I don't assume that a path dep always starts with one of these prefixes, it's possible that what looks like an import path is actually a relative path to a local directory.
+-- One possible improvement is to only perform this check when a module is a replacement for another.
+-- Is this too unix specific?
+isPathDep GoModule{modulePath = ModulePath mP} = any (`isPrefixOf` mP) ["./", "../"]
```
zlav, line 225: I agree this is a crude detection method, but I think this is valid and I do not think this is unix specific. I don't think this is unix specific because path deps are declared using `./` in the `go.mod` file and I would be surprised if that changed in a Windows environment. Either way, we have a higher tolerance for risk with path deps since we plan to fully support them soon.

test/Go/GoListPackagesSpec.hs
```diff
@@ -0,0 +1,241 @@
...
+                            , isMainModule = False
+                            , replacement = Nothing
+                            }
+                    , packageDeps = [ImportPath "C"]
+                    , listError = Nothing
+                    }
+                ]
+      assertOnSuccess result $ \_ (graph, _) -> graph `shouldBe` Graphing.direct xdiff
+
+analysisSpec :: Spec
```
zlav, line 221: Can you add a test case here for 1 dep that requires a replacement and 1 dep that is a path dependency?

zlav (changes requested):
Overall the implementation looks great and it accomplishes what we want. I left some generic comments and a request to add tests. I don't think you need to make any significant changes, I mostly just want my questions addressed before approving.

src/App/Fossa/Config/ListTargets.hs
```diff
@@ -69,10 +69,12 @@ mergeOpts cfgfile _envvars ListTargetsCliOpts{..} = do
...
-  ExperimentalAnalyzeConfig $
-    fmap
-      gradleConfigsOnly
-      (maybeCfg >>= configExperimental >>= gradle)
+  ExperimentalAnalyzeConfig
+    ( fmap
+        gradleConfigsOnly
+        (maybeCfg >>= configExperimental >>= gradle)
+    )
+    False -- This should be ok because its discovery should not work differently than the old Go modules tactic.
```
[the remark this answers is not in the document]
csasarak: No, the new tactic only works on things that have already been discovered, like a `go.mod` file. Do you think `True` is the better option?

test/GraphUtil.hs
```diff
@@ -9,6 +9,8 @@ module GraphUtil (
   expectEdges',
   expectDirect,
   expectDirect',
+  expectGraphEqual,
+  expectGraphEqual',
```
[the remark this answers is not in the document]
csasarak: `expectGraphEqual'` is not used so I doesn't have to be exported. I can remove it.

src/Strategy/Go/GoListPackages.hs
csasarak: The type of `foldl'` requires an initial value: `foldl' :: Foldable t => (b -> a -> b) -> b -> t a -> b` which is what those empty values are. Can you tell me a bit about what you're envisioning?

src/Strategy/Go/GoListPackages.hs
csasarak: You're right, this is a holdover from an earlier version of the code. The only case I've seen this happen is with the special `C` package and that's been filtered out earlier in the process now. If we can't look one up, that means there's something weird with the `go list` output and it outputs a fatal error in that case. I'm going to remove the comment.

src/Strategy/Go/GoListPackages.hs
csasarak: I was unsure when I wrote this, but was guessing that it was a requirement for local path deps to start with `./` or `../` because they'd be impossible to distinguish from regular module paths otherwise in the absence of a flag or something indicated it was a path dep.

I had another look here and it looks like this is the correct way to do it actually: https://go.dev/ref/mod#go-mod-file-replace

src/Strategy/Go/GoListPackages.hs
zlav: Oh right, that makes total sense. Thanks for clarifying.

src/Strategy/Go/GoListPackages.hs
zlav: Talked offline and this makes sense now!

src/App/Fossa/Config/ListTargets.hs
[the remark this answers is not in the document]
zlav: Not blocking on this, but personally I would be a fan of making this `True` so that if we _did_ have any issues with the new strategy in discovery, we find them ahead of time.

zlav (approved):

test/Go/GoListPackagesSpec.hs
csasarak: I've replaced the unit tests and included this as a case in them.

src/App/Fossa/Config/ListTargets.hs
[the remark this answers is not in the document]
csasarak: I can, but I don't think we would actually find them ahead of time. This code is a new tactic (I've been loose with tactic vs. strategy in discussing this, I apologize) inside of the existing Go modules strategy, so the discovery code is exactly the same as what's used when for the old `go mod graph` tactic . 

Essentially the strategy is handed a project with a `go.mod` to investigate and *then* it makes the decision of old vs. new tactic.

jssblck (changes requested):

src/App/Fossa/Config/ListTargets.hs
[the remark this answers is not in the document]
csasarak: Hi Jess, I went ahead and created an enum for this.

src/App/Fossa/Config/Analyze.hs
```diff
@@ -287,6 +289,13 @@ cliParser =
     <*> many skipVSIGraphResolutionOpt
     <*> monorepoOpts
     <*> baseDirArg
+    <*> experimentalPackageCentricGoAnalysis
+
+experimentalPackageCentricGoAnalysis :: Parser Bool
+experimentalPackageCentricGoAnalysis =
+  switch $
+    long "experimental-package-centric-go"
```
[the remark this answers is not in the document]
csasarak: Thanks, I was having trouble coming up with a good name. I've gone with `--experimental-use-v3-go-resolver`.

src/Strategy/Go/GoListPackages.hs
```diff
@@ -0,0 +1,237 @@
...
+    ignoredPackages :: HashSet.HashSet ImportPath
+    ignoredPackages = HashSet.insert (ImportPath "C") stdLibImportPaths
+
+    removeIgnoredPackages :: GoPackage -> GoPackage
+    removeIgnoredPackages g@GoPackage{packageDeps = pDeps} =
+      g{packageDeps = filter (\i -> not $ i `HashSet.member` ignoredPackages) pDeps}
+
+    graphEdges :: (Has Diagnostics sig m, Has (Grapher Dependency) sig m) => GoPackage -> m ()
+    graphEdges GoPackage{importPath, packageDeps, listError} = do
+      maybe (pure ()) warn listError
```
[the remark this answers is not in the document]
csasarak: That is easier to read so I've incorporated it. Thanks!

src/Strategy/Gomodules.hs
```diff
@@ -85,7 +89,11 @@ getDeps project = do
...
-          -- In reality, this is highly unlikely scenario, and should almost never happen.
-          <||> context "analysis using go list" (GoList.analyze' (gomodulesDir project))
+        if usePackageCentricGoList
+          then context "analysis using go list (packages)" (GoListPackages.analyze (gomodulesDir project))
+          else
+            context "analysis using go mod graph" (GoModGraph.analyze (gomodulesDir project))
+              -- Go List tactic is only kept in consideration, in event go mod graph fails.
+              -- In reality, this is highly unlikely scenario, and should almost never happen.
+              -- This tactic uses `go list -m -json all`
+              <||> context "analysis using go list (modules)" (GoList.analyze' (gomodulesDir project))
```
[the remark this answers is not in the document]
csasarak: I've changed this to fall back to the current dynamic tactic if the experimental one fails.

jssblck (approved):
Nice work!

src/Strategy/Gomodules.hs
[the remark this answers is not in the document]
zlav: I don't know if I agree with this. I think the `experimental` flag should resemble the future state as closely as possible, and in that state, we're going to remove `go list`. If there is a scenario where the module method fails and then the package's strategy succeeds, we likely will never hear about it.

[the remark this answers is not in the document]
csasarak: I think in the future we'd have to talk about which tactics we want to use as fallbacks and which we want to keep. The module based tactic also produces wrong results and uses `go list -m` internally.

 I'm not sure I follow your last sentence. In the future, if the package based (what I'm implementing in this PR) succeeds then we wouldn't even try the module method (`go mod graph`), right?

PR #1174, opened 2023-04-11
Try gradle_7

csasarak:
# Overview

I noticed integration tests were failing for Gradle in my Go PR. Strangely they seemed to not work on master locally, or when done in CI on a commit that worked previously. When I tried running them locally the error I saw was about using a jdk that didn't support java 17. I never figured out what changed to break the tests, but when I change the test to use a version of [gradle in nix](https://github.com/NixOS/nixpkgs/blob/nixos-22.11/pkgs/development/tools/build-managers/gradle/default.nix#L129) that uses jdk17 it [appears to work](https://github.com/fossas/fossa-cli/tree/fix-gradle-integration-test).

## Acceptance criteria

* Both gradle integration tests should pass.

## Testing plan

Look at the github actions for this PR.

## Risks

I don't know why this test suddenly started failing around a week ago. My guess is something in the nix package for gradle changed, but I haven't found any changes in Nix that coincide with that time-frame. #1168 Does change the fixture utils in a trivial way, but the test succeeded for that PR and on master. The currently failing master is from when I tried running integration tests again to see if a previously successful revision would fail.

## Checklist

### DIFF
--- a/integration-test/Analysis/GradleSpec.hs
+++ b/integration-test/Analysis/GradleSpec.hs
@@ -18,7 +18,7 @@ import Test.Hspec (Spec, aroundAll, describe, it, shouldBe)
 import Types (DiscoveredProjectType (..))
 
 gradleEnv :: FixtureEnvironment
-gradleEnv = NixEnv ["gradle"]
+gradleEnv = NixEnv ["gradle_7"]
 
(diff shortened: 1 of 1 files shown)


spatten (approved):
That seems safe and sane to me! Ship it :)

PR #1178, opened 2023-04-14
Origin paths support for RPM spec file analysis 

zlav:
# Overview

Collect origin path information for RPM spec file analysis

## Acceptance criteria

- Spec files will upload origin path information that will show users in the UI which file an RPM came from.

## Testing plan

- Compile the CLI locally with this branch.
- Analyze a directory that has 2 spec files that have different RPMs in them.
- See in the UI that the origin paths are present.

## Risks

I changed RPM analysis, so I suppose I could have broken that for some people, but I'm not entirely sure how.

## References

[ANE-483](https://fossa.atlassian.net/browse/ANE-483)

## Checklist

[ANE-483]: https://fossa.atlassian.net/browse/ANE-483?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/RPM.hs
+++ b/src/Strategy/RPM.hs
@@ -13,7 +13,6 @@ module Strategy.RPM (
 
 import App.Fossa.Analyze.Types (AnalyzeProject (analyzeProject'), analyzeProject)
 import Control.Effect.Diagnostics (Diagnostics, Has, context)
-import Control.Effect.Diagnostics qualified as Diag
 import Control.Effect.Reader (Reader)
 import Data.Aeson (ToJSON)
 import Data.List (isSuffixOf)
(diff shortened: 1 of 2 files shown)


src/Strategy/RPM.hs
```diff
@@ -77,11 +76,11 @@ findProjects = walkWithFilters' $ \dir _ files -> do
 
   case specs of
     [] -> pure ([], WalkContinue)
-    _ -> pure ([RpmProject dir specs], WalkContinue)
+    _ -> pure (RpmProject dir <$> specs, WalkContinue)
 
 data RpmProject = RpmProject
   { rpmDir :: Path Abs Dir
-  , rpmFiles :: [Path Abs File]
+  , rpmFiles :: Path Abs File
```
spatten, line 83: Nit: should we rename this to `rpmFile`, since it's just one file now?

spatten (approved):
:+1:

src/Strategy/RPM.hs
zlav: Yes, good catch

saramaebee:
This was the one I'd looked into most closely. Just to make sure I'm following what you did here, instead of having an RPM project have multiple specfiles, you instead are having each specfile count as its own RPM project, and that's what's giving us granularity over which manifest file accounts for each specific dependency, right?

PR #1183, opened 2023-04-20
Update ci actions

csasarak:
# Overview

Update some of our CI actions.

1. Update rust-cache because  it uses `save-state` which will be [deprecated](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) at the end of May. 
2. Move the rust-caching in `build-all.yml` until after the rust tool chain has been installed.
3. Upgrade to `actions/cache@v3`. I initially did this to port `dist-newstyle` caching from hubble which relies on v3 features. I'm shelving that caching for now, but figured we could keep this part.

## Acceptance criteria

* Builds are successful and use the caches they before
* The build-all action correctly uses cached rust code and doesn't throw an error.

## Testing plan

I pushed commits and looked at CI output. If you look at the build logs for the latest commit you should see a rust cache being downloaded and extracted with no warnings. View the errors/warnings from our earlier CI files [here](https://github.com/fossas/fossa-cli/actions/runs/4736825815/jobs/8408861084).

## Risks

None

## Checklist

### DIFF
--- a/.github/workflows/badges.yml
+++ b/.github/workflows/badges.yml
@@ -1,0 +1,19 @@
+name: Update badges
+on:
+  schedule:
+    - cron: "0 0 * * 0" # At 00:00 every Sunday
+  workflow_dispatch:
+
+jobs:
+  downloads:
+    name: update-downloads-badge
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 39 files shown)


zlav (approved):
Looks simple enough to me! CI passing is a good validation that this works as well.

PR #1184, opened 2023-04-21
Do not fall back to module based analysis with new Go strategy.

csasarak:
# Overview

To make the experimental go analysis behave more like it will when it is GA'ed, do not fall back to `go mod graph` based analysis on failure.

## Acceptance criteria

* It should not fall back to module based analysis.

## Testing plan

I had to make the experimental code-path fail unconditionally with `fatal` and verified that it fell back to static analysis when running `fossa analyze --experimental-use-v3-go-resolver`. I then took the `fatal` call away and ran again and verified that the new strategy ran. 

I still need to add an integration test for Go. I think that would be a better context for testing this in an automated fashion.

## Checklist


spatten (approved):
:+1:

PR #1186, opened 2023-05-04
Add debugging docs

jssblck:
# Overview

An internal discussion on tooling ([Slack](https://teamfossa.slack.com/archives/C0155DTGWB1/p1683231147378089)) prompted this:
a reference for tooling to use and an overview of how to debug FOSSA CLI.

Rendered preview: https://github.com/fossas/fossa-cli/blob/debugging-fossa-cli/docs/references/debugging/README.md

## Acceptance criteria

This document hopefully helps internal folks read debug bundles;
ideally maybe it can help with other users too!

## Testing plan

Docs only, no testing plan.

## Risks

Not really risky.

## References

None.

## Checklist

- [n/a] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [n/a] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [n/a] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

### DIFF
--- a/docs/walkthroughs/debugging-your-integration.md
+++ b/docs/walkthroughs/debugging-your-integration.md
@@ -1,16 +1,22 @@
 ### `fossa analyze`: Why wasn't my project found?
 
-If your project wasn't found, make sure you meet the requirements in the [relevant language/build-tool's quick reference](../references/strategies/README.md)
+If your project wasn't found, make sure you meet the requirements in the
+[relevant language/build-tool's quick reference](../references/strategies/README.md)
 
-If your project meets the requirements, it's very likely `fossa analyze` found your project, but dependency analysis failed. To show the failure reason, use the `--debug` flag:
+If your project meets the requirements, it's very likely `fossa analyze` found your project,
+but dependency analysis failed. To show the failure reason, use the `--debug` flag:
 
 ```sh
 fossa analyze --debug
 ```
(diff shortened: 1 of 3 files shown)


docs/walkthroughs/debugging-your-integration.md
```diff
@@ -1,16 +1,22 @@
...
+If your project meets the requirements, it's very likely `fossa analyze` found your project,
+but dependency analysis failed. To show the failure reason, use the `--debug` flag:
 
 ```sh
 fossa analyze --debug
 ```
 
-Note that the output is likely to be very noisy: our project discovery process is very lenient, and can produce many false-positives. False-positive projects almost always fail during the dependency analysis step, so we don't show analysis failures by default.
+Note that the output is likely to be very noisy: our project discovery process is very lenient,
+nd can produce many false-positives.
```
spatten, line 14: missing an "a" on "and"
```suggestion
and can produce many false-positives. 
```

spatten (approved):
This is super helpful! 👍

PR #1188, opened 2023-05-05
Include analysis output in debug bundle

jssblck:
# Overview

Analysis output (the equivalent of running with `-o`) is now included in FOSSA CLI debug bundles, at the `bundleOutput` top level field.

## Acceptance criteria

- We no longer have to guess at analysis output, or ask users for the result of `fossa analyze -o` when we need to know the output dependency graph.
- We no longer need to go to Core to fetch the output for a build if we have the debug bundle available.

## Testing plan

I tested this locally.
We currently don't seem to have tests for debug bundle output, and I think adding them is out of scope for me at the moment. Also I think this change is very non-risky as it's very strongly type defined.

```
; cabal run fossa -- analyze ~/projects/broker -o --debug
; gunzip --to-stdout fossa.debug.json.gz | jq '.bundleOutput.sourceUnits | map(.Manifest)'
[
  "/home/jess/projects/broker/"
]

; cabal run fossa -- analyze --only-target cargo --only-target cabal --debug -o
; gunzip --to-stdout fossa.debug.json.gz | jq '.bundleOutput.sourceUnits | map(.Manifest)'
[
  "/home/jess/projects/fossa-cli/",
  "/home/jess/projects/fossa-cli/"
]

; mkdir empty
; cabal run fossa -- analyze empty --debug -o
; gunzip --to-stdout fossa.debug.json.gz | jq '.bundleOutput'                             
"scan was not successful, no output"

; cabal run fossa -- container analyze postgres:11.6 --debug -o
; gunzip --to-stdout fossa.debug.json.gz | jq '.bundleOutput.image.os'
"debian"
```

## Risks

I don't think this is risky at all.

## References

This is something we've wanted to do for a while, I was prompted [here](https://teamfossa.slack.com/archives/C0155DTGWB1/p1683316539017289?thread_ts=1683313729.875339&cid=C0155DTGWB1), and I thought it'd be easy. It was, so here goes!

## Checklist

Note: I plan to update https://github.com/fossas/fossa-cli/pull/1186 prior to merging it with these changes.

- [n/a] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+- Debug bundle: The raw dependency graph FOSSA CLI discovers is output in the FOSSA Debug Bundle. ([#1188](https://github.com/fossas/fossa-cli/pull/1188))
+
 <!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
 ## v3.7.9
 - License Scanning: Add support for "full file uploads" for CLI-side license scans. ([#1181](https://github.com/fossas/fossa-cli/pull/1181))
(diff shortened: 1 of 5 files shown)


csasarak (approved):
This is going to simplify our debug work a lot, thanks for doing it!

PR #1189, opened 2023-05-13
fix: Fix badge count

elldritch:
## Background

The shields.io downloads badge seems to be undercounting our actual download count. The badge shows 1.1M downloads:

But if you actually use the API to calculate all downloads of all releases, we actually have 5.8M downloads. I suspect this is because Shields is being rate-limited by the GitHub API, and they might not be correctly paginating to view all of our releases.

## This PR

This PR fixes the download count issue by:

1. Using a manually set badge for downloads, instead of using the native Shields GitHub badge.
2. Updating that badge's contents on a weekly basis using a GitHub Actions cron job.

This PR also refactors the badges to have uniform styles.

### DIFF
--- a/.github/workflows/badges.yml
+++ b/.github/workflows/badges.yml
@@ -1,0 +1,17 @@
+name: Update badges
+on:
+  schedule:
+    - cron: "0 0 * * 0" # At 00:00 every Sunday
+  workflow_dispatch:
+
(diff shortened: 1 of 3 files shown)


elldritch:
Force merging this because it's a low-risk change.

PR #1192, opened 2023-05-16
[No Ticket] Early error if invalid version or empty string in `fossa-deps` (1/2)

meghfossa:
# Overview

Ran into triage session over some invalid chars in s3 upload for version attribute in vendor dep, but found that similar issue exist for other attributes. 

This PR, 

- Throws an error early, if version field of vendor dependency in`fossa-deps` contains known incompatible character (ANE-980). This should be permanent fix - adding this as stopgap for now.
- Throws an error early, if required field in `custom-dependencies` in fossa-deps has empty `name`, such as " ", or ""
- Throws an error early, if required field in `custom-dependencies` in fossa-deps has empty `version`, such as " ", or ""
- Throws an error early, if required field in `custom-dependencies` in fossa-deps has empty `license`, such as " ", or ""
- Throws an error early, if required field in `remote-dependencies` in fossa-deps has empty `name`, such as " ", or ""
- Throws an error early, if required field in `remote-dependencies` in fossa-deps has empty `url`, such as " ", or ""
- Throws an error early, if required field in `remote-dependencies` in fossa-deps has empty `version`, such as " ", or ""
- Throws an error early, if required field in `vendored-dependencies` in fossa-deps has empty `name`, such as " ", or ""
- Throws an error early, if required field in `vendored-dependencies` in fossa-deps has empty `path`, such as " ", or ""
- Throws an error early, if required field in `reference-dependencies` in fossa-deps has empty `name`, such as " ", or ""

## Acceptance criteria

- Throws an error early if value of fossa-deps are incompatible (required field of object, are empty or only contain whitespace, rather than waiting for harder diagnose network error

## Testing plan

- `make install-dev`
- Add valid `fossa-deps.yml`, and perform `fossa-dev analyze -o`, it should succeed
- Modify `fossa-deps.yml` making it invalid (only whitespace entry on required field, etc.), and perform `fossa-dev analyze -o`, it should fail

Example of valid `fossa-deps.yaml`
```yaml
referenced-dependencies:
  - name: "name"
    type: gem

custom-dependencies:
  - name: "name"
    version: "version"
    license: "license"

remote-dependencies:
  - name: "name"
    version: "version"
    url: "url"

vendored-dependencies:
  - name: "name"
    path: "."

```
I added unit test for each condition.

## Risks

N/A

## References

This is similar to ane-885

## Checklist

### DIFF
--- a/src/Data/Aeson/Extra.hs
+++ b/src/Data/Aeson/Extra.hs
@@ -72,7 +72,7 @@ neText obj key = do
 onlyNonEmpty :: (ToString a) => Key -> a -> Parser a
 onlyNonEmpty key val =
   if (Text.null . Text.strip . toText . toString $ val)
-    then fail $ "expected field '" <> toString key <> "' to be non-empty, but recieved: '" <> toString val <> "'"
+    then fail $ "expected field '" <> toString key <> "' to be non-empty, but received: '" <> toString val <> "'"
     else pure val
 
 -- | Like 'Data.Aeson.encode', but produces @Text@ instead of @ByteString@
--- a/src/Data/Aeson/Extra.hs
+++ b/src/Data/Aeson/Extra.hs
@@ -1,5 +1,6 @@
 module Data.Aeson.Extra (
   forbidMembers,
+  neText,
   TextLike (..),
   encodeJSONToText,
 ) where
@@ -8,10 +9,11 @@ import Control.Applicative ((<|>))
 import Control.Monad (when)
 import Data.Aeson qualified as Aeson
 import Data.Aeson.KeyMap qualified as Object
-import Data.Aeson.Types (FromJSON (parseJSON), Key, Object, Parser, ToJSON)
+import Data.Aeson.Types (FromJSON (parseJSON), Key, Object, Parser, ToJSON, (.:))
 import Data.Foldable (traverse_)
-import Data.String.Conversion (decodeUtf8, toString, toText)
+import Data.String.Conversion (ToString, decodeUtf8, toString, toText)
 import Data.Text (Text)
+import Data.Text qualified as Text (null, strip)
 
 -- | A Text-like field
 --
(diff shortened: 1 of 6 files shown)


src/Data/Aeson/Extra.hs
```diff
@@ -51,6 +59,22 @@ forbidMembers typename names obj = traverse_ (badMember obj) names
...
+-- >   MyDataType <$> obj `neText` "my-data-field"
+neText :: (ToString a, FromJSON a) => Object -> Key -> Parser a
+neText obj key = do
+  (val :: a) <- obj .: key
+  onlyNonEmpty key val
+
+onlyNonEmpty :: (ToString a) => Key -> a -> Parser a
+onlyNonEmpty key val =
+  if (Text.null . Text.strip . toText . toString $ val)
+    then fail $ "expected field '" <> toString key <> "' to be non-empty, but recieved: '" <> toString val <> "'"
```
spatten, line 75: Nit: spelling

```suggestion
    then fail $ "expected field '" <> toString key <> "' to be non-empty, but received: '" <> toString val <> "'"
```

src/Data/Aeson/Extra.hs
```diff
@@ -51,6 +59,22 @@ forbidMembers typename names obj = traverse_ (badMember obj) names
         fail . toString $
           "Invalid field name for " <> typename <> ": " <> toText name
 
+-- | Parses non-empty value. It considers string of value with only
+-- whitespaces to be empty.
+--
+-- >  parseJSON = withObject "MyDataType" $ \obj ->
+-- >   MyDataType <$> obj `neText` "my-data-field"
+neText :: (ToString a, FromJSON a) => Object -> Key -> Parser a
```
spatten, line 67: I love this helper. I bet there are other places we can and should use it

spatten (approved):
This looks great! Thank you for fixing this so nicely

PR #1195, opened 2023-05-18
[ANE-461] Adds setup.cfg support

meghfossa:
# Overview

This PR, 
- Adds support for `setup.cfg` for `setup.py` analysis.

## Acceptance criteria

- `install_requires` attributes from `setup.cfg` are considered when analyzing via setuptools. 

## Testing plan

```bash
git pull origin && git checkout feat/setup-cfg && make install-dev
mkdir sandbox && cd sandbox && git clone https://github.com/Greg-Myers-SB/localstack.git
fossa-dev analyze -o | jq
```

You should see locators for: 
```
    boto3>=1.20
    click>=7.0
    cachetools>=3.1.1,<4.0.0
    dataclasses; python_version < '3.7'
    localstack-client>=1.31
    localstack-ext>=0.14.1
    plux>=1.3.1
    psutil>=5.4.8,<6.0.0
    python-dotenv>=0.19.1
    pyyaml>=5.1
    rich>=10.7.0
    requests>=2.20.0,<2.26
    semver>=2.10
    stevedore>=3.4.0
    typing-extensions; python_version < '3.8'
    tailer>=0.4.1
```

## Risks
Sometimes users have complicated dependency config in python world where, they may have `setup.cfg` && `setup.py`, and `reqs.txt` - in which case we would be duplicating
found deps,  but this is what happens right now if the users have `setup.py` and `reqs.txt`. 

## References

[ANE-461](https://fossa.atlassian.net/browse/ANE-461)

## Checklist

[ANE-461]: https://fossa.atlassian.net/browse/ANE-461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -8,8 +8,6 @@
 ## v3.7.11
 - `fossa-deps.yml`: Adds strict parsing to so that required field with only whitespace strings are prohibited early. Also throws an error, if incompatible character is used in vendor dependency's version field. ([#1192](https://github.com/fossas/fossa-cli/pull/1192))
 
-
-
 ## v3.7.10
 - License Scanning: Fix a bug where the license scanner did not run on MacOS 13 on M1 Macs ([#1193](https://github.com/fossas/fossa-cli/pull/1193))
 - Debug bundle: The raw dependency graph FOSSA CLI discovers is output in the FOSSA Debug Bundle. ([#1188](https://github.com/fossas/fossa-cli/pull/1188))
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,9 +1,14 @@
 # FOSSA CLI Changelog
 
+<!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+
+## v3.7.12
+- `setup.cfg`: Adds support for setup.cfg, in conjuction with `setup.py`. ([#1195](https://github.com/fossas/fossa-cli/pull/1195))
+
 ## v3.7.11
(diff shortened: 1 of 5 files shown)


Changelog.md
```diff
@@ -1,9 +1,14 @@
...
+<!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+
+## v3.7.12
+- `setup.cfg`: Adds support for setup.cfg, in conjuction with `setup.py`. ([#1195](https://github.com/fossas/fossa-cli/pull/1195))
+
 ## v3.7.11
 - `fossa-deps.yml`: Adds strict parsing to so that required field with only whitespace strings are prohibited early. Also throws an error, if incompatible character is used in vendor dependency's version field. ([#1192](https://github.com/fossas/fossa-cli/pull/1192))
 
-<!-- - title: description ([#](https://github.com/fossas/fossa-cli/pull/#)) -->
+
```
spatten, line 11: [smallest of nits] There are too many blank lines here

src/Strategy/Python/SetupPy.hs
```diff
@@ -39,3 +43,43 @@ installRequiresParser = do
 
     symbol :: Text -> Parser Text
     symbol = L.symbol space
+
+-- | Parses install requirements listed in setup.cfg
+-- Docs: https://setuptools.pypa.io/en/latest/userguide/declarative_config.html
```
spatten, line 48: Can you add an example of what an install_requires block looks like?

Also, and this is totally optional, but I initially thought we needed to support the comma delimited format and then I figured out that we don't after reading a bit more deeply. It might be worth pointing out that only the dangling form is supported because there can be commas in the version specs.

spatten (approved):
Looks good!

PR #1202, opened 2023-05-23
Remove monorepo

jssblck:
# Overview

Removes monorepo from FOSSA CLI, except in a couple places where we currently test for whether a project is monorepo before attempting to wait for its build or uploading a new scan.

Also removed:

- Documentation about the `fossa vps` subcommand.
- The `wiggins` embedded binary.
- API calls, modules, and functions that were only related to monorepo/VPS.

## Acceptance criteria

Monorepo no longer exists in FOSSA CLI, and meanwhile everything else still works.
I'm relying on tests and the type system to prove the latter.

## Testing plan

I built the project and was able to run a simple scan, otherwise relying on tests.

## Risks

No users should be depending on this functionality any longer, so this should not be risky.

## References

https://fossa.atlassian.net/browse/ANE-755

## Checklist

- [x] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [x] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [x] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `# Unreleased` section at the top.
- [x] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).

PR #1203, opened 2023-05-24
Ane 873 - support --policy-id

csasarak:
# Overview

This PR makes it so that the CLI can accept a new policy id argument as part of require policy work.

This will likely merge after #1202  because removing wiggins will simplify this PR.

## Acceptance criteria

* Should be able to set a policy id with `--policy-id <num>` or using the `policyId: <id>` directive in `.fossa.yml`.
* Only one of a policy name or policy id can be set. This rule applies to mixed CLI and config file arguments also.

## Testing plan

1. Set the require policy flag for an organization.
2. Scan a project, making sure to add `--project <name>` with a name that has _never_ been used before. A previously deleted project does not behave as a new project on this endpoint. [ANE-462](https://fossa.atlassian.net/browse/ANE-462) is meant to address this. It should fail with an error about a missing policy.
3. Run fossa analyze with the `--policy-id <number>` with the id of a policy. You can get this by navigating to the policy's page in the web UI and looking at the URL. It should succeed.

Run `fossa analyze --policy name --policy-id 1`, it should fail. Likewise if you put both or one of the directives in `.fossa.yml`. 

## References

[ANE-873](https://fossa.atlassian.net/browse/ANE-873)
Core PR to support policy-id: https://github.com/fossas/FOSSA/pull/10171

## Checklist

[ANE-462]: https://fossa.atlassian.net/browse/ANE-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ANE-873]: https://fossa.atlassian.net/browse/ANE-873?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


docs/references/files/fossa-yml.md
```diff
@@ -114,7 +114,10 @@ The name field sets the projects visible name in the FOSSA dashboard. By default
 The name of the team in your FOSSA organization to associate this project with.
 
 #### `project.policy:`
-The name of the policy in your FOSSA organization to associate this project with.
+The name of the policy in your FOSSA organization to associate this project with. Mutually excludes `project.policyId`.
+
+#### `project.policyId:`
+The id of the policy in your FOSSA organization to associate this project with. Mutually excludes `project.policy`.
```
jssblck, line 120: Please document the exact semantics, since these are both marked "mutually excludes".

Reading the code, I believe that if both are provided, `policy` takes precedence.
If this is correct, please clearly document everywhere we document these flags/settings.

I recommend using wording similar to:
```
#### `project.policy:`
The name of the policy in your FOSSA organization with which to associate this project. 

If this option and `project.policyId` are both provided, 
this option takes precedence and causes `project.policyId` to be ignored.

#### `project.policyId:`
The id of the policy in your FOSSA organization with which to associate this project. 

If this option and `project.policy` are both provided, 
`project.policy` takes precedence and causes this option to be ignored.
```

jssblck: I think it'd also be helpful to tell users how to get the ID of a given policy.

If we don't have existing Core-side docs to point them to for this, please consider writing a quick doc in `docs.fossa.com` to link to.

src/App/Fossa/Config/Common.hs
```diff
@@ -138,6 +141,9 @@ data CacheAction
 defaultTimeoutDuration :: Duration
 defaultTimeoutDuration = Minutes 60
 
+readMWithError :: Read a => String -> ReadM a
+readMWithError errMsg = auto <|> readerError errMsg
```
jssblck, line 145: Do we have to provide a custom error message? I thought `auto` would generate one based on the type provided?

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -259,9 +260,17 @@ instance FromJSON ConfigProject where
...
       <*> obj .:? "labels" .!= []
       <*> obj .:? "releaseGroup"
+      <*> obj .:? "policyId"
+    where
+      parsePolicy obj = do
+        pName <- obj .:? "policy"
+        pId <- obj .:? "policyId"
+        case (pName, pId) of
+          (Just _, Just _) -> fail "Only one of 'policy' or 'policyId' can be set at a time."
+          _ -> pure $ (PolicyName <$> pName) <|> (PolicyId <$> pId)
```
jssblck, line 273: I believe this functionality is different than when provided in the CLI. Please unify the functionality such that FOSSA CLI always either chooses one, or rejects if both are provided.

jssblck (changes requested):

src/App/Fossa/Config/Common.hs
csasarak: It does, and the default isn't terrible but I chose this because the default doesn't actually tell you what's expected. Only that you did something wrong:
```
option --policy-id: cannot parse value `hello'
```

jssblck: Oh, gross. Yeah I like your option better.

src/App/Fossa/Config/ConfigFile.hs
csasarak: Can you talk a bit more about the differences you see? As far as I can tell the CLI does the same thing. Note that for `optparse-applicative` `<|>` combines option parsers in a pick-only-one kind of way which I wasn't expecting initially. Not sure if this is what you were referring to.

csasarak: When I run `cabal run fossa -- analyze --policy hello --policy-id 123` I get an error as expected. It's an ugly error, I couldn't see a way to make it nice, but it does fail.

Are you thinking of this [line](https://github.com/fossas/fossa-cli/pull/1203/files#diff-81606c0dad857cde73069f8f95ee8661c45d303c5a2c0e329eedbf3f26439a90R171)?
```haskell
parsePolicyOptions = optional (policy <|> policyId) -- For Parsers '<|>' tries every alternative and fails if they all succeed.
```
`optparse-applicative`s implementation of `<|>` for `Parser`s accepts the first alternative to succeed, unless more than one is included. Docs [here](https://github.com/pcapriotti/optparse-applicative#alternative). 

I may be missing a case though.

jssblck: Oh, I misunderstood how this worked! My mistake, thanks for clarifying. In that case, I think we're good to go.

csasarak: Thanks! I spent an embarrassing amount of time when implementing this to get that behavior because it didn't ever occur to me that the answer would just be to use `<|>`. It's definitely understandable how it'd be confusing.

docs/references/files/fossa-yml.md
csasarak: I've added a note here about how the CLI variants of these values take precedence over config file ones. From our other discussion here I think that `mutually excludes` is the right language otherwise. 

I will look into making a doc tomorrow on where to get a policy id, right now the main way I know of to get them is from the URL of the policy page.

jssblck (approved):

PR #1204, opened 2023-05-25
ANE 969: first party scans licenses count as targets [3/4]

spatten:
# Overview

When we analyze a build, we fail if no targets are found or all targets are filtered out.

If you are running first-party scans, we want to succeed if any licenses are found.

## Acceptance criteria

- If an analysis with first-party scans finds any licenses, it should succeed
- The rest of the logic remains unchanged

## Testing plan

Make a directory and put a single license in it and an couple of other files with no licenses:

```
mkdir ~/tmp/small
cd ~/tmp/small
cp ~/fossa/themis-license-data/upstream-licenses/mit.LICENSE LICENSE
echo "hello" > hello.txt
echo "bye" > bye.txt
```

Analyze that with and without the `--experimental-force-first-party-scans` flag. It should fail with the flag off, with an error like "No analysis targets found in directory". It should succeed with the flag

```
make install-dev
fossa-dev analyze --output --debug ~/tmp/small # should fail
fossa-dev analyze --output --experimental-force-first-party-scans ~/tmp/small # should succeed
```

Delete the LICENSE file and scan again. It should now fail with the flag:

```
rm ~/tmp/small/LICENSE
fossa-dev analyze --output --experimental-force-first-party-scans ~/tmp/small # should fail
```

## Checklist


src/App/Fossa/Analyze/Upload.hs
```diff
@@ -100,13 +102,15 @@ uploadSuccessfulAnalysis (BaseDir basedir) metadata jsonOutput revision units li
...
-      Just licenses -> do
+    uploadResult <- case (maybeUnits, maybeLicenseUnits) of
+      (Just units, Nothing) -> uploadAnalysis revision metadata units
+      (mUnits, Just licenses) -> do
         org <- getOrganization
         let fullFileUploads = FullFileUploads $ orgRequiresFullFileUploads org
+        let units = maybe [] NE.toList mUnits
         let mergedUnits = mergeSourceAndLicenseUnits units licenses
         runStickyLogger SevInfo $ uploadAnalysisWithFirstPartyLicensesToS3AndCore revision metadata mergedUnits fullFileUploads
+      (Nothing, Nothing) -> fatalText "Attempted to upload with no results" -- This should never happen, as the caller to this function checks for this before calling
```
csasarak, line 113: Since this comment imposes a responsibility on the caller I think that it should be repeated in a comment docstring.

Alternatively, I think this type forbids what you're talking about and makes the type checker force the caller to do what's correct if you accept it as an argument to the function:
```haskell
data ScanUnits =
SourceUnitsOnly (NE.NonEmpty SourceUnit)
| LicenseSourceUnitOnly LicenseSourceUnit
| SourceAndLicenseUnits (NE.NonEmpty SourceUnit) LicenseSourceUnit
```

csasarak (approved):

src/App/Fossa/Analyze/Filter.hs
```diff
@@ -8,34 +8,53 @@ import App.Fossa.Config.Analyze (IncludeAll (..))
...
 import Data.List.NonEmpty (NonEmpty, fromList)
 import Srclib.Converter qualified as Srclib
-import Srclib.Types (SourceUnit)
+import Srclib.Types (LicenseSourceUnit (licenseSourceUnitLicenseUnits), LicenseUnit (licenseUnitName), SourceUnit)
 
 data CountedResult
   = NoneDiscovered
   | FilteredAll
-  | FoundSome (NonEmpty SourceUnit)
+  | FoundDependenciesOnly (NonEmpty SourceUnit)
```
csasarak, line 16: The final three constructors here are isomorphic to `ScanUnits`. Is there a reason to separate them out this way rather than just
```haskell
data CountedResult
= NoneDiscovered
   | FilteredAll
   | CountedScanUnits ScanUnits
```

spatten: You're right, that's much nicer. I'll make that change too

Thank you!

PR #1205, opened 2023-05-25
ANE 966: first party scan docs [4/4]

spatten:
# Overview

Add the documentation and the Changelog entry for first-party scans.

Here is the rendered doc: https://github.com/fossas/fossa-cli/blob/c972065a830f009a5dd0dd7bf7bdde84893c6917/docs/features/first-party-license-scans.md

## Acceptance criteria

- The documentation is sufficient

## Testing plan

Doc-only

## Checklist


jssblck (approved):
This looks good!

[optional] If possible, it'd be nice if we could elevate "why would I do this" in the documentation (maybe with its own header) so that someone skimming can quickly see a reason to use/not use this feature.

PR #1208, opened 2023-05-30
fix failing Gradle integration test

csasarak:
# Overview

Spring boot was using a plugin that Gradle couldn't find anymore. I've upgraded the version of springboot to a version that doesn't have that problem in its settings.Gradle.

## Acceptance criteria

Integration tests should pass now.

## Testing plan

I verified that integration tests completed successfully. 

## Checklist


spatten (approved):
Thank you!

PR #1212, opened 2023-06-01
Add metrics tracking to the PR template

zlav:
# Overview

We've talked a lot about tracking metrics as a team. This is an easy way to get people in the habit of thinking about it.

## Risks

We think too much about useless metrics

## Checklist


jssblck (approved):
:shipit:

jssblck (approved):
:shipit:

PR #1214, opened 2023-06-03
[ANE-610] Adds PDM Support

meghfossa:
# Overview

This PR, adds initial PDM support. 

It does not support, 
- local dependency (these are path dependencies)
- non-git version control sources

## Acceptance criteria

- As a user, I can analyze PDM project when I have `pyproject.toml` and `pdm.lock`
- As a user, I can analyze PDM project when I only have `pyproject.toml`

### Prereq

```bash
git checkout master && git pull origin && git checkout feat/adds-pdm && make install-dev
```

### PDM (with lockfile)

1. Create pdm project
2. Perform `pdm lock` (to create lockfile)
3. Perform `fossa-dev analyze`

Or clone example pdm project: https://github.com/pdm-project/pdm

```yaml
# .fossa.yml
version: 3
paths:
  exclude:
    - ./tests/

```

### PDM (without lockfile)

1. Create pdm project (ensure there is no lockfile)
2. Perform `fossa-dev analyze`

Or clone example pdm project: https://github.com/pdm-project/pdm, and `rm pdm.lock`.

```yaml
# .fossa.yml
version: 3
paths:
  exclude:
    - ./tests/

```

## Risks

N/A

## Metrics

N/A

## References

https://fossa.atlassian.net/browse/ANE-610

## Checklist


docs/references/strategies/languages/python/pdm.md
```diff
@@ -0,0 +1,252 @@
...
+version: 3
+
+paths:
+  exclude:
+    - ./myexample.egg-info
+```
+
+### F.A.Q
+
+1. How do I perform analysis for `pdm`?
```
meghfossa, line 222: /s/ How do I perform analysis for only `pdm`?

src/Strategy/Python/PDM/Pdm.hs
```diff
@@ -0,0 +1,95 @@
+module Strategy.Python.PDM.Pdm (
```
csasarak, line 1: Could you add a short comment describing the relationship of this module to `Strategy.PDM`? The file structure suggests that PDM is a strategy in its own right, but this file's placement suggests it's actually part of the Python strategy. 

Another possibility could be to move this to `Strategy/PDM` if `PDM` should be considered its own strategy distinct from the `Python` strategy.

csasarak (changes requested):
I'm only requesting changes to get a bit more clarity on how the strategies are organized here. Otherwise, this looks good to me.

src/Strategy/Python/PDM/Pdm.hs
meghfossa: Aha - I was following existing convention i.e. poetry (under python), gopkg (under golang), cabal/stack etc. Strategies are coupled with package manager, as opposed to languages. Let me know if you feel strongly about this - happy to change the layout/structure.

src/Srclib/Converter.hs
```diff
@@ -155,6 +158,7 @@ depTypeToFetcher = \case
   UserType -> "user"
   PubType -> "pub"
   SwiftType -> "swift"
+  PathType -> "url"
```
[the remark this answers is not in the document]
csasarak: I'm interested in hearing your thoughts about this as we have upcoming tickets (ANE-383/ANE-893) which are both focused on Go path deps. We were hoping to use them as a model for how to do it for other path deps. In particular a challenge we have to solve is how to present path deps in the UI in a way that isn't confusing or misleading.

csasarak (approved):

src/Srclib/Converter.hs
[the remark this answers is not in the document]
meghfossa: I wasn't aware re ANE-383/ANE-893. I will bring this up in our slack channel. If ANE-383 is already slotted, I will defer some of the conan work for later, so we can have consistent model.

src/Strategy/Python/PDM/Pdm.hs
meghfossa: addressed in https://github.com/fossas/fossa-cli/pull/1214/commits/fb13068ee19125441037fc138e26751d3b8073b3.

PR #1216, opened 2023-06-06
[ANE-1018] Fails for `remote-dep` when it goes above length of 255

meghfossa:
# Overview

This PR, 
- Adds remote dep constraint on character length (only in upload mode)

## Acceptance criteria

- Fatally fails, when `remote-dep`'s length is greater than allowed length.

## Testing plan

**Prereq:**

```bash
git checkout master && git pull origin && git checkout fix/limit-chars-remote-dep && make install-dev
```

**Test Happy Path:**
```yaml
# > cat fossa-deps.yml
remote-dependencies:
  - name: fossa-cli
    url: https://github.com/fossas/fossa-cli/archive/refs/heads/master.zip
    version: "latest"
```

Run: `fossa analyze -p ane1018 -r happy --fossa-api-key some-key` (should succeed)

**Test not so Happy Path:**
```yaml
# > cat fossa-deps.yml
remote-dependencies:
  - name: fossa-cli
    url: https://github.com/fossas/fossa-cli/archive/refs/heads/master.zip
    version: "latest"

  - name: long-dep
    url: https://ohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurlohsuchalongurl.com/asset.zip
    version: "0.0.1"
```
Run: `fossa analyze -p ane1018 -r not-happy --fossa-api-key some-key` (should fail for `fossa-deps.yml`)

ane1018-ex

## Risks

We only fail when we are not in `--output` mode, since we don't know the organization Id at output only mode.

Endpoint Code:
- https://github.com/fossas/FOSSA/blob/develop/modules/BuildMetadataManager/additionalDependencyData.ts#L72
- https://github.com/fossas/FOSSA/blob/develop/modules/UrlPrivateDependencyManager.ts#L67

Long term, we should have have more holistic validation approach, ideally done at endpoint for things that requires some interaction w endpoint. 

## Metrics

N/A

## References

https://fossa.atlassian.net/browse/ANE-1018

## Checklist

### DIFF
--- a/src/App/Fossa/ManualDeps.hs
+++ b/src/App/Fossa/ManualDeps.hs
@@ -31,7 +31,7 @@ import App.Fossa.VendoredDependency (
  )
 import App.Types (FullFileUploads (..))
 import Control.Carrier.FossaApiClient (runFossaApiClient)
-import Control.Effect.Diagnostics (Diagnostics, context, fatalText)
+import Control.Effect.Diagnostics (Diagnostics, context, fatalText, fatal)
 import Control.Effect.FossaApiClient (FossaApiClient, getOrganization)
 import Control.Effect.Lift (Has, Lift)
 import Control.Effect.StickyLogger (StickyLogger)
@@ -53,8 +53,9 @@ import Data.String.Conversion (toString, toText)
 import Data.Text (Text, toLower)
 import Data.Text qualified as Text
 import DepTypes (DepType (..))
+import Diag.Diagnostic (ToDiagnostic (renderDiagnostic))
 import Effect.Exec (Exec)
-import Effect.Logger (Logger)
+import Effect.Logger (Logger, vsep, pretty, indent)
 import Effect.ReadFS (ReadFS, doesFileExist, readContentsJson, readContentsYaml)
 import Fossa.API.Types (ApiOpts, Organization (..))
 import Path (Abs, Dir, File, Path, mkRelFile, (</>))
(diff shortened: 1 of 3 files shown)


jssblck:
I'm fine with merging this if the backend fix isn't going to be swift, but we [_have_ decided](https://teamfossa.slack.com/archives/CTN0VM59S/p1686244609339709?thread_ts=1686241993.895809&cid=CTN0VM59S) to fix this in the backend by simply removing the limit for locator length. 

Given that, can we add documentation in the changelog to say something like "this is temporary until the FOSSA backend supports longer locators" so that we don't give customers whiplash?

src/App/Fossa/ManualDeps.hs
```diff
@@ -148,9 +149,16 @@ toSourceUnit root depsFile manualDeps@ManualDependencies{..} maybeApiOpts vendor
     -- Don't do anything if there are no vendored deps.
     (_, Nothing) -> pure []
 
+  -- Some manual deps, such as remote dependencies in source unit cannot be
+  -- validated without endpoint interactions.
+  rdeps <- case maybeApiOpts of
+    Just apiOpts -> runFossaApiClient apiOpts $ do
+      traverse (\r -> validateRemoteDep r =<< getOrganization) remoteDependencies
```
jssblck, line 156: 🤔 I'm actually not sure about this, but does this result in `getOrganization` being called for every element in the list during `traverse`? If it wasn't monadic I think the answer would be no, but since it is I think it does?

If so please move the `getOrganization` outside of `traverse`.

jssblck (approved):
Please address comment before merging!

PR #1218, opened 2023-06-09
Fix maven pom parent child relationships

csasarak:
# Overview

When scanning a directory with a package (A) that declares another package (B) to be its parent, if the parent (B) is also detected in the POM closure code we wouldn't report (A) as a target because it is no longer a root package - it's a child POM of (B). 

This is confusing to users because it's possible for the top-level maven package in a directory to not be found as a target if its parent is included in a local subdirectory. This was discovered in a project where the `.m2` directory was a subdirectory of the scanned directory. If `.m2` weren't present, the top-level would be a target (`maven@./`) but if it were present it would not be.

This code changes the CLI so what we only establish that parent-child relationship if (A) has (B) as a parent and (B) also declares (A) as a submodule. The consequences of this for existing projects are that we may list more targets for analysis than we did before in the case described above.

## Acceptance criteria

We should consider a maven package a target even if its parent also exists in the scanned directory unless the parent declares that package as a submodule.

## Testing plan

I have attached a [minimal example](https://github.com/fossas/fossa-cli/files/11726771/minimal-example-with-parent.tar.gz) which demonstrates the problem. The example uses `.m2` as the directory, but it could be named anything or located anywhere in the project. 

You can also create an example by finding a [minimal spring boot project](https://github.com/javabycode/spring-boot-maven-example-helloworld.git), running `mvn compile` to make sure it has all deps and then copying `~/.m2` into the project root.

## Risks

The only risk I can think of is that some customers may see the number of scanned targets increased, though the conditions that this bug appeared in are probably uncommon. 

## Metrics

None

## References

- _[ANE-1007](https://fossa.atlassian.net/browse/ANE-1007)

## Checklist

[ANE-1007]: https://fossa.atlassian.net/browse/ANE-1007?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


spatten (approved):
This looks good to me!

PR #1219, opened 2023-06-09
Fix windows build in CI

csasarak:
# Overview

We were getting an error on installing cabal in some CI jobs. I noticed that we are on an older version of the action. This PR upgrades it and fixes the problem. Example [failing job](https://github.com/fossas/fossa-cli/actions/runs/5225443835/jobs/9434910055#step:4:108). 

## Acceptance criteria

CI should for windows should get past the "install cabal" step. 

## Testing plan

I waited to see if CI would get past the "install cabal" step. 

## Checklist


jssblck (approved):
✨

PR #1223, opened 2023-06-14
Revert "Fix maven pom parent child relationships (#1218)"

csasarak:
This reverts commit 0cca671492d37d3b74491adad394e1ad84b390e8. The changelog has only been appended to.

# Overview

Maven pom discovery reversion.

## Checklist


zlav (approved):
Good call 👍 Let's hold off and test some more repos, this is a good win for our integration testing.

PR #1228, opened 2023-06-15
Don't rebuild if we only change the changelog, readme, docs, etc.

csasarak:
# Overview

It's frustrating to have to wait for a full build when only changing the changelog, readme, etc. 

## Acceptance criteria

* Pushing changes to the ignored files in a branch should not trigger a rebuild of everything

## Testing plan

TBD

## Risks

* [Chris to investigate] Are there risks of not having the right version/hash output in our binaries?
* [Chris to investigate] Does pushing a tag still always trigger a build? If so I think that prevents the problems I talked about in the first question.

## Metrics

None

## References

No ticket.

## Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,5 +1,10 @@
 name: Benchmarks
-on: [push]
+on:
+  push:
+    paths-ignore:
+      - Changelog.md
+      - README.md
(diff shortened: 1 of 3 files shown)


csasarak:
This is actually silly-hard on GH so closing.

PR #1229, opened 2023-06-16
[ANE-265] Conan (3/n)

meghfossa:
# Overview

This PR, 

- Adds basic mapper, that  maps `conan` to `archive` dependency.

I expect this to be a throwaway code, since we will change most of this once path dependency work lands.

## Acceptance criteria

Unit test passes

## Testing plan

This PR does not modify any existing strategy or code. I relied on automated testing.

## Risks

N/A

## Metrics

N/A

## References

https://fossa.atlassian.net/browse/ANE-265

## Checklist

### DIFF
--- a/src/Graphing/Enrich.hs
+++ b/src/Graphing/Enrich.hs
@@ -3,7 +3,7 @@ module Graphing.Enrich (
 ) where
 
 import App.Fossa.LicenseScanner (licenseScanSourceUnit)
-import App.Fossa.VendoredDependency (VendoredDependency (..), VendoredDependencyScanMode (SkipPreviouslyScanned))
+import App.Fossa.VendoredDependency (VendoredDependency (..), VendoredDependencyScanMode (SkippingNotSupported))
 import App.Types (FullFileUploads (..))
 import Control.Algebra (Has)
 import Control.Carrier.Lift (Lift)
@@ -36,7 +36,7 @@ import Srclib.Types (Locator (locatorFetcher, locatorProject, locatorRevision))
 -- transformation,
 -- -
 -- * It preserves edges of conan dependencies in the transformation.
--- * It DOES NOT ignores the non-production dependency in transformation.
+-- * TODO: Ignore non-production dependency in transformation
 conanToArchives ::
   ( Has Diagnostics sig m
   , Has (Lift IO) sig m
@@ -63,15 +63,16 @@ conanToArchives rootPath g =
       -- before they can be referenced in graph as archive deps!
       archiveLocators <-
         licenseScanSourceUnit
-          SkipPreviouslyScanned
-          Nothing
+          -- always perform scan as vendor deps's
+          -- manifest are rarely updated with changes in practice
+          SkippingNotSupported
+          Nothing -- Ignore path filters
           (FullFileUploads False)
           rootPath
           vendorDeps
 
-      -- 2. We replace all conan dependencies with
-      -- archive dependency by transforming archive locators to
-      -- archive dependency
+      -- 2. We make archive archive dependencies from locators, and original
+      -- conan dependencies.
       let (failed, archiveDeps) =
             partitionEithers $
               map (locatorToArchiveDep allConanDeps) $
@@ -81,9 +82,9 @@ conanToArchives rootPath g =
         fatal $
           FailedToTransformLocators failed
 
-      -- 3. Now we replace all archive deps with conan deps
-      -- in original graph. If we are unable to find twin of archive dep
-      -- (i.e. conan dep), we fail fatally!
+      -- 3. We replace all conan dependencies with archive dependencies from
+      -- original graph. If we are unable to find twin of archive dep
+      -- (e.g. sourcing conan dep), we fail fatally!
       case fromList <$> archiveToConanDep archiveDeps of
         Left lonelyArchiveDeps -> fatal $ UnableToFindTwinOfArchiveDep lonelyArchiveDeps
         Right registry -> pure $ gmap (\graphDep -> Map.findWithDefault graphDep graphDep registry) g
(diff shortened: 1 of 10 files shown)


src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+  case (null unableToTransformConanDep, transformedVendorDep) of
+    (True, Nothing) -> pure g
+    (False, _) -> fatal $ FailedToTransformConanDependency unableToTransformConanDep
+    (True, Just depsAndVendorDeps) -> do
+      let vendorDeps = NE.map snd depsAndVendorDeps
+
+      -- 1. We upload all vendor deps (transformed from conan deps),
+      -- since conan dependencies have to be scanned
+      -- before they can be referenced in graph as archive deps!
+      archiveLocators <-
```
meghfossa, line 64: Ideally, we do something where `licenseScanSourceUnit` provides `(vendorDep, locator)s` so, even if we do not have version in our dependency, we can confidently perform replacement in the graph. I left it in this implementation, as we always have revision value, and more so unique hash in this revision for conan.

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+import Path (Abs, Dir, Path)
+import Srclib.Converter (fetcherToDepType, toLocator, verConstraintToRevision)
+import Srclib.Types (Locator (locatorFetcher, locatorProject, locatorRevision))
+
+-- | Transform all 'Conan' dependency to 'Archive' dependency.
+-- It performs cli-side license scanning and upload of results. In this,
+-- transformation,
+-- -
+-- * It preserves edges of conan dependencies in the transformation.
+-- * TODO: Ignore non-production dependency in transformation
```
meghfossa, line 39: I will address this in PR which adds analyzer, and discovery for conan.

meghfossa:
@spatten tagged you here since you did all other conan pr reviews

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+      let vendorDeps = NE.map snd depsAndVendorDeps
+
+      -- 1. We upload all vendor deps (transformed from conan deps),
+      -- since conan dependencies have to be scanned
+      -- before they can be referenced in graph as archive deps!
+      archiveLocators <-
+        licenseScanSourceUnit
+          -- always perform scan as vendor deps's
+          -- manifest are rarely updated with changes in practice
+          SkippingNotSupported
```
spatten, line 68: I'm reading this as "in practice we often get code changes without a change in the version of the dependency"? If so, then yeah, this makes total sense.

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+          Nothing -- Ignore path filters
+          (FullFileUploads False)
```
spatten, line 70: I think we should respect the org's settings on this. This is pretty important to some orgs

src/Graphing/Enrich.hs
spatten: If, however, we're using the hash of the contents to get the version, then I think enabling skipping would be a nice performance boost for basically free

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
+module Graphing.Enrich (
```
spatten, line 1: I think this should be called something like `Graphing.EnrichConan`. `Graphing.Enrich` seems too generic, considering that all of these functions are specific to Conan.

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+sequence' :: [Either a b] -> Either [a] [b]
+sequence' r = case partitionEithers r of
+  ([], xs) -> Right xs
+  (xs, _) -> Left xs
+
+toVendorDeps :: [Dependency] -> ([Dependency], [(Dependency, VendoredDependency)])
+toVendorDeps deps = partitionEithers $ map conanDepToVendorDep deps
+
+-- | Trasnforms Conan dependency to Vendor Dependency.
+conanDepToVendorDep :: Dependency -> Either Dependency (Dependency, VendoredDependency)
```
spatten, line 120: I think the behaviour of `conanToVendoredDep`, `locatorToArchiveDep` and `findArchiveTwin` is just complicated enough that we should have either unit tests for them or tests for the unhappy path of `conanToArchives` that shows that they return a `Left` result when we expect them to

spatten (changes requested):
Requesting changes because I think we should support the `FullFileUploads` flag for sure. All other comments are optional

src/Graphing/Enrich.hs
```diff
@@ -0,0 +1,230 @@
...
+    allConanDeps :: [Dependency]
+    allConanDeps = filter isConanDep allDeps
+
+    unableToTransformConanDep :: [Dependency]
+    unableToTransformConanDep = fst $ toVendorDeps allConanDeps
+
+    transformedVendorDep :: Maybe (NonEmpty (Dependency, VendoredDependency))
+    transformedVendorDep = nonEmpty . snd $ toVendorDeps allConanDeps
+
+    archiveToConanDep :: [Dependency] -> Either [Dependency] [(Dependency, Dependency)]
```
csasarak, line 104: [Optional] My understanding here is that the `Left` side of the `Either` is a list of deps with no partner. If that's true I'd recommend making a small newtype to describe the difference:

```haskell
newtype LonelyDeps = LonelyDeps  { unLonelyDeps :: [Dependency] }
  deriving (Eq, Ord)
  deriving newtype (Show) 
```

I think this replacement should only need a few calls to the constructor and `unLonelyDeps` in this module to work.

csasarak (approved):
I agree with Scott on changes, but will approve since he's already requested them. Everything I mention is optional.

spatten (approved):
This all looks great!

PR #1233, opened 2023-06-21
[No Ticket] Log basic HTTP request info in debug bundles.

csasarak:
# Overview

This provides basic support for outputting some tracing info for our HTTP requests. Output is only to the debug bundle. 

This PR itself is pretty minimal in terms of the info it outputs but it makes it much easier to add additional info in the future.

I discovered that several of our subcommands support `--debug` but don't output any debug bundle. They will not be able to benefit from this work. For those I had to add a call to `ignoreDebug` to make the types check. I think I did this correctly but if a reviewer can double-check it would be helpful.

## Acceptance criteria

* It should output the method and URL whenever an HTTP call is made through our api client effect to the debug bundle.
* It should not output this information to stderr/stdout.

## Testing plan

I ran `fossa analyze` and `fossa test`, etc. with the `--debug` flag and checked that it output some information in the debug bundle (where relevant) and didn't output anything to the console.

## Risks

None, in the worst case I think nothing will happen.

## Metrics

None.

## References

None.

## Checklist

### DIFF
--- a/src/App/Fossa/Container/Test.hs
+++ b/src/App/Fossa/Container/Test.hs
@@ -16,6 +16,7 @@ import App.Fossa.Config.Container (
 import App.Fossa.Config.Container qualified as Config
 import App.Fossa.Container.Scan (scanImageNoAnalysis)
 import App.Types (OverrideProject (OverrideProject, overrideBranch, overrideName, overrideRevision), ProjectRevision (..))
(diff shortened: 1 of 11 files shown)


spatten (approved):
This all looks good to me

I tested with a few different commands and flags and everything worked perfectly.

👍

PR #1235, opened 2023-06-29
Ane 956 output rules data

csasarak:
# Overview

This PR makes it so that we fetch and output data from a different endpoint in VSI. It also generates rules from that data and outputs them when `--debug` is used.

I've opted not to list this in the changelog or docs yet because it only affects debug output. 

## Acceptance criteria

* With the `--detect-vendored` and `--debug` flags set the console output outputs a list of rules (`Generated Rules ...`)
* With the `--detect-vendored` and `--debug` flags set the debug bundle has a list of rules (`Generated Rules ...`)

## Testing plan

I made automated tests around the code the parse inferences from the backend and around turning those into a set of rules. 

You can manually test by building this branch and running fossa like this:
`fossa analyze --detect-vendored --debug <project dir>`

I've compared the locators output by the code on this branch with that of the released CLI and they seem to correspond.

## Risks

The new output is not risky. I have also verified that the new endpoint seems to work both against our production SaaS. 
## Metrics

No.

## Checklist

### DIFF
--- a/src/App/Fossa/VSI/Analyze.hs
+++ b/src/App/Fossa/VSI/Analyze.hs
@@ -6,8 +6,7 @@ module App.Fossa.VSI.Analyze (
 ) where
 
 import App.Fossa.VSI.Fingerprint (Combined, fingerprint)
-import App.Fossa.VSI.IAT.Types qualified as IAT
-import App.Fossa.VSI.Types (ScanID (..), VsiRule (..), generateRules)
+import App.Fossa.VSI.Types (ScanID (..), generateRules)
 import App.Fossa.VSI.Types qualified as VSI
 import App.Types (ProjectRevision)
 import App.Util (FileAncestry (..), ancestryDerived, ancestryDirect)
@@ -53,7 +52,7 @@ runVsiAnalysis ::
   Path Abs Dir ->
   ProjectRevision ->
   AllFilters ->
-  m ([VSI.Locator], [IAT.UserDep])
+  m [VSI.VsiRule]
 runVsiAnalysis dir projectRevision filters = context "VSI" $ do
   -- If we try to run with fewer than 2 capabilities, STM will deadlock
   capabilities <- max 2 <$> sendIO getNumCapabilities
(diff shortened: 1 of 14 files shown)


src/App/Fossa/VSI/Analyze.hs
```diff
@@ -78,13 +79,17 @@ runVsiAnalysis dir projectRevision filters = context "VSI" $ do
...
   context "Wait for cloud analysis" $ waitForAnalysis scanID
 
-  discoveredRawLocators <- context "Download analysis results" $ getVsiInferences scanID
-  when (null discoveredRawLocators) $ fatalText "No dependencies discovered with VSI"
+  rules <- context "Download analysis results" $ do
+    inferences <- getVsiInferences scanID
+    let rules = generateRules inferences
+    logDebug . pretty $ "Generated Rules: " <> (decodeUtf8 @Text . encode $ rules)
+    pure rules
+  --    pure . filter (/= "") . VSI.uniqueVsiLocators $ inferences
```
jssblck, line 87:
```suggestion
```

src/App/Fossa/VSI/Types.hs
```diff
@@ -14,16 +15,28 @@ module App.Fossa.VSI.Types (
   isTopLevelProject,
   toDependency,
   parseAnalysisStatus,
+  VsiFilePath (..), -- data constructor only exported for testing
+  VsiRulePath (..), -- data constructor only exported for testing
+  VsiInference (..),
+  -- VsiLocator (..), -- data constructor only exported for testing
```
jssblck, line 21:
```suggestion
```

test/App/Fossa/VSI/TypesSpec.hs
```diff
@@ -0,0 +1,141 @@
...
+  VsiRule (VsiRulePath "/otherProject") (Locator "git" "github.com/otherProject" "2.0.0")
+    : VsiRule (VsiRulePath "/baz") (Locator "git" "github.com/someProject" "1.0.0")
+    : singleRuleExpected
+
+nestedProjectInferences :: [(VsiFilePath, VsiInference)]
+nestedProjectInferences =
+  ( VsiFilePath "/foo/bar/g.c"
+  , VsiInference . Just $ Locator "git" "github.com/facebook/follyNested" "1.0.0"
+  )
+    : singleInference
```
jssblck, line 105: Technically this cannot occur with the current design of VSI, but since it's written it doesn't hurt to support this case!

jssblck (approved):
Nice work! Looks awesome 💯

PR #1237, opened 2023-06-30
[Review after ANE-956] Ane 958 src unit origin paths

csasarak:
# Overview

In #1235 I change the CLI to download full rules for VSI. This PR changes the CLI to report each VSI dependency as a different source unit to FOSSA.

On our current releases when running with `--detect-vendored` you'll get output like this:
Screenshot 2023-06-30 at 2 17 01 PM
As you can see, there is just one result for VSI which actually finds multiple deps for this project. 

This PR changes it so we report a new source unit for every dep that VSI finds:
Screenshot 2023-07-03 at 4 19 23 PM
Including one for the project overall (this will not be shown in FOSSA's UI).

In FOSSA, the deps are now reported with their proper origin paths from within the project. This is show on the right hand side of this screenshot:
Screenshot 2023-06-30 at 2 08 48 PM

Clicking on the dep will show the full path:
Screenshot 2023-06-30 at 2 21 52 PM

## Acceptance criteria

* Each dependency is now its own source unit with an origin path showing where in the project it was found (the VSI Rule that found it)
* The CLI also reports each VSI dep independently.

## Testing plan

I manually tested scanning with VSI against staging and prod. I verified that the deps that were found were the same but the UI reflected the change in source unit reporting.

You can test this using the [demo](https://github.com/fossas/cpp-vsi-demo). 

## Risks

I removed some type safety around paths in the SourceUnit's structure. 

## References

- [ANE-958](https://fossa.atlassian.net/browse/ANE-958): Output rules as source units with origin paths

## Checklist

[ANE-958]: https://fossa.atlassian.net/browse/ANE-958?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Analyze/Log4jReport.hs
+++ b/src/App/Fossa/Analyze/Log4jReport.hs
@@ -72,7 +72,7 @@ import Effect.ReadFS (ReadFS)
 import GHC.Generics (Generic)
 import Graphing (directList, getRootsOf, hasPredecessors, vertexList)
 import Options.Applicative (InfoMod, progDesc)
-import Path (Abs, Dir, File, Path, SomeBase, toFilePath)
+import Path (Abs, Dir, Path, toFilePath)
 import Prettyprinter (Doc, Pretty (pretty), annotate, vsep)
 import Prettyprinter.Render.Terminal (AnsiStyle, Color (Yellow), color)
 import Srclib.Converter (verConstraintToRevision)
@@ -236,7 +236,7 @@ data Log4jVulnerableReportItem = Log4jVulnerableReportItem
   deriving (Show, Eq, Ord)
 
 data VulnerableDependencyOrigin
-  = VulnerableDependencyManifestFiles [SomeBase File]
+  = VulnerableDependencyManifestFiles [Text]
   | VulnerableDependencyRootDependency [Dependency]
   deriving (Eq, Ord)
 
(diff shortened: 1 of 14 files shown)


src/App/Fossa/Analyze/Log4jReport.hs
```diff
@@ -236,7 +236,7 @@ data Log4jVulnerableReportItem = Log4jVulnerableReportItem
   deriving (Show, Eq, Ord)
 
 data VulnerableDependencyOrigin
-  = VulnerableDependencyManifestFiles [SomeBase File]
+  = VulnerableDependencyManifestFiles [Text]
```
jssblck, line 239: Consider using `SomePath` instead, which loosens the restrictions as much as possible while still indicating that we want this to mean a "path" of some kind.

Reference: https://github.com/fossas/fossa-cli/blob/2da715e852e15375094e998fc27f161aad757a52/src/Path/Extra.hs#L19-L24

jssblck (commented):
Obviously not a review since in draft, but this approach looks good. If possible it'd be pretty cool to be able to log the locator a given rule resolved to in the summary, but if that's annoying don't worry about it.

src/App/Fossa/Analyze/Log4jReport.hs
csasarak: Thanks, I have made this change.

zlav (approved):
All changes look good to me! Jess's point about the scan summary is addressed and the approach looks solid to me.

PR #1242, opened 2023-07-25
Add `diagnose` CLI

jssblck:
# Overview

Adds a `diagnose` CLI utility.

Usually we try to ensure that diagnostics are captured in the FOSSA CLI debug bundle.
However sometimes we need information that isn't reasonable to collect on every system
(due to performance, privacy, or other reasons).

Sometimes we also want to validate a diagnostic is useful before we include it in the debug bundle.

These diagnostics are intended to be used when directed by FOSSA support, and moved into the mainline debug bundle if they seem broadly useful.

## Acceptance criteria

- The `diagnose` utility is included in builds.
- The utility can be used via `diagnose walk` to diagnose file walk operations.

## Testing plan

I ran it locally and output looks reasonable:
```
; time cargo run --release --bin diagnose -- walk ~/projects/fossa-cli --trace-spans none --trace-level info
    Finished release [optimized] target(s) in 0.04s
     Running `target/release/diagnose walk /Users/jessica/projects/fossa-cli --trace-spans none --trace-level info`
2023-07-25T01:58:52.206153Z  INFO main{root="/Users/jessica/projects/fossa-cli"}: diagnose::cmd::walk: inspecting root: /Users/jessica/projects/fossa-cli
2023-07-25T01:58:52.618362Z  INFO main{root="/Users/jessica/projects/fossa-cli" root_device_id=16777230}: diagnose::cmd::walk: entries inside '/Users/jessica/projects/fossa-cli' inspected: 30033
2023-07-25T01:58:52.618376Z  INFO main{root="/Users/jessica/projects/fossa-cli" root_device_id=16777230}: diagnose::cmd::walk: directories: 5330; files: 24703
2023-07-25T01:58:52.618379Z  INFO main{root="/Users/jessica/projects/fossa-cli" root_device_id=16777230}: diagnose::cmd::walk: deepest level: 23 ('/Users/jessica/projects/fossa-cli/integration-test/artifacts/gradle/sample/spring-boot-3.0.0-M1/ci/images/releasescripts/src/test/resources/io/spring/concourse/releasescripts/sonatype/artifactory-repo/org/springframework/example/module-three/1.0.0/module-three-1.0.0.jar')
cargo run --release --bin diagnose -- walk ~/projects/fossa-cli --trace-spans  0.11s user 0.39s system 97% cpu 0.513 total
```

## Risks

Not particularly risky.

## Metrics

We could add metrics to track usage of this binary, but I don't know that it's worth the time.

## References

Created primarily to help debug this long-running issue: https://fossa.zendesk.com/agent/tickets/6322

## Checklist

I'm not planning to cut a release for this, nor add it to the changelog, since it isn't really a change to FOSSA CLI. If desired we can keep a separate changelog in the `tools/diagnose/` directory.


jssblck:
This utility was useful in debugging the issue it was intended to debug, let's merge it for future use!

tools/diagnose/src/cmd/walk.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[tracing::instrument(skip(args), fields(root_device_id))]
+pub fn main(root: PathBuf, args: &WalkArgs) -> Result<()> {
+    info!("inspecting root: {}", root.display());
+    trace!("inspecting root: {}", root.display());
+    debug!("walk options: {args:?}");
+
+    let root_device_id = util::device_num(&root).wrap_err("get device ID")?;
+    tracing::Span::current().record("root_device_id", root_device_id);
+    trace!("device_id: {root_device_id}");
```
csasarak, line 53: If we find that the root device is different than one of the walked files, should we run something like `mount -v` in order to get filesystem info? I'm not sure what the windows equivalent is.

If you think it's safe, just unconditionally capturing the output of that command for cross-reference may also work. I don't think there should be anything sensitive in it.

csasarak (approved):
LGTM, I really like the idea of having a place to allow collecting ad-hoc diagnostics or diagnostics that might be too sensitive for some customers but not others. So thanks for implementing this!

What do you see the future of this functionality as being? Rolling into the cli as a subcommand, reimplementing in the CLI, or something else?

jssblck:
> What do you see the future of this functionality as being? Rolling into the cli as a subcommand, reimplementing in the CLI, or something else?

I think for the foreseeable future it should stay separate. I also think if we find things in here that are concretely useful to include in debug bundles we should consider either reimplementing in Haskell or embedding+shelling out to this CLI automatically, rather than e.g. exposing as a subcommand.

tools/diagnose/src/cmd/walk.rs
jssblck: Hmm, I'd rather leave this as is for now, since we aren't certain we need this information, but agreed we can add that if it is something we run into that we want for debugging.

jssblck:
Talked to Chris on Slack, we're happy merging in this state.

PR #1255, opened 2023-08-07
Clarify Python setuptools documentation

csasarak:
#Overview

I found some language that implied that getting python deep deps for setuptools is a matter of parsing, when I think that calling it a parsing issue undersells the the actual complexity of getting an accurate list of deps from setup.py.

## Acceptance criteria

* Should describe why we can't currently get deep deps easily for Python with setuptools jusing `setup.py`

## Testing plan

This is purely a doc change.

## Checklist

### DIFF
--- a/docs/references/strategies/languages/python/setuptools.md
+++ b/docs/references/strategies/languages/python/setuptools.md
@@ -48,11 +48,11 @@ from both files.
 synchronization between them.
 * Since we don't actually run `setup.py` or do our own interpretation of the Python code therein, `install_requires` not defined as a literal array of string literals done in the `setup.py` file will hide the
 true `install_requires` list from our view. For example, we don't have a way to find `install_requires` set up this way:
-```python
-a = ['package1==1.0.0']
-b = ['package2==2.0.0']
-install_requires = a + b
-```
+    ```python
+    a = ['package1==1.0.0']
+    b = ['package2==2.0.0']
+    install_requires = a + b
+    ```
 However, we do catch *variables* named `install_requires`, as long as they are declared earlier in the file than the `install_requires` keyword argument to `setup`.
 * Often, the `requirements.txt` file entirely overlaps the `setup.py` file.  This is almost always by design.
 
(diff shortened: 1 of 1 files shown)


docs/references/strategies/languages/python/setuptools.md
```diff
@@ -46,9 +46,14 @@ from both files.
...
-true `install_requires` list from our view.  However, we do catch *variables* named `install_requires` as well, as
-long as they are declared earlier in the file than the `install_requires` keyword argument to `setup`.
+* Since we don't actually run `setup.py` or do our own interpretation of the Python code therein, `install_requires` not defined as a literal array of string literals done in the `setup.py` file will hide the
+true `install_requires` list from our view. For example, we don't have a way to find `install_requires` set up this way:
+```python
+a = ['package1==1.0.0']
+b = ['package2==2.0.0']
+install_requires = a + b
+```
+However, we do catch *variables* named `install_requires`, as long as they are declared earlier in the file than the `install_requires` keyword argument to `setup`.
```
spatten, line 56: Nit: if you indent these lines by three spaces then they get included in the list item and render a bit more nicely

```suggestion
* Since we don't actually run `setup.py` or do our own interpretation of the Python code therein, `install_requires` not defined as a literal array of string literals done in the `setup.py` file will hide the
   true `install_requires` list from our view. For example, we don't have a way to find `install_requires` set up this way:
   ```python
   a = ['package1==1.0.0']
   b = ['package2==2.0.0']
   install_requires = a + b
   ```
   However, we do catch *variables* named `install_requires`, as long as they are declared earlier in the file than the `install_requires` keyword argument to `setup`.
```

spatten (approved):
Nice find! 🚢

docs/references/strategies/languages/python/setuptools.md
spatten: Sorry, my suggested edits didn't quite work because there was a code-block in the suggested edit, but hopefully the idea still makes sense

csasarak: Thanks for catching that! I think I was able to get it to look better and also did some rewording to make sure `install_requires` is introduced before I start giving more details about it.

PR #1259, opened 2023-08-10
Use a newtype for originpaths

csasarak:
# Overview

This change makes it possible for source unit origin paths to have instances of from/to json which are inverses of each other.

## Acceptance criteria

* The CLI has a `SourceUnit` type with From/To Json instances that and produce/consume for each other.

## Testing plan

I tested the CLI output and verified that the outputted paths are paths like we expect. Since this uses the generated From/To json instances I did not think it was worth adding a test for them being inversions of each other.

## Risks

This is less type safe. I think it's clear that the intention is for an OriginPath to be a path, but nothing really enforces it. Particularly when using `FromJSON`. 

One potential option is to make a `FromJSON` instance which tries to parse an origin path as a file, then as a directory and fails if neither of them work, but otherwise just wraps the string it got. Let me know if you think this is worthwhile.

## References

- _[ANE-1159](https://fossa.atlassian.net/browse/ANE-1159)

## Checklist

[ANE-1159]: https://fossa.atlassian.net/browse/ANE-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Srclib/Types.hs
+++ b/src/Srclib/Types.hs
@@ -47,6 +47,13 @@ instance ToText LicenseScanType where
 instance ToJSON LicenseScanType where
   toJSON = toJSON . toText
 
+-- |This type is meant to represent the paths in a project where a particular set of dependencies were discovered.
+-- For example, in a project directory that has a @go.mod@ the OriginPath might be 'foo/bar/go.mod'.
+-- In a project with VSI dependencies the OriginPath would be the directory the dep was found in, such as 'vendored/zlib/'
+--
+-- OriginPaths were previously `SomeBase File`, however with support for VSI OriginPaths can now be a directory in addition to a file path.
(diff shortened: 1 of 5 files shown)


src/Srclib/Types.hs
```diff
@@ -42,6 +47,20 @@ instance ToText LicenseScanType where
 instance ToJSON LicenseScanType where
   toJSON = toJSON . toText
 
+newtype OriginPath = OriginPath FilePath
```
jssblck, line 50: [nit] Please document the intention of this type.

jssblck (approved):

src/Srclib/Types.hs
csasarak: I've added a comment explaining what an `OriginPath` is and why they are that way.

PR #1261, opened 2023-08-11
Ane 862 notarize mac binaries

csasarak:
# Overview

To make it so that customers can run `fossa` on Macs without getting a warning this PR notarizes the binaries. [Notarization](https://developer.apple.com/documentation/security/notarizing_macos_software_before_distribution) involves uploading the executable(s) to Apple and them scanning them. This creates a record in their database that Mac's can fetch on their first run to verify that they've been scanned.

## Acceptance criteria

Users should not get a message like the following when first running `fossa` or any of the other binaries:

Screenshot 2023-08-14 at 11 05 53 AM

## Testing plan

I downloaded versions of the binaries from our build action and tried to run them. I used this build for testing: https://github.com/fossas/fossa-cli/actions/runs/5837966561

When I run `fossa` from that archive I get the following message now:
Screenshot 2023-08-14 at 11 04 07 AM

The current release version will output a message about the binary not having been scanned for malware the first time it is run.

## References

[ANE-862](https://fossa.atlassian.net/browse/ANE-862)
[Slack](https://teamfossa.slack.com/archives/C0155DTGWB1/p1691616686713359?thread_ts=1684275181.104099&cid=C0155DTGWB1)

## Checklist

[ANE-862]: https://fossa.atlassian.net/browse/ANE-862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -175,7 +175,7 @@ jobs:
         codesign --options runtime -s 'Fossa, Inc.' release/diagnose
 
         # Perform notarization
-        zip notarization-archive.zip release
+        zip -rj notarization-archive.zip release
         xcrun notarytool submit notarization-archive.zip --apple-id "$APPLE_NOTARIZATION_DEV_ID" --password "$APPLE_NOTARIZATION_DEV_PASS" --team-id "$APPLE_TEAM_ID" --wait
         rm notarization-archive.zip
 
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -150,6 +150,9 @@ jobs:
         MACOS_BUILD_CERT_BASE64: ${{ secrets.MACOS_BUILD_CERT_BASE64 }}
         MACOS_BUILD_CERT_P12_PASSWORD: ${{ secrets.MACOS_BUILD_CERT_P12_PASSWORD }}
         MACOS_KEYCHAIN_PASSWORD: ${{ secrets.MACOS_KEYCHAIN_PASSWORD }}
+        APPLE_NOTARIZATION_DEV_PASS: ${{ secrets.APPLE_NOTARIZATION_DEV_PASS }}
+        APPLE_NOTARIZATION_DEV_ID: ${{ secrets.APPLE_NOTARIZATION_DEV_ID }}
+        APPLE_TEAM_ID: ${{ secrets.APPLE_TEAM_ID }}
       run: |
         # create variables
         CERTIFICATE_PATH=$RUNNER_TEMP/build_certificate.p12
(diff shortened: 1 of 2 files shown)


.github/workflows/build-all.yml
```diff
@@ -167,9 +170,21 @@ jobs:
...
         security list-keychain -d user -s $KEYCHAIN_PATH
 
-        codesign -s 'Fossa, Inc.' release/fossa
-        codesign -s 'Fossa, Inc.' release/pathfinder
-        codesign -s 'Fossa, Inc.' release/diagnose
+        # '--options runtime' enables the hardened runtime: https://developer.apple.com/documentation/security/hardened_runtime
+        # The hardened runtime is required for notarization.
+        codesign --options runtime -s 'Fossa, Inc.' release/fossa
+        codesign --options runtime -s 'Fossa, Inc.' release/pathfinder
+        codesign --options runtime -s 'Fossa, Inc.' release/diagnose
```
jssblck, line 177: The company name is FOSSA (in all caps).
```suggestion
        codesign --options runtime -s 'FOSSA, Inc.' release/fossa
        codesign --options runtime -s 'FOSSA, Inc.' release/pathfinder
        codesign --options runtime -s 'FOSSA, Inc.' release/diagnose
```

jssblck (approved):
This works:

Screenshot 2023-08-14 at 10 25 41 AM

But does not yield the expected results for a user upon clicking "Open":
Screenshot 2023-08-14 at 10 25 45 AM

Are we missing a `chmod` or something? We should ensure that these binaries open in a terminal. For example, here is what appears when I double click another binary:
Screenshot 2023-08-14 at 10 32 57 AM

csasarak:
@jssblck I've added a `chmod +x` to the signing step, though we typically do that before generating the release-archives archive. 

I've had a look at how to get the double-click behavior to work and I think it may be a bug in gatekeeper that causes that warning. This [thread](https://developer.apple.com/forums/thread/706379) has some information from an Apple staffer under the section "Tool Blocked by Gatekeeper". Unfortunately it looks like just notarizing isn't always enough to make Gatekeeper accept the executable when double-clicked. 

I am interested in how customers were getting this issue though. I found out that the reason we don't see anything from GateKeeper when using `install-latest.sh` is that not every program that downloads a file sets the [quarantine attribute ](https://eclecticlight.co/2020/10/29/quarantine-and-the-quarantine-flag/). cURL it seems does not. 

Executing the program from the command-line the first time it's run should show the window I showed above with the "OK" button. After doing that, double-clicking in finder should work the way you expect. If you feel strongly about being able to double-click to run these executables the first time they're seen, I think the best way will be to try to generate a `.app` or `.dmg` and have the user go through an install step first. 

I'm going to ask the customer if running these programs via the CLI first to get through gatekeeper is a problem for them.

PR #1266, opened 2023-08-29
Ane 967 additional report fmts

csasarak:
# Overview

This adds support in the CLI for some of the formats already supported by Core.

As part of this I tried to add a bit more documentation for the report command and in the process advanced the index-state in order to be able to use a feature in one of the newer versions of `optparse-applicative`.

## Acceptance criteria

* The CLI can fetch attribution reports in every format currently supported by FOSSA. Except PDF.

PDFs aren't included in this because you can't download them directly from FOSSA.

## Testing plan

I added one test which is meant to break if we ever add a format to the format data structure and don't support it fully in our output format parser. 

I then did manual tests by fetching each report format for a project.

## References

- [ANE-957](https://fossa.atlassian.net/browse/ANE-957)

## Checklist

[ANE-957]: https://fossa.atlassian.net/browse/ANE-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## v3.8.10
 
+- Reports: Can now export reports formatted as CycloneDX (json/xml), CSV, HTML, and JSON SPDX. ([#1266](https://github.com/fossas/fossa-cli/pull/1266))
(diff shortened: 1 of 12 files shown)


spatten (approved):
This all looks good to me 👍

PR #1267, opened 2023-08-30
Dep upgrades

csasarak:
# Overview

This PR upgrades a number of our deps. Haskell packages on Hackage use the [PVP](https://pvp.haskell.org/). I've mostly only changed the packages that had a major version change or were not using something like `^>= A.B.C.D`.

I haven't upgraded a few deps that interact with `codec-rpm`. The 9.4 branch has some changes that would facilitate this upgrade but I didn't get a chance to bring them in here and test them. 

## Acceptance criteria

* We have upgraded some of our packages.

## Testing plan

I checked that our unit tests were completing and ran the CLI against several test projects. Everything seemed to be working normally.

I also built hubble against this branch and it built fine. Unit tests completed successfully.

## Risks

I've switched us away from `cryptonite` to `crypton` because cryptonite is [unmaintained](https://github.com/yesodweb/wai/pull/931#issue-1743179602).

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## v3.8.10
 
+- Reports: Can now export reports formatted as CycloneDX (json/xml), CSV, HTML, and JSON SPDX. ([#1266](https://github.com/fossas/fossa-cli/pull/1266))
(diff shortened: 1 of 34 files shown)


spatten (approved):
Good call on testing against Hubble!

This feels pretty safe, seeing as it compiles and the tests pass 👍

PR #1268, opened 2023-08-30
Prevent infinite recursion from Pom file property interpolation

zlav:
# Overview

This address an issue where Maven static analysis hangs do to infinite recursion when pom file analysis attempts to interpolate a dependencies version property from a parent block (a parent pom or `dependencyManagement`). 

The issue arises when there is a block like the following `<junit5.version>${junit5.version}</junit5.version>`. 

Property interpolation works as follows:
- A dependency has the version `${junit5.version}`
- We clean (`splitTheProperty`) the maven property up to give us `junit5.version`
- We find a reference for `junit5.version` in a parent pom or `dependencyManagement` section
- That block looks like `<junit5.version>${junit5.version}</junit5.version>`
- We clean (`splitTheProperty`) the maven property up to give us `junit5.version`
- We find a reference for `junit5.version` in a parent pom or `dependencyManagement` section
- ...

## Acceptance criteria

- The user's project can successfully scan
- Infinite loops are no longer possible in the Pom file analysis

## Testing plan

Method 1
- Build the binary and ensure it can successfully analyze the user project in the Zendesk ticket

Method 2
- Add tests to ensure an infinite loop is handled
- Ensure they pass with the changes made
- Ensure they fail without the changes made

## Risks

I really can't think of any since the only this prevents is an infinite loop.

## References

https://fossa.zendesk.com/agent/tickets/6791

## Checklist


test/Maven/PomStrategySpec.hs
```diff
@@ -26,6 +26,10 @@ spec = do
     it "should interpolate multiple properties" $ do
       interpolateProperties pom "${project.groupId}${project.artifactId}" `shouldBe` "MYGROUPMYARTIFACT"
 
+    it "should not infinitely recurse when interpolating a property that is interpolated to itself" $ do
+      let pom' = pom{pomProperties = Map.singleton "project.groupId" "${project.groupId"}
+      interpolateProperties pom' "${project.groupId}" `shouldBe` "OTHERGROUP"
+
```
spatten, line 32: Two things need to be fixed here:

1) `"${project.groupId"` => `"${project.groupId}"`
2) the expected groupId should be `${project.groupId}`.

This should work:

```suggestion
it "should not infinitely recurse when interpolating a property that is interpolated to itself" $ do
      let pom' = pom{pomProperties = Map.singleton "project.groupId" "${project.groupId}"}
      interpolateProperties pom' "${project.groupId}" `shouldBe` "project.groupId"
```

```
cabal test unit-tests --test-option=--color --test-option="--match=should not infinitely recurse when interpolating a property that is interpolated to itself"
```

spatten (approved):
:+1: after a minor fix to the test

test/Maven/PomStrategySpec.hs
zlav: Didn't know about `--test-option`, that makes it a lot easier to test locally. Thanks!

PR #1270, opened 2023-09-06
Point users to `licenseScanPathFilters` docs

jssblck:
# Overview

Adds a doc link to `licenseScanPathFilters` from `.fossa.yml` docs.


spatten (approved):
👍

PR #1274, opened 2023-09-12
Keyword Search & Custom license Search Documentation

spatten:
# Overview

Documentation for the Keyword Search and Custom License Search features.

This will be the last PR merged for this set of features, so I included the change to the changelog as well.

You can see the rendered files here: https://github.com/fossas/fossa-cli/blob/55c7ed4129b9fc44e7ae7eb0b75381ad25f0f03f/docs/features/custom-license-and-keyword-searches.md

## Acceptance criteria

The Keyword Search and Custom License Search features are fully documented

## Checklist


docs/features/custom-license-and-keyword-search-regular-expression-syntax.md
```diff
@@ -0,0 +1,285 @@
+<!--
```
csasarak, line 1: Is it worth just linking to this rather than copying the text?

docs/features/custom-license-and-keyword-searches.md
```diff
@@ -0,0 +1,211 @@
...
+![A Custom License in a FOSSA Licensing report](../assets/custom-license-scan-report.png)
+
+Note that the name will be used as the license's name in the UI, so it is important to use names that are understandable to someone looking at license issues and viewing a report.
+
+If your project is set to raise issues for a license of type "Custom License", then an issue will be raised for any custom licenses found.
+
+## Regular expression format
+
+The regular expressions in Custom License and Keyword searches use Rust's regular expression syntax. Here are a few examples. You can also view the [full regular expression syntax documentation](./custom-license-and-keyword-search-regular-expression-syntax.md).
+
```
csasarak, line 62: This may be really obvious to folks used to worked with regexes, but I'd would add an explicit example calling out the fact that a simple string like "Match Me Exactly" is also a valid regular expression. Possibly repeat this where you first use the words "regular expression" and as a heading here.

You imply it in the first example, but I could see someone who hasn't worked with regexes before but wants to match something simple coming in and thinking they have to learn regex. Personally, I'd like customers to not use regex features as much as humanly possible in order to avoid support tickets that boil down to "Why doesn't my regex work?"

docs/features/custom-license-and-keyword-searches.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+You can also set the `ignoreOrgWideCustomLicenseScanConfigs` flag to `true` in your `.fossa.yml` file. For example:
+
+```yaml
+version: 3
+ignoreOrgWideCustomLicenseScanConfigs: true
+```
+
+
```
csasarak, line 209: I recommend adding a troubleshooting section with a link to https://regex101.com/ advising them to try out their regexes on non-sensitive data with the "Rust" setting selected. Hopefully this will help folks to make sure that their regexes are correct before concluding that we are wrong.

docs/references/subcommands/analyze.md
```diff
@@ -67,6 +67,14 @@ The Vendored Dependencies feature allows you to scan for licenses directly in yo
 | `--force-vendored-dependency-scan-method` | Force the vendored dependency scan method. The options are 'CLILicenseScan' or 'ArchiveUpload'. 'CLILicenseScan' is usually the default unless your organization has overridden this. |
 | `--force-vendored-dependency-rescans`     | Force vendored dependencies to be rescanned even if the revision has been previously analyzed by FOSSA. This currently only works for CLI-side license scans.                         |
 
+### Custom License Searches
+
+The "Custom License Searches" feature allows you to define regular expressions that are used to search through your codebase. If a match to the regular expression is found, then a custom license is reported for that match.
```
csasarak, line 72:
```suggestion
The "Custom License Searches" feature allows you to define text strings or regular expressions that are used to search through your codebase. If a match to the text string or regular expression is found, then a custom license is reported for that match.
```
This is more wordy, but maybe this is useful to be less intimidating for users not familiar with regexes who may not even need all that functionality?

docs/references/subcommands/analyze.md
```diff
@@ -67,6 +67,14 @@ The Vendored Dependencies feature allows you to scan for licenses directly in yo
 | `--force-vendored-dependency-scan-method` | Force the vendored dependency scan method. The options are 'CLILicenseScan' or 'ArchiveUpload'. 'CLILicenseScan' is usually the default unless your organization has overridden this. |
 | `--force-vendored-dependency-rescans`     | Force vendored dependencies to be rescanned even if the revision has been previously analyzed by FOSSA. This currently only works for CLI-side license scans.                         |
 
+### Custom License Searches
```
csasarak, line 70: Could you link to this section directly somewhere in the TOC? https://github.com/fossas/fossa-cli/tree/master/docs#table-of-contents 

I know we already link to the analyze.md doc, but I think it's worth surfacing sections like this higher. I haven't gotten around to trying to rewrite the TOC but having more topics listed will make that work easier in the future. I wouldn't do anything fancy, just maybe dump the link somewhere?

csasarak (changes requested):
I've asked for a few more items in the docs mainly aimed at making the feature seem a bit more accessible to users who don't need full regular expression support. 

The only thing really holding up approval for me is a small `troubleshooting` section with some tips on how to validate that a regex is written correctly before filing a support ticket with us.

docs/features/custom-license-and-keyword-search-regular-expression-syntax.md
spatten: There are two reasons why we're copying rather than just linking

1) We want to be able to link directly to something that we own 
2) I made some minor edits to the file to remove the example Rust code that would probably just be confusing for most people reading this

csasarak: That makes sense, thanks!

docs/features/custom-license-and-keyword-searches.md
spatten: Good idea! I added a section describing this, and also added some info on what characters you'll need to escape if you want to match them exactly

docs/features/custom-license-and-keyword-searches.md
spatten: Done!

csasarak (approved):

PR #1275, opened 2023-09-13
Conda pip support

csasarak:
# Overview

This PR provides partial support for pip package specifiers in `environment.yml`.

What it doesn't support are URLs like would be used in pip.

## Acceptance criteria

It is possible to analyze conda projects with an `environment.yml` like:
```
name: example-environment

dependencies:
  - click=8.1.3
  - pip:
    - pytest-httpserver==1.0.6 
```

## Testing plan

I tested manually. I also refactored the tests a little bit for EnvironmentYmlSpec and added this as a case.

## References

[Slack thread](https://teamfossa.slack.com/archives/C0155DTGWB1/p1694524113935939)

## Checklist

### DIFF
--- a/src/Strategy/Conda/EnvironmentYml.hs
+++ b/src/Strategy/Conda/EnvironmentYml.hs
@@ -1,5 +1,7 @@
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE DerivingStrategies #-}
+{-# LANGUAGE ViewPatterns #-}
 
 module Strategy.Conda.EnvironmentYml (
   analyze,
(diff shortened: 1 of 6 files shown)


src/Strategy/Conda/EnvironmentYml.hs
```diff
@@ -73,9 +87,36 @@ getCondaDepFromText rcd =
...
+  deriving (Eq, Ord, Show)
+
+instance FromJSON CondaPkg where
+  parseJSON cd = (Pkg <$> parseJSON cd) <|> withObject "CondaPipPkg" parseOtherPkgManager cd
+    where
+      parseOtherPkgManager :: Object -> Parser CondaPkg
+      parseOtherPkgManager (KeyMap.toList -> [("pip", deps)]) = trace "Pip package!" $ withArray "Pip Packages" parseReqs deps
+      parseOtherPkgManager obj = fail ("Expected object with a single key (\"pip\") and list of deps. Got: " <> show obj)
+
+      parseReqs :: Array -> Parser CondaPkg
```
spatten, line 105: In the Slack thread and in this PR's description, you said that this would fail for URLs. If this is true, should we add a comment giving an example of what fails and noting that it does not work yet?

test/Conda/EnvironmentYmlSpec.hs
```diff
@@ -6,27 +8,16 @@ import Data.ByteString qualified as BS
...
-    { dependencyType = CondaType
-    , dependencyName = "name"
-    , dependencyVersion = Just (CEq "version1")
-    , dependencyLocations = []
-    , dependencyEnvironments = mempty
-    , dependencyTags = Map.empty
-    }
+import Text.RawString.QQ (r)
 
 dependencyTwo :: Dependency
```
spatten, line 22: I don't think `dependencyTwo` is used outside of `epectedGraph`. Maybe just move it into `expectedGraph`?

spatten (approved):
Looks great! Two small comments, but nothing at all blocking

src/Strategy/Conda/EnvironmentYml.hs
csasarak: Good idea, I've added some text about this.

test/Conda/EnvironmentYmlSpec.hs
csasarak: You're right, I've inlined it.

saramaebee (approved):

PR #1285, opened 2023-09-22
Millhone CLI: commit

jssblck:
# Overview

Implements the `commit` subcommand, which generates a `fossa-deps` file from the output of `millhone analyze`.

## Acceptance criteria

Users are able to generate a `fossa-deps` file.

## Testing plan

In one terminal, run Millhone. I ran it with the in-memory DB for convenience:

```shell
foundation/millhone on ⑆ millhone/nits [$] is 📦 v0.2.0 via 🦀 v1.72.0
❯ cargo run -- --db-in-memory
    Finished dev [unoptimized + debuginfo] target(s) in 0.16s
     Running `/Users/jessica/projects/foundation/target/debug/millhone --db-in-memory`
  2023-09-21T23:42:11.659337Z  INFO millhone: creating server, options: Application { port: 3000, tracing: TracingConfig { trace_level: Info, trace_spans: Off, trace_format: Text, trace_colors: Auto }, app_db: AppDbConfig { app_db_host: NonEmptyString("localhost"), app_db_port: 5432, app_db_name: NonEmptyString("millhone"), app_db_user: NonEmptyString("fossa"), app_db_password: Secret([REDACTED alloc::string::String]), app_db_max_pool: 10, app_db_min_pool: 10, app_db_connection_timeout_seconds: 30 }, corpus_db: CorpusDbConfig { corpus_db_nodes: ["localhost:9042"], corpus_db_keyspace: "millhone", corpus_db_username: None, corpus_db_password: None }, db_in_memory: true }
    at millhone/src/main.rs:198

  2023-09-21T23:42:11.669588Z  INFO millhone: booting server, addr: 0.0.0.0:3000
    at millhone/src/main.rs:202
```

I then copied the test files from `foundation-libs/snippets` and ran these steps:
```shell
millhone ingest \
  --direct-endpoint http://localhost:3000 \
  --api-key-id testing \
  --api-secret testing \
  --locator 'git+github.com/fossas/cpp-vsi-demo$76987d9d364bc2c73897549a71d13371a32cd3e7' \
  --kind signature \
  --kind body \
  --kind full \
  --transform space \
  ~/projects/cpp-vsi-demo/example-internal-project

millhone analyze \
  --direct-endpoint http://localhost:3000 \
  --api-key-id testing \
  --api-secret testing \
  --kind signature \
  --kind body \
  --kind full \
  --transform space \
  -o millhone_analyze_output \
  src 

millhone commit \
  --analyze-output-dir millhone_analyze_output \
[...]

PR #1287, opened 2023-09-25
Millhone: Parallelization

jssblck:
# Overview

Updates the `analyze` and `ingest` commands to support parallelization.
I did not parallelize `commit` as I think that is not overall not worth the effort, it's not nearly so performance sensitive.

Additionally:
- Transforms all fatal errors in `analyze` into warnings.
- Makes default tracing less verbose.
- Adds more detail to verbose tracing (`--trace-level trace`).
- Minor reorganization for the code.

## Acceptance criteria

Everything works as before, but now with more parallelization.

My testing project is rather small so I only saw modest gains, but in general my CPU use went from 70%-ish to 200%-ish during these two subcommands.

## Testing plan

I performed the steps in https://github.com/fossas/fossa-cli/pull/1285 and saw no difference in functionality.

No tests were added for this change; in Rust safe parallelization is statically guaranteed by the type system and borrow checker.

## Risks

None

## Metrics

None

## References

https://fossa.atlassian.net/browse/ANE-1186

## Checklist

### DIFF
--- a/extlib/millhone/src/cmd/analyze.rs
+++ b/extlib/millhone/src/cmd/analyze.rs
@@ -1,4 +1,7 @@
-use std::{path::PathBuf, collections::{HashMap, HashSet}};
+use std::{
+    collections::{HashMap, HashSet},
+    path::PathBuf,
+};
 
 use clap::Parser;
 use getset::Getters;
@@ -95,7 +98,7 @@ pub fn main(endpoint: &BaseUrl, opts: Subcommand) -> Result<(), Report> {
             } else {
                 debug!(path = %path.display(), "skipped: not a file");
                 None
-            } 
+            }
         })
         // Extract snippets from each file in parallel.
         .filter_map(|path| -> Option<(PathBuf, HashSet<ContentSnippet>)> {
--- a/extlib/millhone/src/cmd/analyze.rs
+++ b/extlib/millhone/src/cmd/analyze.rs
@@ -123,12 +123,10 @@ pub fn main(endpoint: &BaseUrl, opts: Subcommand) -> Result<(), Report> {
             let fingerprint = found.snippet().fingerprint();
             let matching_snippets = client
                 .lookup_snippets(fingerprint)
-                .wrap_err_with(|| format!("lookup snippet for fingerprint '{fingerprint}'"))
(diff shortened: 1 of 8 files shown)


extlib/millhone/src/cmd/analyze.rs
```diff
@@ -36,123 +38,157 @@ pub struct Subcommand {
...
-        }
+    WalkDir::new(root)
+        // Follow symlinks; loops are yielded as errors automatically.
+        .follow_links(true)
+        // Not chosen for a specific reason, just seems reasonable.
+        .max_depth(1000)
+        // Just make the walk deterministic (per directory anyway).
+        .sort_by_file_name()
+        .into_iter()
+        .inspect(|_| total_count_entries.increment())
```
spatten, line 83: Is there a difference between `inspect` and `tap`? It looks to me like they do the same thing, but `inspect` is a standard method on iterators, so it's idiomatic to use them there. Is that right?

spatten (approved):
Looks good! I just had the one question, but that's to assuage my curiousity

extlib/millhone/src/cmd/analyze.rs
jssblck: `inspect` is an iterator adaptor (so for a given `Iterator<Item = T>`, its closure receives `&T`), whereas `tap` takes the value (so for a given `T`, its closure receives `&T`).

This looks really similar but the effect on iterators is such that for e.g. `vec![1].into_iter().tap(|x| {})`, `x` is actually `&impl Iterator<Item = usize>`, _not_ what you're actually after (which is `&usize`, and accessible via `inspect`).

jssblck: We can also see this in the different syntax:

- `tap`: `fn tap(self, func: impl FnOnce(&Self)) -> Self`
- `inspect`: `fn inspect<F: FnMut(&Self::Item)>(self, f: F) -> Inspect<Self, F>` (also note `Inspect` implements `Iterator`).

Without going too much into it, from there we can see that `tap` actually takes ownership of `self` (which is whatever it's being invoked on). Consequently, if invoked on an iterator, it gets the whole iterator and not just one item.

PR #1295, opened 2023-10-09
Ane 1024 locators in fossa deps

JeffreyHuynh1:
# Overview

Currently, there’s a bit of tedium involved in mapping locators to their appropriate values in a fossa-deps file, and if you’re dealing with multiple types of dependencies it can make macros or using find+replace a little more involved.

Adding support for locators to fossa-deps would be a much-appreciated QoL boost, as it would simplify the reproduction process significantly. Getting locators is often the first step when investigating an issue.

## Acceptance criteria

- Someone can paste a series of locators into a fossa-deps file and scan their project with minimal effort.

```
locator-dependencies:
- pip+Django$2.1.7
```

- Unlinked markdown doc in fossa cli

   - Put under an internal subfolder for debug specific functionality

   - Create a notion page for fossa-cli reference and link to that.

## Testing plan

- Added locator dependencies on the success case ( 2 cases of success)
- Added failing case where there is an error on empty locator dependency
- Added notion page for locator dependency

## References

- [ANE-1024](https://fossa.atlassian.net/browse/ANE-1024)

## Checklist

[ANE-1024]: https://fossa.atlassian.net/browse/ANE-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/ManualDeps.hs
+++ b/src/App/Fossa/ManualDeps.hs
@@ -169,8 +169,8 @@ toSourceUnit root depsFile manualDeps@ManualDependencies{..} maybeApiOpts vendor
         (ManualJSON path) -> tryMakeRelative root path
         (ManualYaml path) -> tryMakeRelative root path
 
-  pure
-    $ SourceUnit
+  pure $
+    SourceUnit
       { sourceUnitName = renderedPath
       , sourceUnitManifest = renderedPath
       , sourceUnitType = "user-specific-yaml"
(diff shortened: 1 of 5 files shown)


src/App/Fossa/ManualDeps.hs
```diff
@@ -16,7 +16,8 @@ module App.Fossa.ManualDeps (
   findFossaDepsFile,
   readFoundDeps,
   getScanCfg,
-) where
+)
+where
```
jssblck, line 20: [curious] why?

test/App/Fossa/ManualDepsSpec.hs
```diff
@@ -54,6 +56,10 @@ theWorks = ManualDependencies references customs vendors remotes
       [ VendoredDependency "vendored" "path" Nothing
       , VendoredDependency "versioned" "path/to/dep" (Just "2.1.0")
       ]
+    locators =
+      [ Locator "fetcher-1" "one" Nothing
+      , Locator "fetcher-2" "two" $ Just "1.0.0"
+      ]
```
jssblck, line 62: [optional] This is an invalid locator, and if support accidentally provides an invalid locator to import it may be confusing as to why it isn't working.

Consider validating the provided fetcher using `fetcherToDepType`, such that if the result is `Nothing` FOSSA CLI emits a `logWarn` that the provided locator may not be valid.

Note: I don't think the contract of `Locator`'s `FromJSON` instance should change to do this, I think this should be special-cased in `ManualDeps`.

jssblck (commented):
lgtm after CLA, comments are optional

src/App/Fossa/ManualDeps.hs
JeffreyHuynh1: That was auto generated when I ran `make fmt`

jssblck (approved):

PR #1296, opened 2023-10-09
[ANE-458] machine-readable fossa list-target command

meghfossa:
# Overview

This PR adds `--format` options to `fossa list-targets` command:
- `legacy` (current format)
- `ndjson` 
- `text` (same format as `--only-target` and `--exclude-target`)

```bash
user@user-MacBook-Pro code % fossa-dev list-targets example-projects --format=ndjson 
{"path":"php/","type":"composer"}
{"path":"cabal/multi-home-cabal-example/","type":"cabal"}
{"path":"fossa-deps/remote-dependencies/","type":"cabal"}
{"path":"fossa-deps/vendored-dependencies/","type":"cabal"}
{"path":"fossa-deps/unarchived-vendored-dependencies/","type":"cabal"}
{"path":"cocoapods/simple-cocoapods-example/","type":"cocoapods"}
{"path":"swift-ios-xcode-proj/using-swift-pm-example.xcodeproj/","type":"swift"}

...
```

```bash
user@user-MacBook-Pro code % fossa-dev list-targets example-projects --format=text
composer@php/
cabal@fossa-deps/vendored-dependencies/
cabal@fossa-deps/remote-dependencies/
cabal@fossa-deps/unarchived-vendored-dependencies/
cabal@cabal/multi-home-cabal-example/
cocoapods@cocoapods/simple-cocoapods-example/
...
```

## Acceptance criteria

On Windows, macOs, and Linux, as a user,

- [x]  I can use `fossa list-targets --format=ndjson` to retrieve line separated valid analysis target in json,
- [x] I can use `fossa list-targets --format=text` to retrieve valid analysis target in text format
    - It renders in same format as one accepted by `fossa analyze --only-target`, and `fossa analyze --exclude-target`
    - I can copy paste single line of output from `fossa list-targets` to `fossa analyze --only-target=<LINE>` and it correctly applies filter 
- [x] I can use `fossa list-targets --format=legacy`
- [x] When no format option is provided for `fossa list-targets`, it uses `legacy` format by default

## Testing plan

1. `git pull origin && git checkout feat/machine-readable-list-targets && make install-dev`
2. `fossa-dev list-targets`
- Check default case, it should produce same output as current release fossa-cli
4. `fossa-dev list-targets --format=ndjson`
- It should print json target per line, if target is not present, it shall not include it in json object
5. `fossa-dev list-targets --format=text`
- It should print target per line

I used, 
1. Multi-module gradle project (to check project with `target` case): https://docs.gradle.org/current/samples/zips/sample_building_java_applications_multi_project-groovy-dsl.zip
2. Example-project repository (to check project without targets): https://github.com/fossas/example-projects

## Risks

N/A

## Metrics

None

## References

- [ANE-458](https://fossa.atlassian.net/browse/ANE-458): Implement machine-readable fossa list-targets command

## Checklist

- [x] ~I added tests for this PR's change (or explained in the PR description why tests don't make sense).~

[ANE-458]: https://fossa.atlassian.net/browse/ANE-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # FOSSA CLI Changelog
 
-## v3.8.18
+## Unreleased
 
 - `fossa list-targets`: list-target command supports `--format` option with: `ndjson`, `text`, and `legacy`. ([#1296](https://github.com/fossas/fossa-cli/pull/1296))
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.8.18
+
+- `fossa list-targets`: list-target command supports `--format` option with: `ndjson`, `text`, and `legacy`. ([#1296](https://github.com/fossas/fossa-cli/pull/1296))
+
 ## v3.8.17
 
 Integrates FOSSA snippet scanning into the main application.
--- a/src/App/Fossa/ListTargets.hs
+++ b/src/App/Fossa/ListTargets.hs
@@ -8,6 +8,7 @@ module App.Fossa.ListTargets (
 import App.Fossa.Analyze.Discover (DiscoverFunc (DiscoverFunc), discoverFuncs)
 import App.Fossa.Config.Analyze (ExperimentalAnalyzeConfig)
 import App.Fossa.Config.ListTargets (
+  ListTargetOutputFormat (..),
   ListTargetsCliOpts,
   ListTargetsConfig (..),
   mkSubCommand,
@@ -33,11 +34,11 @@ import Control.Effect.Debug (Debug)
 import Control.Effect.Lift (Lift, sendIO)
 import Control.Effect.Stack (Stack)
 import Control.Effect.Telemetry (Telemetry)
-import Data.Aeson (ToJSON)
+import Data.Aeson (ToJSON, encode, object, (.=))
 import Data.Aeson.Extra (encodeJSONToText)
 import Data.Foldable (for_, traverse_)
-import Data.Set qualified as Set
 import Data.Set.NonEmpty (toSet)
+import Data.String.Conversion (decodeUtf8, toText)
 import Discovery.Filters (AllFilters)
 import Discovery.Projects (withDiscoveredProjects)
 import Effect.Exec (Exec)
@@ -50,6 +51,7 @@ import Effect.Logger (
   color,
   logDebug,
   logInfo,
+  logStdout,
   logWarn,
  )
 import Effect.ReadFS (ReadFS)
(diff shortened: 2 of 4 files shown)


src/App/Fossa/ListTargets.hs
```diff
@@ -97,14 +100,49 @@ runAll ::
...
   , Has Telemetry sig m
   ) =>
+  ListTargetOutputFormat ->
   Path Abs Dir ->
   m ()
-runAll basedir = traverse_ single discoverFuncs
+runAll outputFmt basedir = traverse_ single discoverFuncs
   where
-    single (DiscoverFunc f) = withDiscoveredProjects f basedir (printSingle basedir)
+    single (DiscoverFunc f) = withDiscoveredProjects f basedir formatter
```
meghfossa, line 108: [note] We can use the `Output (Discovered Project a)` to sink across thread boundaries as mentioned in ticket's implementation note, but I chose to `logStdout`'s concurrent output for simplicity - but happy to change it!

Ref: https://hackage.haskell.org/package/concurrent-output-1.10.19/docs/System-Console-Concurrent.html#v:outputConcurrent

csasarak: This seems OK to me as long as the data is output atomically and not mangled.

src/App/Fossa/ListTargets.hs
```diff
@@ -97,14 +100,50 @@ runAll ::
...
+renderNdJson basedir project = do
+  let maybeRel = makeRelative basedir (projectPath project)
+  let render txt = logStdout . decodeUtf8 $ (encode txt) <> "\n"
+  case maybeRel of
+    Nothing -> pure ()
+    Just rel ->
+      case projectBuildTargets project of
+        ProjectWithoutTargets ->
+          render $ object ["type" .= projectType project, "path" .= toFilePath rel]
+        FoundTargets targets -> for_ (Set.toList $ toSet targets) $ \target -> do
```
csasarak, line 126: Sets are Foldable so I think that `Set.toList` is unnecessary, `for_` should work just fine with it.

src/App/Fossa/ListTargets.hs
```diff
@@ -97,14 +100,50 @@ runAll ::
...
+  let maybeRel = makeRelative basedir $ projectPath project
+  let render txt = logStdout $ txt <> "\n"
+
+  case maybeRel of
+    Nothing -> pure ()
+    Just rel ->
+      case projectBuildTargets project of
+        ProjectWithoutTargets -> do
+          render $ toText (projectType project) <> "@" <> toText (toFilePath rel)
+        FoundTargets targets -> for_ (Set.toList $ toSet targets) $ \target -> do
```
csasarak, line 142: I think `for_` should work find without the added `toList` because Sets are Foldable.

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.8.17
```
csasarak, line 3: We generally try to release most new versions to Winget now. I'll send you a Slack with a bit of information.

csasarak (approved):
LGTM, mostly minor comments.

PR #1300, opened 2023-10-11
[FDN-12] VSI Dep Paths on windows

spatten:
# Overview

[Delivers FDN-12](https://fossa.atlassian.net/browse/FDN-12) — Investigate "Could not Parse Rule Path" for GlobalDots (Fujitsu)

The bug was that when you ran `fossa analyze --detect-vendored` on a Windows machine and got matches to vendored binaries, the CLI would fail with an error like this:

```
Could not parse rule path: "/Framework/Common/fjlic/private/WINx64/"
```

This was caused by an attempt to call `parseAbsDir` on the rule path on a Windows machine, which failed.

It turned out that there was no real need to parse the path in this case, so I renamed `Srclib.toSourceUnit` to `Srclib.projectToSourceUnit`, and made that call another function that took in the path as Text. Then I called that second function from the VSI code paths.

## Acceptance criteria

- This fixes the problem

## Testing plan

Do the following on a Windows machine:

```
mkdir tester
cd tester
mkdir vendor/make
<download make-4.4 from https://ftp.gnu.org/gnu/make/ and untar it into this directory>
cd ../..
fossa analyze --detect-vendored
```

Previous to this we would get the following error:

```
Could not parse rule path: "/vendor/make/"
```

## References

Two Slack threads:

- Me asking for help on reproducing this: https://teamfossa.slack.com/archives/C054BCAU5N1/p1696964780382109 
- Original bug report from Deepak: https://teamfossa.slack.com/archives/C043EM3L96Z/p1696871963231889

## Checklist

### DIFF
--- a/docs/contributing/HACKING.md
+++ b/docs/contributing/HACKING.md
@@ -94,7 +94,7 @@ If you installed HLS in the old, complicated way, you can safely remove it.  HLS
 
 You should also set the `FOSSA_SKIP_EMBED_FILE_IN_HLS` environment variable for HLS. This prevents HLS from embedding binaries, which helps to avoid a giant memory footprint for HLS.
 
-You can also tell the Fourmolu plugin to use an external config. This ensures that it picks up our `fourmolu.yaml` file.
+You can also tell the Fourmolu plugin to use an external config. This will tell the language server to use an external executable that you've installed, which ensures that it picks up our `fourmolu.yaml` file.
 
 In VSCode, this is done by adding this to your `settings.json`:
 
--- a/docs/contributing/HACKING.md
+++ b/docs/contributing/HACKING.md
@@ -94,12 +94,15 @@ If you installed HLS in the old, complicated way, you can safely remove it.  HLS
 
 You should also set the `FOSSA_SKIP_EMBED_FILE_IN_HLS` environment variable for HLS. This prevents HLS from embedding binaries, which helps to avoid a giant memory footprint for HLS.
 
+You can also tell the Fourmolu plugin to use an external config. This ensures that it picks up our `fourmolu.yaml` file.
+
 In VSCode, this is done by adding this to your `settings.json`:
 
 ```json
     "haskell.serverEnvironment": {
         "FOSSA_SKIP_EMBED_FILE_IN_HLS": true,
     },
+    "haskell.plugin.fourmolu.config.external": true,
 ```
 
 ## Linting
--- a/src/App/Fossa/VSIDeps.hs
+++ b/src/App/Fossa/VSIDeps.hs
@@ -32,7 +32,7 @@ import Path (Abs, Dir, Path, SomeBase (..), parseAbsDir)
 import Path.Extra (SomePath (SomeDir))
 import Srclib.Converter qualified as Srclib
 import Srclib.Types (AdditionalDepData (..), SourceUnit (..), SourceUserDefDep)
-import System.FilePath (makeValid)
+import System.FilePath (normalise)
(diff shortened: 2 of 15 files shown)


src/App/Fossa/VSIDeps.hs
```diff
@@ -57,21 +66,37 @@ analyzeVSIDeps dir projectRevision filters skipResolving = do
...
-      case parseAbsDir (toString vsiRulePath) of
-        Just ruleDir -> pure $ toSourceUnit (toProject ruleDir dependencies) Nothing
-        Nothing -> fatal $ "Could not parse rule path: " <> show vsiRulePath
+
+#ifdef mingw32_HOST_OS
+-- The path we get back from Core is a Unix-style path, so we need to turn it into a Windows-style path by
+-- prepending a drive (e.g. "C:\") to it and running `normalise` on it.
+-- For example, if we get a path like "/Framework/Common/fjlic/private/WINx64/", we want to end up with "C:\Framework\Common\fjlic\private\WINx64\"
+ruleToSourceUnit :: (Has Diagnostics sig m, Has FossaApiClient sig m, Has ReadFS sig m) => VSI.SkipResolution -> VSI.VsiRule -> m (SourceUnit)
+ruleToSourceUnit skipResolving VSI.VsiRule{..} = do
```
jssblck, line 82: Note that the VSI results _appear_ to be absolute paths, but are actually relative to the scan directory (Sherlock considers `/` the "root of the scan", not the "root of the drive").

Does this account for this fact? Specifically, if we scan a project at `C:\Users\me\Documents\projects\foo`, and get back a path like `/bar/baz`, the full path should be `C:\Users\me\Documents\projects\foo\bar\baz`, not `C:\bar\baz`.

src/App/Fossa/VSIDeps.hs
```diff
@@ -57,21 +66,37 @@ analyzeVSIDeps dir projectRevision filters skipResolving = do
...
+  resolvedGraph <- resolveGraph [vsiRuleLocator] skipResolving
+  dependencies <- fromEither $ Graphing.gtraverse VSI.toDependency resolvedGraph
+  cwd <- getCurrentDir
+  let drive = takeDrive $ fromAbsDir cwd
+  let absPath = normalise $ joinDrive drive $ toString vsiRulePath
+  case parseAbsDir absPath of
+    Just ruleDir -> pure $ toSourceUnit (toProject ruleDir dependencies) Nothing
+    Nothing -> fatal $ "Could not parse rule path: " <> (show vsiRulePath) <> ", normalized: " <> normalise (toString vsiRulePath)
```
jssblck, line 90: Instead of `fatal`, can we just warn about it and use the scan path as the source unit path? I really want to avoid issues here from preventing users from using the feature at all. Same comment in the unix version below.

jssblck (approved):

docs/contributing/HACKING.md
```diff
@@ -94,12 +94,15 @@ If you installed HLS in the old, complicated way, you can safely remove it.  HLS
 
 You should also set the `FOSSA_SKIP_EMBED_FILE_IN_HLS` environment variable for HLS. This prevents HLS from embedding binaries, which helps to avoid a giant memory footprint for HLS.
 
+You can also tell the Fourmolu plugin to use an external config. This ensures that it picks up our `fourmolu.yaml` file.
```
spatten, line 97: I know it's not related to this PR, but I thought I'd sneak this in while I had it in my head

csasarak (approved):

PR #1303, opened 2023-10-13
Implementation to add custom fossa-deps file on analyze run

JeffreyHuynh1:
# Overview

As a user, I want to provide path to fossa-deps file during analysis, similar to how I can use -c | --config to provide custom config file. This is so I can run analysis across multiple fossa-deps file.  

## Acceptance criteria

As a user, I can provide different path to fossa-deps file during analysis.

For example, I should be able to use custom/configs/zlib.fossa-deps.yml as fossa-deps file during analysis

```
/custom
 - /libs
 - /zlib
 - /configs
    - zlib.fossa-deps.yml
    - libs.fossa-deps.yml
```

## Testing plan

- Tested and confirmed we are pulling the targeted fossa-deps config file with flag `--fossa-deps-config 
Paths and file formats tested:
  - a.yml (relative path and absolute path)
  - ../fossa-deps.yaml
  - ../z.json
  
 - When ran with no flag, it just goes through the normal experience and searches for the fossa-deps-config in the analyze directory

## Risks

- Flag name
- Implementation 

## References

- [ANE-979](https://fossa.atlassian.net/browse/ANE-979)

## Checklist

[ANE-979]: https://fossa.atlassian.net/browse/ANE-979?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -299,7 +299,7 @@ analyze cfg = Diag.context "fossa-analyze" $ do
         then do
           logInfo "Running in VSI only mode, skipping manual source units"
           pure Nothing
-        else do Diag.context "fossa-deps" . runStickyLogger SevInfo $ analyzeFossaDepsFile basedir customFossaDepsPath maybeApiOpts vendoredDepsOptions
+        else Diag.context "fossa-deps" . runStickyLogger SevInfo $ analyzeFossaDepsFile basedir customFossaDepsPath maybeApiOpts vendoredDepsOptions
   maybeLernieResults <-
     Diag.errorBoundaryIO . diagToDebug $
       if filterIsVSIOnly filters
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -269,7 +269,7 @@ analyze cfg = Diag.context "fossa-analyze" $ do
       skipResolutionSet = Config.vsiSkipSet $ Config.vsiOptions cfg
       vendoredDepsOptions = Config.vendoredDeps cfg
       grepOptions = Config.grepOptions cfg
-      customFossaDepsPath = Config.customFossaDepsPath cfg
+      customFossaDepsFile = Config.customFossaDepsFile cfg
 
   -- additional source units are built outside the standard strategy flow, because they either
   -- require additional information (eg API credentials), or they return additional information (eg user deps).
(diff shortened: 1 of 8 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -269,11 +268,16 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
   -- additional source units are built outside the standard strategy flow, because they either
   -- require additional information (eg API credentials), or they return additional information (eg user deps).
-  vsiResults <- Diag.errorBoundaryIO . diagToDebug $
-    Diag.context "analyze-vsi" . runStickyLogger SevInfo . runFinally $ do
+  vsiResults <- Diag.errorBoundaryIO
+    . diagToDebug
+    $ Diag.context "analyze-vsi"
+      . runStickyLogger SevInfo
+      . runFinally
+    $ do
```
JeffreyHuynh1, line 280: Don't know why these are re-formatting when I am using `make fmt`. Auto format occurs at different points in this pr as well fyi

JeffreyHuynh1:
Questions:

- Is there any changes that I need to make to any docs about these new changes?
- I created a new function in ManualDeps.hs, but looking that ManualDepsSpec.hs it's really only testing results of json decoding. Should I be adding some test cases for the new function I created, or is it sufficient to just leave it as is bc there isn't any test cases for all the other functions in ManualDeps.hs?

src/App/Fossa/ManualDeps.hs
```diff
@@ -80,17 +82,39 @@ analyzeFossaDepsFile ::
...
+retrieveCustomFossaDepsFile fossaDepsPath = do
+  let extension = takeExtension fossaDepsPath
+  file <- (validateFile fossaDepsPath)
+
+  if extension == ".yml" || extension == ".yaml"
+    then pure $ Just $ ManualYaml file
+    else
+      if extension == ".json"
+        then pure $ Just $ ManualJSON file
+        else pure Nothing
```
meghfossa, line 116: We should fail fatally here - since we are silently ignoring, if user has provided some non compliant file. I think for `--config`, we fail fatally. Prefer if replicate behaviour here: `fatalText "expected ... but, received: "`

meghfossa (approved):
- Make sure to add this in changelog!
- Some comments regarding flag name
- Some comments regarding failing fatally if provided deps file does not exist or compliant! 

🚀

PR #1306, opened 2023-10-19
Ghc dist new style cache

csasarak:
# Overview

GHC 9.4 better supports caching of dist-newstyle artifacts. This PR is an attempt to take advantage of that to make our build times less.

From this PR you can expect:
* We cache `dist-newstyle` so builds should be quite a bit quicker.
* The cabal install check in `lint.yml` is removed. I don't think it really serves much purpose and it was an additional build (we do four builds build on every CI push currently) I don't think we need.
* Benchmarks have been changed to be something that needs to be triggered manually and should not block builds. I'm not aware of anywhere we really check that the benchmarks are good so it just seemed like we were burning cycles for nothing.
* The CLI no longer reports if there was a dirty worktree in `--version` output. See "Risks" below.

Credit goes to the platform team who wrote versions of this that I was able to build off of.

## Testing plan

Test cases:

[x] It builds when a new tag is pushed and generates a CLI which reports the tag as the version.
[x] It uses both the cabal cache and the dist-newstyle cache across systems.
[x] Changes to vendored binaries (resulting in a new hash I think?) also result in recompilation of `EmbeddedBinary.hs`
      - I checked this locally. It works, but I had to add the bins to `extra-source-files` in our cabal file. 

To review, you can look at the times for each of our builds. They should be in the 10 -15 minute range. Additionally, in the logs you should see the build step only building a few of our modules rather than all of them. Feel free to push some minor code changes into this branch to view the behavior.

## Risks

The main risks are associated with not triggering template Haskell recompilation when necessary. I have gotten cabal/GHC to reliably notice changes to dependent binaries. This functionality relies on the cached build products having had those files available when building. This is pretty much always the case in our CI since building and fetching all dependent binaries is a prerequisite for building. 

Getting cabal/GHC to notice a new git tag and recompile Version.hs is tricky. See [comments](https://github.com/fossas/fossa-cli/pull/1306/files#diff-7682a9fd2942d8a81f55c30abd512b277b1b1eb7c5f16b2c4d88323de90e55d0R164). This is admittedly a hack, but necessary for this PR to move forward.

Aside from the above being a dollar-store solution, I also had to remove the `(dirty)` string from our `--version` output. The reason is that by adding a comment to make `Version.hs` compile unconditionally, the build process makes the worktree dirty, so even release builds would report their version as dirty. There may be a way to work around this by writing code to ignore changes to `Version.hs` when figuring out dirtiness, but I didn't think it was worth it to further complicate things for this unless a reviewer thinks we need it. Personally I don't think "(dirty)" provides much value to anyone and think it's worth trading for faster builds. I checked and neither the CLI nor hubble seem to look for the word. I'd like a reviewer to sign off on this since it technically is a change observable by customers. 

Telemetry now will also report environment differently: If the CLI was built on master _and_ has a version tag we'll consider as a production build. This isn't perfect, but I think will be more accurate than determining if it was built with a dirty work-tree. Another option could be to set this explicitly as an env var we read during compilation and use that. 

Finally, because we will now be saving a cache for every commit it means that we'll be generating more cache data towards our 10 GB limit. I think this is OK because many of the caches will be long stale by the time GH decides to start evicting. Each commit will generate about 1 GB in just `dist-newstyle` caches. Rust caches and the cabal store are generally more durable, shareable and altogether are ~1 GB.  As an estimate, I'd say that leaves us space for ~6 or 7 simultaneous and active branches of development without stepping on each other. 

### Caveats

`master` can't use caches from child branches though the reverse is true. This means that builds on master _may_ still wind up being longer than they were on the branch. Instead, releases and builds on `master` will fall back to whatever the most recent caches on `master` were. 

Likewise, tags on a branch are considered sibilings to branches so they'll fall back to using master caches.

Even with the above, I think this PR is an improvement on the status quo.

Note: When this merges initially and the first time we release it will trigger full builds. This is because master cannot rely on caches from child branches and none exist on master yet.

## Checklist

- [ ] Github actions is my favorite programming language.


.github/workflows/bench.yml
```diff
@@ -49,17 +53,48 @@ jobs:
         cargo -V || echo "no cargo"
 
     - uses: Swatinem/rust-cache@v2
-    
+
+    # Compute cache key and save to a temporary file.
+    #
+    # We compute the cache key based on the solved install plan instead of just
+    # hashing the `.cabal` file, since there are many kinds of changes that will
+    # cause `.cabal` to change (e.g. adding new source modules).
```
jssblck, line 61: Do such changes not always result in changes to the install plan? I'd think they would.

.github/workflows/bench.yml
```diff
@@ -74,7 +109,6 @@ jobs:
       env:
         RUN_CMD: cabal build benchmarks
       run: |
-        cabal update
```
jssblck, line 77: Why would we not want to do this?

spectrometer.cabal
```diff
@@ -3,9 +3,23 @@ name:               spectrometer
...
+-- were originally built with some file of that name present.
+-- I think this has always been the case even without this change.
+-- Run `cabal clean` to force a rebuild if you first built a version of fossa CLI
+-- without binaries to embed.
 extra-source-files:
   scripts/*.gradle
   scripts/*.jar
+  vendor-bins/index.gob.xz
+  vendor-bins/lernie
+  vendor-bins/themis-cli
```
jssblck, line 22: Make sure to add Rust binaries to this as well.

jssblck (approved):

.github/workflows/bench.yml
csasarak: The install plan only specifies the dependent packages that need to be there for the build. For spectrometer, it just lists the various spectrometer targets like fossa, pathfinder, the library, etc. Adding a new source module won't change the build plan. Adding a new 3rd party dep, for instance, would.

.github/workflows/bench.yml
csasarak: It's done as part of the "Calculate cabal cache key" step now.

spectrometer.cabal
csasarak: Good catch. It's just millhone/berkeleydb in addition, correct? 

I added both the variant with `.exe` and the one without.

jssblck: Yep, correct!

.github/workflows/bench.yml
jssblck: Ah, gotcha 👍🏻

PR #1309, opened 2023-10-20
[No Ticket] Updates registry api to account for docker index's new behaviour for `HEAD` api calls

meghfossa:
# Overview

Container Registry API implementation by docker.io has changed recently, and is leading to 404 errors, when `fossa-cli` performs `HEAD` to assert if the resource exists. 

Current `fossa-cli`'s implementations is consistent with documented SPEC: https://distribution.github.io/distribution/spec/api/#existing-manifests, but now docker.io will throw 404, if the `Accept` header does not include exact media-type of manifest. 

## Acceptance criteria

- can pull image from multi-image manifest matching host's platform from registry without HTTP error.

## Testing plan

I added automated test  to cover this scenario/findings. 

1. Ensure you do not have docker running
2. `git checkout feat/use-accept-headers-in-head-req && make install-dev`
3. perform `fossa-dev container analyze -o redis:latest` (this should not throw 404)

## Risks

N/A

## Metrics

N/A

## References

This is something I discovered via bulk testing to de-risk tar errors.

## Checklist

- ~If this PR introduced a user-visible change, I added documentation into `docs/`.~
- ~If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json`. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).~
- ~If I made changes to a subcommand's options, I updated `docs/references/subcommands/<subcommand>.md`.~


src/Control/Carrier/ContainerRegistryApi/Authorization.hs
```diff
@@ -144,11 +144,13 @@ getAuthToken ::
...
   Manager ->
   -- | Manager to use for requests
+  Maybe [Text] ->
+  -- | Content-Types for Accept Header
   Maybe AuthToken ->
   -- | Existing Token (if any)
   m (Maybe AuthToken)
-getAuthToken cred reqAttempt manager token = do
-  let request' = applyAuthForExistingToken $ reqAttempt{method = "HEAD"}
+getAuthToken cred reqAttempt manager accepts token = do
```
meghfossa, line 152: Previously container registry api (as implemented by index.docker.io) did not require `HEAD` calls to have same `Accept` header as manifest's type (specifically for multi-platform). Although, this is not explicitly noted in registry v2 SPEC, if not provided, it throws 404 error. Note that, other public registry do not exhibit this behaviour so far, but seems to work regardless.

test/Container/RegistryApiSpec.hs
```diff
@@ -68,6 +68,16 @@ registryApiSpec :: Spec
 registryApiSpec =
   describe "Container Registry API" $
     describe "Public Registry APIs" $ do
+      describe "Microsoft registry" $ do
```
spatten, line 71: When I run this on my laptop, I see this:

```
make test-cabal ARGS="Container Registry API"
```

```
 test/Test/Effect.hs:125:33: 
  1) Container.RegistryApi, Container Registry API, Public Registry APIs, Microsoft registry, should get manifest with digest
       expected: RepoDigest "sha256:bd36eac352dbb2d27a63f4714b9063fc8c7e9c615e430006dbeb7e3448cd5780"
        but got: RepoDigest "sha256:d24e8ccc34df12a86f907d27d85bfa732f30c771bfd9bd7e4ebf4ba66f18b241"
```

Is this because something changed on the registry or because I'm downloading something different because I'm on macOS or ???

spatten (approved):
Looks good to me!

I had one question about a failing test when I run this locally, but I don't think that's blocking

test/Container/RegistryApiSpec.hs
meghfossa: No os or arch shouldn't matter - there was an actual update to the tag (they overwrote the image@tag) - updated with more stale example now

spatten: :+1:

PR #1316, opened 2023-11-01
Fix install-latest.sh when fossa binary already exists

spatten:
# Overview

[Delivers ANE-1262](https://fossa.atlassian.net/browse/ANE-1262)

If you run `install-latest.sh` on macOS and the fossa binary already exists on the system, then the `cp` command will create a binary that fails to run.

This PR fixes that by using `mv` instead of `cp`.

## Acceptance criteria

`install-latest.sh` can be safely run even if `fossa` is already installed

## Testing plan

Run this multiple times in a row:

```
./install-latest.sh
fossa -V
```

## Risks

I think that this is low risk. It's a small change. But if it breaks things it will affect a lot of customers

## References

[ANE-1262](https://fossa.atlassian.net/browse/ANE-1262)

## Checklist

[ANE-1262]: https://fossa.atlassian.net/browse/ANE-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/install-latest.sh
+++ b/install-latest.sh
@@ -54,7 +54,8 @@ execute() {
       binexe="${binexe}.exe"
     fi
     log_debug "installing binary: $binexe"
-    cp "${srcdir}/${binexe}" "${BINDIR}/${binexe}" 2> /dev/null || sudo cp "${srcdir}/${binexe}" "${BINDIR}/"
+    # Use mv instead of cp, as copying when a binary already exists in that spot can create an invalid binary on macOS
+    mv "${srcdir}/${binexe}" "${BINDIR}/${binexe}" 2> /dev/null || sudo cp "${srcdir}/${binexe}" "${BINDIR}/"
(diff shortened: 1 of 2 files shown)


install-latest.sh
```diff
@@ -54,7 +54,8 @@ execute() {
       binexe="${binexe}.exe"
     fi
     log_debug "installing binary: $binexe"
-    cp "${srcdir}/${binexe}" "${BINDIR}/${binexe}" 2> /dev/null || sudo cp "${srcdir}/${binexe}" "${BINDIR}/"
+    # Use mv instead of cp, as copying when a binary already exists in that spot can create an invalid binary on macOS
```
spatten, line 57: This is the only change in the PR. The rest of the changes are just removing whitespace

meghfossa (approved):

PR #1318, opened 2023-11-06
[ANE-1092] Path Dependency Support

meghfossa:
# Overview

This PR is part of the chain, and will not be merged until `CORE` pr is merged!

- modifies `go list` tactic, such that `path+` dependencies are supported!
- adds support for `path+` dependencies 

Sibling PR: https://github.com/fossas/FOSSA/pull/11311

I will put up second PR, once CORE PR is approved, to avoid waste work. Second PR will include docs, and minor optimization to avoid rescanning, scanned dependency, as well as better handling for `checkForEmptyUpload` 

## Acceptance criteria

As a user, 
- [ ] I can successfully scan go project, which has vendor replace directives
- [ ] I can successfully scan go project, which has vendor replace directives
- [ ] `fossa-cli` does not license scan path dependencies, if path dependencies are not supported by endpoint
- [ ] `fossa-cli` does not license scan path dependencies, if output mode is used (`-o` flag)
- [ ] `fossa-cli` can license scan path dependencies, for matching data
- [ ] `fossa-cli` can license scan path dependencies, for fullfile data

## Testing plan

- I relied on automated testing.
- Testing plan will be included in second/final pr of the chain.

However, if you wish you can load the endpoint from: https://github.com/fossas/FOSSA/pull/11311

1. Create example: (refer to repro case here: https://fossa.atlassian.net/browse/ANE-383)
2. Build `make install-dev`
3. Perform `fossa-dev analyze -e http://localhost:9578 --fossa-api-key <KEY> repro`
4. You should see path dependencies on the endpoint

## Risks

N/A

## References

https://fossa.atlassian.net/browse/ANE-1092

## Checklist


JeffreyHuynh1 (approved):
LGTM

src/App/Fossa/Analyze.hs
```diff
@@ -373,13 +384,13 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
           (Just firstParty, Nothing) -> Just firstParty
-  let result = buildResult includeAll additionalSourceUnits filteredProjects licenseSourceUnits
+  let result = buildResult includeAll additionalSourceUnits filteredProjects' licenseSourceUnits
   let keywordSearchResultsFound = (maybe False (not . null . lernieResultsKeywordSearches) lernieResults)
 
   -- If we find nothing but keyword search, we exit with an error, but explain that the error may be ignorable.
   -- We do not want to succeed, because nothing gets uploaded to the API for keyword searches, so `fossa test` will fail.
   -- So the solution is to still fail, but give a hopefully useful explanation that the error can be ignored if all you were expecting is keyword search results.
-  case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectResults filteredProjects additionalSourceUnits licenseSourceUnits) of
+  case (keywordSearchResultsFound, checkForEmptyUpload includeAll filteredProjects' filteredProjects' additionalSourceUnits licenseSourceUnits) of
```
spatten, line 393: I don't understand why we're sending `filteredProjects'` twice into `checkForEmptyUpload`. It feels like a bug, maybe not introduced in this PR, but could we fix it here if it's easy?

If that's coming in the future PR, that's all good. Just wanted to double-check

src/App/Fossa/PathDependency.hs
```diff
@@ -0,0 +1,258 @@
...
+            , dependencyName = name'
+            , dependencyVersion = Just $ CEq version'
+            }
+
+-- * Helpers
+
+canResolve :: Dependency -> Bool
+canResolve d = dependencyType d == UnresolvedPathType
+
+isValidDep :: Bool -> Dependency -> Bool
```
spatten, line 194: Can we rename this to `isValidPathDep` and `canResolve` to `canResolvePathDep` or something like that? I think this would reduce uncertainty that these functions are only for path deps.

src/App/Fossa/PathDependency.hs
```diff
@@ -0,0 +1,258 @@
...
+      let uniqDeps = NE.nub pathDeps
+
+      -- 2. Get hash from the dependencies, and check
+      -- dependencies that can be skipped!
+      depsWithMetadata <- catMaybes . NE.toList <$> traverse (metadataOf baseDir) uniqDeps
+      let needToScanDepsWithMetadata = depsWithMetadata
+
+      -- 3. We scan and upload dependencies, and retrieve transformed
+      -- dependencies!
+      scannedDeps <- traverse (scanAndUpload Nothing fullFile projectRevision) needToScanDepsWithMetadata
```
spatten, line 121: Edit: I think this is coming in the future PR. If it is, then ignore this

~~In order to speed up subsequent scans, it would be nice to filter to only path deps that are unknown to FOSSA here. For vendored deps, we do this by hitting an API and asking it if Core knows about the revisions for the locators we're thinking of scanning, and then only scanning the ones that Core does not know about. There's also a flag to force re-scanning.~~

~~This is all described here: https://github.com/fossas/fossa-cli/blob/master/docs/features/vendored-dependencies.md#performance~~

src/Strategy/Python/PDM/Pdm.hs
```diff
@@ -137,7 +137,7 @@ reqDepType :: Req -> DepType
 reqDepType (NameReq{}) = PipType
 reqDepType (UrlReq _ _ url _) =
   if "file://" `isPrefixOf` URI.render url
-    then PathType
+    then UnresolvedPathType
```
spatten, line 140: Wait, is this all we have to do to turn on path dep support for Python PDM projects? That's so nice!

spatten: Also, if it is we should probably have something about PDM projects in the test plan

spatten (approved):
Looks great!

I think that you're planning to address most of these comments in a future PR, so ignore them if that's true

PR #1320, opened 2023-11-08
[ANE-1166] Support lock-file v6 for `pnpm`

meghfossa:
# Overview

This PR, 

- Adds support for `pnpm` lock-file: `6.0`, which is default format with pnpm `8.x`.

Sibling PR: https://github.com/fossas/example-projects/pull/10

## Acceptance criteria

As a user, 
- [ ] I can analyze `pnpm` project using lock-file, with`6.0` 
- [ ] I can analyze `pnpm` workspace project using lock-file, with`6.0` 

## Testing plan

I added automated tests. 

- You can also test by performing `fossa analyze -o` against any `pnpm` project using `8.x`

For example, 
```bash
# Make fossa-cli
git pull origin && git checkout feat/pnpm-lock-v6 && make install-dev

# Get cases
cd ..
git clone -b feat/adds-pnpm-v6 git clone https://github.com/fossas/example-projects

# Analyze
fossa-dev analyze ./example-projects/javascript/pnpm-v6/ -o | jq
fossa-dev analyze ./example-projects/javascript/pnpm-v6-workspace/ -o | jq
```

Let's take popular project:
```bash
git clone https://github.com/vitejs/vite.git --depth 0
cd viite && fossa-dev analyze -o --only-target pnpm@./
```

There is also example: https://fossa.atlassian.net/browse/ANE-1166

## Risks

N/A

## Metrics

N/A

## References

https://fossa.atlassian.net/browse/ANE-1166

## Checklist

### DIFF
--- a/src/Strategy/Node/Pnpm/PnpmLock.hs
+++ b/src/Strategy/Node/Pnpm/PnpmLock.hs
@@ -11,7 +11,9 @@ import Data.Aeson.Extra (TextLike (..))
 import Data.Foldable (for_)
 import Data.Map (Map, toList)
 import Data.Map qualified as Map
+import Data.Maybe (listToMaybe)
 import Data.Set qualified as Set
+import Data.String.Conversion (toString)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Data.Yaml (FromJSON, Object, Parser, (.!=), (.:), (.:?))
@@ -91,7 +93,7 @@ import Path (Abs, File, Path)
 --      @
 --
 --    * Key of `packages` refer (e.g. "/buffer@4.9.2") denotes name of dependency and resolved version using '@' separator
---        - For dependency resolved via registry resolver, format is: "/${dependencyName}@${resolvedVersion}${peerDepsInBrackets}".
+--        - For dependency resolved via registry resolver, format is: "/${dependencyName}@${resolvedVersion}${peerDepsInParenthesis}".
 --      @
 --      >   /ieee754@1.1.13:
 --      >      resolution: {integrity: sha512...}
@@ -116,16 +118,14 @@ data PnpmLockfile = PnpmLockfile
   }
   deriving (Show, Eq, Ord)
 
-data PnpmLockFileVersion = PnpmLock5 | PnpmLock6 deriving (Show, Eq, Ord)
+data PnpmLockFileVersion
+  = PnpmLock4Or5
+  | PnpmLock6
+  deriving (Show, Eq, Ord)
 
 instance FromJSON PnpmLockfile where
   parseJSON = Yaml.withObject "pnpm-lock content" $ \obj -> do
-    rawLockFileVersion :: TextLike <- obj .:? "lockfileVersion" .!= (TextLike mempty)
-    let lockFileVersion =
-          if Text.isPrefixOf "6" (unTextLike rawLockFileVersion)
-            then PnpmLock6
-            else PnpmLock5
-
+    rawLockFileVersion <- getVersion =<< obj .:? "lockfileVersion" .!= (TextLike mempty)
     importers <- obj .:? "importers" .!= mempty
     packages <- obj .:? "packages" .!= mempty
 
@@ -146,7 +146,13 @@ instance FromJSON PnpmLockfile where
             then Map.insert "." virtualRootWs importers
             else importers
 
-    pure $ PnpmLockfile refinedImporters packages lockFileVersion
+    pure $ PnpmLockfile refinedImporters packages rawLockFileVersion
+    where
+      getVersion (TextLike ver) = case (listToMaybe . toString $ ver) of
+        (Just '4') -> pure PnpmLock4Or5
+        (Just '5') -> pure PnpmLock4Or5
+        (Just '6') -> pure PnpmLock6
+        _ -> fail ("expected lockfileVersion: 4.x, 5.x or 6.x but, got: " <> show ver)
 
 data ProjectMap = ProjectMap
   { directDependencies :: Map Text ProjectMapDepMetadata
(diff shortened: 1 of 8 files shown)


src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -76,17 +77,55 @@ import Path (Abs, File, Path)
...
+--    * Key of `packages` refer (e.g. "/buffer@4.9.2") denotes name of dependency and resolved version using '@' separator
+--        - For dependency resolved via registry resolver, format is: "/${dependencyName}@${resolvedVersion}${peerDepsInBrackets}".
+--      @
+--      >   /ieee754@1.1.13:
+--      >      resolution: {integrity: sha512...}
+--      >      dev: false
+--      >
+--      >   /@clerk/nextjs@4.22.1(next@13.4.10)(react-dom@18.2.0)(react@18.2.0):
+--      >       resolution: {integrity: sha512...}
+--      @
```
jssblck, line 102: This says that peer dependencies are in brackets, but they are surrounded by parenthesis in the example. Which is correct? Please update accordingly.

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -76,17 +77,55 @@ import Path (Abs, File, Path)
...
+--
+--    * If project has set peerDependencies to be not auto installed, pnpm
+--      by default, does not include them in the lockfile. So, no additional
+--      work is required for newly introduced `settings.autoInstallPeers` field.
```
jssblck, line 106: I don't understand what this has to do with our usage. Can you clarify?

Specifically:
- Does FOSSA CLI need to do anything special to correctly report peer deps?
- If not, why not? If so, why so, and what does it need to do?

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -76,17 +77,55 @@ import Path (Abs, File, Path)
...
 
+data PnpmLockFileVersion = PnpmLock5 | PnpmLock6 deriving (Show, Eq, Ord)
+
 instance FromJSON PnpmLockfile where
   parseJSON = Yaml.withObject "pnpm-lock content" $ \obj -> do
+    rawLockFileVersion :: TextLike <- obj .:? "lockfileVersion" .!= (TextLike mempty)
+    let lockFileVersion =
+          if Text.isPrefixOf "6" (unTextLike rawLockFileVersion)
+            then PnpmLock6
+            else PnpmLock5
```
jssblck, line 127: This construction means that we'll parse lockfiles at v4 and below like v5, but it also means we'll parse lockfiles at v7 and above like v5 which seems incorrect.

Please compare the version in a way that properly accounts for this, unless there is a good reason for this functionality.

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -121,6 +160,18 @@ instance FromJSON ProjectMap where
...
+  { version :: Text
+  }
+  deriving (Show, Eq, Ord)
+
+instance FromJSON ProjectMapDepMetadata where
+  -- This is v5 lock format
+  parseJSON (Yaml.String r) = pure $ ProjectMapDepMetadata r
+  -- This is v6 lock format
+  parseJSON (Yaml.Object obj) = ProjectMapDepMetadata <$> obj .: "version"
+  parseJSON _ = fail "Expected to contain string or object with 'version' field!"
```
jssblck, line 173: Please print the thing that failed to parse properly to simplify later debugging.

e.g.:
```suggestion
  parseJSON other = fail ("Invalid format; expected pure string or an object with a `version` field, got: " <> show other)
```

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -294,6 +372,14 @@ buildGraph lockFile = withoutLocalPackages $
     withoutSymConstraint :: Text -> Text
     withoutSymConstraint version = fst $ Text.breakOn "_" version
 
+    -- We do not care about peer dependency version
+    -- in the version postFix, for final transformation.
+    --
+    -- >> withoutPeerDepSuffix "1.2.0" = "1.2.0"
+    -- >> withoutPeerDepSuffix "1.2.0(babel@1.0.0)" = "1.2.0"
```
jssblck, line 379: Please clarify why this is the case in the comment.

jssblck (changes requested):
The v4/v7 handling is the only hard blocker here unless there's good reason for it working the way it does (or I'm misunderstanding).

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -107,11 +149,17 @@ instance FromJSON PnpmLockfile where
             then Map.insert "." virtualRootWs importers
             else importers
 
-    pure $ PnpmLockfile refinedImporters packages
+    pure $ PnpmLockfile refinedImporters packages rawLockFileVersion
+    where
+      getVersion (TextLike ver) = case (listToMaybe . toString $ ver) of
```
meghfossa, line 154: We never supported version `<=3` previously, and there is no v7 as of yet.

jssblck: I think we should compare this as a number, and not die on v7 and above by default.
In other words, in the future v7 changes but in a way that doesn't render our parser incompatible, we shouldn't fail to parse it.

meghfossa: patched with: https://github.com/fossas/fossa-cli/pull/1320/commits/26204b4dfc6f6372c2c3885bb7c7fea77f1feadd

jssblck (approved):

PR #1323, opened 2023-11-10
[ANE-1250] Adds `fossa init` command

meghfossa:
# Overview

This PR, 
- Adds `fossa init` command
- Changes `fossa-deps`, parsing such that empty `fossa-deps` file is valid fossa-deps file

## Acceptance criteria

As a user, 
- I can create example, `.fossa.yml.example` and `fossa-deps.yml.example` with `fossa init` command

## Testing plan

```bash
git checkout feat/fossa-init-cmd-is-back
make install-dev
mkdir sandbox && cd sandbox
fossa-dev init

# current cli
fossa analyze -o  # fails because empty fossa-deps is not allowed
echo 'numpy' > reqs.txt
fossa analyze -o  # fossa-deps analysis fails, because empty fossa-deps is not allowed
rm -rf reqs.txt

# with this change!
fossa-dev analyze -o  # you should see that analysis fails, because we find no targets, not because of empty fossa-deps
echo 'numpy' > reqs.txt
fossa analyze -o  # fossa-deps analysis succeeds
```

## Risks

N/A

## Metrics

N/A

## References

https://fossa.atlassian.net/browse/ANE-1250

## Checklist


src/App/Fossa/Init/.fossa.yml
```diff
@@ -0,0 +1,69 @@
+# # fossa-cli configuration file.
```
meghfossa, line 1: Double commented, so users can `Ctrl+A` and uncomment in their code editors, while still retaining helpful links!

JeffreyHuynh1 (approved):

src/App/Fossa/Init/.fossa.yml
```diff
@@ -0,0 +1,69 @@
...
+# #   - https://support.fossa.com/
+# #
+# # -----
+
+version: 3
+# server: https://app.fossa.com
+# apiKey: a1b2c3
+#
+# project:
+#   id: github.com/fossas/fossa-cli
```
spatten, line 18: IMO, each entry in this file should have a short comment explaining what it's for.

E.g.:

```
# # Your FOSSA API key. You can obtain this by going to the settings => Integrations => API section of the FOSSA app. This will be found at `<server>/account/settings/integrations/api_tokens`.
```

src/App/Fossa/Init/fossa-deps.yml
```diff
@@ -0,0 +1,42 @@
...
+# # To learn more:
+# # - https://github.com/fossas/fossa-cli/blob/master/docs/references/files/fossa-deps.md
+# #
+# # If you need assistance or would like to file a defect, contact support at:
+# # - https://support.fossa.com/
+# #
+# # -----
+#
+#
+# referenced-dependencies:
```
spatten, line 21: Same comment here: each entry should have a short explanation of what it does and a link to the docs if it's too much to explain in a short comment.

src/App/Fossa/Init.hs
```diff
@@ -0,0 +1,102 @@
...
+
+runInit :: (Has Diagnostics sig m, Has (Lift IO) sig m, Has ReadFS sig m, Has Logger sig m) => m ()
+runInit = do
+  dir <- getCurrentDir
+  mkFossaYml dir
+  mkFossaDeps dir
+
+mkFossaYml :: (Has Diagnostics sig m, Has (Lift IO) sig m, Has ReadFS sig m, Has Logger sig m) => Path Abs Dir -> m ()
+mkFossaYml baseDir = do
+  (fossaYml, fossaYmlExist) <- checkFile (baseDir </> fossaYmlFile)
```
spatten, line 51: I know this wasn't in the ACs of the ticket, but for broker we made `broker init` also output a `.example` file. In this case it would be `.fossa.yml.example`. 

This contains exactly the same contents as the generated .fossa.yml. The benefit is that the user has the unchanged version for reference while they are editing the non-example version.

We also always overwrite the example version with the current contents. This allows us to show the user new features even there is an existing .fossa.yml or fossa-deps.yml.

spatten (approved):
I made a few comments, but I'm fine with merging this as is and continuing work in another PR. They both increase the scope of this ticket quite a bit.

IMO, making the files we generate self-documenting (by adding comments explaining each line) is a requirement for this ticket to be marked as done, but generating the example files is a nice to have.

src/App/Fossa/Init/.fossa.yml
```diff
@@ -0,0 +1,196 @@
...
+#
+#   # An external link that will appear in the FOSSA UI for this specific project.
+#   # NOTE:
+#   #   Can only be set when creating a project (running fossa analyze for the first time),
+#   #   Otherwise, they will be silently ignored.
+#   link: fossa.com
+#
+#   # The URL of your project that will appear in FOSSA.
+#   # Intended to be the URL to the repository of this project.
+#   url: github.com/fossas/fossa-clif
```
spatten, line 70: I think there's an extra `f` at the end of this URL

src/App/Fossa/Init/.fossa.yml
```diff
@@ -0,0 +1,196 @@
...
+#
+#
+# # Vendored Dependencies Configuration
+# # See documentation for details: https://github.com/fossas/fossa-cli/blob/master/docs/features/vendored-dependencies.md
+# vendoredDependencies:
+#   # If true, forces a re-scan of all vendored dependencies on every run. Default: false
+#   forceRescans: false
+# 
+#   # Determines whether vendored dependencies are scanned using "Archive Upload" or "CLI License Scan" method.
+#   # Possible values: ArchiveUpload, CLILicenseScan. Default: CLI License Scan (organization default)
```
spatten, line 120: [nit]: the default should be `CLILicenseScan`, without the spaces.

Also, I think we should expand more on "organization default". Maybe something like this (stolen from the documentation):

> The default is usually "CLILicenseScan", but your organization may have opted to default to "ArchiveUpload".

spatten (approved):
Re-reviewing for the changes to the example .fossa.yml.

It looks amazing! Two minor nits from me, but overall this is a great change

Yedfree (approved):

PR #1331, opened 2023-11-16
Ane 908 support maven filtering by scope

JeffreyHuynh1:
# Overview

Users should be able to filter Maven dependency scopes in the .fossa.yml either by inclusion or exclusion.

## Acceptance criteria

User can provide maven dependency scopes that they would like to filter by in their fossa.yml

- What does this feature look like when it is finished?
```
maven:
  scope-only:
    - compile
    - runtime
## OR
  scope-exclude:
    - provided
    - system
    - test
```

## Testing plan

Maven.CommonSpec
- Tested various cases of graph filtering based on provided scope-only and scope-exclude lists

Manually tested graph filtering in all 4 strategies and ensured filtering aligned with provided config values

## Risks

Filtering logic

## References

- [ANE-908](https://fossa.atlassian.net/browse/ANE-908)
-  [Zendesk-5913](https://fossa.zendesk.com/agent/tickets/5913)
- [Zendesk-5641](https://fossa.zendesk.com/agent/tickets/5641)
- [Zendesk-5032](https://fossa.zendesk.com/agent/tickets/5032)

## Checklist

[ANE-908]: https://fossa.atlassian.net/browse/ANE-908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Strategy/Maven/Common.hs
```diff
@@ -0,0 +1,35 @@
...
+mavenDependencyToDependency :: MavenDependency -> Dependency
+mavenDependencyToDependency MavenDependency{..} = dependency
+
+filterMavenDependencyByScope :: Set Text -> Set Text -> Graphing MavenDependency -> Graphing MavenDependency
+filterMavenDependencyByScope scopeIncludeSet scopeExcludeSet = Graphing.shrink isMavenDependencyIncluded
+  where
+    isMavenDependencyIncluded :: MavenDependency -> Bool
+    isMavenDependencyIncluded MavenDependency{..} = case (Set.null dependencyScopes, Set.null scopeIncludeSet, Set.null scopeExcludeSet) of
+      (False, True, True) -> True
+      (False, False, True) -> dependencyScopes `Set.isSubsetOf` scopeIncludeSet
```
JeffreyHuynh1, line 30: As it stands, if a dependency has a multi scope of : ["compile" , provided"] 

and the scope inclusion set is ["compile"]

Then the dependency would be filtered out. Should this be the intended behavior here?

src/App/Fossa/Config/ConfigFile.hs
```diff
@@ -257,6 +259,12 @@ newtype ExperimentalGradleConfigs = ExperimentalGradleConfigs
   {gradleConfigsOnly :: Set Text}
   deriving (Eq, Ord, Show)
 
+data MavenScopeConfigs = MavenScopeConfigs
```
meghfossa, line 262: [Question] Looking at the ticket, it says we want to support EITHER only-scope or exclude-scopes, (similar to how you have it in docs), but this type would mean, user can defined both. 

If we do not want user to define both, we should use Sum Type to restrict possible scenario - it's either include or exclude! 
```
data MavenScopeConfig = MavenScopeOnlyConfig (Set Text) | MavenScopeExcludeConfig (Set Text)
```

docs/references/files/fossa-yml.md
```diff
@@ -303,6 +303,33 @@ Found target: yarn@prod/docker
...
+
+#### scope-exclude:
+
+The list of `exclude` scopes that you would like to exclude from scanning
+
+
+```yaml
+version: 3
+
+maven:
```
meghfossa, line 322: [Required] You should update spec file `fossa-yml.v3.schema.json`. You can use the schema validator online to confirm your schema works for changes. I use: https://www.jsonschemavalidator.net/ 

These files are used by various IDEs (and plugins) to provide auto complete, when user is editing .fossa-* files

src/App/Fossa/Config/Analyze.hs
```diff
@@ -447,6 +451,17 @@ mergeStandardOpts maybeConfig envvars cliOpts@AnalyzeCliOpts{..} = do
     <*> pure grepOptions
     <*> pure customFossaDepsFile
 
+collectMavenScopeFilters ::
+  ( Has Diagnostics sig m
+  ) =>
+  Maybe ConfigFile ->
+  m MavenScopeFilters
+collectMavenScopeFilters maybeConfig = do
+  let cfgMavenScopeFilters = maybe mempty collectConfigMavenScopeFilters maybeConfig
```
meghfossa, line 460: [Prefer, Optional] you can simplify by doing: `pure $ maybe mempty collectConfigMavenScopeFilters maybeConfig`

src/App/Fossa/Container/Sources/DockerArchive.hs
```diff
@@ -163,6 +163,7 @@ analyzeLayer systemDepsOnly filters capabilities osInfo layerFs tarball = do
           (projectResults, ()) <-
             runTarballReadFSIO layerFs tarball
               . runReader noExperimental
+              . runReader (mempty :: MavenScopeFilters)
```
meghfossa, line 166: [Optional] use `noMavenScopeFiltering` and provide comment to signify why we do not care about maven scope filtering here!

src/Strategy/Maven.hs
```diff
@@ -64,29 +66,47 @@ getDeps ::
...
   m DependencyResults
 getDeps (MavenProject closure) = do
+  includeScopeFilters <- asks includeScope
+  excludeScopeFilters <- asks excludeScope
+  let includeScopeFilterSet = scopes includeScopeFilters
+      excludeScopeFilterSet = scopes excludeScopeFilters
+
   (graph, graphBreadth) <- context "Maven" $ getDepsDynamicAnalysis closure <||> getStaticAnalysis closure
+  let filteredGraph = filterMavenDependencyByScope includeScopeFilterSet excludeScopeFilterSet graph
+      newGraph = gmap mavenDependencyToDependency filteredGraph
```
meghfossa, line 81: [Prefer] Maybe `finalizedGraph` instead of `newGraph`. I think `new` is slightly confusing.

src/Strategy/Maven/Pom.hs
```diff
@@ -64,19 +65,30 @@ data MavenLabel
...
+              }
+      MavenDependency dep (Set.fromList [])
+
+    applyLabelToMavenDep :: MavenLabel -> MavenDependency -> MavenDependency
+    applyLabelToMavenDep lbl mavenDep = do
+      case lbl of
+        MavenLabelScope scope ->
+          mavenDep{dependency = depWithLabel, dependencyScopes = Set.insert scope (dependencyScopes mavenDep)}
+        _ -> mavenDep{dependency = depWithLabel}
+      where
```
meghfossa, line 90: [Prefer] You should avoid (as much as possible) of having multiple where clauses, for readability, 

```haskell
someMethod = ...
    where
        someCall = ... 
            where
                someOtherCall
```

src/Strategy/Maven/Pom.hs
```diff
@@ -64,19 +65,30 @@ data MavenLabel
...
+        depWithLabel = applyLabel lbl (dependency mavenDep)
```
meghfossa, line 91: [Prefer, Optional] provide type inference for readability

src/App/Fossa/ListTargets.hs
```diff
@@ -81,6 +81,7 @@ listTargetsMain ListTargetsConfig{..} = do
     . withTaskPool capabilities updateProgress
     . runAtomicCounter
     . runReader experimental
+    . runReader (mempty :: MavenScopeFilters)
```
meghfossa, line 84: [Optional, Prefer] comment on why we do not care about scope filters for list-targets

src/Discovery/Filters.hs
```diff
@@ -204,6 +238,12 @@ comboInclude = FilterCombination
 comboExclude :: [TargetFilter] -> [Path Rel Dir] -> FilterCombination Exclude
 comboExclude = FilterCombination
 
+setInclude :: (Set Text) -> FilterSet Include
```
meghfossa, line 241: [Optional] avoid braces for single arg. I think we should add this on hlint.

test/Maven/CommonSpec.hs
```diff
@@ -0,0 +1,162 @@
...
+      -- Dep()
+      -- After filtering with includeSet = Set("compile") and excludeSet = Set("test")
+      -- Dep()
+      let dep = createDepFromScopes []
+      let includeSet = Set.fromList ["compile"]
+      let excludeSet = Set.fromList ["test"]
+      let graph = Graphing.directs [dep]
+      let graph' = filterMavenDependencyByScope includeSet excludeSet graph
+
+      expectDirect [dep] graph'
```
meghfossa, line 51: should this not be empty, since we specified `scope-only: compile`

meghfossa (changes requested):
In the documentation, we explicitly say we can either have `scope-only` OR `scope-exclude`, operative word being `OR` - but the current code allows both. 

If we want to _allow both_, we should document following:

1. what happens when `scope-only: a` and `scope-exclude: a`
2. what happens to a dependency which has `scope: a, b`, for `scope-only: a` 
3. what happens to a dependency which has `scope: a, b`, for `scope-exclude: a`

docs/references/files/fossa-yml.md
JeffreyHuynh1: good catch thank you for that!

src/App/Fossa/Container/Sources/DockerArchive.hs
JeffreyHuynh1: Ack will reformat. 

Question, the reason why I set MavenScopeFilters to mempty was because I was following the same notation as noExperiemental where they have the experiemental configs set empty. Is it safe to assume that mavenScopeFiltering isn't allowed when analyzing from a docker archive or did I misinterpret this?

src/Strategy/Maven/Pom.hs
JeffreyHuynh1: That's a good point. What are the cases in which having multiple where clauses is acceptable? Just curious bc I see it done [here](https://github.com/fossas/fossa-cli/blob/bcd1142515684b565ccc8f67f4e91d61cc2e6d5b/src/Strategy/Maven/Pom.hs#L98C1-L98C1)

src/Strategy/Maven/Pom.hs
JeffreyHuynh1: Removing function in where clause so no need to provide type inference here

src/Strategy/Maven.hs
```diff
@@ -64,29 +66,47 @@ getDeps ::
...
 
 getDeps' ::
   ( Has (Lift IO) sig m
   , Has Diagnostics sig m
+  , Has (Reader MavenScopeFilters) sig m
   ) =>
   MavenProject ->
   m DependencyResults
 getDeps' (MavenProject closure) = do
+  includeScopeFilters <- asks includeScope
```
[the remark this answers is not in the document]
JeffreyHuynh1: Good point!

src/App/Fossa/Config/ConfigFile.hs
JeffreyHuynh1: I spoke to David for clarification on the AC. 

He wrote:
- Hey! I think we can do both only and exclude in the path and target filtering, so it would make sense if we could do both to keep the behavior? In practice, I’d imagine that only one or the other is used
-  i think its 100% fine if theyre mutually exclusive.


With that being I think it makes sense to make it EITHER scope-only or scope-exclude. I don't see the benefit of supporting both and it will declutter the code substantially

test/Maven/CommonSpec.hs
JeffreyHuynh1: good point, I think you're right about this

docs/references/strategies/languages/maven/maven.md
```diff
@@ -33,6 +33,24 @@ The mavenplugin and treecmd tactic can result in transitive dependencies which d
 ```
 `httpclient` will appear as a transitive dependency in the FOSSA UI, but it will not have any paths. There are a few things that contribute to this happening. `httpclient`'s only listed parent is `restassured` which is a `test` dependency, however, `httpclient` is a `compile`. This tells us that `httpclient` has another parent in the graph, but we are unable to determine where.
 
+## Filtering by Maven Dependency Scope 
+
+You can use [configuration file](../../../files/fossa-yml.md) to provide maven dependency scopes that you would like to filter. You can filter by either inclusion or exclusion. When both scope-only and scope-exclude are provided, scope-only will take precedence and be used for filtering.
```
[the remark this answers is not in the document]
JeffreyHuynh1: This is true and have tested this as well!

Yedfree (approved):

PR #1334, opened 2023-11-24
[ANE-949] Deep / Transitive dependencies for setuptool projects

jagonalez:
# Overview

PR adds support for determining deep / transitive dependencies for setuptool python projects (seutp.py, setup.cfg, requirements.txt). This is done by shelling out to `python -m pip list` and `python -m pip show`. `pip list` will give us a list of installed packages, where `pip show` will give us package dependencies. If neither python nor pip are installed then we fall back to current implements.

For req*.txt - only transitive dependencies for packages found in a req*.txt file will be reported. 
For setup.py/setup.cfg, we attempt to parse the name, if a name is found we attempt to find a package installed with the same name and add transitives dependencies for those dependencies. If we cannot parse the name, or do not find a package installed with the same name, we fall back to only reporting the direct dependencies parsed in the config file.

## Acceptance criteria

- transitive dependencies will be reported for python packages which have been installed in an environment
- if no packages are installed then only direct dependencies are reported

## Testing plan

- ...

## Risks

if users are not running fossa-cli within a virtual environment then there is a chance that we will report incorrect transitive dependencies, pip list ignores the require virtual env flag.

## Metrics

## References

_Example:_

## Checklist


src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+toPackages packagesInfo = do
+  map (toPackage [] packagesInfo) packagesInfo
+  where
+    toPackage seen metadata (PackageMetadata name version requires) =
+      Package
+        name
+        version
+        (mapMaybe (findPackageByName seen metadata) requires)
+
+    findPackageByName seen metadata pkgName = do
```
meghfossa, line 112: [blocking] [Preferred] We should use cannonical python package names!

```
echo 'nUmPy' > reqs.txt
pip install -r reqs.txt #install numpy package!
```

src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+data InstalledPackage = InstalledPackage
+  { name :: Text
+  , version :: Text
+  }
+  deriving (Show, Eq, Ord, Generic, FromJSON)
+
+data PackageMetadata = PackageMetadata Text Text [Text]
+  deriving (Show, Eq, Ord)
+
+data Package = Package
```
meghfossa, line 39: [Optional][Minor] use `PythonPackage` instead of `Package` since we are exporting this symbol.

src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+  , pkgVersion :: Text
+  , requires :: [Package]
+  }
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+pythonPip :: [Text] -> Command
+pythonPip args =
+  Command
+    { cmdName = "python"
+    , cmdArgs = ["-m", "pip", "--require-virtualenv"] <> args
```
meghfossa, line 50: [Optional] Add code comment with args for future readability/docs
```
cmdArgs = [
"-m", 
"pip", 
"--require-virtualenv" # Allow pip to only run in a virtual environment; exit with an error otherwise.
] <> args
```

docs/references/strategies/languages/python/setuptools.md
```diff
@@ -11,6 +11,19 @@ approach to dependency management in python projects.
 
 ## Analysis: requirements.txt
 
+For the CLI to identify transitive dependencies from a requirements.txt file,
```
meghfossa, line 14: [Prefer] We include some text around commands we are using to infer deep dependencies: `pip show`, etc.

src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+    toPackage seen metadata (PackageMetadata name version requires) =
+      Package
+        name
+        version
+        (mapMaybe (findPackageByName seen metadata) requires)
+
+    findPackageByName seen metadata pkgName = do
+      let pkg = find (\(PackageMetadata name _ _) -> name == pkgName) metadata
+      case pkg of
+        Just p -> do
```
meghfossa, line 115: [Optional] [Minor] I don't think we need `do`

src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+    -- Requires that a space trails the identifier.
+    ident :: Parser Text
+    ident = lexeme $ toText <$> takeWhileP Nothing (/= '\n')
+
+-- | Maps package metadata into a package which contains the package's transitive dependencies.
+toPackages :: [PackageMetadata] -> [Package]
+toPackages packagesInfo = do
+  map (toPackage [] packagesInfo) packagesInfo
+  where
+    toPackage seen metadata (PackageMetadata name version requires) =
```
meghfossa, line 106: [Optional] [Prefer] include type annotation for where block functions, for readability. 

```
toPackage :: ...
findPackageByName :: ...
```

src/Strategy/Python/Util.hs
```diff
@@ -37,17 +48,56 @@ reqToDependency req =
...
+buildGraphSetupFile maybePackages pyPackageName pyReqs cfgPackageName cfgReqs = do
+  Graphing.gmap reqToDependency $ do
+    case maybePackages of
+      Nothing -> Graphing.fromList (pyReqs ++ cfgReqs)
+      Just packages -> do
+        run . evalGrapher $ do
+          addDeps packages pyPackageName pyReqs
+          addDeps packages cfgPackageName cfgReqs
+  where
+    addDeps packages maybeName reqs = do
```
meghfossa, line 67: [Optional] [Minor] Include type annotations for readability!

src/Strategy/Python/Pip.hs
```diff
@@ -0,0 +1,126 @@
...
+  { pkgName :: Text
+  , pkgVersion :: Text
+  , requires :: [Package]
+  }
+  deriving (Show, Eq, Ord, Generic, FromJSON, ToJSON)
+
+pythonPip :: [Text] -> Command
+pythonPip args =
+  Command
+    { cmdName = "python"
```
meghfossa, line 49: [Question] Lot's of older linux distro, have `python` defaulting to `python2`, and `python3` defaulting to `python3`. How do we address this scenario. For instance, if customer is using older distro, we will likely not report edges, since `python` will default to `python2`? We should either add this in FAQ. We should also provide workaround for our customers as well.

https://stackoverflow.com/questions/64801225/python-or-python3-what-is-the-difference

src/Strategy/Python/Setuptools.hs
```diff
@@ -36,11 +38,13 @@ import Types (
   GraphBreadth (Partial),
  )
 
-discover :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters) sig m) => Path Abs Dir -> m [DiscoveredProject SetuptoolsProject]
-discover = simpleDiscover findProjects mkProject SetuptoolsProjectType
+discover :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters) sig m, Has Exec sig m) => Path Abs Dir -> m [DiscoveredProject SetuptoolsProject]
+discover = do
```
meghfossa, line 42: [Optional] I think we can unalive the `do`

meghfossa (commented):
Added some comments as I'm working through it - would be able to wrap up review later today.

src/Strategy/Python/Pip.hs
meghfossa: I can't think of any other language ecosystem we support which has multiple versions of interpreter installed on system. I think adding docs should suffice. On worst case scenario, customers can modify path variable for fossa cmd.

src/Strategy/Python/Pip.hs
jagonalez: good catch - these are technically the canonical names, but since pip is case insensitive when comparing the package names i'll modify the logic to lowercase each package name. This way we report the canonical python name for transitive dependencies.

src/Strategy/Python/Pip.hs
jagonalez: I believe the work around is to use a virtualenv. I've updated the docs to encourage customers to use a virtual environment. Since we're using `python -m pip` that should allow them to specify the python version in the virtual env.

jagonalez: Oh and I did test this with python2.7, `pip list` and `pip show` do work with python2.7 so we should still report edges.

docs/references/strategies/languages/python/setuptools.md
```diff
@@ -58,7 +94,7 @@ true `install_requires` list from the CLI's view. For example, FOSSA CLI does no
 
 ## Examples
```
meghfossa, line 95: [Optional] Let's add this in FAQ explicitly (I found that this helps a lot in triaging/confirming behaviour for on-call and sales/support team)

## F.A.Q

### Why is `fossa-cli` reporting incorrect transitive dependencies or edges?

`fossa-cli` uses `python -m pip list` and `python -m pip show` command to infer
transitive dependencies, and it's edges. It's paramount that you are in project's [venv](https://docs.python.org/3/tutorial/venv.html)
prior to invoking `fossa analyze` command. If you are not in project's virtual environment, 
`fossa-cli` will use global python environment to infer transitive dependencies, 
edges, and dependency versions.

For example, you should always invoke `fossa-cli` in project's python's virtual environment. 
```bash
python -m venv ~/envs/my-venv
source ~/envs/my-venv/bin/activate # this command may differ depending on your python version!
pip install -r requirements.txt

fossa analyze
```

If you are unable to get any transitive dependencies, ensure you can run following command in project's directory: 
- `python -m pip list --format json` 
- `python -m pip show`

meghfossa (changes requested):
- Gr8 work, this will have noticeable impact for our python users!
- One more change (mostly on discovery end)!

Other bits, 
- Add integration test (since this will be heavily used by our python users)
```haskell
-- integration-test/Analysis/SetuptoolsSpec.hs

-- Here is templated fixture, it may be easier to simply 
-- perform fixture/test without below pattern, if you are
-- running into issue with nixEnv, and buildCommand

-- To test this locally:
-- cabal test integration-tests --test-show-details=streaming --test-option=--format=checks --test-option=--match --test-option="wildCardForTest"
pythonDynamic :: AnalysisTestFixture (Setuptools.SetuptoolsProject)
pythonDynamic =
  AnalysisTestFixture
    "pythonDynamic" 
    Setuptools.discover
    LocalEnvironment -- You may have to use nix env and build cmd
    Nothing -- Build command pip install, and virtual env, etc, 
    $ FixtureArtifact
      "https://github.com/pallets/flask/archive/refs/tags/2.0.2.tar.gz"
      [reldir|python/setuptools/flask/|]
      [reldir|flask-2.0.2/|]

spec :: Spec
spec = do
  testSuiteDepResultSummary pythonDynamic SetuptoolsProjectType $
  DependencyResultsSummary 
     numDeps 
     numDirectDeps 
     numEdges 
     numManifestFiles 
     Complete
  ```

- Update changelog before merge!
- [Optional] Include example project in [example-project](https://github.com/fossas/example-projects) repo

PR #1342, opened 2023-12-08
Fix cabal cache store

csasarak:
# Overview

I did not realize that newer versions of cabal [moved the cabal cache directory](https://github.com/haskell/cabal/blob/master/release-notes/cabal-install-3.10.1.0.md#significant-changes) in cases where one didn't exist already when I upgraded `cabal`. Because of this, none of our Linux builds seem to be using a cabal cache. You can see an example [here](https://github.com/fossas/fossa-cli/actions/runs/7095962102/job/19313707689). No cache is found:

Screenshot 2023-12-08 at 3 00 25 PM

But no cache is stored either:
Screenshot 2023-12-08 at 3 00 44 PM

 Cabal on linux now will use `.cabal/*` if one already exists, but otherwise defer to the [XDG_CACHE_STATE](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) for caching dependency builds data:
> $XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used. 

Since there are no Linux [caches](https://github.com/fossas/fossa-cli/actions/caches) anymore, this branch aside, cabal is always placing built packages in `~/.local/state/cabal` which is why caching fails with the above warning when the build tries to save the data at `~/.cabal/store`. 

This PR edits our CI to use that directory for the cabal store on Linux. Windows/MacOS are unaffected because they get their cabal store dir from `actions/setup-haskell`. 

## Acceptance criteria

Linux builds, including benchmarks, integration tests, and lint use cabal caches correctly again on Linux.

## Testing plan

I made this PR and then checked the builds.  For a Linux build [here](https://github.com/fossas/fossa-cli/actions/runs/7147648446/job/19467529346?pr=1342) you can see the the cache is downloaded in the `Cache Cabal Store` step and the build only builds first party code. In an [earlier build](https://github.com/fossas/fossa-cli/actions/runs/7146627153/job/19464685526) you can see where the cache was saved.

Compare this with a [linux build from master](https://github.com/fossas/fossa-cli/actions/runs/7145795590/job/19462213569), which both fails to use a cache for the build and fails to save one at the end. 

## Checklist


JeffreyHuynh1:
This looks good to me! Seems like we're just updating the path of the cache and I looked through the links for the caching portion of the builds and everything checks out.

JeffreyHuynh1:
Only question is should we be documenting this somewhere so that others are aware?

JeffreyHuynh1 (approved):

csasarak:
> Only question is should we be documenting this somewhere so that others are aware?

That's a really good question to ask, @JeffreyHuynh1 . Do you have anything in mind? 

I can't really think of a message that would have helped me solve the problem faster or that I think will make this less likely to happen in the future.

PR #1348, opened 2023-12-19
Parallelize integration tests

csasarak:
# Overview

After caching dist-newstyle and improving build times, the slowest part of our CI process became the integration tests. The test portion of the integration tests would generally take over [ten](https://github.com/fossas/fossa-cli/actions/runs/7213720369/job/19654202611) [minutes](https://github.com/fossas/fossa-cli/actions/runs/7199926736/job/19612716630) to [finish](https://github.com/fossas/fossa-cli/actions/runs/7185975894/job/19570381637).

The changes on this branch usually take less than ten. See [this]() build for an example. Older builds on this branch have different config values, but generally show a [similar](https://github.com/fossas/fossa-cli/actions/runs/7267136295/job/19800283112) [speedup](https://github.com/fossas/fossa-cli/actions/runs/7266941904/job/19799732211). The overall effect here is that integration tests should go just about as fast as our build steps. I expect our full CI pipeline to _generally_ be able to complete in 15 minutes or less if this change merges.

When comparing runtimes for this PR branch to master, make sure to only compare the "Run all integration tests" step. Because of changes to `dist-newstyle` caching comparing the overall time it takes the whole workflow to complete against commits to master is an apples to oranges comparison. 

## Acceptance criteria

Integration tests should run faster.

## Testing plan

I ran integration tests multiple times in and got success consistently.

## Risks

Risks are that there are tests which step on each other's state. One place where I noticed this was when extracting tools to temporary directories. Occasionally tests would clobber each other's work when using time-stamps to add uniqueness. I've tried to address this specific instance but issues like that could appear in other places. Because bugs of this kind of difficult to reproduce it's possible there are additional bugs in our integration test-suite I haven't addressed here.

I think though that this risk is OK. In the worst case, disabling the parallelism is pretty easy. It's also possible to switch individual portions of the test-suite back to running sequentially with the `sequential` function. I'd like some advice from a reviewer if they think this risk is worth it. 

## Checklist

### DIFF
--- a/.github/workflows/integrations-test.yml
+++ b/.github/workflows/integrations-test.yml
@@ -122,4 +122,4 @@ jobs:
 
     - name: Run all integration tests
       run: |
-        cabal test --project-file=cabal.project.ci.linux --test-show-details=direct --test-option=--times integration-tests --test-options="-j 12"
+        cabal test --project-file=cabal.project.ci.linux --test-show-details=direct --test-option=--times integration-tests --test-options="-j 32"
--- a/.github/workflows/integrations-test.yml
+++ b/.github/workflows/integrations-test.yml
@@ -122,4 +122,4 @@ jobs:
 
     - name: Run all integration tests
       run: |
-        cabal test --project-file=cabal.project.ci.linux --test-show-details=direct --test-option=--times integration-tests
+        cabal test --project-file=cabal.project.ci.linux --test-show-details=direct --test-option=--times integration-tests --test-options="+RTS -N -RTS"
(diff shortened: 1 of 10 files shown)


spatten (approved):
This all looks good and safe to me and gives a nice speedup.

The only thing I think we should add is a comment explaining how to disable parallel execution of the integration tests. Is it as simple as removing `--test-options="+RTS -N -RTS"` from the `cabal test` invocation?

Something like:

```
# `--test-options="+RTS -N -RTS"` enables parallel execution of the integration tests. To disable parallel execution, just remove it.
```

csasarak:
> The only thing I think we should add is a comment explaining how to disable parallel execution of the integration tests. Is it as simple as removing `--test-options="+RTS -N -RTS"` from the `cabal test` invocation?

Good idea, I've added a comment describing this.

PR #1349, opened 2023-12-20
1/2 Removing SCM warnings

JeffreyHuynh1:
# Overview

We are currently showing SCM warnings, but we want to avoid this in certain situations.

TO DO:
- unit tests
- info level logging 

## Acceptance criteria

SCM warnings never result in an error/warning because the CLI works fine without them, it just uses a date for a revision.

If the SCM warning would normally occur and the user has not provided a version with --revision, output an info message noting that we didn’t get a revision with --revision and couldn’t automatically discover one from SCM and will use a date instead.

The above should apply to both fossa analyze, fossa test, and fossa report .

## References

- [ANE-1255](https://fossa.atlassian.net/browse/ANE-1255)

## Checklist

[ANE-1255]: https://fossa.atlassian.net/browse/ANE-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Config/Common.hs
+++ b/src/App/Fossa/Config/Common.hs
@@ -302,10 +302,16 @@ collectRevisionData (BaseDir basedir) maybeConfig cacheStrategy cliOverride = do
   let override = collectRevisionOverride maybeConfig cliOverride
   case cacheStrategy of
     ReadOnly -> do
-      inferred <- inferProjectFromVCS basedir <||> inferProjectCached basedir <||> inferProjectDefault basedir
+      maybeInferred <- inferProjectFromVCS basedir
+      inferred <- case maybeInferred of
+        Nothing -> inferProjectCached basedir <||> inferProjectDefault basedir
+        Just inferredFromVCS -> pure inferredFromVCS
       pure $ mergeOverride override inferred
     WriteOnly -> do
-      inferred <- inferProjectFromVCS basedir <||> inferProjectDefault basedir
+      maybeInferred <- inferProjectFromVCS basedir
+      inferred <- case maybeInferred of
+        Nothing -> inferProjectDefault basedir
+        Just inferredFromVCS -> pure inferredFromVCS
       let revision = mergeOverride override inferred
       saveRevision revision
       pure revision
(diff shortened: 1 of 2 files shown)


src/App/Fossa/Config/Common.hs
```diff
@@ -302,10 +302,16 @@ collectRevisionData (BaseDir basedir) maybeConfig cacheStrategy cliOverride = do
...
-      inferred <- inferProjectFromVCS basedir <||> inferProjectCached basedir <||> inferProjectDefault basedir
+      maybeInferred <- inferProjectFromVCS basedir
+      inferred <- case maybeInferred of
+        Nothing -> inferProjectCached basedir <||> inferProjectDefault basedir
+        Just inferredFromVCS -> pure inferredFromVCS
       pure $ mergeOverride override inferred
     WriteOnly -> do
-      inferred <- inferProjectFromVCS basedir <||> inferProjectDefault basedir
+      maybeInferred <- inferProjectFromVCS basedir
+      inferred <- case maybeInferred of
```
csasarak, line 312: Case analysis can often be simplified with the [`maybe`](https://hackage.haskell.org/package/base-4.19.0.0/docs/Prelude.html#v:maybe) function:
```haskell
maybe :: b -> (a -> b) -> Maybe a -> b
```
The type describes what it does: It takes a default value, and a function that can be applied inside a `Maybe` to produce a value of that type. If the `Maybe a` is `Nothing` then it returns the default, otherwise it applies the function to `a`.

```haskell
inferred <- maybe (inferProjectDefault baseDir) pure (inferProjectFromVCS basedir)
```
should be equivalent. Keep in mind that Haskell is lazy, so `maybe` is able to return a value in the `Just` case without evaluating the default argument. If you look at the cases in its [source code](https://hackage.haskell.org/package/base-4.19.0.0/docs/src/Data.Maybe.html#maybe) it may be clearer why. 

Not directly useful here, but good to know is also the [`fromMaybe`](https://hackage.haskell.org/package/base-4.19.0.0/docs/Data-Maybe.html#v:fromMaybe) function.

csasarak (changes requested):
I've gone through this and left some commentary. Please let me know if you want to chat about any of my comments or alternate solutions. I think the solution with `errorBoundary` might be the way to go here, but also want to make sure you're fully on-board with it as well.

PR #1350, opened 2023-12-20
Kill pathfinder

csasarak:
# Overview

This PR:

* Kills the pathfinder binary.
* Moves the `LicenseAnalyzeProject` into its own module. This typeclass is for declared license scanning something which I think we've discussed wanting in the future, so I didn't delete.

I've put these changes in an `Unreleased` block in our readme because while this is a user-visible change I don't think that it warrants a new release. I can change this if a reviewer disagrees since we generally try to release things ASAP.

## Acceptance criteria

Pathfinder is gone.

## Risks

A customer somewhere was using `pathfinder` and we didn't know. It hasn't been updated for a very long time though so they can still get it from an old release if this is the case.

## References

- [ANE-1239](https://fossa.atlassian.net/browse/ANE-1239): Kill Pathfinder

## Checklist

[ANE-1239]: https://fossa.atlassian.net/browse/ANE-1239?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App/Fossa/Analyze/LicenseAnalyze.hs
```diff
@@ -1,4 +1,4 @@
-module App.Pathfinder.Types (
+module App.Fossa.Analyze.LicenseAnalyze (
```
JeffreyHuynh1, line 1: Do you think it's worthwhile to write a comment here as to why we're keeping `LicenseAnalyzeProject` and moving it into its own module?

JeffreyHuynh1:
All the changes here look good! 

I was a little curious about PathFinder though because I've never worked with it. What was the intended purpose of PathFinder and why is it that we're deleting it now? I understand that there hasn't been any recent work done with PathFinder, but aside from that, that's the only thing I know about it.

JeffreyHuynh1 (approved):

src/App/Fossa/Analyze/LicenseAnalyze.hs
csasarak: Good question, and yes. I've done this.

csasarak:
> I was a little curious about PathFinder though because I've never worked with it. What was the intended purpose of PathFinder and why is it that we're deleting it now?

It was an executable that would check manifest files (like `Cargo.toml`, `pom.xml`, etc.) for declared licenses. I'm not exactly sure what the history of it being a separate binary and not a part of regular `fossa` is.

PR #1352, opened 2023-12-20
Fix Conflicting Types for "revision" in fossa-yml schema

gilfaizon:
# Overview

I've been spending a lot of time in the .fossa.yml file lately, and the erroneous red squiggly has become increasingly bothersome.

This change prevents the "Expected Array" error message from appearing when using the `revision` key.

When pasting our [example](https://github.com/fossas/fossa-cli/blob/master/docs/references/files/fossa-yml.md) .fossa.yml deps file, the following occurs:
image

Ideally, we'd have no errors, as this is valid and accepted by fossa-cli.

When attempting to use an array, the CLI returns the following error:
```
[ERROR] ----------
  An issue occurred

  >>> Relevant errors

    Error

      Error parsing file: /path/to/.fossa.yml.

          Aeson exception:
          Error in $.revision: parsing ConfigRevision failed, expected Object, but encountered Array
```

After the change:
image

## Acceptance criteria

Using the revision key does not throw an error, everything else works as intended.

## Testing plan

1. Checkout `master`
2. Point your VSC/IDE yaml schemas to the appropriate file.
    - (This can be done by adding the following to your settings.json)
```json
        "yaml.schemas": {
            "file:/path/to/fossa-cli/docs/references/files/fossa-deps.schema.json": "fossa-deps.yml",
            "file:/path/to/fossa-cli/docs/references/files/fossa-yml.v3.schema.json": ".fossa.yml"
        },
```
3. Create a .fossa.yml with the following:
```yaml
version: 3

revision:
  commit: "12345"
  branch: master
```
4. Observe the error
5. Checkout this branch
6. Reload VSC
7. Error resolved!

## Risks

I think this change is fine because the type is declared in [another location](https://github.com/fossas/fossa-cli/blob/master/docs/references/files/fossa-yml.v3.schema.json#L104-L106), but I don't know if the best practice is to change the type at this (PR) location to `object` or to remove it completely. Still, I guess that's what code reviews are for!

## Checklist


gilfaizon:
> I don't know if the best practice is to change the type at this (PR) location to object or to remove it completely.

This made me curious, so I tested it.

Before the latest commit, arrays are "allowed" in the config but will throw an error in the CLI.
image

After the latest commit, arrays will throw errors, hopefully minimizing the situation where the CLI 
will error fatally:
image

I think this is the best-case scenario.

spatten (approved):
This looks like the correct change to me.

Thanks for fixing this!

PR #1354, opened 2024-01-02
Removing SCM warnings

JeffreyHuynh1:
# Overview

We are currently showing SCM warnings, but we want to avoid this in certain situations.

## Acceptance criteria

SCM warnings never result in an error/warning because the CLI works fine without them, it just uses a date for a revision.

If the SCM warning would normally occur and the user has not provided a version with --revision, output an info message noting that we didn’t get a revision with --revision and couldn’t automatically discover one from SCM and will use a date instead.

The above should apply to both fossa analyze, fossa test, and fossa report .

## Testing plan

Tested all the different error scenarios for SCM warnings and ensured that the errors weren't being populated for the various sub commands (analyze, test, report)

```
InvalidRemote -> "Missing 'origin' git remote"
GitConfigParse err -> "An error occurred when parsing the git config: " <> pretty err
MissingGitConfig -> "Missing .git/config file"
MissingGitHead -> "Missing .git/HEAD file"
InvalidBranchName branch -> "Invalid branch name: " <> pretty branch
MissingBranch branch -> "Missing ref file for current branch: " <> pretty branch
MissingGitDir -> "Could not find .git directory in the current or any parent directory"
```

Also made sure that the info level logging was being produced. See screenshot for occurrence of info level logging.

Screenshot 2024-01-02 at 1 38 49 PM

The errors from the SCM warnings are only being shown in debug mode as seen here:

Screenshot 2024-01-02 at 3 18 23 PM

## References

- [ANE-1255](https://fossa.atlassian.net/browse/ANE-1255)

## Checklist

[ANE-1255]: https://fossa.atlassian.net/browse/ANE-1255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App/Fossa/Config/Common.hs
```diff
@@ -302,17 +308,36 @@ collectRevisionData (BaseDir basedir) maybeConfig cacheStrategy cliOverride = do
...
+      ) =>
+      m InferredProject ->
+      m InferredProject
+    inferVCSInfo nextStep = do
+      vcsInfo <- errorBoundary $ inferProjectFromVCS basedir
+      case vcsInfo of
+        Failure emittedWarns errGroup ->
+          do
+            logDebug (renderFailure emittedWarns errGroup "")
+            logInfo "Unable to infer project revision from version control system. The project revision will be set to the current timestamp."
```
JeffreyHuynh1, line 333: Wanted thoughts on this logging message. Debating if I should keep 'Unable to infer project revision from version control system' because I don't want to give users the impression that our code is failing to do something as it might raise concerns and further questions from customers.

src/App/Fossa/Config/Common.hs
```diff
@@ -302,17 +308,36 @@ collectRevisionData (BaseDir basedir) maybeConfig cacheStrategy cliOverride = do
...
+      inferred <- inferVCSInfo $ inferProjectCached basedir <||> inferProjectDefault basedir
       pure $ mergeOverride override inferred
     WriteOnly -> do
-      inferred <- inferProjectFromVCS basedir <||> inferProjectDefault basedir
+      inferred <- inferVCSInfo $ inferProjectDefault basedir
       let revision = mergeOverride override inferred
       saveRevision revision
       pure revision
+  where
+    inferVCSInfo ::
```
JeffreyHuynh1, line 319: decided to put this function under a 'where' clause as opposed to having it in a 'let'  bc I felt like it was a little bit neater / readable

src/App/Fossa/Config/Common.hs
csasarak: I think this is a good question to ask and a great thing to think about. 

In this case, I would avoid calling it an "Error" since we already use that term elsewhere for problems serious enough to call into question whether we're getting the best results - our to-be-changed usage of "Error" when outputting warnings notwithstanding. This issue really doesn't affect results at all, only how we report the scan results in FOSSA. I would do something like this: "Unable to infer project revision from VCS, using current timestamp as the revision."

csasarak (approved):
LGTM. The only change I'd like to see is for you to change the language of the output message like we chatted about. Let me know if you want to discuss more.

PR #1356, opened 2024-01-05
Use `-batch` and `-no-colors` for sbt for compatibility with older SBT.

csasarak:
# Overview

An older version of `sbt`, version `1.2.8`, and the sbt dependency-graph-plugin don't understand the double-hyphen long-option option syntax. For example, they require `-batch` rather than `--batch`. 

## Acceptance criteria

* We can scan the new project from this [PR](https://github.com/fossas/example-projects/pull/13).

## Testing plan

I manually tested using current `fossa` and the version from this branch. Analysis succeeds using the plugin on this branch, but not the version from master.

## Risks

There is the risk that some newer version of `sbt` doesn't respect both syntaxes. I think that's unlikely. 

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## v3.8.30
+
+- Fix an issue with long-option syntax for older versions of `sbt` ([#1356](https://github.com/fossas/fossa-cli/pull/1356))
+
 ## v3.8.29
 - Prevents showing SCM warnings in fossa analyze, test, and report [#1354](https://github.com/fossas/fossa-cli/pull/1354)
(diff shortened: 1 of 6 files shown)


meghfossa (approved):
[Optional] Add integration with fixture: https://github.com/fossas/example-projects/pull/13/files

May need to modify `NixEnv` to allow for older versions: https://lazamar.co.uk/nix-versions/?channel=nixpkgs-unstable&package=sbt

csasarak:
Good idea, thanks! I learned that you can configure SBT at the project level and it should download the appropriate version. I think I won't have to edit the nixenv.

csasarak:
I worked on adding an integration test for the very old `sbt` but it was increasing the time it took for our integration tests to run by over a factor of 2. I think the risks are pretty low here so I'm going to move forward with merging.

PR #1360, opened 2024-01-11
debug: Add extra debug logging along Gradle analysis path

elldritch:
# Overview

This PR adds extra debug logging along the Gradle analysis path, to help solve user issues for ZD-7543.

## Acceptance criteria

Enough logging to understand where dependencies are being filtered out.

## Testing plan

I didn't, seems low risk, especially since the logging carrier explicitly ignores `SevDebug` when not in debug mode.

## Risks

I'm worried about performance implications, but `IgnoreDebugC` seems to have already taken that into account.

## References

- https://fossa.atlassian.net/browse/FDN-97
- https://fossa.zendesk.com/agent/tickets/7543

## Checklist

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -106,6 +106,7 @@ import Effect.Exec (Exec)
 import Effect.Logger (
   Logger,
   Severity (..),
+  logDebug,
   logInfo,
   logStdout,
  )
@@ -353,8 +354,10 @@ analyze cfg = Diag.context "fossa-analyze" $ do
           forkTask $ do
             res <- Diag.runDiagnosticsIO . diagToDebug . stickyLogStack . withEmptyStack $ Archive.discover (runAnalyzers filters) basedir ancestryDirect
             Diag.withResult SevError SevWarn res (const (pure ()))
+  logDebug $ "Unfiltered project scans: " <> pretty (show projectScans)
 
   let filteredProjects = mapMaybe toProjectResult projectScans
+  logDebug $ "Filtered project scans: " <> pretty (show projectScans)
 
   maybeEndpointAppVersion <- case destination of
     UploadScan apiOpts _ -> runFossaApiClient apiOpts $ do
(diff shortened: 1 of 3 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -390,6 +394,7 @@ analyze cfg = Diag.context "fossa-analyze" $ do
           (Just firstParty, Nothing) -> Just firstParty
   let keywordSearchResultsFound = (maybe False (not . null . lernieResultsKeywordSearches) lernieResults)
   let outputResult = buildResult includeAll additionalSourceUnits filteredProjects' licenseSourceUnits
+  logDebug $ "Source unit result: " <> pretty (show outputResult)
```
csasarak, line 397: The source units are already output in in the debug bundle under `bundleOutput` [here](https://github.com/fossas/fossa-cli/blob/master/src/App/Fossa/Analyze/Debug.hs#L124) I think. Is there a reason to output them here in addition to that? 

More generally, I'd consider putting some of these data as top-level keys in the [`DebugBundle`](https://github.com/fossas/fossa-cli/blob/master/src/App/Fossa/Analyze/Debug.hs#L165) or sub-keys of `bundleOutput`. They'd be easier to locate I think if that were the case.

csasarak (approved):
I'm approving to not block since I think there's low risk here. But please have a look at my comments. Thanks for adding these!

src/App/Fossa/Analyze.hs
```diff
@@ -353,8 +354,10 @@ analyze cfg = Diag.context "fossa-analyze" $ do
           forkTask $ do
             res <- Diag.runDiagnosticsIO . diagToDebug . stickyLogStack . withEmptyStack $ Archive.discover (runAnalyzers filters) basedir ancestryDirect
             Diag.withResult SevError SevWarn res (const (pure ()))
+  logDebug $ "Unfiltered project scans: " <> pretty (show projectScans)
 
   let filteredProjects = mapMaybe toProjectResult projectScans
+  logDebug $ "Filtered project scans: " <> pretty (show projectScans)
```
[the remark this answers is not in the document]
elldritch: Good catch! Ugh, this is what I get for using AI autocomplete.

src/App/Fossa/Analyze.hs
elldritch: Good suggestion. I've removed this extraneous message.

Is it alright to merge the others as messages for now? I'd like to get this into user hands ASAP, and looking at the plumbing around `collectDebugBundle` it looks like it might take some more thoughtful plumbing to get analysis messages out of the `Result`.

elldritch: Merging given Slack conversation.

PR #1362, opened 2024-01-16
Ane 797 static only flag

csasarak:
# Overview

This PR adds a way to enable static-only analysis as a CLI option. Formerly this was available only to library consumers. Additionally, it provides a way to control which kind of analysis is run as a parameter to the functions in `src/App/Fossa/Analyze.hs`. 

I changed the CLI to output a message in the scan summary stating that only static tactics were used so that this will be more obvious to support staff:

Screenshot 2024-01-16 at 4 26 26 PM

It also adds a new name for `analyseProject'`, `analyseProjectStaticOnly` which better describes the purpose and intent of that function. A compiler warning encourages consumers of `spectrometer` to use the new name. 

I didn't audit each of our `AnalyzeProject` instances to make sure that their implementations of `analyzeProject'` are actually static only. I don't think this is really needed, but if a reviewer disagrees I can do that.

## Acceptance criteria

* Customers and consumers of `spectrometer` have a way to run the CLI but only use static strategies.

## Testing plan

A simple way to see this in action is with a `Go` project:
```
fossa analyze --static-only-analysis example-projects/
```
Go doesn't have any static tactics so it fails with an error whereas running without the flag operates successfully. 

Maven on the other hand has both a static and a dynamic strategy. Using our example-projects repository we can scan with dynamic enabled:
```
fossa analyze example-projects/maven/example-maven-project/  -o
```
You'll notice in the output that the maven plugin was installed and run. Also, the message about static only analysis is not present.

Running against the same project with the static only flag:
```
fossa analyze --static-only-analysis example-projects/maven/example-maven-project/ -o
```
Does not show the plugin running and also outputs the message about doing a static only analysis. 

## Risks

The main risk is that CI customers may try to use static analysis and get poor results that they then ask us about. I've tried to point users of this feature in the right direction to understand why this is the case. 

## References

[ANE-797](https://fossa.atlassian.net/browse/ANE-797)

## Checklist

[ANE-797]: https://fossa.atlassian.net/browse/ANE-797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/references/strategies/README.md
+++ b/docs/references/strategies/README.md
@@ -123,8 +123,9 @@ See the linked documentation above for details.
 
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
-While we recommend running the tool with all possible strategies enabled if you need to limit FOSSA CLI's run to only use static strategies we offer the `--static-only-analysis` flag. 
+Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI's run to only only use static strategies the CLI offers the `--static-only-analysis` flag. 
 This flag makes the the FOSSA CLI not attempt to use any third party tools, such as `npm`, `pip`, etc.
+FOSSA CLI may still use FOSSA-developed tools that ship with the CLI binary, however.
 
 It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that has both.
 
--- a/docs/references/strategies/README.md
+++ b/docs/references/strategies/README.md
@@ -123,7 +123,10 @@ See the linked documentation above for details.
 
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
-It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that
+While we recommend running the tool with all possible strategies enabled if you need to limit FOSSA CLI's run to only use static strategies we offer the `--static-only-analysis` flag. 
+This flag makes the the FOSSA CLI not attempt to use any third party tools, such as `npm`, `pip`, etc.
+
+It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that has both.
 
 ### Strategies by type
 
--- a/docs/references/strategies/README.md
+++ b/docs/references/strategies/README.md
@@ -124,7 +124,7 @@ See the linked documentation above for details.
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
 Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI to only use static strategies the CLI offers the `--static-only-analysis` flag.
-This flag prevents the FOSSA CLI from using any third party tools, such as `npm`, `pip`, maven plugins, etc.
+This flag prevents FOSSA CLI from using any third party tools, such as `npm`, `pip`, maven plugins, etc.
 With this option enabled, strategies that don't offer a way to analyze statically will fail with an error.
 
 It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that has both.
--- a/docs/references/strategies/README.md
+++ b/docs/references/strategies/README.md
@@ -123,8 +123,9 @@ See the linked documentation above for details.
 
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
-Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI's run to only only use static strategies the CLI offers the `--static-only-analysis` flag. 
-This flag makes the the FOSSA CLI not attempt to use any third party tools, such as `npm`, `pip`, etc.
+Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI's run to only use static strategies the CLI offers the `--static-only-analysis` flag. 
+This flag prevents the FOSSA CLI from using any third party tools, such as `npm`, `pip`, maven plugins, etc.
+With this option enabled, strategies that don't offer a way to analyze statically will fail with an error.
 FOSSA CLI may still use FOSSA-developed tools that ship with the CLI binary, however.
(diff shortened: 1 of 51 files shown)


docs/references/strategies/README.md
```diff
@@ -123,7 +123,12 @@ See the linked documentation above for details.
 
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
-It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that
+Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI's run to only use static strategies the CLI offers the `--static-only-analysis` flag.
```
jssblck, line 126:
```suggestion
Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI to only use static strategies the CLI offers the `--static-only-analysis` flag.
```

docs/references/strategies/README.md
```diff
@@ -123,7 +123,12 @@ See the linked documentation above for details.
 
 Languages supported by FOSSA CLI can have multiple strategies for detecting dependencies, one primary strategy that yields ideal results and zero or more fallback strategies. Within this list of strategies, we have the concept of _static_ and _dynamic_ strategies. Static strategies parse files to find a dependency graph (example: parse a `package-lock.json` file). Dynamic strategies are required when analyzing package managers that do not offer complete lockfiles, such as Gradle or Go. Dynamic strategies require a working build environment to operate in.
 
-It is important to note that neither type of strategy has an inherent benefit when detecting dependencies. If a supported language has only a static or only a dynamic strategy, this does not mean it is less supported than a language that
+Running the tool with all possible strategies enabled is recommended, but if you need to limit FOSSA CLI's run to only use static strategies the CLI offers the `--static-only-analysis` flag. 
+This flag prevents the FOSSA CLI from using any third party tools, such as `npm`, `pip`, maven plugins, etc.
```
jssblck, line 127: [nit] FOSSA CLI is a name
```suggestion
This flag prevents FOSSA CLI from using any third party tools, such as `npm`, `pip`, maven plugins, etc.
```

docs/references/strategies/README.md
```diff
@@ -123,7 +123,12 @@ See the linked documentation above for details.
...
+With this option enabled, strategies that don't offer a way to analyze statically will fail with an error.
+FOSSA CLI may still use FOSSA-developed tools that ship with the CLI binary, however.
```
jssblck, line 129:
```suggestion
```
[nit] this doesn't seem like user-useful information. Static strategies don't imply "nothing executes", just that we're statically analyzing.

docs/references/subcommands/analyze.md
```diff
@@ -114,14 +114,15 @@ We support the following archive formats:
...
-| [`--detect-vendored`](./analyze/detect-vendored.md)              | Enable the vendored source identification engine. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md).                            |
-| [`--detect-dynamic './some-binary`](./analyze/detect-dynamic.md) | Analyze the binary at the provided path for dynamically linked dependencies. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md). |
+| [`--detect-vendored`](./analyze/detect-vendored.md)                               | Enable the vendored source identification engine. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md).                            |
+| [`--detect-dynamic './some-binary`](./analyze/detect-dynamic.md)                  | Analyze the binary at the provided path for dynamically linked dependencies. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md). |
+| [`--static-only-analysis`](../strategies/README.md#static-and-dynamic-strategies) | Do not use third-party tools when analyzing projects.
```
jssblck, line 125: [curious] why did we decide on this flag over `--strict`?

src/App/Fossa/Analyze/ScanSummary.hs
```diff
@@ -113,6 +119,14 @@ instance Pretty ScanCount where
       , pretty numWarnings <> " analysis " <> plural "warning" "warnings" numWarnings
       ]
 
+staticOnlyAnalysisMessage :: Doc a
+staticOnlyAnalysisMessage =
+  vcat
+    [ "Performed static-only analysis!!!"
```
jssblck, line 125: I appreciate the enthusiasm, but it reads very oddly in the otherwise quite dry FOSSA CLI output 😆
```suggestion
    [ "FOSSA CLI was constrained to static-only analysis."
```

src/App/Fossa/Analyze/ScanSummary.hs
```diff
@@ -113,6 +119,14 @@ instance Pretty ScanCount where
...
+    , "Results may be different in comparison to an analysis using build tools."
+    , "See " <> pretty staticAndDynamicStrategies <> " for more information."
```
jssblck, line 127: [nit] Due to the length of GitHub URLs, prefer to place it at the end:
```suggestion
    , "More information: " <> pretty staticAndDynamicStrategies
```
This way if the terminal wraps the URL the user isn't having to deal with that while parsing the sentence.

src/App/Fossa/Analyze/Types.hs
```diff
@@ -113,9 +113,20 @@ data DiscoveredProjectIdentifier = DiscoveredProjectIdentifier
...
 class AnalyzeProject a where
-  -- | Analyze a project with any tactic.
+  -- | Analyze a project with any tactic including dynamic ones.
   analyzeProject :: AnalyzeTaskEffs sig m => FoundTargets -> a -> m DependencyResults
 
   -- | Analyze a project with only static tactics.
-  analyzeProject' :: AnalyzeStaticTaskEffs sig m => FoundTargets -> a -> m DependencyResults
+  -- Static here generally means not using any external utilities.
+  -- An exception is berkeleydb which calls a program embedded in the CLI.
+  analyzeProjectStaticOnly :: AnalyzeStaticTaskEffs sig m => FoundTargets -> a -> m DependencyResults
```
jssblck, line 124: I think we should be very precise: static concretely means "statically analyzing". Whether we execute external tools is not relevant to this equation.

src/App/Fossa/Analyze/Types.hs
```diff
@@ -113,9 +113,20 @@ data DiscoveredProjectIdentifier = DiscoveredProjectIdentifier
...
+  -- This is required to satisfy the MINIMAL pragma below and for compatibility with existing consumers of AnalyzeProject.
+  -- Please don't use this default implementation for any new code.
+  analyzeProjectStaticOnly = analyzeProject'
```
jssblck, line 127: Why not just refactor callsites and drop this?

src/App/Fossa/Config/Analyze.hs
```diff
@@ -291,6 +301,7 @@ cliParser =
     <*> flagOpt ForceNoFirstPartyScans (long "experimental-block-first-party-scans" <> help "Block first party scans. This can be used to forcibly turn off first-party scans if your organization defaults to first-party scans.")
     <*> flagOpt IgnoreOrgWideCustomLicenseScanConfigs (long "ignore-org-wide-custom-license-scan-configs" <> help "Ignore custom-license scan configurations for your organization. These configurations are defined in the \"Integrations\" section of the Admin settings in the FOSSA web app")
     <*> optional (strOption (long "fossa-deps-file" <> help "Path to fossa-deps file including filename (default: fossa-deps.{yaml|yml|json})"))
+    <*> flagOpt StaticOnlyTactics (long "static-only-analysis" <> help "Do not use dynamic tools for discovering dependencies.")
```
jssblck, line 304: Again here let's be very precise about specifically what this means: we're statically analyzing the project. Whether we use external tools to do so is irrelevant.

src/App/Fossa/Analyze/Types.hs
jssblck: wait it looks like we did just refactor the callsites- why even have `analyzeProject'` still?

jssblck (changes requested):

src/App/Fossa/Analyze/Types.hs
csasarak: You're right. I was initially thinking that maybe `hubble` was using that name and didn't want to force them to do more work. I will drop it.

docs/references/subcommands/analyze.md
csasarak: I thought it was descriptive.

We also have a ticket to implement a `--strict` mode. I thought the goal for that would be to turn all warnings into errors and constrain the CLI to using only the "best" tactic for a given strategy. See this [ticket](https://fossa.atlassian.net/browse/ANE-236). I think we should reserve `--strict` for later. 

If you meant to write `--static` I think that would be fine. An invocation would be like `fossa analyze ... --static ...` which describes that it's the analysis that would be made static.

jssblck: oh wait you're right, I was just misremembering what `--strict` was meant to be!

src/App/Fossa/Analyze/Types.hs
csasarak: I've deleted `analyzeProject'`.

PR #1363, opened 2024-01-17
Adds basic reachability integration in cli (Part 1/N)

meghfossa:
# Overview

This is part of stack PR. This PR will not be merged, until all stacked PR's are merged into this branch.

This PR, 
- Adds basic command line tool for reachability (reachability/extlib)
- Adds hidden `fossa reachability <PATH>` command for debugging

## Acceptance criteria

- Automated build works
- `fossa reachability <PATH>` can be executed (It will fail in part 1 PR)

## Testing plan

```bash
git checkout master && git pull origin && git checkout feat/reachability-cli && make install-dev
```

after which,
```bash
fossa-dev reachability

# Note that this command will fail, as this PR does not add implementation
# so error is to be expected! You should see something like following in error message:
#
# Command error log:
#            DEBUG [reachability] Reading from file system: "/Users/megh/code/fossa-cli/"
#            DEBUG [reachability] Parsing file system: "/Users/megh/code/fossa-cli/"
#            thread 'main' panicked at extlib/reachability/src/lib.rs:31:9:
#            not yet implemented
#            note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

## Risks
N/A

## Metrics
N/A

## References
N/A

## Checklist

### DIFF
--- a/extlib/reachability/Cargo.toml
+++ b/extlib/reachability/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "reachability"
+version = "0.0.1"
+edition = "2021"
+
+[features]
+default = []
+
+[dependencies]
+tikv-jemallocator = { version = "0.5.4", optional = true }
+clap = { version = "4.0.0", features = ["derive"] }
+simple_logger = { version = "4.3.3", features = ["stderr", "colors"], default-features = false }
+log = "0.4.17"
+snippets = { git = "https://github.com/fossas/lib-snippets", version = "0.2.0", features = ["lang-java-11"] }
+stable-eyre = "0.2.2"
+serde = { version = "1.0.144", features = ["derive"] }
+serde_json = "1.0.85"
--- a/extlib/reachability/src/lib.rs
+++ b/extlib/reachability/src/lib.rs
@@ -1,0 +1,33 @@
+//! Parses language source code to create call graphs.
+//!
+//! # Supported Parser
+//! - Java
+use log::debug;
+use serde::Serialize;
+use snippets::content::Content;
+use stable_eyre::{eyre::Context, Result};
+use std::io::{self, Read};
+use std::path::PathBuf;
+
+#[derive(Debug, Serialize)]
+pub struct Reachability {}
+
+impl Reachability {
+    pub fn from_stdin() -> Result<Self> {
+        let mut stdin = String::new();
+        io::stdin()
+            .read_to_string(&mut stdin)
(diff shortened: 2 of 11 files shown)


extlib/reachability/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+use snippets::content::Content;
+use stable_eyre::{eyre::Context, Result};
+use std::io::{self, Read};
+use std::path::PathBuf;
+
+#[derive(Debug, Serialize)]
+pub struct Reachability {}
+
+impl Reachability {
+    pub fn from_stdin() -> Result<Self> {
```
meghfossa, line 16: I have intentionally left out implementation. This will be added in second PR.

extlib/reachability/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[features]
+default = []
+
+[dependencies]
+tikv-jemallocator = { version = "0.5.4", optional = true }
+clap = { version = "4.0.0", features = ["derive"] }
+simple_logger = { version = "4.3.3", features = ["stderr", "colors"], default-features = false }
+log = "0.4.17"
+snippets = { git = "https://github.com/fossas/lib-snippets", version = "0.2.0", features = ["lang-java-11"] }
+stable-eyre = "0.2.2"
```
spatten, line 15: I'm worried that we are using too many error handling libraries.

Broker uses `error-stack`. Lernie and Sherlock use `this-error`. Reachability-toolkit uses `anyhow`.

Now we're adding `eyre` into the mix.

If there's a good reason to use `stable-eyre` instead of one of the other three, then that's fine. But I wanted to double-check

spatten (approved):
I have one question about the proliferation about error handling libraries that we use, but I don't think that's blocking.

Other than that, looks great!

extlib/reachability/Cargo.toml
meghfossa: I copied the dependencies and code structure from `berkelydb`, but I can modify it to recent rust code one's.

spatten: Hmm. Maybe we should merge this as-is, but have a quick conversation about whether we want to standardize this a bit in standup tomorrow.

I don't think it blocks this PR; we can just make the change later if we decide to

meghfossa:
Superseded by: https://github.com/fossas/fossa-cli/pull/1372

PR #1364, opened 2024-01-17
ANE-1254: Error Structure

JeffreyHuynh1:
# Overview

As a fossa user, errors should be structured.

## Acceptance criteria

- Errors have a structure similar to broker:

Screenshot 2024-01-28 at 10 24 43 PM

## Testing plan

- Manually tested new error structure implementation 

Here are how errors are structured now: 

Screenshot 2024-01-28 at 5 59 47 PM

Screenshot 2024-01-28 at 6 00 00 PM

You can also test by triggering an error.

## Risks

- Changes to renderDiagnostic. Instead of rendering errors to Doc AnsiStyle, this implementation renders errors to an Errata object. 

## References

- [ANE-1254](https://fossa.atlassian.net/browse/ANE-1254)

## Checklist

[ANE-1254]: https://fossa.atlassian.net/browse/ANE-1254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JeffreyHuynh1:
There was a pretty big change in this pr that changed the way we render errors to our diagnostic. Instead of rendering everything as Doc AnsiStyle, it has been changed to return an Errata object. Currently, we aren't necessarily doing much with the Errata module aside from using the source location and basic rendering features. 

However, Errata allows you to highlight code blocks and add styling using pointers / underlines. I  think that this might be beneficial in cases where we come across errors when parsing things like our fossa-deps / .fossa config files. This would work probably need to done later down the line. 

I would like to get thoughts on this implementation and if it's too big of a change.

JeffreyHuynh1:
Additionally, we used to capture contextual information about error failures through our errCtx carrier. It held a lot of information such as troubleshooting, documentation references, and other contextual information. In order, to structure and label our errors similar to broker. I created additional carriers for errHelp, errDoc,  and errSupport to allow us to easily categorize the various types of contextual info that can be attached to an error.

src/Data/Error.hs
```diff
@@ -0,0 +1,134 @@
+{-# LANGUAGE ImplicitParams #-}
```
spatten, line 1: This is the only place in the codebase where we use `ImplicitParams`. I'm assuming that it's necessary, but wanted to double-check. Also, the documentation for CallStack says this [here](https://hackage.haskell.org/package/base-4.17.2.0/docs/GHC-Stack-Types.html#t:CallStack):

```
NOTE: The intrepid user may notice that HasCallStack is just an alias for an implicit parameter ?callStack :: CallStack . This is an implementation detail and should not be considered part of the CallStack API, we may decide to change the implementation in the future.
```

I don't really know if this is something to be worried about, but I'd like to have someone who does give this a thumbs up

src/Effect/Exec.hs
```diff
@@ -186,24 +188,19 @@ data ExecErr
...
-    else
-      vsep
-        [ "Command execution failed: "
-        , ""
-        , indent 4 details
-        , ""
-        , reportDefectMsg
-        ]
+    then do
+      let header = "Could not find executable: `" <> cmdName cmdFailureCmd
```
spatten, line 195: [minor] Missing a back-tick after the command name
```suggestion
      let header = "Could not find executable: `" <> cmdName cmdFailureCmd <> "`"
```

src/Effect/Exec.hs
```diff
@@ -440,13 +437,10 @@ selectBestCmd workdir CandidateAnalysisCommands{..} = selectBestCmd' (NE.toList
...
 instance ToDiagnostic CandidateCommandFailed where
-  renderDiagnostic CandidateCommandFailed{..} =
-    pretty $
-      "Command "
-        <> show failedCommand
-        <> " not suitable: running with args "
-        <> show failedArgs
-        <> " resulted in a non-zero exit code"
+  renderDiagnostic CandidateCommandFailed{..} = do
+    let header = "Command: " <> failedCommand <> " not suitable"
```
spatten, line 441: It would be nice to surround the failedCommand and the failedArgs with quotes or put them on a newline so that we more clearly distinguish the command and/or args from the help text. This is out of scope for this ticket, as it's an improvement to an existing error message, but I thought I'd mention it while I noticed it

src/Strategy/Bundler.hs
```diff
@@ -132,7 +136,10 @@ analyzeGemfileLock :: (Has ReadFS sig m, Has Diagnostics sig m) => BundlerProjec
 analyzeGemfileLock project =
   warnOnErr AllDirectDeps
     . warnOnErr MissingEdges
-    . errCtx (BundlerMissingLockFile $ bundlerGemfile project)
+    . errCtx (BundlerMissingLockFileCtx $ bundlerGemfile project)
```
spatten, line 139: I think this is a widespread problem, but I encountered it while testing this, so I'll leave the comment here.

I created a ruby project and deleted the `Gemfile.lock` file. This creates a warning in the scan summary.

On master, the scan summary does not contain colour codes. On this branch it does.

The colour codes are helpful if you `cat` the scan-summary, but if you open it in an editor they don't get rendered properly. Here's what it looks like in VS Code:

CleanShot 2024-02-02 at 16 55 07@2x

Do we want to remove the colour codes when generating scan-summary.txt?

spatten (commented):
I'm about 3/4 of the way through reviewing, and running out of steam 😅 

I'm going to submit this for now, and continue reviewing on Monday

Note to self: I'm at `src/Strategy/Conan/ConanGraph.hs`

src/Strategy/Bundler.hs
spatten: This isn't a problem with the colour codes, it's just that VSCode doesn't display color codes by default. 

Installing the ANSI colors extension gives you a nice view of the file: https://marketplace.visualstudio.com/items?itemName=iliazeus.vscode-ansi

CleanShot 2024-02-02 at 17 01 25@2x

test/Control/Carrier/DiagnosticsSpec.hs
```diff
@@ -6,7 +6,8 @@ import Control.Carrier.Telemetry (withoutTelemetry)
 import Control.Effect.Diagnostics (ToDiagnostic (..), fatal)
 import Data.Text (Text)
 import Effect.Exec (ExitCode (..))
-import Effect.Logger (ignoreLogger, logInfo, pretty)
+import Effect.Logger (ignoreLogger, logInfo)
```
spatten, line 9: This is a scarily small ratio of test change to code change. 

It could be that there are a bunch of tests for error cases but the changes in this PR don't affect those tests. In that case, the error code is being exercised by the tests and all is good. 

If there are no tests on most of these errors currently, we should fix this up. I'm fine with doing that in a separate PR, as this one is huge.

If you've thought about this and come to the conclusion that this is too hard, that's reasonable, but it should be documented in the PR description.

spatten (commented):
This is a great change, and I love the look of the new error structure. 

However, I have a few questions:

1) Is the use of `ImplicitParams` reasonable. I'd like to get a sanity check from someone else on the team on this. If you've already run this by someone else, then that counts as a sanity check

2) Are the colour codes in `scan-summary.txt` okay, or should we strip them out when writing to `scan-summary.txt`

3) Is there any way we can add tests for this change. As I said in the comment below, I'm fine with these coming in a separate PR

I'd like to get answers on these before I approve. But it's looking great other than those questions!

src/Data/Error.hs
JeffreyHuynh1: So we can get rid of `ImplicitParams` by changing the code to be 

```
import GHC.Stack (CallStack, getCallStack, SrcLoc(..))

-- getSourceLocation returns SourceLocation with the filePath, line, col of the call site
getSourceLocation :: CallStack -> SourceLocation
getSourceLocation callStack = case getCallStack callStack of
  (_, loc) : _ -> SourceLocation (srcLocFile loc) (srcLocStartLine loc) (srcLocStartCol loc)
  _ -> SourceLocation "Unknown" 0 0
```

I know @csasarak had comments about this as well. So pulling him in to get his thoughts.

src/Strategy/Bundler.hs
JeffreyHuynh1: After playing around with this, in order to remove the ANSI color codes from the doc we need to do some regex to parse the doc as a string and remove the color codes that way. 

I'm wondering if we should just keep it as is and have something in the `HACKING.md` to direct ppl to download the extension you listed so that we can view the colors in VSCode

test/Control/Carrier/DiagnosticsSpec.hs
JeffreyHuynh1: Yea I'm going to create a separate ticket to add more tests. I agree, I was pretty shocked when I saw the state of our test coverage here.

src/Strategy/Bundler.hs
spatten: Unfortunately I don't think putting a comment in HACKING.md will help much. The scan-summary is used by users, and the audience for HACKING.md is people who are submitting patches to the CLI.

Let's ask in the #team-analysis channel about this. I really don't know if users look at this file very often, but I think they might and this might be worth fixing.

There is a package to do this, but I haven't looked into it other than seeing that it exists: https://hackage.haskell.org/package/strip-ansi-escape

JeffreyHuynh1: Updated to remove the ANSI color codes form ScanSummary 

Screenshot 2024-02-09 at 4 36 15 PM

src/Data/Error.hs
csasarak: When we spoke earlier, my comments were more that I think it's annoying to have to call `getSourceLocation` to get a source location, it'd be neat if it could just be automatic in order to make the work of future programmer's simpler. I don't know that any of my ideas are really much simpler than this though. If I'm reading it correct it's also optional to give `Errata` a location - you can just give it no blocks. Is that correct?

Regarding `ImplicitParams` itself, I think it's not ideal because it's relying on a global dynamic variable to get its info, but I also don't really know of a better way to do this. I would recommend using `HasCallStack` though instead of the `ImplicitParams` directly for the reasons Scott pointed out.

JeffreyHuynh1: Yea the location is optional, I was just using it in the areas where the errors cause our code to exit

src/Diag/Result.hs
```diff
@@ -146,6 +152,27 @@ data ErrCtx where
...
+  ErrSupport :: ToDiagnostic diag => diag -> ErrSupport
+
+instance Show ErrSupport where
+  showsPrec p (ErrSupport s) = showParen (p > 10) $ showString "ErrSupport " . diagToShowS s
+
+-- | Some error doc type. Right now, this just requries a ToDiagnostics instance for the type.
+data ErrDoc where
+  ErrDoc :: ToDiagnostic diag => diag -> ErrDoc
+
+instance Show ErrDoc where
```
csasarak, line 173: Generally with Haskell, the expectation is that `read . show = id` which it does not in this case. I would let the compiler derive the `Show` instance and then use either `toString` or `toText` as appropriate.

csasarak: On second thought, I think it doesn't matter in this case. There's also no guarantee it will be derivable I think. Sorry!

JeffreyHuynh1:
More diagnostic test cases will be added in a separate ticket!

spatten (approved):
This looks great. 🚢

PR #1365, opened 2024-01-18
Container scanning: `bdb` and `sqlite` strategies handle alternate locations

jssblck:
# Overview

Updates berkeleydb and sqlite strategies in containers to support rpm databases in non-standard locations.

## Acceptance criteria

Users with BerkeleyDB and Sqlite RPM databases in non-standard locations now have reported dependencies.

## Testing plan

I don't have an image to test this with, as the user reporting this issue uses a private image.
I believe this will fix the issue because in the logs for the image, we see the following line:
```
[BuildLayer] InsertUpdate usr/lib/sysimage/rpm/Packages
```

Which indicates that there is a BerkeleyDB database file in this non-standard location, that would have worked had we supported that location.

I believe this change is safe because we already use this exact method for the NDB strategy, this just also utilizes the same method in the other two strategies.

## Risks

The main risk is that this introduces extra noise by parsing files that should not be parsed.

## Metrics

None.

## References

https://fossa.atlassian.net/browse/FDN-87

## Checklist

### DIFF
--- a/src/Strategy/BerkeleyDB.hs
+++ b/src/Strategy/BerkeleyDB.hs
@@ -72,9 +72,21 @@ findProjects osInfo = walkWithFilters' $ \dir _ files -> do
   case findFileNamed "Packages" files of
     Nothing -> pure ([], WalkContinue)
     Just file -> do
-      if (Text.isInfixOf "var/lib/rpm/" $ toText . toFilePath $ file)
+      if isSupportedPath file
         then pure ([BerkeleyDatabase dir file osInfo], WalkContinue)
         else pure ([], WalkContinue)
+  where
+    -- The standard location for this is '/var/lib/rpm/',
+    -- but some distros in some versions symlink this elsewhere.
+    --
+    -- For maximal compatibility while still being reasonably confident that this is the 'Packages.db'
+    -- for the system RPM install, this function just checks whether the file is a child of any directory that
+    -- contains the word 'rpm'.
+    --
+    -- We may want to consider making walk work with symlinks (so that we are confident we're using the right file)
+    -- or unconditionally trying any 'Packages.db' named file.
+    isSupportedPath :: Path Abs File -> Bool
+    isSupportedPath = Text.isInfixOf "rpm" . toText . toFilePath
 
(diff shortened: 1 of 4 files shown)


src/Strategy/BerkeleyDB.hs
```diff
@@ -72,9 +72,21 @@ findProjects osInfo = walkWithFilters' $ \dir _ files -> do
   case findFileNamed "Packages" files of
     Nothing -> pure ([], WalkContinue)
     Just file -> do
-      if (Text.isInfixOf "var/lib/rpm/" $ toText . toFilePath $ file)
+      if isSupportedPath file
         then pure ([BerkeleyDatabase dir file osInfo], WalkContinue)
         else pure ([], WalkContinue)
+  where
+    -- The standard location for this is '/var/lib/rpm/',
```
csasarak, line 79: It looks like the new directory is something the Fedora and distros downstream of it are doing: https://fedoraproject.org/wiki/Changes/RelocateRPMToUsr

Would it be worthwhile to just check for just the one alternate case? It looks like the change isn't arbitrary and chosen by distros either, it has buy-in from upstream `rpm`: http://lists.rpm.org/pipermail/rpm-ecosystem/2021-December/000781.html

I only looked into this/ask because searching for a path with `rpm` in it feels a little bit arbitrary. Assuming I had a filename of `Packages` or `rpmdb.sqlite` and got a `False` out of `isSupportedPath` I don't think I would really be that confident that it was _not_ an rpm db.

csasarak (approved):
LGTM. I left a comment about how you do the check, but don't require any changes.

src/Strategy/BerkeleyDB.hs
jssblck: Since we already do it this way today, are you okay with leaving this as-is? Context: I want to reduce the uncertainty of this change as much as possible, and I think the best way to do that is to just exactly copy what we're already doing for the NDB strategy.

csasarak: I am, thanks!

PR #1368, opened 2024-01-22
Ane 1249 fossa feedback

JeffreyHuynh1:
# Overview

We currently have no way to get feedback from engineers about how their FOSSA-CLI experience is, we only poll in the UI.

We should implement a command, fossa feedback and subsequently (in a separate ticket) automatically prompt people to give us feedback after a certain amount of criteria is met.

## Acceptance criteria

- Users have the ability to give feedback about the FOSSA CLI from the FOSSA CLI

## Testing plan

run `fossa feedback`

Screenshot 2024-02-06 at 10 28 50 AM

## References

_Example:_

- [ANE-1231](https://fossa.atlassian.net/browse/ANE-1231)

## Checklist

[ANE-1231]: https://fossa.atlassian.net/browse/ANE-1231?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -111,7 +111,7 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
     runInit = pure $ putStrLn "The 'experimental-license-scan' has been deprecated and renamed to 'license-scan'. Please use the 'license-scan' command instead."
 
 feedbackCommand :: Mod CommandFields (IO ())
-feedbackCommand = command "feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your FOSSA CLI experience, submit feature requests, and report bugs/issues")
+feedbackCommand = command "experimental-feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your FOSSA CLI experience, submit feature requests, and report bugs/issues")
   where
     feedbackPrompt :: Parser (IO ())
     feedbackPrompt =
--- a/src/App/Fossa/Main.hs
+++ b/src/App/Fossa/Main.hs
@@ -111,7 +111,7 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
     runInit = pure $ putStrLn "The 'experimental-license-scan' has been deprecated and renamed to 'license-scan'. Please use the 'license-scan' command instead."
 
 feedbackCommand :: Mod CommandFields (IO ())
-feedbackCommand = command "feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your fossa-cli experience, submit feature requests, and report bugs/issues")
+feedbackCommand = command "feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your FOSSA CLI experience, submit feature requests, and report bugs/issues")
   where
     feedbackPrompt :: Parser (IO ())
     feedbackPrompt =
@@ -126,14 +126,8 @@ feedbackCommand = command "feedback" (info feedbackPrompt $ progDescDoc $ format
             , "* Submit feature requests to:"
             , newlineTrailing . newlinePreceding $ indent 4 "support@fossa.com"
             , "* Provide feedback on overall cli experience at:"
-            , newlineTrailing . newlineTrailing . newlinePreceding $ indent 4 "https://docs.google.com/forms/d/e/1FAIpQLSdfz4KX_j6lpCF0zICQaUd-Rjn_Vj-IelCSmqOii__fqyYVmg/viewform"
+            , newlineTrailing . newlineTrailing . newlinePreceding $ indent 4 "https://go.fossa.com/cli-feedback"
             ]
-
--- putStrLn "At FOSSA, we are committed to delivering an exceptional user experience and are continously working towards improving our product."
--- putStrLn "Your feedback is crucial in shaping our ongoing efforts to innovate and provide an even better user experience!"
--- putStrLn ("\n * Report bugs and issues at: " <> toString supportUrl)
--- putStrLn ("\n * Submit feature requests to: support@fossa.com")
--- putStrLn ("\n * Provide feedback on overall cli experience at: https://docs.google.com/forms/d/e/1FAIpQLSdfz4KX_j6lpCF0zICQaUd-Rjn_Vj-IelCSmqOii__fqyYVmg/viewform")
 
 decodeSubCommand :: (GetSeverity a, GetCommonOpts a, Show b, ToJSON b) => SubCommand a b -> Mod CommandFields (IO ())
 decodeSubCommand cmd@SubCommand{..} = command commandName $ info (runSubCommand cmd) commandInfo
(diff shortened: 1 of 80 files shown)


src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,31 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
     runInit :: Parser (IO ())
     runInit = pure $ putStrLn "The 'experimental-license-scan' has been deprecated and renamed to 'license-scan'. Please use the 'license-scan' command instead."
 
+feedbackCommand :: Mod CommandFields (IO ())
+feedbackCommand = command "feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your fossa-cli experience, submit feature requests, and report bugs/issues")
```
jssblck, line 114: [nit] the proper name for the cli is "FOSSA CLI"

src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,31 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
...
+            , newlineTrailing . newlinePreceding $ indent 4 "support@fossa.com"
+            , "* Provide feedback on overall cli experience at:"
+            , newlineTrailing . newlineTrailing . newlinePreceding $ indent 4 "https://docs.google.com/forms/d/e/1FAIpQLSdfz4KX_j6lpCF0zICQaUd-Rjn_Vj-IelCSmqOii__fqyYVmg/viewform"
+            ]
+
+-- putStrLn "At FOSSA, we are committed to delivering an exceptional user experience and are continously working towards improving our product."
+-- putStrLn "Your feedback is crucial in shaping our ongoing efforts to innovate and provide an even better user experience!"
+-- putStrLn ("\n * Report bugs and issues at: " <> toString supportUrl)
+-- putStrLn ("\n * Submit feature requests to: support@fossa.com")
+-- putStrLn ("\n * Provide feedback on overall cli experience at: https://docs.google.com/forms/d/e/1FAIpQLSdfz4KX_j6lpCF0zICQaUd-Rjn_Vj-IelCSmqOii__fqyYVmg/viewform")
```
jssblck, line 136: Leftover?

src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,31 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
...
+          vsep
+            [ newlinePreceding "At FOSSA, we are committed to delivering an exceptional user experience and are continously working towards improving our product."
+            , "Your feedback is crucial in shaping our ongoing efforts to innovate and provide an even better user experience!"
+            , ""
+            , "* Report bugs and issues at:"
+            , newlineTrailing . newlinePreceding $ indent 4 $ pretty supportUrl
+            , "* Submit feature requests to:"
+            , newlineTrailing . newlinePreceding $ indent 4 "support@fossa.com"
+            , "* Provide feedback on overall cli experience at:"
+            , newlineTrailing . newlineTrailing . newlinePreceding $ indent 4 "https://docs.google.com/forms/d/e/1FAIpQLSdfz4KX_j6lpCF0zICQaUd-Rjn_Vj-IelCSmqOii__fqyYVmg/viewform"
```
jssblck, line 129: Where does this go?

src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,31 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
...
+      pure $
+        putDoc $
+          vsep
+            [ newlinePreceding "At FOSSA, we are committed to delivering an exceptional user experience and are continously working towards improving our product."
+            , "Your feedback is crucial in shaping our ongoing efforts to innovate and provide an even better user experience!"
+            , ""
+            , "* Report bugs and issues at:"
+            , newlineTrailing . newlinePreceding $ indent 4 $ pretty supportUrl
+            , "* Submit feature requests to:"
+            , newlineTrailing . newlinePreceding $ indent 4 "support@fossa.com"
```
jssblck, line 127: The support site basically just has them send an email, should we join these two bullets?

src/App/Fossa/Main.hs
jssblck: Thinking about this more, I think this url being like this is not a good look. We should short link at least - we have go.fossa.com short links in route53

src/App/Fossa/Main.hs
JeffreyHuynh1: haha yea oops

src/App/Fossa/Main.hs
JeffreyHuynh1: I think having the support site is nice in the sense that the default ticket form already gives details on what to provide to the support team.

i.e. debug bundle, additional photos , etc

It also seems more structured than a standard email

src/App/Fossa/Main.hs
JeffreyHuynh1: It goes to a google form that I created

Yea I like the idea of having a short link. Is it not a good look in terms of a security standpoint or is it due to readability?

src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,25 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
     runInit :: Parser (IO ())
     runInit = pure $ putStrLn "The 'experimental-license-scan' has been deprecated and renamed to 'license-scan'. Please use the 'license-scan' command instead."
 
+feedbackCommand :: Mod CommandFields (IO ())
+feedbackCommand = command "experimental-feedback" (info feedbackPrompt $ progDescDoc $ formatStringToDoc "Provide feedback on your FOSSA CLI experience, submit feature requests, and report bugs/issues")
```
JeffreyHuynh1, line 114: Talked with zach about running this command as an experiment. Is the process of doing so just having the command name tagged with "experimental"?

src/App/Fossa/Main.hs
jssblck: I agree; in that case wouldn't we just always send them to the support site?

JeffreyHuynh1: The current ticket forms look like they're more geared for bug reporting. So If we were to consolidate and send users to the support site for both feature requests and bug reporting, then I think it would be best to create a new feature request ticket template. I would need to look into this as I'm not sure how to do that as of rn . 

Screenshot 2024-02-05 at 1 37 02 PM

jssblck: Hmm, okay. Let's ask Brian which he prefers. I'm good with whatever you and he work out.

src/App/Fossa/Main.hs
jssblck: Yes: https://github.com/fossas/fossa-cli/tree/master/docs/references/experimental#what-denotes-an-experimental-feature

jssblck: That being said, why should this be `experimental`? This isn't something we anticipate being scripted ever, right, so no need to worry about maintaining it forever. cc @zlav

src/App/Fossa/Main.hs
```diff
@@ -107,6 +110,25 @@ experimentalLicenseScanCommand = command "experimental-license-scan" (info runIn
...
+  where
+    feedbackPrompt :: Parser (IO ())
+    feedbackPrompt =
```
jssblck, line 117: Related to https://github.com/fossas/fossa-cli/pull/1368/files#r1478940928 - consider testing whether in a TTY and explicitly exiting with an error code if not. 

This way the only way a user can run it in a script is if they intentionally pretend to be in a TTY, ensuring that we don't have to consider scripting backwards compatibility ever for this subcommand.

jssblck (approved):
This looks good after these comments are resolved. I have feedback about the form but will deliver that in slack.

PR #1371, opened 2024-01-26
Ane 1252 preflight check

JeffreyHuynh1:
# Overview

As a fossa user, I want fossa to perform pre-flight mode so I’m confident that my API key, and configuration and dependency files are valid.

## Acceptance criteria

- If fossa-cli, has invalid API key for action, it should halt prior to analysis or any actions that requires API key

- If fossa-cli, has invalid fossa-deps, it should halt prior to analysis or any actions

- If fossa-cli, has invalid fossa-yml, it should halt prior to analysis or any action

- If fossa-cli, should be able to connect to access [app.fossa.com](http://app.fossa.com/) for any command that requires VALID API Key (or halt)

If fossa-cli, should be able to write to temp directory in the system

## Testing plan

PreflightChecksSpec:
- mock failed and successful calls to getOrganization 

Manually tested related commands with invalid api key

## References

- [ANE-1252](https://fossa.atlassian.net/browse/ANE-1252)

## Checklist

[ANE-1252]: https://fossa.atlassian.net/browse/ANE-1252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JeffreyHuynh1:
The ticket specified to validate the .fossa.yml and fossa-deps.yml configs. 

.fossa config is validated at parsing and an error will be thrown if the config isn't correct. Fossa-deps is only used in `fossa analyze` so I just moved that function call earlier on in the analyze work flow. As a result, I didn't include the validation for .fossa and fossa-deps it in the preflightChecks function to avoid duplicate calls to validate the config files. 

Was wondering if this makes sense, or if I should just add that validation to the preflight checks?

src/App/Fossa/Analyze.hs
```diff
@@ -282,6 +283,18 @@ analyze cfg = Diag.context "fossa-analyze" $ do
       shouldAnalyzePathDependencies = resolvePathDependencies $ Config.experimental cfg
       allowedTactics = Config.allowedTacticTypes cfg
 
+  manualSrcUnits <-
```
JeffreyHuynh1, line 286: Moving call to extract / validate fossa-deps earlier on in the analyze work flow

src/App/Fossa/Container/Test.hs
```diff
@@ -55,7 +56,9 @@ test ::
   ) =>
   ContainerTestConfig ->
   m ()
-test ContainerTestConfig{..} =
+test ContainerTestConfig{..} = do
+  _ <- ignoreDebug $ runFossaApiClient apiOpts preflightChecks
```
meghfossa, line 60: looks like you are using ignoreDebug in all invocation of `runFossaApiClient apiOpts preflightChecks`, maybe something like,

```Haskell
guardWithPreflightChecks :: ApiOpts -> ....
guardWithPreflightChecks apiOpts = ignoreDebug $ runFossaApiClient apiOpts preflightChecks

preflightChecks ::
  ( Has Diagnostics sig m
  , Has (Lift IO) sig m
  , Has FossaApiClient sig m
  ) =>
  m ()
preflightChecks = context "preflight-checks" $ do
```

src/App/Fossa/PreflightChecks.hs
```diff
@@ -0,0 +1,51 @@
...
+import Effect.Logger (pretty, vsep)
+import Path (
+  File,
+  Path,
+  Rel,
+  fromAbsFile,
+  mkRelFile,
+  (</>),
+ )
+import Path.IO
```
meghfossa, line 24: [Prefer] Don't do blanket import.

src/App/Fossa/PreflightChecks.hs
```diff
@@ -0,0 +1,51 @@
...
+module App.Fossa.PreflightChecks (
+  preflightChecks,
+) where
+
+import App.Docs (apiKeyUrl)
+import Control.Carrier.Diagnostics (Diagnostics, errCtx)
+import Control.Carrier.Stack (context)
+import Control.Effect.Diagnostics (ToDiagnostic, fatalOnIOException)
+import Control.Effect.FossaApiClient (FossaApiClient, getOrganization)
+import Control.Effect.Lift
```
meghfossa, line 12: [Prefer] Don't do blanket import.

meghfossa (approved):

PR #1375, opened 2024-02-07
Warn when no analysis targets are found

JeffreyHuynh1:
# Overview

Current the CLI returns an error when we scan a project without any analysis targets. Change it so that we return a warning instead of an error.

## Acceptance criteria

Emit a warning when there are no analysis targets

## Testing plan

run fossa analyze && fossa test on an empty folder

Before:
Screenshot 2024-02-07 at 12 21 45 PM

After:

`fossa analyze`  results

Screenshot 2024-02-07 at 1 58 07 PM

`fossa test` results

Screenshot 2024-02-07 at 1 53 13 PM

## References

[Slack thread](https://teamfossa.slack.com/archives/C039KE5ERNE/p1707240999977229)

_Example:_

- [ANE-1434](https://fossa.atlassian.net/browse/ANE-1434)
## Checklist

[ANE-1434]: https://fossa.atlassian.net/browse/ANE-1434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,6 @@
 # FOSSA CLI Changelog
 
-## v3.8.37
+## v3.9.0
 - Emits a warning instead of an error when no analysis targets are found ([#1375](https://github.com/fossas/fossa-cli/pull/1375))
 
 ## v3.8.36
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## v3.8.37
+- Emits a warning instead of error when no analysis targets are found ([#1375](https://github.com/fossas/fossa-cli/pull/1375))
+
 ## v3.8.36
 - `fossa feedback`: Allow users to provide feedback on their cli experience ([#1368](https://github.com/fossas/fossa-cli/pull/1368))
 - Add preflight checks to validate API key, connection to FOSSA app, and ability to write to temp directory in relevant commands
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # FOSSA CLI Changelog
 
 ## v3.8.37
-- Emits a warning instead of error when no analysis targets are found ([#1375](https://github.com/fossas/fossa-cli/pull/1375))
+- Emits a warning instead of an error when no analysis targets are found ([#1375](https://github.com/fossas/fossa-cli/pull/1375))
 
 ## v3.8.36
 - `fossa feedback`: Allow users to provide feedback on their cli experience ([#1368](https://github.com/fossas/fossa-cli/pull/1368))
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -408,7 +408,6 @@ analyze cfg = Diag.context "fossa-analyze" $ do
   let keywordSearchResultsFound = (maybe False (not . null . lernieResultsKeywordSearches) lernieResults)
   let outputResult = buildResult includeAll additionalSourceUnits filteredProjects' licenseSourceUnits
 
-  -- In the case that we don't find any analysis targets, emit a warning
   scanUnits <-
     case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectScans filteredProjects' additionalSourceUnits licenseSourceUnits) of
       (False, NoneDiscovered) -> do
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -96,6 +96,7 @@ import Data.Aeson qualified as Aeson
 import Data.ByteString.Lazy qualified as BL
 import Data.Flag (Flag, fromFlag)
 import Data.Foldable (traverse_)
+import Data.Functor (($>))
 import Data.List.NonEmpty qualified as NE
 import Data.Maybe (fromMaybe, mapMaybe)
 import Data.String.Conversion (decodeUtf8, toText)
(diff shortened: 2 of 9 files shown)


src/App/Fossa/Analyze.hs
```diff
@@ -411,12 +411,15 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
-    (_, CountedScanUnits scanUnits) -> doUpload outputResult iatAssertion destination basedir jsonOutput revision scanUnits
+
+  -- In the case that we don't find any analysis targets, we only want to emit a warning.
+  void . recover . Diag.warnOnErr ErrNoProjectsDiscovered $
+    case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectScans filteredProjects' additionalSourceUnits licenseSourceUnits) of
+      (False, NoneDiscovered) -> Diag.fatal ErrNoProjectsDiscovered
+      (True, NoneDiscovered) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
+      (False, FilteredAll) -> Diag.fatal ErrFilteredAllProjects
+      (True, FilteredAll) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
+      (_, CountedScanUnits scanUnits) -> doUpload outputResult iatAssertion destination basedir jsonOutput revision scanUnits
```
jssblck, line 422: this does technically accomplish the goal of making the CLI not exit with an error, but I think the actual goal of this ask is that we still create a project in FOSSA (that just has no dependencies)

src/App/Fossa/Analyze.hs
```diff
@@ -411,12 +411,15 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
-    (False, NoneDiscovered) -> Diag.fatal ErrNoProjectsDiscovered
-    (True, NoneDiscovered) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
-    (False, FilteredAll) -> Diag.fatal ErrFilteredAllProjects
-    (True, FilteredAll) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
-    (_, CountedScanUnits scanUnits) -> doUpload outputResult iatAssertion destination basedir jsonOutput revision scanUnits
+
+  -- In the case that we don't find any analysis targets, we only want to emit a warning.
+  void . recover . Diag.warnOnErr ErrNoProjectsDiscovered $
+    case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectScans filteredProjects' additionalSourceUnits licenseSourceUnits) of
+      (False, NoneDiscovered) -> Diag.fatal ErrNoProjectsDiscovered
```
csasarak, line 418: Would changing `Diag.fatal` to `Diag.warn` be sufficient? 

This code first emits an error, then it immediately checks to see if that error is a specific type so it can be changed into a warning. I think it's more straightforward to just emit the warning directly.

jssblck (changes requested):

csasarak (changes requested):

src/App/Fossa/Analyze.hs
JeffreyHuynh1: I gotcha, I missed that part. Thanks for the clarification!

src/App/Fossa/Analyze.hs
```diff
@@ -411,8 +411,11 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
   -- So the solution is to still fail, but give a hopefully useful explanation that the error can be ignored if all you were expecting is keyword search results.
+  -- In the case that we don't find any analysis targets, we want to emit a warning and upload empty source units.
   case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectScans filteredProjects' additionalSourceUnits licenseSourceUnits) of
-    (False, NoneDiscovered) -> Diag.fatal ErrNoProjectsDiscovered
+    (False, NoneDiscovered) -> do
+      Diag.warn ErrNoProjectsDiscovered
+      doUpload outputResult iatAssertion destination basedir jsonOutput revision $ SourceUnitOnly []
     (True, NoneDiscovered) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
     (False, FilteredAll) -> Diag.fatal ErrFilteredAllProjects
     (True, FilteredAll) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
```
jssblck, line 421: Looks great, we should apply this to all the other cases too.
Probably this looks like:
1. Change the `case` branches to just emit warnings, and evaluate to the `scanUnits` to upload
2. After the `case` statement, do the upload with the returned `scanUnits`

jssblck (changes requested):

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## v3.8.37
```
meghfossa, line 3: [Prefer] that we bump the minor version, as I believe this change constitutes a significant enough modification in what constitutes a successful analysis.

Prior to this, the convention was that Fossa would exit fatally if it could not find anything to report. This was to avoid scenarios where someone installs `fossa analyze && fossa test`, and CI would pass, but no "test" was being conducted (as the CLI was not discovering anything). This sometimes led to the incorrect assumption that code is compliant, when in reality FOSSA couldn't detect anything (while customer believing that we found dependencies, and found no issues).

JeffreyHuynh1: That's a good point. I'm open to that , let me post in the group channel to get the others' thoughts as well.

jssblck (approved):
Excellent! ✨ 

One note: I didn't get a warning when I ran it on an empty project, unless I ran in debug mode, since we don't print warnings by default anymore.

```
❯ cabal run fossa -- analyze ~/projects/scratch/empty 
Warning: The package list for 'hackage.haskell.org' is 98 days old.
Run 'cabal update' to get the latest list of available packages.
Warning: Requested index-state 2024-01-02T10:02:41Z is newer than
'hackage.haskell.org'! Falling back to older state (2023-11-01T19:52:38Z).
Resolving dependencies...
        Unable to infer project revision from VCS, using current timestamp as the revision.

Using project name: `empty`
Using revision: `2024-02-08T17:52:08Z`
Using branch: `No branch (detached HEAD)`

    View FOSSA Report:
    https://app.fossa.com/projects/custom%2b24357%2fempty/refs/branch/master/2024-02-08T17:52:08Z

```

I wonder if it's worth upgrading the warning to an info level. May be worth asking in the group chat, but I'm also good to have it merge as-is.

src/App/Fossa/Analyze.hs
```diff
@@ -408,15 +408,22 @@ analyze cfg = Diag.context "fossa-analyze" $ do
...
-    (True, NoneDiscovered) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
-    (False, FilteredAll) -> Diag.fatal ErrFilteredAllProjects
-    (True, FilteredAll) -> Diag.fatal ErrOnlyKeywordSearchResultsFound
-    (_, CountedScanUnits scanUnits) -> doUpload outputResult iatAssertion destination basedir jsonOutput revision scanUnits
+  scanUnits <-
+    case (keywordSearchResultsFound, checkForEmptyUpload includeAll projectScans filteredProjects' additionalSourceUnits licenseSourceUnits) of
+      (False, NoneDiscovered) -> do
+        Diag.warn ErrNoProjectsDiscovered
+        pure emptyScanUnits
+      (True, NoneDiscovered) -> do
```
csasarak, line 416: [Optional]:

```haskell
do   Diag.warn ErrOnlyKeywordSearchResultsFound
        pure emptyScanUnits
```
Is equivalent to:
```haskell
Diag.warn ErrOnlyKeywordSearchResultsFound $> emptyScanUnits
```
The type of `$>` is `$> :: f a -> b -> f b`. Use whatever seems more readable to you. But `do` notation is technically not necessary here because while diagnostics _are_ a monad this code isn't using anything its applicative interface doesn't already provide.

csasarak (approved):
Looks good!

src/App/Fossa/Analyze.hs
JeffreyHuynh1: Oh nice, I like this better actually

PR #1378, opened 2024-02-09
Explain why test deps may not be removed using go mod tidy.

csasarak:
# Overview

FOSSA will generally not report Go module test deps. This scenario isn't called out specifically in our documentation about how `go.mod` is advisory rather than defines the exact set of deps a go modules project will use. I added a quick paragraph. 

## Acceptance criteria

* We have some guidance ready and available for when a test dep doesn't appear in a Go modules project.

## Checklist

### DIFF
--- a/docs/references/strategies/languages/golang/gomodules.md
+++ b/docs/references/strategies/languages/golang/gomodules.md
@@ -235,3 +235,24 @@ not using backoff!
 
 As a concrete step towards resolving this sort of discrepancy, we recommend running `go mod tidy` on projects regularly;
 this command should synchronize the `go.mod` file with the actual state of the project.
+
+#### Test Dependencies
+
+Sometimes the above procedure may uncover a dependency that is not reported by FOSSA but that is also not removed by `go mod tidy`.
+The other reason that a dependency may appear in `go.mod` but not in FOSSA is that it is a test dependency.
+`go.mod` itself does not label dependencies as being used only in tests, but FOSSA's Go module strategy can identify these and will exclude them.
+To verify that a direct dependency is not a test-only dependency, we recommend searching the project source code for where a package from the module is imported.
+Generally, for a declaration in `go.mod` like:
+
+```
+	github.com/prometheus/client_golang v1.12.2
+```
+
+You can search for its imports using a command like:
+
+```sh
+$ find <project_directory> -name \*.go -exec grep -Hn "github.com/prometheus/client_golang" {} \;
+```
+
+Then verify that it appears in at least one Go source file _not_ ending in `_test.go`.
+You can read more about how tests are defined in Go [here](https://go.dev/doc/tutorial/add-a-test).


docs/references/strategies/languages/golang/gomodules.md
```diff
@@ -235,3 +235,24 @@ not using backoff!
...
+Generally, for a declaration in `go.mod` like:
+
+```
+	github.com/prometheus/client_golang v1.12.2
+```
+
+You can search for its imports using a command like:
+
+```sh
+$ find <project_directory> -name \*.go -exec grep -Hn "github.com/prometheus/client_golang" {} \;
```
csasarak, line 254: I thought a bit about writing a tiny script to do this procedure and report back if a dep looks like a test dep. I thought it might be overkill right now, but let me know if you disagree.

docs/references/strategies/languages/golang/gomodules.md
```diff
@@ -235,3 +235,24 @@ not using backoff!
 
 As a concrete step towards resolving this sort of discrepancy, we recommend running `go mod tidy` on projects regularly;
 this command should synchronize the `go.mod` file with the actual state of the project.
+
+#### Test Dependencies
+
+Sometimes the above procedure may uncover a dependency that is not reported by FOSSA but that is also not removed by `go mod tidy`.
+The other reason that a dependency may appear in `go.mod` but not in FOSSA is that it is a test dependency.
+`go.mod` itself does not label dependencies as being used only in tests, but FOSSA's Go module strategy can identify these and will exclude them.
```
jssblck, line 243:
```suggestion
`go.mod` itself does not label dependencies as being used only in tests, but FOSSA's Go module strategy identifies and excludes test-only dependencies.
```

docs/references/strategies/languages/golang/gomodules.md
meghfossa: This would be immense, we've had ticket where customer had a list of dependency that they were expecting in final analysis, but all of them were test deps. Going dep by dep in debug bundle was quite time sink. Having script would greatly reduce these sort of on-call tickets.

jssblck: I agree, let's leave this simple. If it were me I would point users to `rg` so that the same advice is applicable cross-platform and is simpler, but up to you.

e.g.:
> You can search for its imports using a command like the below. Here we use [ripgrep](https://github.com/BurntSushi/ripgrep?tab=readme-ov-file#ripgrep-rg) as `rg`, but any similar tool should suffice:
> ```sh
> rg 'github.com/prometheus/client_golang' -F -l
> ```

docs/references/strategies/languages/golang/gomodules.md
```diff
@@ -235,3 +235,24 @@ not using backoff!
...
+```
+
+You can search for its imports using a command like:
+
+```sh
+$ find <project_directory> -name \*.go -exec grep -Hn "github.com/prometheus/client_golang" {} \;
+```
+
+Then verify that it appears in at least one Go source file _not_ ending in `_test.go`.
+You can read more about how tests are defined in Go [here](https://go.dev/doc/tutorial/add-a-test).
```
jssblck, line 258: [optional] since we expect users to read this when they have test-only deps, i recommend focusing on the affirmative case to make it easier to read for them.
```suggestion
If the import only appears in source files that end in `_test.go`, it is a test-only dependency.
You can read more about how tests are defined in Go [here](https://go.dev/doc/tutorial/add-a-test).
```

docs/references/strategies/languages/golang/gomodules.md
```diff
@@ -235,3 +235,24 @@ not using backoff!
...
+	github.com/prometheus/client_golang v1.12.2
+```
+
+You can search for its imports using a command like:
+
+```sh
+$ find <project_directory> -name \*.go -exec grep -Hn "github.com/prometheus/client_golang" {} \;
+```
+
+Then verify that it appears in at least one Go source file _not_ ending in `_test.go`.
```
jssblck, line 257: 🤔 what if their non-`_test.go` file is only called by a `_test.go` file? what do we report & what does Go consider it as in that scenario?

jssblck (approved):

docs/references/strategies/languages/golang/gomodules.md
csasarak: I made a quick example project and we report that the dep in the non-test file is a regular dependency and not just a test dep, even though in practice it is only used in a test-suite. I think that's because the non`_test.go` file is still part of a larger package and so could theoretically be called from non-test code. Strangely, this did not change even when I made the function called from the test code private to the package.

I think this means that the find/grep method is generally consistent with what we report and what Go considers packages to be, but please ask questions if the above doesn't make much sense.

docs/references/strategies/languages/golang/gomodules.md
csasarak: The cross-platform point is good. I will add a line about `rg` being a good option if `find`/`grep` aren't available. But the reason I went with a `find`/`grep` solution first is that they are generally available in the base-system of just about anything vaguely POSIX-y where `rg` is not. I'd like something a user can just copy-paste and use without the additional tool installation step be available.

Megh, I'll try to put one together and time-box to about a half-hour. In the worst case, I think this method is something support folks can do manually.

docs/references/strategies/languages/golang/gomodules.md
csasarak: This is a great point, thank you!

docs/references/strategies/languages/golang/gomodules.md
jssblck: Yeah, that tracks. I think Go's code pruning isn't granular enough to notice this and this is roughly what I thought would be the case. Thanks for checking!

docs/references/strategies/languages/golang/gomodules.md
csasarak: @meghfossa I've added a script if you want to look it over. For now, I think I won't mention it in the docs. I'd rather a support person be guiding them so that they can have more context.

PR #1381, opened 2024-02-15
Ane 1425 maven analysis hanging

JeffreyHuynh1:
# Overview

Maven analysis continues to run without ever stopping

## Acceptance criteria

Ensure that maven analysis does not run forever.

## Testing plan

Added a test case to include a cyclic graph in GraphingSpec.hs

## References

- [ANE-1425](https://fossa.atlassian.net/browse/ANE-1425)

## Checklist

[ANE-1425]: https://fossa.atlassian.net/browse/ANE-1425?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Graphing.hs
```diff
@@ -425,20 +444,24 @@ color graph extractSet update origin extractProperty nodesToColor = gmap applyCo
...
---       6    \
---             5
+--   Here is an example to retrieve all reachable of nodes from 1, where the reachable nodes are not 2:
 --
--- reachableSuccessorsWithCondition [(1, 2), (1, 3), (2, 4), (3, 6), (4, 5)]  1  (\x -> x `Set.notMember conditionalSet) conditionalSet= Set.fromList [2] =
+-- >  reachableSuccessorsWithCondition [(1, 2), (1, 3), (2, 4), (3, 6), (4, 5)]  1  (\x -> x `Set.notMember conditionalSet) conditionalSet= Set.fromList [2]
+-- >  Returns: Set (3, 6)
 --
--- Set (3, 6)
+--   In the context of Maven submodule filtering (Maven/Common.hs), `reachableSuccessorsWithCondition` returns a submodule's dependencies given that the dependency is not a submodule.
```
csasarak, line 464: This description should be edited to explain what the function does without reference to Maven because `src/Graphing.hs` is mostly agnostic to what's actually in the graphs it manipulates. What you're really doing is saying that we ignore cycles when finding successors, correct?

src/Graphing.hs
```diff
@@ -396,26 +396,45 @@ subGraphOf n (Graphing gr) =
...
+--
+--   d2 and d3 after computation:
+-- >    d2 = SimpleDep                 d3 = SimpleDep
+-- >      { name : "d2"                   { name : "d3"
+-- >      , colorTag : "green"            , colorTag : "red"
+-- >      , colors : Set("blue")          , colors : Set("blue")
+-- >      }                               }
+--
+--   In the context Maven submodule filtering (Maven/Common.hs), `color` is used to attach submodule information to nodes.
+--   This is later used to determine which nodes to filter based on the provided submodule filters.
```
csasarak, line 437: Avoid coupling an explanation of this function (which is generalized to all `Ord a`) to uses of the functions which are more specialized. A reader should be able to understand how `color` works without having to think about how Maven specifically uses it. If this is just an example for illustrative purposes then I guess it's OK if you make it clearer that it's an example.

Recommendation: This information is more relevant and useful in `Maven/Common.hs`. Please move or delete it.

src/Strategy/Maven.hs
```diff
@@ -177,12 +177,24 @@ getStaticAnalysis submoduleTargets closure = do
   filteredGraph <- applyMavenFilters submoduleTargets allSubmodules graph
   pure (filteredGraph, graphBreadth)
 
-applyMavenFilters :: Has (Reader MavenScopeFilters) sig m => Set Text -> Set Text -> Graphing MavenDependency -> m (Graphing Dependency)
+applyMavenFilters ::
+  ( Has Diagnostics sig m
```
csasarak, line 181: Even though the type is long I think it still looks better on one line. Did the formatter do this? It doesn't when I add the `Diagnostics` constraint. Maybe we can pair on seeing what's up if it was the formatter.

src/Strategy/Maven/Common.hs
```diff
@@ -55,7 +64,7 @@ filterMavenSubmodules includedSubmoduleSet completeSubmoduleSet graph = do
     -- reachableNodesOnCondition returns all of the submodule's children, so we need to union the set with the submodule to get a complete set (origin + children)
     reachableNodesFromSubmodule :: Text -> Set Text
     reachableNodesFromSubmodule mavenDep = do
-      let children = reachableSuccessorsWithCondition edgeList mavenDep notSubmodule completeSubmoduleSet
+      let children = reachableSuccessorsWithCondition edgeList mavenDep notSubmodule completeSubmoduleSet $ Set.fromList []
```
csasarak, line 67: Please use `Set.empty`.

src/Strategy/Maven/Common.hs
```diff
@@ -28,6 +28,15 @@ data MavenDependency = MavenDependency
 mavenDependencyToDependency :: MavenDependency -> Dependency
 mavenDependencyToDependency MavenDependency{..} = dependency
 
+-- | Filter all submodules (including their dependencies) that are not in `includedSubmoduleSet`.
+--
+--   We use `reachableSuccessorsWithCondition` to retrieve a submodule's dependencies so long as
+--   the dependency is not a submodule itself. This is important in the case that a dependency belongs to multiple submodules
```
csasarak, line 34: [nit] "so long as the dependency is not also a submodule in the project"

It makes it more obvious what the relationship between the submodules is, but this could also be less obvious to me just be because I'm not as familiar with this code as you are right now. I'll defer to your judgement whether the suggestion actually provides clarity.

csasarak (changes requested):
This is a great improvement. If you address my questions I can approve.

src/Strategy/Maven.hs
JeffreyHuynh1: The formatter didn't do this, I just thought that the convention  that we had in our code base , from what I've seen, was to have this portion as multi line when we have more than one constraint.

src/Graphing.hs
JeffreyHuynh1: That's correct the work in this pr is to ignore cycles when finding successors.

I should've made note of this, but I am also reformatting the comments for the maven submodule functions and the graphing functions that I added. This was based on a previous discussion that I had with Jess bc it was mentioned that it was hard to distinguish the purpose of these functions and it's intended use

src/Strategy/Maven.hs
csasarak: I think it's somewhat a matter of taste, I'm not aware of a hard rule around more than one constraint. You should keep this if you think it's easier to read, that's more important than following a rule.

JeffreyHuynh1: I'm okay with having this as a one liner!

src/Graphing.hs
csasarak: ah ok, that's really reasonable. When I read this initially I thought it was suggesting that this was made especially rather than simply as an example. Thanks!

src/Strategy/Maven/Common.hs
```diff
@@ -28,6 +28,75 @@ data MavenDependency = MavenDependency
 mavenDependencyToDependency :: MavenDependency -> Dependency
 mavenDependencyToDependency MavenDependency{..} = dependency
 
+-- | Filter all submodules (including their dependencies) that are not in `includedSubmoduleSet`.
```
csasarak, line 31: This is really well written and makes the why and how really clear!

test/GraphingSpec.hs
```diff
@@ -430,17 +430,28 @@ reachableSuccessorsWithConditionSpec = do
...
+
+  --   1 -> 2 -> 3
+  --        |    |
+  --        5 <- 4
+  let cyclicGraph :: Graphing Int = Graphing.directs [1] <> Graphing.edges [(1, 2), (2, 3), (3, 4), (4, 5), (5, 2)]
+
   let condition :: Int -> Set Int -> Bool
       condition x excluded = x `Set.notMember` excluded
+  let emptySet :: Set Int
+      emptySet = Set.fromList []
```
csasarak, line 442: You can just use `Set.empty` inline.

test/GraphingSpec.hs
```diff
@@ -430,17 +430,28 @@ reachableSuccessorsWithConditionSpec = do
   --       6    \
   --             5
   let graph :: Graphing Int = Graphing.directs [1, 2] <> Graphing.edges [(1, 2), (1, 3), (2, 4), (3, 6), (4, 5)]
+
+  --   1 -> 2 -> 3
+  --        |    |
+  --        5 <- 4
+  let cyclicGraph :: Graphing Int = Graphing.directs [1] <> Graphing.edges [(1, 2), (2, 3), (3, 4), (4, 5), (5, 2)]
+
   let condition :: Int -> Set Int -> Bool
```
csasarak, line 439: I know you didn't write this, but the `condition` function is exactly the same as `Set.notMember` and the name is even less descriptive. I suggest inlining it.

csasarak (approved):
Looks good!

PR #1390, opened 2024-03-04
[No Ticket] Workflow updates

csasarak:
# Overview

* Upgrades several packages to use newer version so that they will be on `node20` per https://github.blog/changelog/2023-09-22-github-actions-transitioning-from-node-16-to-node-20/
* Changes uses of `set-output` to use environment variables.
* Uses `dtolnay/rust-toolchain` to install Rust.
* Fixes badges to use a GH token env var.

## Acceptance criteria

Our builds still work. 

## Testing plan

I tested these by running our CI. The artifacts for the `create-release` step look correct to me. 

## Checklist

### DIFF
--- a/.github/workflows/badges.yml
+++ b/.github/workflows/badges.yml
@@ -8,6 +8,8 @@ jobs:
   downloads:
     name: update-downloads-badge
     runs-on: ubuntu-latest
+    env:
+      GH_TOKEN: ${{ github.token }}
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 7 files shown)


spatten (approved):
Looks good to me!

I was a little worried about switching away from `action-rs/toolchain`, but this Reddit thread convinced me: https://old.reddit.com/r/rust/comments/vyx4oj/actionsrs_organization_became_unmaintained/

It seems like we should make that change through all of our repos

PR #1399, opened 2024-03-21
Ane 926 release group

JeffreyHuynh1:
# Overview

The CLI should offer a way to create release groups from the command line. Creating and deleting other org-level objects should be done through their own command.

## Acceptance criteria

- There is a way to create and delete a release group from the CLI. 

- Emit audit logs on release group creation in CORE

## Testing plan

example .fossa.yml:
```
releaseGroup:
  title: example-title
  release: example-release-title
  releaseGroupProjects:
    - projectLocator: custom+1/git@github.com/example
      projectRevision:   12345 
      projectBranch: main
  licensePolicy: license-policy-name 
  securityPolicy: security-policy-name
  qualityPolicy: quality-policy-name
  teams:
      - team1
      - team2
```

Create a release group:

- `fossa release-group create --title example-title --release example-release-title --project-locator custom+1/git@github.com/example --project-revision 12345 --project-branch main`
- `fossa release-group create -c /path/to/.fossa.yml/ `

Add projects to a release group:
- `fossa release-group add-projects  --title example-title --release example-release-title --project-locator custom+1/git@github.com/example --project-revision 12345 --project-branch main`
- `fossa release-group add-projects -c /path/to/.fossa.yml/ `

Delete a release group:
- `fossa release-group delete  --title example-title`

Delete a release group release:
- `fossa release-group delete-release  --title example-title --release example-release-title`

## References

- [ANE-926](https://fossa.atlassian.net/browse/ANE-926)

## Checklist

[ANE-926]: https://fossa.atlassian.net/browse/ANE-926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App/Fossa/Config/Common.hs
```diff
@@ -209,8 +213,25 @@ metadataOpts =
...
+      pure $ removeReleaseGroupMetadata projectMetadata
+  where
+    removeReleaseGroupMetadata :: ProjectMetadata -> ProjectMetadata
+    removeReleaseGroupMetadata project = project{projectReleaseGroup = Nothing}
+
+    deprecationMessage :: Text
+    deprecationMessage =
+      renderIt $
+        vsep
+          [annotate (color Red) "Release group options for this command have been deprecated. Refer to `fossa release-group` subcommands to interact with FOSSA release groups."]
```
JeffreyHuynh1, line 234: Thinking more about this, not sure if we want to deprecate this on analyze. There was a customer that was asking to update the release projects upon scan and if we deprecate this, they wouldn't be able to do this.

src/App/Fossa/ReleaseGroup/DeleteRelease.hs
```diff
@@ -0,0 +1,43 @@
...
+
+  releaseGroups <- getReleaseGroups
+  maybeReleaseGroupId <- retrieveReleaseGroupId releaseGroupTitle releaseGroups
+  case maybeReleaseGroupId of
+    Nothing -> fatalText $ "Release group `" <> releaseGroupTitle <> "` not found"
+    Just releaseGroupId -> do
+      let releaseGroupIdText = toText releaseGroupId
+
+      releases <- getReleaseGroupReleases releaseGroupIdText
+      when (length releases <= 1) $ fatalText "You are not permitted to delete a release when there is only one release in your release group"
```
csasarak, line 37: [nit] This is insignificant here so not change needed, but `length` is pretty inefficient on lists in Haskell. They're linked lists so `length` has to traverse the whole thing to get a length. You can avoid this by pattern matching.

src/App/Types.hs
```diff
@@ -83,6 +89,54 @@ data ProjectRevision = ProjectRevision
...
+
+instance ToJSON ReleaseGroupRevision where
+  toJSON ReleaseGroupRevision{..} =
+    object
+      [ "title" .= releaseGroupTitle
+      , "release" .= releaseGroupReleaseRevision
+      , "licensePolicy" .= releaseGroupLicensePolicy
+      , "securityPolicy" .= releaseGroupSecurityPolicy
+      , "teams" .= releaseGroupTeams
+      ]
```
csasarak, line 110: [Optional] It's annoying that you have to write this. GHC has the [disambiguate record fields](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/disambiguate_record_fields.html) extension which I *think* would allow you to just name the fields the same thing as their JSON counterparts and then you can generate the To/From JSON instances with Generics.

I understand that it's a departure from how we've done things, but I think it's safe to do. `DisambiguateRecordFields` will be on by default in the `GHC2024` language edition. The one caveat is that to define two record types with the same field names, the records have to go in different modules. This might make it not worth doing the work for now.

src/App/Fossa/ReleaseGroup/Create.hs
```diff
@@ -0,0 +1,74 @@
...
+  teams <- getTeams
+
+  maybeLicensePolicyId <- retrievePolicyId (releaseGroupLicensePolicy releaseGroupRevision) LICENSING policies
+  maybeSecurityPolicyId <- retrievePolicyId (releaseGroupSecurityPolicy releaseGroupRevision) SECURITY policies
+  maybeQualityPolicyId <- retrievePolicyId (releaseGroupQualityPolicy releaseGroupRevision) QUALITY policies
+  maybeTeamIds <- retrieveTeamIds (releaseGroupTeams releaseGroupRevision) teams
+
+  let req = CreateReleaseGroupRequest (releaseGroupTitle releaseGroupRevision) (releaseGroupReleaseRevision releaseGroupRevision) maybeLicensePolicyId maybeSecurityPolicyId maybeQualityPolicyId maybeTeamIds
+  res <- createReleaseGroup req
+  logStdout $ "Created release group with id: " <> toText (groupId res)
```
meghfossa, line 53: [optional] You should provide permalink to release group, after create execution. This applies to rest of applicable command as well.

docs/references/files/fossa-yml.md
```diff
@@ -150,6 +168,55 @@ The `labels` field allows you to add labels to projects so that you can classify
...
+<img src="../../assets//project-revision-example.png">
+
+#### `releaseGroup.releaseGroupProjects.projectBranch:`
+The branch associated with the project.
+
+#### `releaseGroup.licensePolicy:`
+The name of the license policy associated with the release group.
+
+#### `releaseGroup.securityPolicy:`
+The name of the security policy associated with the release group.
```
meghfossa, line 212: [optional] We should include docs link here like we do in `.fossa.yml.example`. This applies to other policies as well

src/Fossa/API/Types.hs
```diff
@@ -944,3 +954,149 @@ instance FromJSON AnalyzedPathDependency where
...
+      , "projects" .= updatedProjects
+      ]
+
+data UpdateReleaseProjectRequest = UpdateReleaseProjectRequest
+  { updateReleaseProjectLocator :: Text
+  , updateReleaseProjectRevisionId :: Text
+  , updateReleaseProjectBranch :: Text
+  , -- Existence of this field signifies to core that the project exists and the release project just needs to be updated.
+    -- If this field is empty it means that we are adding a new project to the release.
+    maybeReleaseGroupId :: Maybe Int
```
csasarak, line 1062: The `maybe` bit is described by the type and unneeded.

test/App/Fossa/Config/ReleaseGroup/CreateSpec.hs
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+spec :: Spec
+spec = do
+  currDir <- runIO getCurrentDir
+  let scanDir = currDir </> fixtureDir
+      absFilePath = scanDir </> $(mkRelFile ".fossa.yml")
+  describe "mergeOpts" $ do
+    it' "should use values from create opts when both create opts and config values are present" $ do
+      createConfig <- mergeOpts (Just $ configFile absFilePath) (EnvVars Nothing False False Nothing Nothing mempty) createOpts
```
csasarak, line 186: Please use a name for the `EnvVars` data.

docs/references/files/fossa-yml.md
```diff
@@ -26,6 +26,24 @@ project:
...
     - test-project
 
+releaseGroup:
+  title: release-group-title
+  release: release-group-release
+  releaseGroupProjects:
+    - projectLocator: custom+123/git@github.com/fossas/fossa-cli
+      projectRevision: "12345"
+      projectBranch: master 
+    - projectLocator: custom+123/git@github.com/example
```
csasarak, line 36: We discussed project id vs. locator with respect to modifying projects and I think we should do the same here. This is technically an entirely new config block so there's no compatibility issues but users are already used to using `projectId` elsewhere. 

I think it'd be good to allow/prefer `projectId` and make `projectLocator` available for people who really need it.

csasarak (changes requested):

src/App/Fossa/ReleaseGroup/Create.hs
```diff
@@ -0,0 +1,74 @@
...
+  -- Having release groups with the same name is a current functionality. However, we want to refrain from creating release groups with the same
+  -- name on the CLI as it will make it harder to determine which release groups users want to modify in our other release group commands.
+  when (isJust maybeReleaseGroupId) $
+    fatalText $
+      "Release Group `" <> releaseGroupTitle releaseGroupRevision <> "` already exists"
+
+  policies <- getPolicies
+  teams <- getTeams
+
+  maybeLicensePolicyId <- retrievePolicyId (releaseGroupLicensePolicy releaseGroupRevision) LICENSING policies
```
meghfossa, line 46: [question] why are we doing policy resolution logic in cmdline opposed to backend?

src/App/Fossa/ReleaseGroup/Create.hs
```diff
@@ -0,0 +1,74 @@
...
+  maybeSecurityPolicyId <- retrievePolicyId (releaseGroupSecurityPolicy releaseGroupRevision) SECURITY policies
```
meghfossa, line 47: [nit] you should log these steps, right now in cmdline all we see is, for long time while 5 or 6 requests are performed sequentially. You can use `StickyLogger` or `context` 

```
Running FOSSA release-group create
```

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1596,3 +1618,123 @@ uploadReachabilityContent signedUrl bs = fossaReq $ do
...
+  , Has Diagnostics sig m
+  , Has Debug sig m
+  ) =>
+  ApiOpts ->
+  Text ->
+  Text ->
+  m ()
+deleteReleaseGroupRelease apiOpts releaseGroupId releaseId = fossaReq $ do
+  (baseUrl, baseOpts) <- useApiOpts apiOpts
+  void $ req DELETE (releaseGroupReleaseURLEndpoint baseUrl releaseGroupId releaseId) NoReqBody ignoreResponse baseOpts
```
meghfossa, line 1681: [blocker] use `context` for explainability, and debugging. This applies to all requests here

```
response <-
    context ("Retrieving a signed S3 URL for reachability content upload") $
      req POST (signedReachabilityContentURLEndpoint baseUrl .....
```

meghfossa (changes requested):
I see that chris already added comments - here are some of them from my review; Also test-plan's `.fossa.yml` is incorrect - e.g. `projectLocator`

src/App/Fossa/ReleaseGroup/Create.hs
JeffreyHuynh1: Core wanted me to use their existing endpoints for my work and they only have endpoints to retrieve policies by id, so I had to do policy resolution here.

docs/references/files/fossa-yml.md
JeffreyHuynh1: I think allowing projectId makes sense when modifying projects as that field was already present in the `project` config block. However, in this case, because we are introducing a new config block I would rather constrain users to using `projectLocator` and make it simple. I think if we get requests to include `projectId` as an option, then we can go ahead and add it, but for now I would like to keep it as is. What do you think?

src/App/Fossa/ReleaseGroup/Create.hs
```diff
@@ -0,0 +1,74 @@
...
+createMain CreateConfig{..} = do
+  logInfo "Running FOSSA release-group create"
+
+  releaseGroups <- getReleaseGroups
+  maybeReleaseGroupId <- retrieveReleaseGroupId (releaseGroupTitle releaseGroupRevision) releaseGroups
+  -- Having release groups with the same name is a current functionality. However, we want to refrain from creating release groups with the same
+  -- name on the CLI as it will make it harder to determine which release groups users want to modify in our other release group commands.
+  when (isJust maybeReleaseGroupId) $
+    fatalText $
+      "Release Group `" <> releaseGroupTitle releaseGroupRevision <> "` already exists"
```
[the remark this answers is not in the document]
JeffreyHuynh1: As this error will show up in the errata header I think it looks better without punctuation IMO

e.g. 
Error: Release group `example` already exists

vs 

Error: Release group `example already exists.

src/App/Fossa/Config/ReleaseGroup/AddProjects.hs
```diff
@@ -0,0 +1,97 @@
...
+  , releaseOpts :: Maybe Text
+  , projectsOpts :: Maybe [ReleaseGroupProjectOpts]
+  }
+  deriving (Eq, Ord, Show, Generic)
+
+cliParser :: Parser AddProjectsOpts
+cliParser =
+  AddProjectsOpts
+    <$> releaseGroupCommonOpts
+    <*> optional (strOption (applyFossaStyle <> long "config" <> short 'c' <> helpDoc configHelp))
```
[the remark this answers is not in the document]
JeffreyHuynh1: I agree there is lot of shared options between our subcommands. I think it might be worth creating a separate ticket (I can create this) to break out all the shared options and put them into a common module. What do you think?

src/App/Fossa/Config/ReleaseGroup/Create.hs
```diff
@@ -0,0 +1,114 @@
...
+collectReleaseGroupRevision maybeConfig CreateOpts{..} = do
+  let licensePolicy = licensePolicyOpts <|> (maybeConfig >>= configReleaseGroup >>= configReleaseGroupLicensePolicy)
+      securityPolicy = securityPolicyOpts <|> (maybeConfig >>= configReleaseGroup >>= configReleaseGroupSecurityPolicy)
+      qualityPolicy = qualityPolicyOpts <|> (maybeConfig >>= configReleaseGroup >>= configReleaseGroupQualityPolicy)
+      -- NOTE: teamsOpts and projectsOpts default to Just [] when it is not set through CLI flags.
+      --       Convert these to Nothing so we can try to extract from the config file.
+      teams = case teamsOpts of
+        Nothing -> (maybeConfig >>= configReleaseGroup >>= configReleaseGroupTeams)
+        Just [] -> (maybeConfig >>= configReleaseGroup >>= configReleaseGroupTeams)
+        Just teamsOpts' -> Just teamsOpts'
```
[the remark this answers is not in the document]
JeffreyHuynh1: I have changed our parser to use [some](https://hackage.haskell.org/package/base-4.16.3.0/docs/Control-Applicative.html#v:some) instead of `many`. This won't be an issue anymore.

src/App/Fossa/ReleaseGroup/AddProjects.hs
```diff
@@ -0,0 +1,68 @@
...
+  maybeReleaseGroupId <- retrieveReleaseGroupId title releaseGroups
+  case maybeReleaseGroupId of
+    Nothing -> fatalText $ "Release group `" <> title <> "` not found"
+    Just releaseGroupId -> do
+      let releaseGroupIdText = toText releaseGroupId
+
+      releases <- getReleaseGroupReleases releaseGroupIdText
+      release <- retrieveReleaseGroupRelease (releaseTitle releaseGroupReleaseRevision) releases
+      let releaseId = releaseGroupReleaseId release
+      logDebug $ "The releases : " <> pretty (pShow (releases))
```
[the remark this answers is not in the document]
JeffreyHuynh1: This was logging that I used for personal testing and forgot to delete. I've removed this.

src/App/Fossa/Config/Common.hs
JeffreyHuynh1: That's true, you would be able to run `analyze` and then update the release projects with `release-group add-projects`, but I think the customer was looking to only run `fossa analyze` to have their release group project updated.

[slack thread](https://teamfossa.slack.com/archives/C043EM3L96Z/p1710800449974419?thread_ts=1710450155.589159&cid=C043EM3L96Z)

csasarak: It sounds like deprecating them would not totally prevent the customer from updating a project in a release group. I think it's OK in that case to say that they'll need to use these new commands even if they have a preference for `analyze`.

src/App/Fossa/Config/Common.hs
```diff
@@ -482,6 +497,14 @@ commonOpts =
           , boldItalicized "Default: " <> coloredBoldItalicized Green "full"
           ]
 
+configHelp :: Maybe (Doc AnsiStyle)
```
csasarak, line 500: I don't see a reason for these to be `Maybe` since it doesn't seem like there's any possibility that they will be `Nothing`. This forces consumers of that data to handle the case of `Nothing` regardless though.

src/App/Fossa/Config/ReleaseGroup/AddProjects.hs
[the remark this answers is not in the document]
csasarak: I'm fine with a ticket, but I was only asking for sharing for the config values defined in this PR. That we can't get sharing for everything doesn't mean it isn't good to do it where it's possible and relatively easy.

src/App/Fossa/Config/ReleaseGroup/AddProjects.hs
```diff
@@ -0,0 +1,90 @@
...
+mergeOpts ::
+  ( Has Diagnostics sig m
+  ) =>
+  Maybe ConfigFile ->
+  EnvVars ->
+  AddProjectsOpts ->
+  m AddProjectsConfig
+mergeOpts maybeConfig envVars cliOpts@AddProjectsOpts{..} = do
+  apiOpts <- collectApiOpts maybeConfig envVars releaseGroupCommon
+  title <- mergeReleaseGroupTitle titleOpts (maybeConfig >>= configReleaseGroup >>= configReleaseGroupTitle)
```
csasarak, line 81: The `maybeConfig >>=` can be put into a common binding like you did elsewhere.

src/App/Fossa/ReleaseGroup/Common.hs
```diff
@@ -0,0 +1,31 @@
...
+import Fossa.API.Types (ReleaseGroup (..), ReleaseGroupRelease (..))
+
+retrieveReleaseGroupId :: Has Diagnostics sig m => Text -> [ReleaseGroup] -> m (Maybe Int)
+retrieveReleaseGroupId title releaseGroups = do
+  let filteredReleaseGroups = filter (\rg -> releaseGroupTitle rg == title) releaseGroups
+  case filteredReleaseGroups of
+    [] -> pure Nothing
+    [releaseGroup] -> pure . Just $ releaseGroupId releaseGroup
+    (_ : _ : _) ->
+      errHelp ("Navigate to the FOSSA web UI to rename your release groups so that they are unqiue" :: Text)
```
csasarak, line 18:
```suggestion
      errHelp ("Navigate to the FOSSA web UI to rename your release groups so that they are unique." :: Text)
```

csasarak: Relatedly, as you find things that you think should be constrained in FOSSA's data-model to be unique, can you make tickets if you haven't yet?

src/App/Fossa/ReleaseGroup/Common.hs
```diff
@@ -0,0 +1,31 @@
...
+        $ "Multiple instances of release groups with title `" <> title <> "` found. Unable to determine which release group to modify."
+
+retrieveReleaseGroupRelease :: Has Diagnostics sig m => Text -> [ReleaseGroupRelease] -> m ReleaseGroupRelease
+retrieveReleaseGroupRelease releaseTitle releases = do
+  let filteredReleaseGroups = filter (\r -> releaseGroupReleaseTitle r == releaseTitle) releases
+  case filteredReleaseGroups of
+    [] -> fatalText $ "Release `" <> releaseTitle <> "` not found"
+    [release] -> pure release
+    (_ : _ : _) ->
+      errHelp ("Navigate to the FOSSA web UI to rename your release group releases so that they are unqiue" :: Text)
```
csasarak, line 29:
```suggestion
      errHelp ("Navigate to the FOSSA web UI to rename your release group releases so that they are unique" :: Text)
```

csasarak (changes requested):
This looks pretty good. I can approve after you add some clarification in the docs on the differences in semantics between `projectId` and `projectLocator` as well as some description that `projectId`s are only unique for the purposes of analyze.

src/App/Fossa/Config/Common.hs
JeffreyHuynh1: These help docs are of type `Maybe` due to [Options.Applicative](https://hackage.haskell.org/package/optparse-applicative-0.18.1.0/docs/Options-Applicative.html) `helpDoc` function. It required a`Maybe` type.

src/App/Fossa/Config/Common.hs
JeffreyHuynh1: That sounds good to me

src/App/Fossa/Config/ReleaseGroup/AddProjects.hs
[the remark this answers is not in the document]
JeffreyHuynh1: You right, extracted common cli opts to the Common module.

docs/references/files/fossa-yml.md
JeffreyHuynh1: Sounds good I updated the docs!

docs/references/files/fossa-yml.md
```diff
@@ -339,3 +411,26 @@ maven:
...
+
+> NOTE: Projects uploaded through the CLI will have `custom` embedded into the project locator. 
+
+<img src="../../assets/project-locator-example.png">
+
+### When do I use project ID vs project locator?
+
+Project ID should be used when referencing projects uploaded via the CLI. In most cases, providing just a project ID is sufficient. However, when referencing projects uploaded through avenues outside the CLI (Quick Import, Archive Upload, etc), providing just a project ID fails. This is due to the nature of FOSSA's project naming conventions as project locators are constructed differently depending on how a project was uploaded. There is no way to tell at runtime which project a user is referencing when only given a project ID (i.e. Is this a project uploaded via the CLI or Archive Upload?).
+
+In `fossa release-group` project locator is used to reference projects that will be added to your release group's release. By using project locator, all FOSSA project upload flows are covered and the CLI is able to determine the specific project a user is trying to add to a release.
```
JeffreyHuynh1, line 436: Let me know if this makes sense or if I should word this differently.

csasarak (approved):

PR #1405, opened 2024-04-04
[No Ticket] Adds attributions to release assets

meghfossa:
# Overview

This PR, 
- adds attribution report to release assets

## Acceptance criteria
- attribution report is included in release assets

## Testing plan

I don't think this can be tested end/end. 

## Risks
N/A

## Metrics

N/A

## References

N/A

## Checklist

### DIFF
--- a/.github/workflows/report.yml
+++ b/.github/workflows/report.yml
@@ -6,14 +6,16 @@ jobs:
   add-attributions-to-release:
     name: add-attributions-to-release ${{ github.event.release.tag_name }}
     runs-on: ubuntu-latest
+
+    # release changes require contents write
     permissions:
-      contents: write # release changes require contents write
+      contents: write 
 
     steps:
       - uses: actions/checkout@v4
       - name: Install fossa-cli
         run: |
-          curl -H 'Cache-Control: no-cache' https://raw.githubusercontent.com/fossas/fossa-cli/master/install-latest.sh | bash
+          ./install-latest.sh -d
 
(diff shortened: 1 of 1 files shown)


.github/workflows/report.yml
```diff
@@ -0,0 +1,31 @@
...
+          ./install-latest.sh -d
+
+      # since this is only invoked after the release is published,
+      # we can safely presume that fossa has ran dependency scan on the commit
+      # from 'dependency-scan' job!
+      #
+      # docs: https://cli.github.com/manual/gh_release_upload
+      - name: Persist attributions to release
+        run: |
+          fossa --format cyclonedx-json attribution > attribution.bom.json
```
meghfossa, line 27: I think we should be uploading sboms with releases ....

spatten (commented):
I like the idea!

Do we know what happens if this fails? I.e. should we make it so failing doesn't block releases until we're sure this works?

meghfossa:
> Do we know what happens if this fails? I.e. should we make it so failing doesn't block releases until we're sure this works?

This only runs after the release has been published (not draft) - and it adds additional asset into existing release. Even if it fails - it should not impact existing release

spatten:
> This only runs after the release has been published (not draft) - and it adds additional asset into existing release. Even if it fails - it should not impact existing release

Perfect! Let's ship it then

spatten (approved):
:+1:

PR #1406, opened 2024-04-11
Support Ruby GIT remote dependencies

zlav:
# Overview

We have not been handling Git remote dependencies in Ruby projects. We have been stripped the remote and making them `GemType` deps. This PR handles this by making them `GitType` deps and also handling an edge case where some dependencies in the `DEPENDENCIES` section have a `!` suffix if they have a remote other than the main one listed for RubyGems.

## Acceptance criteria

The main criteria was testing on the impacted user's project and seeing that the dependencies were correctly detected. This has been solved

## Testing plan
There were two things I needed to test and I went about it the same way
1. `GIT` remote deps are detected properly
2. Dependencies in `DEPENDENCIES` with `!` suffix have it stripped **and** that this does not negatively impact detecting other dependencies.

Testing this was relatively simple
1. Ensure that the impacted user's project scanned successfully
2. Search for other popular and random other projects on GitHub through "site:github.com inurl:Gemfile.lock" and ensure that they scan properly.

## Risks

Before this change, it was impossible to detect Git remote deps, so there should be no risk there except for user's not expecting GIT dependencies to show up.
The riskiest part is handling the `!` suffix convention. The convention has documentation publicly and is used by several repositories (the linked ANE ticket and the one linked in the code comment). The main risk here is that there are other impacts of handling the "!" suffix that I am not aware of. As of now, I am fully convinced this is the correct move, based on reviewing ~50 GitHub repositories and observing no conflicts.

## References

- [ANE-1624](https://fossa.atlassian.net/browse/ANE-1624): Issue where GIT remote deps in Ruby projects aren't showing up

## Checklist

[ANE-1624]: https://fossa.atlassian.net/browse/ANE-1624?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JeffreyHuynh1 (changes requested):
I looked at the failing integration test and it looks like we're missing 4 deps (expecting 210 deps but got 206 deps). After seeing this, I looked at the Gemfile.lock that we're testing against and saw that it's probably from these deps:

```
mysql2 (~> 0.5)!
queue_classic!
rails!
redis-namespace!
websocket-client-simple!
```

Couple of things:
- `mysql2 (~> 0.5)!` , there is '!' here but it still gets resolved to a Gem dep. Should this be resolved to a GitType? 
   - The current logic doesn't account for the '!' suffix when it's after the version
- Your changes just change the dep from GemType -> GitType if there is a '!' suffix, I don't think it should affect the number of dependencies found in our integration tests, or at least that's what I assume from a quick glance. Can we look more into this. 
   -  Update:  Looked more at the ticket and Sara noted that we have a part of code that parses the `GIT` section [here](https://github.com/fossas/fossa-cli/blob/master/src/Strategy/Ruby/GemfileLock.hs#L138-L144) . I think that the drop in the number of deps makes sense then bc it has already been accounted for. You would just need to change the dep value [here ](https://github.com/fossas/fossa-cli/blob/133a06358151a0d6292947707ddc8b3b3b2d3033/integration-test/Analysis/RubySpec.hs#L27) .

src/Strategy/Ruby/GemfileLock.hs
```diff
@@ -89,9 +91,9 @@ toDependency pkg = foldr applyLabel start
         }
 
     applyLabel :: GemfileLabel -> Dependency -> Dependency
-    applyLabel (GemfileVersion ver) dep = dep{dependencyVersion = Just (CEq ver)}
+    applyLabel (GemfileVersion ver) dep = dep{dependencyVersion = (dependencyVersion dep) <|> Just (CEq ver)}
```
JeffreyHuynh1, line 94: [prefer]: 
```suggestion
    applyLabel (GemfileVersion ver) dep = dep{dependencyVersion = dependencyVersion dep <|> (Just . CEq) ver}
```

- Remove paren
- Changed to Just (CEq ver) -> (Just . CEq) ver , to keep similar convention to your changes on line 96

src/Strategy/Ruby/GemfileLock.hs
```diff
@@ -257,8 +259,12 @@ dependenciesSectionParser = L.nonIndented scn $
     _ <- chunk "DEPENDENCIES"
     pure $ L.IndentMany Nothing (pure . DependencySection) findDependency
 
+-- The "!" suffix check checks for a Bundler convention that uses !'s to signify a dep is from another remote.
+-- We already check to se if the dep is part of another remote, so we can remove it.
```
JeffreyHuynh1, line 263:
```suggestion
-- Check for Bundler convention that use !'s to signify if a dep is from another remote.
-- We already check to see if the dep is part of another remote, so we can remove it.
```

zlav:
@JeffreyHuynh1 Yea I think you nailed it on the integration test differences. The `!` previously caused the dependencies to be duplicated in the CLI's integration calculations.

JeffreyHuynh1 (approved):
This is looking good! Only had one question that wasn't addressed, for a dependency formatted like `mysql2 (~> 0.5)!`, is this intended to be categorized as a GitDep or GemDep? Currently, it's still being categorized as a GemDep.

zlav:
Screenshot 2024-04-18 at 12 40 16 PM
@JeffreyHuynh1 where are you seeing that? When I scan Rails 6.1.4 with this branch I see it show up as GitType

PR #1412, opened 2024-04-16
increase the timeouts for report API calls to 600 seconds

spatten:
# Overview

No Ticket

See https://teamfossa.slack.com/archives/C043EM3L96Z/p1713304985656999

We were having a problem running `fossa attribution report --format html ...` for a project that was taking a long time to run the report.

The fix is to increase the timeout when we hit the reports API.

## Acceptance criteria

We can generate the report without timing out.

## Testing plan

```
make install-dev
```

Do this in an empty directory with no git repo in it or its parents:

```
fossa-dev report attribution --format html --project 'projectname from slack message' --revision 'revision from slack message'
```

Repeat with `--format json`.

Note that it takes > 30 seconds to run, but it does not take forever.

The previous behavior was that it would make the API call, timeout after 30 seconds, and then try again. I don't know if it ever gave up or if it just kept trying forever.

## Risks

The report in question takes a bit over a minute to generate. Increasing the timeout to 5 minutes should be enough, but if you think it should be larger (10 minutes?) I'm totally willing to bump it up.

## References

https://teamfossa.slack.com/archives/C043EM3L96Z/p1713304985656999

## Checklist

### DIFF
--- a/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
+++ b/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
@@ -1183,8 +1183,8 @@ getAttributionJson apiOpts ProjectRevision{..} = fossaReq $ do
             =: True
           <> "dependencyInfoOptions[]"
             =: packageDownloadUrl
-          -- Large reports can take over a minute to generate, so increase the timeout to 10 minutes
-          <> responseTimeoutSeconds 600
+          -- Large reports can take over a minute to generate, so increase the timeout to 5 minutes
+          <> responseTimeoutSeconds 300
   orgId <- organizationId <$> getOrganization apiOpts
   response <- req GET (attributionEndpoint baseUrl orgId (Locator "custom" projectName (Just projectRevision)) ReportJson) NoReqBody jsonResponse opts
   pure (responseBody response)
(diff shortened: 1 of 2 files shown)


jssblck:
Why even have a default timeout? Users can specify one via `--timeout` if they want, does it actually benefit us in any way to have timeouts otherwise?

spatten:
> Why even have a default timeout? Users can specify one via `--timeout` if they want, does it actually benefit us in any way to have timeouts otherwise?

The default comes from the Network.HTTP.Req library. Search for "responseTimeout" on this page: https://hackage.haskell.org/package/req-3.13.2/docs/Network-HTTP-Req.html#g:12

jssblck:
In that case, is there a benefit to not setting the timeout to some huge value like `600` or even `3600`?

spatten:
> In that case, is there a benefit to not setting the timeout to some huge value like `600` or even `3600`?

Yeah, I'm open to that. I initially had it at 600 seconds and thought it was maybe too much, but I'm happy to go with 600 seconds.

I think we set it to 600 seconds already for one of the VSI endpoints.

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -1183,6 +1183,8 @@ getAttributionJson apiOpts ProjectRevision{..} = fossaReq $ do
             =: True
           <> "dependencyInfoOptions[]"
             =: packageDownloadUrl
+          -- Large reports can take over a minute to generate, so increase the timeout to 10 minutes
+          <> responseTimeoutSeconds 600
```
csasarak, line 1187: This seems fine, but I think it would be good to have a top-level `defaultResponseTimeout` and refer to that rather than write 600 in both places.

csasarak (approved):

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
spatten: The default is actually 30 seconds (from the library, and we are overriding it here.

The default comes from the Network.HTTP.Req library. Search for "responseTimeout" on this page: https://hackage.haskell.org/package/req-3.13.2/docs/Network-HTTP-Req.html#g:12

PR #1419, opened 2024-04-26
[No Ticket]: Change GitHub actions to use macos-12

JeffreyHuynh1:
# Overview

Previously we were using macos-latest for our macos runners. However, there are migrations to use ARM instead of AMD for macos-latest.

It has caused this test to fail:

- https://github.com/fossas/fossa-cli/actions/runs/8852343990

You can see that github has updated macos-latest here:
- https://github.blog/changelog/2024-04-01-macos-14-sonoma-is-generally-available-and-the-latest-macos-runner-image/

Changing to macos-12 to be conservative and based on the suggestion here:
- https://github.com/actions/runner-images/blob/main/README.md#latest-migration-process

## Testing plan

- Test via GH actions

## Checklist

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
-        os: ['windows-latest', 'ubuntu-latest', 'macos-latest']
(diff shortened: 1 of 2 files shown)


csasarak (approved):
This is working for me. After merging please download the binaries and test again before cutting a release.

PR #1420, opened 2024-04-30
Poetry: Support v1.5.0 or greater

meghfossa:
# Overview

This PR adds partial support for `dependency groups`.

Previously, FOSSA CLI presumed, that when category was not provided, dependency was production dependency. This assumption is incorrect starting with v1.5.0 of poetry. This PR provides partial fix.

## Acceptance criteria

- Poetry lock file generated by Poetry v1.5.0 can be accurately analyzed
 
## Testing plan

I relied on automated tests - but you can perform following.

```bash
git checkout master && git pull origin && git checkout temp/megh-git-poetry-dev
cabal run fossa -- analyze -o ./sandbox 
```

Place items from https://fossa.atlassian.net/browse/ANE-1281 in sandbox. (with these files you should see 0 deps)

## Risks

N/A

## References

https://fossa.atlassian.net/browse/ANE-1281

## Checklist


src/Strategy/Python/Poetry.hs
```diff
@@ -180,21 +187,68 @@ analyze PoetryProject{pyProjectToml, poetryLock} = do
 
 -- | Use a `pyproject.toml` to set the direct dependencies of a graph created from `poetry.lock`.
 setGraphDirectsFromPyproject :: Graphing Dependency -> PyProject -> Graphing Dependency
-setGraphDirectsFromPyproject graph pyproject = Graphing.promoteToDirect isDirect graph
+setGraphDirectsFromPyproject graph pyproject = graph'
   where
+    graph' :: Graphing Dependency
+    graph' = hydrateDepEnvs $ Graphing.promoteToDirect isDirect graph
```
csasarak: [nit] Prefer a more specific name here, like `envHydratedGraph`. nbd since it's a very local name though.

src/Strategy/Python/Poetry.hs
```diff
@@ -180,21 +187,68 @@ analyze PoetryProject{pyProjectToml, poetryLock} = do
...
+    labelOptionalDepsIfPoetryGt1_5 g = if isLockLt1_5 then g else Graphing.gmap markEmptyEnvAsOptionalDep g
+
+    isLockLt1_5 :: Bool
+    isLockLt1_5 =
+      "0" `Text.isPrefixOf` lockVersion
+        || "1.0" `Text.isPrefixOf` lockVersion
+        || "1.1" `Text.isPrefixOf` lockVersion
+        || "1.2" `Text.isPrefixOf` lockVersion
+        || "1.3" `Text.isPrefixOf` lockVersion
+        || "1.4" `Text.isPrefixOf` lockVersion
```
csasarak, line 215:
```suggestion
    isLockLt1_5 :: Bool
    isLockLt1_5 = any (`Text.isPrefixOf` lockVersion) ["0", "1.0", "1.1", "1.2", "1.3", "1.4"]
```

csasarak: I haven't tested that myself but you should be able to use a formulation like that ^^^

src/Strategy/Python/Poetry/Common.hs
```diff
@@ -154,20 +176,29 @@ toCanonicalName :: Text -> Text
...
     canonicalPkgName pkg = PackageName $ toCanonicalName $ unPackageName $ poetryLockPackageName pkg
 
+    canonicalPkgName' :: PackageName -> PackageName
+    canonicalPkgName' = PackageName . toCanonicalName . unPackageName
+
+    canonicalProdPkgNames :: [PackageName]
+    canonicalProdPkgNames = map canonicalPkgName' prodPkgs
+
+    isProductionDirectDep :: PoetryLockPackage -> Bool
+    isProductionDirectDep pkg = canonicalPkgName pkg `elem` canonicalProdPkgNames
```
csasarak, line 192: Prefer something like this which is potentially more descriptive than using `'`. Also, Please make `canonicalProdPkgNames` a `Set` since this could potentially be large and expensive to use `elem` with.

```suggestion
     
    canonicalPkgName :: PackageName -> PackageName
    canonicalPkgName = PackageName . toCanonicalName . unPackageName
    
    lockCanonicalPackageName :: PoetryLockPackage -> PackageName
    lockCanonicalPackageName = canonicalPkgName . poetryLockPackageName

    canonicalProdPkgNames :: Set [PackageName]
    canonicalProdPkgNames = Set.fromList . map canonicalPkgName $ prodPkgs

    isProductionDirectDep :: PoetryLockPackage -> Bool
    isProductionDirectDep pkg = canonicalPkgName pkg `Set.member` canonicalProdPkgNames
```

src/Strategy/Python/Poetry/Common.hs
```diff
@@ -154,20 +176,29 @@ toCanonicalName :: Text -> Text
 toCanonicalName t = toLower $ replace "_" "-" (replace "." "-" t)
 
 -- | Maps poetry lock package to map of package name and associated dependency.
-toMap :: [PoetryLockPackage] -> Map.Map PackageName Dependency
-toMap pkgs = Map.fromList $ (\x -> (canonicalPkgName x, toDependency x)) <$> (filter supportedPoetryLockDep pkgs)
+toMap :: [PackageName] -> [PoetryLockPackage] -> Map.Map PackageName Dependency
+toMap prodPkgs pkgs = Map.fromList $ (\x -> (canonicalPkgName x, toDependency x)) <$> (filter supportedPoetryLockDep pkgs)
```
csasarak, line 180: Could this be renamed to something like `makePackageToLockDependencyMap`? I know you didn't write this but `toMap` tells me very little about the purpose of this function at call-sites.

src/Strategy/Python/Poetry/Common.hs
```diff
@@ -200,16 +231,19 @@ toMap pkgs = Map.fromList $ (\x -> (canonicalPkgName x, toDependency x)) <$> (fi
           ref <- poetryLockPackageSourceReference lockPkgSrc
           if poetryLockPackageSourceType lockPkgSrc /= "legacy" then Just ref else Nothing
 
-    toDepEnvironment :: PoetryLockPackage -> DepEnvironment
+    toDepEnvironment :: PoetryLockPackage -> Set.Set DepEnvironment
```
csasarak, line 234: I think a name like `pkgEnvironments` is more descriptive.

src/Strategy/Python/Poetry/PyProject.hs
```diff
@@ -113,10 +115,38 @@ data PyProjectPoetry = PyProjectPoetry
...
+    -- test group only. Note that any dependencies from these groups are excluded
+    -- by default. refer to: https://github.com/kowainik/tomland/issues/336
+    groupDevDependencies :: Map Text PoetryDependency
+  , groupTestDependencies :: Map Text PoetryDependency
   }
   deriving (Show, Eq, Ord)
 
+allPoetryProductionDeps :: PyProject -> Map Text PoetryDependency
+allPoetryProductionDeps project = case pyprojectPoetry project of
+  Just (PyProjectPoetry{dependencies}) -> dependencies
```
csasarak, line 134: Not important for this ticket, but in the future we should probably make the `PyProject` type be `PyProject a` where tool specific things live inside the `a`. that way we can specify that a given set of code is only valid for `PyProject Poetry` rather than relying on runtime checks (checking if a field is `Just` or `Nothing` ) like we do here.

csasarak (changes requested):

csasarak (approved):

PR #1425, opened 2024-05-02
Don't put files in subdir when extracting

jssblck:
# Overview

Updates the `EmbeddedBinary` logic to suffix the `fossa-vendor` directory name with a UUID instead of nesting a UUID folder inside.

This provides complete isolation to multiple callers, such that even if they (or a system process) cleans up the parent after they're done they don't mess with other callers.

## Acceptance criteria

Integration tests pass again

## Testing plan

Relying on integration tests- if this breaks anything with extraction we'll get way more errors, if this fixes the issue observed in other PRs we'll get none.

## Risks

None

## Metrics

None

## References

Resolves https://fossa.atlassian.net/browse/ANE-1742

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## v3.9.16
+- Updates parallel embedded binary extractions to be more properly isolated ([#1425](https://github.com/fossas/fossa-cli/pull/1425)).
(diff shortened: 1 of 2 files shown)


csasarak (approved):
This is great, thanks for tracking that down!

PR #1428, opened 2024-05-10
Fix Mac ARM release process

csasarak:
# Overview

I made a mistake in my last PR where I didn't change the command to make the final release archives correctly. It was putting the intel binaries into the arm archive. 

I also addressed an issue that came up with code-signing while testing the release action. On ARM the liblzma dylib fails during the verification of code signing when running `fossa` because the library is not signed by either Apple or us:

```
dyld[11361]: Library not loaded: /opt/homebrew/opt/xz/lib/liblzma.5.dylib
  Referenced from: <339C808F-7D5F-3897-BEE5-FF154705274E> /Users/ec2-user/arm_macos/fossa
  Reason: tried: '/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (code signature in <1E29DFAD-90C2-3D94-889D-7A212125B52B> '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (no such file), '/opt/homebrew/opt/xz/lib/liblzma.5.dylib' (code signature in <1E29DFAD-90C2-3D94-889D-7A212125B52B> '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' (code signature in <1E29DFAD-90C2-3D94-889D-7A212125B52B> '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' (no such file), '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' (code signature in <1E29DFAD-90C2-3D94-889D-7A212125B52B> '/opt/homebrew/Cellar/xz/5.4.6/lib/liblzma.5.dylib' not valid for use in process: mapping process and mapped file (non-platform) have different Team IDs)
```

On my Intel system, the liblzma dylib seems to be built-in and present in `/usr/local/lib` which seems to be trusted. But on ARM it is from homebrew. I spent some time investigating solutions hoping to find one that would allow the homebrew liblzma to pass verification but could not. The solution I came up with is to [disable the library verification](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_cs_disable-library-validation) using `entitlements.plist`. This is less than ideal, but I could not find another way to make this work that wouldn't involve changing the way we distribute our OS X binaries to bundle in liblzma. 

You can find a copy of the signed and notarized arm binaries with the entitlements [here](https://github.com/fossas/fossa-cli/actions/runs/9056923270).

## Acceptance criteria

It puts the ARM binaries into the arm archive.

## Testing plan

I pushed dummy tags `v9.9.9` and used these to test the release action. I then downloaded and ran the fossa arm binaries on an EC2 instance. If whoever reviews this is using an apple silicon system it would be good if they could also test the signed binaries I link to above.

## Risks

There is the additional entitlement for Mac OS X binaries on arm. I think the actual risk here is small, but it is a risk. 

## Checklist


JeffreyHuynh1 (approved):
I've tested the linked arm binaries and it works well on my machine

PR #1429, opened 2024-05-10
Revert "Add Mac arm64 executable (#1426)"

csasarak:
This reverts commit 75363fc11aa959ffed8cc7c879382d0e3533c9a3.

# Overview

During the initial go of implementing this I missed some issues with codesigning and notarization on Mac that need to be resolved. This PR reverts the mac arm release so that it doesn't get in the way of any other releases.

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -19,7 +19,7 @@ jobs:
 
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


JeffreyHuynh1 (approved):
LGTM as this is just a revert!

PR #1431, opened 2024-05-22
CPAL license

jssblck:
# Overview

Changes the FOSSA CLI license to CPAL, an OSI-approved license that is very close to MPL.


spatten (approved):
Looks good to me 👍

PR #1433, opened 2024-05-22
Update issueDashURL to contain correct revision branch

JeffreyHuynh1:
# Overview

`fossa test` reports the issue URLs incorrectly. It points to the master branch even when last analysis was performed on a different branch. 

## Acceptance criteria

issueDashURL should contain the correct branch. 

## Testing plan
To view current state of our issue urls in `fossa test`

- `fossa analyze /path/to/project/` a project on a branch that is NOT `master`
- `fossa test /path/to/project/` 
- You will see that the issueDashURL will contain `/refs/branch/master` 

To view updated issue urls

- `git checkout fix-issueUrl` 
- `fossa test /path/to/project/` 

## Risks

I have decided to update the `issueDashURL` in the CLI rather than in Core's /api/cli/:locator/issues endpoint for the following reasons:
- There is no way to generate the `issueDashURL` with the correct branch in the /api/cli/:locator/issues endpoint without breaking backwards compatibility.
    - Updating the API to accept the project's branch would be necessary to ensure all users generate URLs with correct branch info. Without explicitly passing the project's branch to the API, it would be impossible to associate a revision's scan with the correct branch.

## References

[ANE-1771](https://fossa.atlassian.net/jira/software/c/projects/ANE/boards/66?selectedIssue=ANE-1771)

## Checklist

[ANE-1771]: https://fossa.atlassian.net/browse/ANE-1771?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Test.hs
+++ b/src/App/Fossa/Test.hs
@@ -20,6 +20,7 @@ import App.Fossa.PreflightChecks (PreflightCommandChecks (TestChecks), preflight
 import App.Fossa.Subcommand (SubCommand)
 import App.Types (
   ProjectRevision (projectName, projectRevision),
+  projectBranch,
  )
 import Control.Algebra (Has)
 import Control.Carrier.Debug (ignoreDebug)
@@ -30,7 +31,7 @@ import Control.Effect.Lift (Lift)
 import Control.Timeout (timeout')
 import Data.Aeson qualified as Aeson
 import Data.String.Conversion (decodeUtf8)
-import Data.Text (Text)
+import Data.Text (Text, replace)
 import Data.Text.Extra (showT)
 import Effect.Logger (
   Logger,
@@ -41,7 +42,7 @@ import Effect.Logger (
   pretty,
   vsep,
  )
-import Fossa.API.Types (Issues (..))
+import Fossa.API.Types (Issue, Issues (..), issueDashURL)
 
 testSubCommand :: SubCommand TestCliOpts TestConfig
 testSubCommand = Config.mkSubCommand testMain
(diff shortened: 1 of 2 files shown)


src/App/Fossa/Test.hs
```diff
@@ -119,3 +125,15 @@ testMain config = do
 
     renderJson :: (Has (Lift IO) sig m, Has Logger sig m) => Issues -> m ()
     renderJson = logStdout . decodeUtf8 . Aeson.encode
+
+    -- `issueDashURL` will have the branch set to `master` for all issues. The url will contain: `/refs/branch/master`
+    -- Update `issueDashURL` for all issues with the current revision's branch.
```
zlav, line 130: Can you add a specific example of before and after? I think this is close, but just write out something like `/refs/branch/differentbranch`

zlav (commented):
This PR looks good, but I'm concerned about the assumption that we can get the branch information locally from the user's machine and use that to expect the URL exists within FOSSA. The reason I say this is because we're using `projectBranch revision` which is what the CLI is submitting to Core, and then we are modifying the response URL from Core to include the branch that we found locally.

This feels brittle to me. I buy your reasoning about being backwards compatible in Core, but I do think that core is the correct place to fix this. I'm open to hearing otherwise.

My only other idea that doesn't break backwards compatibility is to return 1 branch associated with the revision, since that is strictly more likely to be accurate than returning master. (Partially because we don't support [linear git history](https://fossa.atlassian.net/browse/ANE-963), so must revisions only belong to 1 branch)

Another idea would be to validate that the branch, revision, and URL are correct by making some other API call to core.

JeffreyHuynh1:
Making the changes for correct issue dash url in Core. Here's the pr: https://github.com/fossas/FOSSA/pull/12973

PR #1434, opened 2024-05-24
Fix license file matches

JeffreyHuynh1:
# Overview

The changes from pr: [Support true forced rebuilds for vendored dependencies, refactor types](https://github.com/fossas/fossa-cli/pull/1423/files) introduced a bug that caused the contents of the license file matches to be empty. 

image-20240515-194646

Before the changes, the license file match contents appeared as: 

Screenshot 2024-05-23 at 5 05 01 PM

After some investigation, it was discovered that the revision's license scan type was being set to `full_files` . This shouldn't have been the case because the given repro case should've had the scan type set to `match_data` 

Note: Refer to the linked ANE ticket for repro case

## Acceptance criteria

- Ensure that the contents of the license file matches are not empty

## Testing plan

- `fossa analyze`  on the project linked in the ANE ticket
- Check the license file match contents 
- Check that the license scan type is set to the proper type: `match_data`

## Risks

- No risk

## References

- [ANE-1769](https://fossa.atlassian.net/browse/ANE-1769)

## Checklist

[ANE-1769]: https://fossa.atlassian.net/browse/ANE-1769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## 3.9.19
 - Updated the license to CPAL, an OSI-approved license similar to MPL ([#1431](https://github.com/fossas/fossa-cli/pull/1431)).
+- Fixes file matches for license scans ([#1434](https://github.com/fossas/fossa-cli/pull/1434)).
 
 ## v3.9.18
 - Resolves an issue where `vendored-dependencies` were rescanned locally, but not in the FOSSA service,
--- a/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
+++ b/src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
@@ -98,7 +98,7 @@ import App.Support (
  )
 import App.Types (
   DependencyRebuild,
-  FileUpload (FileUploadFullContent),
+  FileUpload (FileUploadMatchData),
   Policy (..),
   ProjectMetadata (..),
   ProjectRevision (..),
@@ -947,7 +947,7 @@ archiveBuildUpload' apiOpts rebuild archive = context ("archive: '" <> toText ar
   (baseUrl, baseOpts) <- useApiOpts apiOpts
 
   let opts = "dependency" =: True <> "rawLicenseScan" =: True
-  let archiveProjects = ArchiveComponents [archive] rebuild FileUploadFullContent
+  let archiveProjects = ArchiveComponents [archive] rebuild FileUploadMatchData
 
   -- The response appears to either be "Created" for new builds, or an error message for existing builds.
   -- Making the actual return value of "Created" essentially worthless.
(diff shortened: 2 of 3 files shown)


src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
```diff
@@ -947,7 +947,7 @@ archiveBuildUpload' apiOpts rebuild archive = context ("archive: '" <> toText ar
   (baseUrl, baseOpts) <- useApiOpts apiOpts
 
   let opts = "dependency" =: True <> "rawLicenseScan" =: True
-  let archiveProjects = ArchiveComponents [archive] rebuild FileUploadFullContent
```
JeffreyHuynh1, line 950: Based on the previous code: https://github.com/fossas/fossa-cli/pull/1423/files#diff-a812ed3f7085f40e35cb15511136c5dea108406ba8c21818dde5a4029b9302d5L942

I've changed this to FileUpload  to `FileUploadMatchData`

src/Fossa/API/Types.hs
```diff
@@ -151,7 +151,9 @@ instance ToJSON ArchiveComponents where
   toJSON ArchiveComponents{..} =
     object
       [ "archives" .= archiveComponentsArchives
-      , "fullFiles" .= archiveComponentsUpload
```
JeffreyHuynh1, line 154: Based on the previous code: https://github.com/fossas/fossa-cli/pull/1423/files#diff-78dfaf45d3d47ee1eb7bc8a678e57e40022f78c8c394da5c9adb61582e3c63c4L154

I've changed the json value for `archiveComponentsUpload` to be set to a boolean.

src/Control/Carrier/FossaApiClient/Internal/FossaAPIV1.hs
jssblck: Since this is an archive upload, why isn't `FullContent` appropriate here?

JeffreyHuynh1: Ah, I changed this to be `FileUploadMatchData` due to the fact that `archiveProjects` was set to `FullFileUploads False` previously. But after looking at Scott's comments on your initial pr, it seems that this was intended bc we shouldn't ever do an archive upload without full files. 

That's my fault, changing this back to `FileUploadFullContent`

Changelog.md
```diff
@@ -1,7 +1,8 @@
 # FOSSA CLI Changelog
 
-## Unreleased
+## 3.9.19
```
jssblck, line 3: lgtm, let's yank 3.9.18 after releasing

jssblck (approved):

PR #1440, opened 2024-06-14
Ane 1808 suppress reachability notice

csasarak:
# Overview

Reachability analysis is quite noisy, and many customers don't have it enabled for their organization anyways. This PR changes it so that:

1. Reachability analysis is skipped when the organization doesn't have reachability enabled. This means that "reachability not supported" messages as well as the entire reachability summary isn't printed out.
2. For an organization that does have it enabled, the messages will be the same. These are noisy, but we can refine them a bit in the future.
3. It still outputs a single message about reachability being turned off at the end of analysis regardless of whether reachability is enabled or not.
4. It will still try to do analysis and print out the analysis log messages with the `-o` option. We have no way of knowing if reachability is enabled in that case so it defaults to doing it. 

## Acceptance criteria

The output isn't nearly as noisy.

## Testing plan

I tested by analyzing projects while having reachability analysis enabled or disabled in my organization.

## Risks

None.

## Metrics

## References

[ANE-1808](https://fossa.atlassian.net/browse/ANE-1808)

## Checklist

[ANE-1808]: https://fossa.atlassian.net/browse/ANE-1808?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+
+- Reachability: For organizations that don't have reachability turned on suppress messages about it. ([#1440](https://github.com/fossas/fossa-cli/pull/1440))
+
 ## 3.9.22
 - Fixes release group flags for `fossa analyze` and `fossa container analyze`  ([#1439](https://github.com/fossas/fossa-cli/pull/1439))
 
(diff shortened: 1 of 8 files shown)


JeffreyHuynh1 (approved):
I've tested this locally and it's working as intended. I'm still seeing the logs statements below at the end of analysis as well,  it looks good!

```
Organization: (39599) does not support reachability. Skipping reachability analysis upload.
For reachability, refer to: https://docs.fossa.com/docs/reachability
```

PR #1442, opened 2024-06-20
Ane 1711 millhone jar cli program [1/3]

csasarak:
# Overview

This PR adds functionality to millhone for analyzing a container tar file for Jars and reporting fingerprints. It is a lightly adapted version of code we already had in another project. It isn't really set up yet to fingerprint arbitrary jars but should be adaptable for that purpose quite easily.

I am going to stack a couple PRs. This one is just the Rust side of things. For a review, I'm mainly looking for a sanity check on what I have so far. 

## Acceptance criteria

* This isn't missing any obvious piece of data that you think we will need or has egregious problems.

## Testing plan

My testing has been manual so far. I've tested with the following images: `cassandra`, `jetty`, `gradle`, and `quay.io/keycloak/keycloak:25.0.1`.

To test with an image from the root cli directory:
```sh
docker pull quay.io/keycloak/keycloak:25.0.1
docker save quay.io/keycloak/keycloak:25.0.1 > keycloak.tar
cargo run --bin millhone -- --log-to stderr analyze-jars keycloak.tar | jless
```

Note that this will take a bit of time as written. Building with `--release` significantly improves performance for this operation so that shouldn't be concern.

For now, I haven't added an automated test because I want to try to get everything working manually end-to-end with the CLI.

## Risks

None. This is new functionality and doesn't touch any existing Millhone work.

## References

[ANE-1711](https://fossa.atlassian.net/browse/ANE-1711)

## Checklist

[ANE-1711]: https://fossa.atlassian.net/browse/ANE-1711?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+
+#[tracing::instrument]
+pub fn main(opts: Subcommand) -> Result<(), Report> {
+    let tar_filename = opts.image_tar_file();
+    let jar_analysis = JarAnalysis {
+        discovered_jars: jars_in_container(opts.image_tar_file())
+            .with_context(|| format!("Analyze Jar Archive {:?}", tar_filename))?,
+    };
+
+    serde_json::to_writer(std::io::stdout(), &jar_analysis).context("Serialize Results")
```
jssblck, line 81: prefer to wrap `stdout` in a bufwriter for improved performance:
```suggestion
    let mut stdout = BufWriter::new(std::io::stdout());
    serde_json::to_writer(&mut stdout, &jar_analysis).context("Serialize Results")
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// Encountered when a manifest file is found but cannot be parsed.
+    #[error("Parsing manifest file {0} {1}")]
+    ParseManifest(PathBuf, serde_json::Error),
+
+    #[error("Reading buffer {0}")]
+    ReadBuffer(std::io::Error),
+
+    #[error("Generating fingerprint {0}")]
+    GenerateFingerprint(fingerprint::Error),
+}
```
jssblck, line 71: We're using `stable_eyre` elsewhere, why do we even want an error enum here vs ad-hoc `.context()`?

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+        if !layers.contains(path.as_ref()) {
+            debug!(?path, "skipped: not a layer file");
+            continue;
+        }
+
+        let layer = path.to_path_buf();
+        // Layers should have a form like blob
+        let layer_discoveries = jars_in_layer(entry)?;
+        discoveries.insert(layer, layer_discoveries);
+    }
```
jssblck, line 108: [nit, optional] if we used `stable_eyre` for this function, it'd read a lot cleaner:
```suggestion
    let mut image = unpack(image_path)?;
    for entry in image.entries().context("iterate entries")? {
        let entry = entry.context("read entry")?;
        let path = entry.path().context("read path")?;
        if !layers.contains(path.as_ref()) {
            debug!(?path, "skipped: not a layer file");
            continue;
        }

        let layer = path.to_path_buf();
        // Layers should have a form like blob
        let layer_discoveries = jars_in_layer(entry)
            .with_context(|| format!("read layer '{layer}'"))?;
        discoveries.insert(layer, layer_discoveries);
    }
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+                    discoveries.push(DiscoveredJar { fingerprint, path });
+                    Ok(())
+                }
+                Err(e) => Err(Error::GenerateFingerprint(e)),
+            }
+        });
+
+        if let Err(e) = res {
+            warn!("Fingerprinting: {:?}", e);
+        }
```
jssblck, line 150: [nit] emitting the `warn` where it is omits the path from the context; prefer emitting inline unless that was desired:
```suggestion
        let path = path.to_path_buf();
        info_span!("jar", ?path).in_scope(|| {
            debug!("fingerprinting");
            let entry = buffer(entry)?;

            match Combined::from_buffer(entry) {
                Ok(fingerprint) => discoveries.push(DiscoveredJar { fingerprint, path })
                Err(e) => warn!("failed to fingerprint: {e:?}"),
            }
        });
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+#[tracing::instrument]
+fn list_container_layers(layer_path: &PathBuf) -> Result<HashSet<PathBuf>, Error> {
+    let mut layers = HashSet::new();
+
+    let mut container = unpack(layer_path)?;
+    for entry in wrap_tar_err!("list entries", container.entries())? {
+        let entry = match entry {
+            Ok(entry) => entry,
+            Err(e) => {
+                warn!("Reading entry {:?}", e);
```
jssblck, line 165: Warnings are emitted to users, and unlike errors aren't nested, so make sure they're relatively complete sentences. Also, prefer to inline variables in format statements.
```suggestion
                warn!("failed to read entry: {e:?}");
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        // There's only one manifest file.
+        break;
+    }
+
+    Ok(layers)
+}
+
+#[track_caller]
```
jssblck, line 198: I'm pretty sure `track_caller` does nothing useful in this context
```suggestion
```

jssblck (approved):

extlib/millhone/src/cmd/analyze_jars.rs
csasarak: I think I started down the road of using an enum and then kept going. I think ad-hoc `.context()` is fine since there's nowhere I actually inspect the results. I will make this change.

csasarak: I think I've resolved this and other feedback. Going to wait to merge until more is done on the second PR. Thanks for having a look!

extlib/millhone/src/cmd.rs
```diff
@@ -20,6 +20,7 @@ use typed_builder::TypedBuilder;
 use walkdir::DirEntry;
 
 pub mod analyze;
+pub mod analyze_jars;
```
jssblck, line 23: [nit, optional] i'd call this "analyze container" or something to make it more clear

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,188 @@
...
+    /// Jars and fingerprints associated with each layer in a jar file.
+    discovered_jars: HashMap<LayerPath, Vec<DiscoveredJar>>,
+}
+
+#[tracing::instrument]
+pub fn main(opts: Subcommand) -> Result<()> {
+    let tar_filename = opts.image_tar_file();
+    let jar_analysis = JarAnalysis {
+        discovered_jars: jars_in_container(opts.image_tar_file())
+            .with_context(|| format!("Analyze Jar Archive {:?}", tar_filename))?,
```
jssblck, line 65: [nit, optional] similarly i'd call this something like "analyze container"
also nitty and optional but remember that errors are commonly rendered like `error: {cause}: {cause}: ...`, so making them all lower case usually reads better
```suggestion
            .with_context(|| format!("analyze container: {:?}", tar_filename))?,
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/// Extracts the container (saved via `docker save`) and finds JAR files inside any layer.
+/// For each one found, fingerprints it and reports all those fingerprints along with their
+/// layer and path.
+#[tracing::instrument]
+fn jars_in_container(image_path: &PathBuf) -> Result<HashMap<LayerPath, Vec<DiscoveredJar>>> {
+    // Visit each layer and fingerprint the JARs within.
+    info!("inspecting container");
+    // May have to make two copies of this Archive, idk if it can be iterated twice.
```
jssblck, line 79: [nit] it looks like this was answered below (in `unpack`'s comment)
```suggestion
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,188 @@
...
+        if !path.to_string_lossy().ends_with(".jar") {
+            debug!(?path, "skipped: not a jar file");
+            continue;
+        }
+
+        let path = path.to_path_buf();
+
+        info_span!("jar", ?path).in_scope(|| -> Result<()> {
+            debug!("fingerprinting");
+            let entry = buffer(entry).context("Read jar file")?;
```
jssblck, line 127: [nit] due to the common ways errors are rendered, i recommend keeping the messages in `context` all lower case
```suggestion
            let entry = buffer(entry).context("read jar file")?;
```

extlib/millhone/src/cmd/analyze_jars.rs
```diff
@@ -0,0 +1,188 @@
...
+            Err(e) => {
+                warn!("failed to read entry: {e:?}");
+                continue;
+            }
+        };
+
+        let path = match entry.path() {
+            Ok(path) => path,
+            Err(e) => {
+                warn!("Entry path {:?}", e);
```
jssblck, line 158: [nit] inline args to format macros; also warnings _are not_ reported with error report mechanisms so should contain more explanation:
```suggestion
                warn!("Failed to read entry path: {e:?}");
```

jssblck (approved):

PR #1448, opened 2024-07-11
[ANE-1809] - Partial distroless container support

csasarak:
# Overview

Right now, if the CLI tries to do a container scan with no OS or OS Release information set, it will fail. This PR changes that the CLI can still return results for things that aren't system deps inside of distroless containers, e.g. Jars and any of our static programming language analyses.

System deps will still not be supported because we don't have a way to recognize distribution package locators that don't have the OS or OS release information. The team has some ideas of how to address that, but isn't really feasible to fix in the short-term, where this provides some value and paves the way for that later work. 

## Acceptance criteria

Containers with no linux distro information can be scanned and partial results uploaded.

## Testing plan

Please see this [related PR](https://github.com/fossas/FOSSA/pull/13234) for information on reproducing.

## Risks

There is potentially an explainability problem for customers - why does FOSSA find my npm/pip/etc. dependencies but not system deps?. I think this is still better than the current status quo though which is "Why doesn't FOSSA find any deps at all?"

## References

_Example:_

- [ANE-1809](https://fossa.atlassian.net/browse/ANE-1809)

## Checklist

[ANE-1809]: https://fossa.atlassian.net/browse/ANE-1809?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App/Fossa/Container/Sources/DockerArchive.hs
```diff
@@ -122,8 +122,9 @@ analyzeFromDockerArchive systemDepsOnly filters withoutDefaultFilters tarball =
   let imageBaseLayer = baseLayer image
       baseDigest = layerDigest imageBaseLayer
   osInfo <-
-    context "Retrieving OS Information" $
-      runTarballReadFSIO baseFs tarball getOsInfo
+    context "Retrieving OS Information"
+      . warnThenRecover @Text "Could not retrieve OS info"
```
csasarak, line 126: Should I log on stdout if we can't get system info? Warnings only print to the terminal if they use `--debug`.

spatten: Hmm. I think this is good as is, but you're right that it would be better to log to stdout.

The log to stdout should be clear that "this is expected but we just wanted you to know in case it was confusing to you".

I was thinking we could link to the documentation, but the thing you'd like to is short enough that it's better to just include it in the message

docs/references/subcommands/container/scanner.md
```diff
@@ -213,6 +213,11 @@ It will then report them to FOSSA which will try to match the Jar files to the p
 This process relies on there being a back-end that can perform that analysis.
 SaaS customers should have this functionality available but on-prem customers may need to contact FOSSA support to have it enabled.
 
+### Distroless Containers
```
spatten, line 216: The ToC didn't get updated with this new header. You need to add this between the "Container Jar Analysis"  and "Supported Container Package Managers" sections:

```
    - [Distroless Containers](#distroless-containers)
```

spatten (approved):
Looks good to me! Sorry this took so long

docs/references/subcommands/container/scanner.md
csasarak: Good catch, I've updated this.

PR #1455, opened 2024-07-31
[ANE-1639] add metadata to vendored dependencies

jagonalez:
# Overview

This PR adds a metadata field for vendored dependencies. This will enable vendored dependencies to be specified with a homepage and description.

## Acceptance criteria

A homepage and description is supported for vendored dependencies

## Testing plan

- create a directory `bar` with a vendored dependency
- create a fossa-deps.yaml file like the following:
```
vendored-dependencies:
- name: metadata-test
  path: bar/
  version: 1.10.0
  metadata:
    homepage: https://fossa.com
    description: "metadata-test"
```
- run `cabal run fossa -- analyze /path/to/folder`
- navigate to the project created in FOSSA, and click on dependencies, if you click on the dependency you should see the homepage and description populated with your values

## Risks

n/a

## Metrics

n/a

## References

https://fossa.atlassian.net/browse/ANE-1639

## Checklist


src/Fossa/API/Types.hs
```diff
@@ -164,6 +164,8 @@ instance ToJSON ArchiveComponents where
 data Archive = Archive
   { archiveName :: Text
   , archiveVersion :: Text
+  , archiveDescription :: Maybe Text
+  , archiveHomePage :: Maybe Text
```
csasarak, line 168: [Optional] A type like this is a bit more descriptive:
```haskell
newtype ArchiveDescription = ArchiveDescription Text
  deriving (Eq, ToJSON, ToText)

newtype ArchiveHomePage = ArchiveHomePage Text
  deriving (Eq, ToJSON, ToText)
  
data Archive = Archive
  { archiveName :: Text
  , archiveVersion :: Text
  , archiveDescription :: Maybe ArchiveDescription
  , archiveHomePage :: Maybe ArchiveHomePage
```

The main benefits here are that it prevents accidentally reordering the values when using `Archive` positionally †. It also means that the type of `getMetadata` includes a bit more documentation on the intended use of the function. You should be able to use derivations so you don't have to write any new code for the `newtype`s.

 † I have [opinions](https://github.com/fossas/fossa-cli/pull/1435#discussion_r1621219423)™ around using data constructors positionally, but it's pervasive in our code-base so maybe consider it but don't feel like you have to actually make changes around this idea.

test/App/Fossa/ManualDepsSpec.hs
```diff
@@ -53,8 +53,9 @@ theWorks = ManualDependencies references customs vendors remotes locators
       , RemoteDependency "url-dep-two" "1.2.4" "www.url2.tar.gz" Nothing
       ]
     vendors =
-      [ VendoredDependency "vendored" "path" Nothing
-      , VendoredDependency "versioned" "path/to/dep" (Just "2.1.0")
+      [ VendoredDependency "vendored" "path" Nothing Nothing
+      , VendoredDependency "versioned" "path/to/dep" (Just "2.1.0") Nothing
+      , VendoredDependency "metadata" "path" (Just "1.1.0") (Just (DependencyMetadata (Just "description for vendored") (Just "we don't validate homepages - vendored")))
```
csasarak, line 58: Should we validate these to ensure that they are URLs? If they send an invalid one up to Core would it just put invalid data in the UI/report? I guess more generally, where would we rather the failure happen in this case?

csasarak (approved):
Looks good.

test/App/Fossa/ManualDepsSpec.hs
jagonalez: Good question - as far as I know, we only display the homepage in the UI / reports, I'm not too concerned about invalid URLs being reported but it might warrant a discussion with the team.

csasarak: I think it's reasonable as-is if that's the case. On second thought thinking about my initial question I think it's hopefully obvious to the customer that it's their fault if one of these urls is messed up.

PR #1457, opened 2024-08-12
[ANE-695] gradle: check if configuration is deprecated

jagonalez:
# Overview
- PR adds support for ignoring deprecated configurations in gradle 6-7, which the gradle dependencies task does [see here](https://github.com/gradle/gradle/blob/v6.0.0/subprojects/diagnostics/src/main/java/org/gradle/api/tasks/diagnostics/AbstractDependencyReportTask.java#L110)

- In gradle 6+, gradle introduced [deprecated configurations](https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:deprecated-configurations), however including the configurations at first only induced a warning. In the `gradle dependencies` task, those configurations are ignored when attempting to resolve and list dependencies. This PR changes our gradle script to match what the `gradle dependencies` task does.

## Acceptance criteria

- deprecated configurations are ignored for gradle projects in v6-7.

## Testing plan

- use this repo: https://github.com/jagonalez/fossa-gradle-test (this repo is using a maven plugin which has dependencies in an ignored configuration).
- build the project
- `cabal build fossa`
- run `cabal run fossa -- analyze ~/dev/fossa-gradle-test -o | jq  ".projects[0].graph.deps | length"`
- you should get 33 (these are the same as what is reported from `gradlew dependencies`
- on the master branch we're getting 38 deps

## Risks

- should be minimal

## Metrics

## References

_Example:_

- [ANE-695](https://fossa.atlassian.net/browse/ANE-695)

## Checklist

[ANE-695]: https://fossa.atlassian.net/browse/ANE-695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jssblck (changes requested):
The code changes look reasonable, but what is the reasoning for why this is the right path to take? I know you linked to some Gradle code, but I'm not sure how that relates to specifically what we're doing.

Please add that reasoning in as a comment (so we know "why", not just "what") and document this behavior in the gradle strategy for users.

jagonalez:
> I know you linked to some Gradle code, but I'm not sure how that relates to specifically what we're doing.

The code I linked is for how Gradle reports dependencies when running `gradle dependencies`. In Gradle 6 - 7, gradle deprecated configurations: https://docs.gradle.org/current/userguide/declaring_dependencies.html#sec:deprecated-configurations. 

If you compare `gradle dependencies` vs the fossa-cli output when a gradle project has a deprecated config - we are reporting more dependenices than `gradle dependencies`. IMO we should be reporting the same dependencies that gradle does.

jssblck (approved):

PR #1460, opened 2024-08-16
Ghc 9.8 upgrade

csasarak:
# Overview

This PR is to upgrade us to be on GHC 9.8. There are a few reasons I think this is good: 

1. 9.8 is still being developed and is considered [stable](https://gitlab.haskell.org/ghc/ghc/-/wikis/GHC-status). 9.4 is considered stable and is recommended by ghcup, but will not have any future releases.
2. Bindists for aarch that are built against musl are officially available starting with GHC 9.8.2. This makes a release for Linux ARM much more doable on GHC 9.8.2 and above than GHC 9.4.

[Release Notes](https://downloads.haskell.org/ghc/9.6.1/docs/users_guide/9.6.1-notes.html) and [9.4 -> 9.6 migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.6).
[Release Notes](https://www.haskell.org/ghc//blog/20231009-ghc-9.8.1-released.html) and [9.6 -> 9.8 migration guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.8).

Very few of the changes to GHC or boot libraries seem to affect us day to day. Things that will:
* The removal of `Alternative` instances for `Either` ([docs](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.6#transformers-06-error-a--alternative-either-a-and-error-a--monadplus-either-a-instances-removed)) which was easily handled because the `Semigroup` instance for `Either` behaves the right way. 
* There is a new compiler warning for partial functions `head`/`tail` as proposed [here](https://github.com/haskell/core-libraries-committee/blob/main/guides/warning-for-head-and-tail.md). I've turned these off for our CI builds because we already check for these using `hlint` to do this in a more refined way. They will be on for developers, but as warnings.

## Acceptance criteria

* Fossa CLI builds with GHC 9.8 in CI.
* All automated tests work.

## Testing plan

* I used automated testing. 
* I downloaded the mac intel binary and verified that both embedded binaries not in this repository were callable.

## Risks

I don't think there are any that can be addressed further than I have. 

## References

[ANE-1757](https://fossa.atlassian.net/browse/ANE-1757)

## Checklist

[ANE-1757]: https://fossa.atlassian.net/browse/ANE-1757?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JeffreyHuynh1 (approved):
I don't see any issues with the upgrade and the passing tests give me confidence that it's a safe change! Also looked over the release notes for the subsequent upgrades and they're pretty minimal changes.

PR #1461, opened 2024-08-21
NuGet strategy consolidation

JeffreyHuynh1:
# Overview

All of the CLI’s nuget strategies run everytime fossa analyze is run. This causes duplicate information to be uploaded on almost every run. The worst part of this is that one of the strategies analyzes a lockfile with determined versions and the other analyzes a manifest with versions that may not be present in the final build.

From Nugets website

The project.json file maintains a list of packages used in a project, known as a package management format. It supersedes packages.config but is in turn superseded by [PackageReference](https://learn.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files) with NuGet 4.0+.

https://fossa.com/blog/managing-dependencies-net-csproj-packagesconfig/

## Acceptance criteria

- Nuget analysis strategies are not all run and utilize fallbacks to ensure that duplicate dependencies are not uploaded.

- Implementation decisions here do not prevent https://fossa.atlassian.net/browse/ANE-702 from being implemented.

- Consolidate analysis for project.assets.json and package reference files because they have the same entry point, .csproj files. The other strategies can remain the same and be ran independently.

Documentation is updated to display what the fallback order is.

## Testing plan

- Refactored existing test to work with the consolidation and it's working as it did previously.

## Risks

[Question] - Was looking into writing an integration test to test the case where we have both `.csproj` files (triggers PackageReference strategy) and `project.assets.json` files, that way we can test out the consolidation changes in this pr. After looking at our integration tests, it looks like we test against [open source repos](https://github.com/fossas/fossa-cli/blob/f58402ec319a198016894b567eec76c6aea17ae2/integration-test/Analysis/NugetSpec.hs#L22) , the issue with `project.assets.json` files is that they are not typically included in open source repos as they are generated automatically by the .NET SDK during the restore process. We also didn't have an existing integration test for the `project.assets.json` strategy either. I feel like we should still have a test for this but am unsure on the best way to go about it. Should I just clone a repo, generate a `project.assets.json` file and push those changes to a new repo so that we can test against for our tests? 

## References

- [ANE-698](https://fossa.atlassian.net/browse/ANE-698)

## Checklist

[ANE-698]: https://fossa.atlassian.net/browse/ANE-698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/Strategy/NuGet.hs
```diff
@@ -0,0 +1,91 @@
...
+  WalkStep (WalkContinue),
+  fileName,
+  findFileNamed,
+  walkWithFilters',
+ )
+import Effect.ReadFS (ReadFS)
+import GHC.Generics (Generic)
+import Path (Abs, Dir, File, Path, parent)
+import Strategy.NuGet.PackageReference qualified as PackageReference
+import Strategy.NuGet.PackageReference qualified as ProjectAssetsJson
```
spatten, line 34: You're importing the same package twice but calling it different things. Shouldn't the second one of these be imported from `Strategy.NuGet.ProjectAssetsJson`?

src/Strategy/NuGet.hs
```diff
@@ -0,0 +1,91 @@
...
+  deriving (Eq, Ord, Show, Generic)
+
+instance ToJSON NuGetProject
+
+instance AnalyzeProject NuGetProject where
+  analyzeProject _ = getDeps
+  analyzeProjectStaticOnly _ = getDeps
+
+getDeps :: (Has ReadFS sig m, Has Diagnostics sig m) => NuGetProject -> m DependencyResults
+getDeps project = context "NuGet" (getAssetsJsonDeps project <||> getPackageReferenceDeps project)
```
spatten, line 85: It feels weird that we're just running `getAssetsJsonDeps` on the file even if it's not named `project.assets.json`. Should we switch on the filename, or add a new field to the `Nuget` type that tells us which type of analysis we should be doing?

spatten (changes requested):
I think we need some tests that test the new behaviour works and actually runs the correct strategy depending on whether a `project.assets.json` file or a `.csproj` file is found during discovery.

We should have a test plan that tests these two cases as well, so we can manually validate that this works.

src/Strategy/NuGet.hs
james-fossa: I've picked this up and am working from the assumption that this was a typo and meant to be `Strategy.NuGet.ProjectAssetsJson`

src/Strategy/NuGet.hs
james-fossa: I changed this to call `getAssetsJsonDeps` only on `project.assets.json`. I think that we should probably rely on the file in question living where it does to avoid some perverse edge case like "this other JSON file I found is a valid `project.assets.json` file but dotnet will not treat it as such. I hold this opinion lightly, will revert if desired.

docs/references/strategies/languages/dotnet/nuget.md
```diff
@@ -1,17 +1,22 @@
-# Quick reference: NuGet
+# NuGet Analysis
 
-## Requirements
+| Strategy                                 | Direct Deps | Transitive Deps | Edges |
```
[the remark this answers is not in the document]
james-fossa: Thanks for the extension link; done and done

docs/features/manual-dependencies.md
```diff
@@ -35,12 +35,10 @@ Note: When parsed, YAML considers text that could be a decimal number (such as 1
 Supported dependency types:
 
 - `bower` - Bower dependencies that are typically found at [bower.io](https://registry.bower.io).
-<!-- markdown-link-check-disable-next-line -->
 - `cargo` - Rust dependencies that are typically found at [crates.io](https://crates.io/).
 - `carthage` - Dependencies as specified by the [Carthage](https://github.com/Carthage/Carthage) package manager.
 - `composer` - Dependencies specified by the PHP package manager [Composer](https://getcomposer.org/), which are located on [Packagist](https://packagist.org/).
 - `cpan` - Dependencies located on the [CPAN package manager](https://www.cpan.org/).
-<!-- markdown-link-check-disable-next-line -->
 - `cran` - Dependencies located on the [CRAN](https://cran.r-project.org/) like repository.
```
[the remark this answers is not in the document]
james-fossa: Not knowing a reason to re-enable, I've added these back

docs/features/manual-dependencies.md
```diff
@@ -54,6 +54,7 @@ Supported dependency types:
 - `pub` - Dart dependencies found at [pub.dev](https://www.pub.dev/).
 - `pypi` - Python dependencies that are typically found at [Pypi.org](https://pypi.org/).
 - `swift` - Swift dependencies using the [Swift Package Manager](https://www.swift.org/package-manager/).
+<!-- markdown-link-check-disable-next-line -->
```
spatten, line 57: Okay, now I'm reversing my previous comment. This comment is not on current master, and I don't think this PR should change that

docs/references/strategies/languages/dotnet/nuget-qr.md
```diff
@@ -0,0 +1,17 @@
+# Quick reference: NuGet
```
spatten, line 1: This file is new in this PR and not referenced anywhere in the documentation. There are other `<language>-qr.md` files in the repo, but I think it's confusing to have both `nuget.md` and `nuget-qr.md`. 

So, I know you didn't add it, but I think we should delete it

integration-test/Analysis/NugetSpec.hs
```diff
@@ -38,7 +55,20 @@ testServiceStackForPkgConfig =
       it "should find targets" $ \(result, _) -> do
         length result `shouldBe` 4
 
+testDotnetCoreTwoExampleForPackageAssetsJson :: Spec
```
spatten, line 58: Should this be named `testDotnetCoreTwoExampleForProjectAssetsJson`?

spatten (approved):
This looks great! I had a few, minor and optional, nits. Nothing blocking approval

PR #1463, opened 2024-08-30
Strict mode

JeffreyHuynh1:
# Overview

We should allow users to enforce that the first strategy within our list of strategies passes. This is to ensure ‘ideal’ results and that no fallback strategies are ran. 

## Acceptance criteria

- When running fossa analyze --strict : 

   - Turn warnings into fatalities

   - Ensure that the first strategy in the list passes

## Testing plan

Manually testing: 
- `fossa analyze --strict --debug` (I scanned a Maven project)
-  In the plugin strategy , emit a fatal error e.g. fatal.MissingDeepDeps
-  See that we do not fallback to the other strategies (deptree, pom, etc) and analysis halts

## Risks

Would like to add some automated tests but wanted to get some opinions. Would it be sufficient to just create a test for `guardStrictMode` bc that's the main logic that powers `strict mode`? Unsure on how to best create tests for strict mode for specific languages / package managers.

## References

- [ANE-236](https://fossa.atlassian.net/browse/ANE-236)

## Checklist

[ANE-236]: https://fossa.atlassian.net/browse/ANE-236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


csasarak (changes requested):
This looks pretty good! I think the main changes I'd ask for here are some clarity around how `--static-only-analysis` and `--strict` interact and a few comments.

csasarak:
One thing I forgot to mention is that I think this should be pretty easy to write tests for. I think that if you go to the tests for each of these strategies there is likely one that gets results from the "ideal" analysis. You can run the analysis using strict mode and just check that the output equals the output of the other test. I'd like it if you could try to make a few of these and see what the effort is like. Let me know if you'd like any help.

carloskcheung:
------------------------
Carlos Cheung - Georgetown Club of Northern California, Board Member
Twitter: @Carloskcheung | 415.515.1063 |
http://www.linkedin.com/in/carloscheung
"Learn to work then you'll always work; Work to learn then you'll always
learn"

integration-test/Analysis/FixtureExpectationUtils.hs
```diff
@@ -91,17 +93,17 @@ withProjectOfType result (projType, projPath) =
...
 
-testSuiteDepResultSummary :: (AnalyzeProject a, Show a, Eq a) => AnalysisTestFixture a -> DiscoveredProjectType -> DependencyResultsSummary -> Spec
-testSuiteDepResultSummary fixture projType depResultSummary =
-  aroundAll (withAnalysisOf fixture) $
+testSuiteDepResultSummary :: (AnalyzeProject a, Show a, Eq a) => Mode -> AnalysisTestFixture a -> DiscoveredProjectType -> DependencyResultsSummary -> Spec
+testSuiteDepResultSummary mode fixture projType depResultSummary =
+  aroundAll (withAnalysisOf mode fixture) $
     describe (toString $ testName fixture) $ do
       it "should find targets" $ \(result, extractedDir) ->
         expectProject (projType, extractedDir) result
```
JeffreyHuynh1, line 109: @csasarak Added some tests like how you suggested. Is this approach somewhat similar to what you had in mind?

csasarak: I was thinking more of something that first ran the test with `NonStrict` and then one that ran it with `Strict` and compared the results. Ideally, they should be the same for these tests because we always are testing the ideal case regardless of the flag. The idea here is that if the `NonStrict` test is always doing ideal analysis, then the `StrictMode` version should return the same results. Basically what I'm trying to think about is "How do we know that setting `Strict` actually does anything?". 

However, on second thought I think maybe the answer is to make `guardStrictMode` just a little bit smarter and test it in isolation in our unit tests:

```haskell
guardStrictMode :: (Has Diagnostics sig m) => Mode -> [m a] -> m a
guardStrictMode _ [] = fatalText "Strict mode: No Possible actions to guard against."
guardStrictMode NonStrict (action1:rest) = foldr (<||>) action1 rest
guardStrictMode Strict (action1:_) = action1 <||> fatalText "Strict mode enabled, skipping other actions"
```
with usage:
```haskell
guardStrictMode mode [tactic1, tactic2, tactic3, ...]
```
Testing then can be something more trivial:

```haskell
do it' "test1" $ do
       res <- recover $ guardStrictMode Strict [fatalText 1, pure 2]
       res `shouldBe'` Nothing
     it' "test2" $ do
       res <- recover $ guardStrictMode NonStrict [fatalText 1, pure 2]
       res `shouldBe'` Just 2
```
I haven't actually run the above spec, but I think it'd be something similar to that. I think this strikes a good balance by having at least some testing around the `Mode` behavior while also not requiring re-running expensive integration tests multiple times.

docs/references/subcommands/analyze.md
```diff
@@ -128,11 +128,12 @@ We support the following archive formats:
...
+| [`--detect-vendored`](./analyze/detect-vendored.md)                               | Enable the vendored source identification engine. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md).                                                                     |
+| [`--detect-dynamic './some-binary`](./analyze/detect-dynamic.md)                  | Analyze the binary at the provided path for dynamically linked dependencies. For more information, see the [C and C++ overview](../strategies/languages/c-cpp/c-cpp.md).                                          |
+| [`--static-only-analysis`](../strategies/README.md#static-and-dynamic-strategies) | Do not use third-party tools when analyzing projects.                                                                                                                                                             |
+| `--strict`                                                                        | Enforces strict analysis to ensure the most accurate results for a strategy by rejecting fallbacks. When ran with `--statitic-only-analysis`, the most optimal static strategy will be applied without fallbacks. |
```
csasarak, line 136:
```suggestion
| `--strict`                                                                        | Enforces strict analysis to ensure the most accurate results by rejecting fallbacks. When run with `--static-only-analysis`, the most optimal static strategy will be applied without fallbacks. |
```

The removal of "accurate results for a strategy" is because you refer to a strategy as if it's a collection of different techniques in the first sentence, but then use it to refer to a component of that collection in the second. I don't feel too strongly about it, but if you think that might be confusing (or I'm just wrong), feel free to leave it as is.

csasarak: Could you also add a paragraph or two about this in the strategies [documentation](https://github.com/fossas/fossa-cli/blob/master/docs/references/strategies/README.md?plain=1#L131)? I think a section `### Strict Analysis` would be sufficient. A good thing to include would be an example of where the decision matters, such as Maven e.g. "For Maven, we will try an analysis using the available `mvn` tool, but if that fails FOSSA will try to statically analyze available pom files. With strict analysis, FOSSA will not attempt the static analysis and report failure if using `mvn` doesn't work."

csasarak (approved):
Looks good. Please address my comments, but I don't think anything there should need me to look this over again unless you want me to.

integration-test/Analysis/FixtureExpectationUtils.hs
csasarak: Hey, I was thinking a little bit more about this and I don't think there's anything more you need to do - I think that as long as you've manually tested this it's OK.

PR #1468, opened 2024-09-10
Upgrade zlib.

csasarak:
# Overview

upgrade zlib.

## Checklist

### DIFF
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -167,7 +167,7 @@ common deps
     , yaml                         ^>=0.11.8
     , yarn-lock                    ^>=0.6.5
     , zip                          ^>=2.0.0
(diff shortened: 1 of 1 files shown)


spatten (approved):
Looks good and safe to me!

Any reason we're seeing the warnings for the use of `head` in this PR even though we haven't changed that code? I don't think I've seen those before

csasarak:
They're new as of the GHC upgrade. GHC itself added some warnings for usages of those functions (in addition to hlint) with the reasoning that the compiler discouraging usage of partial functions is easier for new users of GHC. I write a little bit about it in #1460. I had trouble configuring it so that the warning would be suppressed entirely short of putting pragmas in every place they're used.

JeffreyHuynh1 (approved):

PR #1469, opened 2024-09-10
Convert MSB errors to warnings

jssblck:
# Overview

When a user has configured a project to be linked to a binary via MSB, and then they lose access to the project (either because a different user is running the scan, or the project has been deleted, or some other reason) the scan currently fails:

```shell
Failed to resolve dependencies for the following FOSSA projects:
	custom+19518/some-project$some-revision

You may not have access to the projects, or they may not exist (see the warnings below for details).
If desired you can use --experimental-skip-vsi-graph to skip resolving the dependencies of these projects.
```

The user then has to use `--experimental-skip-vsi-graph` to resolve this issue.
This PR alters the error to be a warning instead of a fatal error.

The `--experimental-skip-vsi-graph` flag is still there so that users using it don't encounter problems, but it is now largely irrelevant (it would only suppress this warning).

## Acceptance criteria

This error is now a warning.

## Testing plan

This is proven in the type system, I don't believe it requires testing.
No tests currently exist for this functionality, so I didn't alter any existing tests.

## Risks

No risk.

## Metrics

None

## References

https://fossa.atlassian.net/browse/ANE-2005

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,10 @@
 # FOSSA CLI Changelog
 
+
+## 3.9.33
+
+- MSB: Failure to resolve a linked project no longer causes the scan to fail ([[#1469](https://github.com/fossas/fossa-cli/pull/1469)]).
+
 ## 3.9.32
 
 - Platform Support: Add a binary for ARM64 Linux environments. ([#1465](https://github.com/fossas/fossa-cli/pull/1465))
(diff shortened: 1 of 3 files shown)


spatten (approved):
This looks good. There's a failing test that needs to be fixed or removed, but the code itself is good to ship!

PR #1470, opened 2024-09-11
Update grafana image hash.

csasarak:
# Overview

One of our unit tests broke because we were expecting the wrong digest hash. I'm really not sure what happened here since no changes we made seem to have touched this. I first noticed this with #1468 to master - it succeeded in the PR but failed when merged to master.

When I `docker pull grafana/grafana:8.1.7-ubuntu` I get an `amd64` image which has the same hash as we were originally expecting. This is despite the fact that the test requests "arm". When I explicitly request arm with `docker pull --platform=arm grafana/grafana:8.1.7-ubuntu` I get an image downloaded with the correct manifest.

Testing with the `redis` image shows the same thing.

This implies to me that in the past we requested the arm image in our unit test but the remote server was sending the amd64 image instead and now for some reason the remote is sending the actual arm image. The difference in digest hashes are what broke our tests.

In any case, this should allow tests to pass on master when it merges.

## Acceptance criteria

* Tests and build succeed.

## Testing plan

I looked at the results of CI.

## Checklist

### DIFF
--- a/src/Control/Carrier/ContainerRegistryApi.hs
+++ b/src/Control/Carrier/ContainerRegistryApi.hs
@@ -105,7 +105,6 @@ import Network.HTTP.Conduit qualified as HTTPConduit
 import Network.HTTP.Types.Header (ResponseHeaders)
 import Path (Abs, Dir, File, Path, filename, mkRelFile, toFilePath, (</>))
 import Path.Internal (Path (..))
-import Debug.Trace (traceShowM)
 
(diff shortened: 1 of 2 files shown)


csasarak:
I think that maybe "arm" wasn't a recognized platform so docker may have just downloaded an image compatible with the platform it was running on. But now "arm" might be a valid value for platform which broke the tests?

PR #1475, opened 2024-10-10
Make "invalid project permission" error message clearer

ryanlink:
This adds detail on how to fix "invalid project permission" error.

# Overview

We get a lot of support tickets from team admins/editors getting "invalid project permission" who don't understand that they need to specify a team with `fossa analyze`. This makes resolving the error clearer to the end user to reduce support ticket volume.

## Acceptance criteria

This adds explanation to the error message, instructing the user to add `--team 'team-name'` to the command or specify a team in the `.fossa.yml`.

## Testing plan

I have not tested this as I haven't yet built fossa-cli locally. A reviewer could build this locally and test that the error message looks OK when trying to analyze a project as a team admin without specifying a team name. Otherwise, it's a straightforward text change.

## Risks

I haven't tested that the longer error message looks OK in the command output. 

## Checklist

### DIFF
--- a/src/App/Fossa/PreflightChecks.hs
+++ b/src/App/Fossa/PreflightChecks.hs
@@ -160,7 +160,7 @@ instance ToDiagnostic ProjectPermissionErr where
   renderDiagnostic CreateTeamProjectPermissionErr =
     Errata (Just projectPermissionErrHeader) [] $ Just "You do not have permission to create projects for the specified team"
   renderDiagnostic CreateProjectOnlyToTeamPermissionErr =
-    Errata (Just projectPermissionErrHeader) [] $ Just "You only have permission to create projects for your team"
+    Errata (Just projectPermissionErrHeader) [] $ Just "You only have permission to create projects for your team. Please specify a team in .fossa.yml or using the --team flag"
 
 releaseGroupPermissionErrHeader :: Text
 releaseGroupPermissionErrHeader = "Invalid release group permission"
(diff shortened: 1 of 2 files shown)


src/App/Fossa/PreflightChecks.hs
```diff
@@ -160,7 +160,7 @@ instance ToDiagnostic ProjectPermissionErr where
   renderDiagnostic CreateTeamProjectPermissionErr =
     Errata (Just projectPermissionErrHeader) [] $ Just "You do not have permission to create projects for the specified team"
   renderDiagnostic CreateProjectOnlyToTeamPermissionErr =
-    Errata (Just projectPermissionErrHeader) [] $ Just "You only have permission to create projects for your team"
+    Errata (Just projectPermissionErrHeader) [] $ Just "You only have permission to create projects for your team. Please specify a team in .fossa.yml or using the --team flag"
```
csasarak, line 163: I actually think this is more appropriate as part of the "help" text for this error: https://github.com/fossas/fossa-cli/blob/b4b091a89df5b9b6cf1d9f038fd28df855a8cdc8/src/App/Fossa/PreflightChecks.hs#L115

Can you incorporate its content there?

csasarak (changes requested):
This seems OK to me, but I think the message belongs in the "Help" section of the error block. It's located [here](https://github.com/fossas/fossa-cli/blob/b4b091a89df5b9b6cf1d9f038fd28df855a8cdc8/src/App/Fossa/PreflightChecks.hs#L115).

src/App/Fossa/PreflightChecks.hs
ryanlink: Good call-out. Done! I also added periods to the Errata msgs, and `(s)` to `team` since you can be an admin/editor of multiple teams.

csasarak (approved):

PR #1481, opened 2024-11-14
Fix Container Registry specs

spatten:
# Overview

The hashes on two docker images that we use in our tests changed, breaking the test expectations.

This PR just updates the hashes to the new ones.

Here's a link to a [failing test run](https://github.com/fossas/fossa-cli/actions/runs/11826754782/job/33003921836?pr=1480).

## Acceptance criteria

The tests pass

## Risks

I think it's safe to just change the hashes, as this is just a test expectation. No code changes were involved: it fails on current master and it passed yesterday, so I'm assuming that this is a change in the images that we're testing against. I did not do any investigation as to why the images changed.

## Checklist

### DIFF
--- a/test/Container/RegistryApiSpec.hs
+++ b/test/Container/RegistryApiSpec.hs
@@ -150,10 +150,9 @@ dhImageWithDigest =
 grafanaMultiArchImage :: Text
 grafanaMultiArchImage = "grafana/grafana:8.1.7-ubuntu"
 
--- TODO
 grafanaMultiArchImageDigest :: RepoDigest
 grafanaMultiArchImageDigest =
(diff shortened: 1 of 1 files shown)


csasarak (approved):
I think this is fine for now to get things moving, but these hashes seem to match up with the hashes for the arm64 versions of these images. You can see it with `docker pull --platform linux/arm64 <image>` and looking in the docker desktop UI. 

I'm not sure why it's selecting those, so maybe there's a bug? I couldn't see in the log a clear indication of what architecture the failing build was running on.

csasarak:
Huh, for some reason the arm test was expecting the x86 sha before this change. If anything, this breakage indicates that something is working better now for some reason.

PR #1482, opened 2024-11-20
Clean up FOSSA dependency scan action

jssblck:
# Overview

Cleans up the dependency scan job to make it read a little more nicely.


spatten (approved):
Looks good to me!

PR #1485, opened 2024-11-27
Update strategies doc

jssblck:
# Overview

It was brought to our attention that the docs for nuget strategies were incorrect: they said the nuget strategies are dynamic but they're actually static.

This PR updates that listing and overhauls the table in an attempt to make the dynamic vs static relationship more clear. Note: I did not double check every strategy; I just went off what we currently listed in that doc.

[Rendered preview here.](https://github.com/fossas/fossa-cli/tree/docs/nuget/docs/references/strategies#strategies-by-type)

Screenshot 2024-11-27 at 1 02 00 PM

### DIFF
--- a/docs/references/strategies/README.md
+++ b/docs/references/strategies/README.md
@@ -145,40 +145,49 @@ Invoke strict analysis with the `--strict` flag when running `fossa analyze`.
 
 ### Strategies by type
 
-> If the FOSSA CLI is forced to utilize a fallback strategy, meaning it did not detect ideal results, a warning is emitted in the scan summary after running `fossa analyze`.
+> [!NOTE]
+> Dynamic strategies require a working build environment for analysis.
+>
+> If a given package manager has a dynamic strategy with a static fallback, that means the static fallback provides worse results,
+> so it is only used when the dynamic strategy fails. If the package manager only has static strategies, that means dynamic analysis
+> is not required for ideal results.
 
-| Language/Package Manager                                                                                                                        | Dynamic   | Static    | Detect Vendored Code | Primary Strategy |
-|-------------------------------------------------------------------------------------------------------------------------------------------------|-----------|-----------|----------------------|------------------|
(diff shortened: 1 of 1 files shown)


docs/references/strategies/README.md
```diff
@@ -145,40 +145,49 @@ Invoke strict analysis with the `--strict` flag when running `fossa analyze`.
...
+> [!TIP]
+> If FOSSA CLI is forced to utilize a fallback strategy, meaning it did not detect ideal results,
+> a warning is emitted in the scan summary after running `fossa analyze`.
+
+> [!WARNING]
+> "Custom" strategies work very differently than the standard package manager based analysis; read their docs for more details.
+
+| Language/Package Manager                                                                                                                        | Kind of analysis             | Detect Vendored Code |
+|-------------------------------------------------------------------------------------------------------------------------------------------------|------------------------------|----------------------|
+| [C#](https://github.com/fossas/fossa-cli/tree/master/docs/references/strategies/languages/dotnet)                                               | Static                       | ❌                    |
```
csasarak, line 164: [Question] Is Nuget always what's used for C#? I don't need this resolved for approval but it feels like the package manager or NuGet is the right thing to put here vs. the language.

csasarak (approved):

docs/references/strategies/README.md
jssblck: it's a little strange. first: no, nuget isn't the only package manager for C# or .NET generally. But inside FOSSA CLI we group all the tactics for .NET under the `Nuget` module, e.g. `Strategy.NuGet.Paket`.

i do think it probably makes more sense to refer to the package manager in this table like we do for other languages.

jssblck: I've gone ahead and updated these to match the other package managers in the table: C# was the only odd one out.

csasarak: I think that looks good, thanks!

PR #1486, opened 2024-12-03
Expose a static only Go modules tactic.

csasarak:
# Overview

We initially didn't expose a static only analysis method in the CLI for Go modules despite there being a pretty low-effort path to providing one. This PR makes a truly static Go analysis method and exposes it for use with the `--static-only-analysis` flag. 

## Acceptance criteria

It is possible to analyze go projects statically.

## Testing plan

Compare running `fossa analyze --static-only-analysis` on this branch and in a release version. Static Go analysis fails using the current release version:

Screenshot 2024-12-02 at 6 16 30 PM
But succeeds in the one on this branch:

Screenshot 2024-12-02 at 6 34 52 PM

## Risks

The main risk is that projects which didn't produce a result in the past may start to. This is the correct thing to do, but may result in more questions.

## References

[slack thread](https://teamfossa.slack.com/archives/C0155DTGWB1/p1732903477682909)

## Checklist

### DIFF
--- a/src/Strategy/Go/Gomod.hs
+++ b/src/Strategy/Go/Gomod.hs
@@ -14,6 +14,7 @@ module Strategy.Go.Gomod (
   PackageVersion (..),
   parsePackageVersion,
   gomodParser,
+  analyzeStatic,
 ) where
 
 import Control.Algebra (Has)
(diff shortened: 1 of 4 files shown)


src/Strategy/Go/Gomod.hs
```diff
@@ -396,6 +397,14 @@ analyze' file = do
     pure ()
   pure (graph, Partial)
 
+-- | This variant of analyze will not attempt to fill in transitive dependencies.
+analyzeStatic :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m (Graphing Dependency, GraphBreadth)
```
jssblck, line 401: If this isn't in the user facing docs already please add it!

jssblck (approved):

src/Strategy/Go/Gomod.hs
csasarak: Good idea, I've added it.

PR #1491, opened 2025-01-08
`diagnose walk` interactive mode

jssblck:
# Overview

Adds an interactive mode to `diagnose walk`. Run `diagnose walk --mode tui` to have it output the stats of the walk in a tui instead of as log messages.

I made the changes here while troubleshooting a ticket involving symlink loops.


tools/diagnose/src/cmd/walk.rs
```diff
@@ -43,9 +55,187 @@ pub enum SymlinkStrategy {
...
+                    warnings.push_front(msg);
+                    warnings_count += 1;
+                }
+
+                continue;
+            }
+        };
+
+        if filters.is_match(&path) {
+            skipped.push_front(path.to_string());
```
csasarak, line 131: [nit] Isn't `path` already a string here?

jssblck: no, it's a `Cow<'_, str>`

csasarak (approved):
LGTM! It was neat to learn about `ratatui` too.

PR #1493, opened 2025-01-13
[ANE-2123] A fix around listing excluded dirs

james-fossa:
# Overview
As it stands, the `paths/excludes` aren't quite universally applied.
Moreover, in the case that the calling user does not have permissions to list the contents of an excluded directory, that directory can inadvertently end up raising an exception by calling `ls` during the discovery phase.

This PR uses `walkWithFilters'` more heavily, as it takes into account `paths/excludes` from the `.fossa.yml` file.

It also changes `walkWithFilters'` to avoid inadvertently calling `listDir` on an excluded (and potentially forbidden) directory.

## Acceptance criteria

When users run `fossa analyze --output` on a directory which contains:
- a directory for which the user does not have read permissions
- a .fossa.yml file excluding that directory

Then the run should not output permission denied errors.

## Testing plan

Set up a directory structure like:

```
├── .fossa.yml
├── foo
│   ├── bar  [error opening dir]
│   ├── baz
│   │   └── bork
│   ├── bortz
└── package.json
```

by running something like:

```sh
mkdir ane-2123
mkdir ane-2123/foo
mkdir ane-2123/foo/bar
chmod a-r ane-2123/foo/bar
touch ane-2123/package.json
```

and placing a .fossa.yml file like:
```
version: 3

targets:
  include:
    - npm

paths:
  exclude:
    - foo/bar
```

Then, run `cabal run fossa -- analyze -e http://localhost:9578 --output --debug ../ane-2123`.

You should see logs from the various discovery functions, but no more permission denied errors.

## Risks

1. I haven't manually tested the changes to swift or maven projects yet (I wanted to get a sense of if this was plausible before so doing at least.) This change has an effect on whether excluded dirs are reflected in swift or maven projects.
2. This changes the semantics of excluded dirs from "don't call the handler on this dir" to "don't even look at this." I think that's a sensible definition for an excluded dir, but there could be cases where people expect the former behavior.

## Metrics

## References

_Example:_

[ANE-2123](https://fossa.atlassian.net/browse/ANE-2123)

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,7 @@
 # FOSSA CLI Changelog
 
 ## 3.9.43
-- Fix a bug where `paths.exclude` may be `ls`'d as part of walking the file system, which raises errors if the user does not have permissions for the excluded path.
+- Discovery: Fix a bug where directories in paths.exclude may still be accessed during discovery which causes an error when users don't have permission to read those directories.
 
 ## 3.9.42
 - Licensing: Adds support for the Text-Tabs+Wrap License
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## 3.9.43
+- Fix a bug where `paths.exclude` may be `ls`'d as part of walking the file system, which raises errors if the user does not have permissions for the excluded path.
+
 ## 3.9.42
 - Licensing: Adds support for the Text-Tabs+Wrap License
 
--- a/src/Discovery/Walk.hs
+++ b/src/Discovery/Walk.hs
@@ -98,7 +98,7 @@ pathFilterIntercept filters base dir subdirs act = do
     -- skipDisallowed needs to look at either:
     --  * WalkStep.WalkContinue
     --  * WalkStep.WalkSkipSome [Text]
-    -- and add on any missing disalloewd subdirs
+    -- and add on any missing disallowed subdirs
     skipDisallowed :: (Applicative m) => m (o, WalkStep) -> m (o, WalkStep)
     skipDisallowed =
       fmap $ \ (o, action) ->
--- a/src/Discovery/Walk.hs
+++ b/src/Discovery/Walk.hs
@@ -14,10 +14,11 @@ module Discovery.Walk (
 ) where
 
 import Control.Carrier.Writer.Church
-import Control.Effect.Diagnostics ( Diagnostics, fatal, context )
+import Control.Effect.Diagnostics (Diagnostics, context, fatal)
 import Control.Effect.Reader (Reader, ask)
 import Control.Monad.Trans
 import Control.Monad.Trans.Maybe
+import Data.Bifunctor (second)
 import Data.Foldable (find)
 import Data.Functor (void)
 import Data.Glob qualified as Glob
@@ -26,10 +27,9 @@ import Data.Maybe (mapMaybe)
 import Data.Set qualified as Set
 import Data.String.Conversion (toString, toText)
 import Data.Text (Text)
-import Discovery.Filters (pathAllowed, AllFilters)
+import Discovery.Filters (AllFilters, pathAllowed)
 import Effect.ReadFS
 import Path
-import Data.Bifunctor (second)
 
 data WalkStep
   = -- | Continue walking subdirectories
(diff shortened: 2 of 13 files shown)


src/Discovery/Walk.hs
```diff
@@ -73,17 +73,46 @@ pathFilterIntercept ::
...
+    disallowedSubdirs :: [Path Abs Dir]
+    disallowedSubdirs = do
+      subdir <- subdirs
+      stripped <- stripProperPrefix base subdir
+      let isAllowed = pathAllowed filters stripped
+      if isAllowed
+        then mempty
+        else [subdir]
+
+    -- skipDisallowed needs to look at either:
```
james-fossa, line 98: This code is "smelly" to me, but I don't know quite what to do about it. If there's a haskell idiom that would seem to apply, I probably just don't know what it is, so enlighten me!

test/Discovery/WalkSpec.hs
```diff
@@ -108,6 +153,31 @@ runWalk ::
   (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> m [Path Abs Dir]
 runWalk = runWalkWithCircuitBreaker 100
 
+runWalkWithFilters' ::
+  ( Has ReadFS sig m
+  , Has Diagnostics sig m
+  ) =>
+  Int -> AllFilters -> Path Abs Dir -> m [Path Abs Dir]
+runWalkWithFilters' maxIters filters startDir =
```
csasarak, line 161: Can you explain the reasoning behind `maxIters`? In a test shouldn't we know if it finishes or not?

csasarak (approved):
Looks good, mostly style comments and a few questions.

src/Discovery/Walk.hs
```diff
@@ -73,17 +73,46 @@ pathFilterIntercept ::
...
+    --  * WalkStep.WalkContinue
+    --  * WalkStep.WalkSkipSome [Text]
+    -- and add on any missing disalloewd subdirs
+    skipDisallowed :: (Applicative m) => m (o, WalkStep) -> m (o, WalkStep)
+    skipDisallowed =
+      fmap $ \ (o, action) ->
+        let skipNames = map (toText . toFilePath . dirname) disallowedSubdirs
```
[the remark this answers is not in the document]
csasarak: One other option: you never actually use the value `disallowedSubdirs` AFAICT except as an input to `skipNames`. You could just put the text conversion into the definition for `disallowedSubdirs`.

src/Discovery/Walk.hs
csasarak: I guess this function doesn't actually depend on `o` so you could factor that out to get `WalkStep -> WalkStep`. Then to run it you would be doing an `fmap . fmap`, the first `fmap` is to push the function into `m` and the second is to manipulate the second argument of the pair. The Functor instance for `(a, b)` is `(b -> c) -> (a, b) -> (a, c)`. But honestly, I think that's overkill.

test/Discovery/WalkSpec.hs
james-fossa: Sure- there felt like a tradeoff between how it works in `runWalk` (default to `maxIters = 100`) and `runWalkWithCircuitBreaker` (passed in as an argument), and since `runWalkWithFilters'` is a special case of `runWalk`, I opted for including the extra lever in case we want it for something else later.

src/Discovery/Walk.hs
```diff
@@ -73,17 +73,46 @@ pathFilterIntercept ::
...
+        in if not (null disallowedSubdirs)
```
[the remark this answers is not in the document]
james-fossa: I see one on Hoogle in a `Data.List.Extra` from hackage but not in our version.
Went ahead and duplicated it for use here, to save time in the future if we think it exists again.
Also took the liberty of adding a `Data.Tuple.Extra` in a similar vein, right now just including `mapSnd`.
Would've added a dependency on `Extra` but there's an HLint warning against using a definition of `second`/`mapSnd` from `Arrow` and the `Extra` package used definition in its implementation.

src/Discovery/Walk.hs
csasarak: I forgot to mention, nothing about your original code seemed smelly to me. My comments were geared at "have you thought about doing it this way?" rather than "wow, that really sucks how about this?"

src/Discovery/Walk.hs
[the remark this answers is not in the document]
james-fossa: Thanks, I reverted that addition once I saw your comment [here](https://github.com/fossas/fossa-cli/pull/1493#discussion_r1913713675) (figured that "even more overkill" was probably not the way to go)
I'll go ahead and reverse the arms of the `if` for the legibility you mention; I agree, it does require less scanning to see both alternatives that way.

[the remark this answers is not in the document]
csasarak: Oh, and one more thing about imports. Given the choice, I've come to think we should qualify by default rather than adding to import lists which is a bit of a departure from how we do things. I mainly think this because it localizes information about where something comes from and also can help tooling e.g. You start typing `Map.` and then the editor can give you options. Starting with nothing and trying to have the editor complete isn't very helpful.

Changelog.md
```diff
@@ -1,5 +1,8 @@
 # FOSSA CLI Changelog
 
+## 3.9.43
+- Fix a bug where `paths.exclude` may be `ls`'d as part of walking the file system, which raises errors if the user does not have permissions for the excluded path.
```
csasarak, line 4: Usually we do things in a format like:

> Discovery: Fix a bug where `paths.exclude` may be `ls`'d as part of walking the file system, which raises errors if the user does not have permissions for the excluded path.

I would also recommend not referencing `ls`. We have Windows users or less technical users who sometimes use the CLI and that may not make sense to them. 

Consider:

>  Discovery: Fix a bug where directories in `paths.exclude` may still be accessed during discovery which causes an error when users don't have permission to read those directories.

james-fossa: Thanks for the context/suggestion, will do

PR #1494, opened 2025-01-13
Add caveat about pnpm v9 lockfile incompatibility

ryanlink:
# Overview

Updating docs to disclaim pnpm v9 incompatibility.

## Acceptance criteria

Docs updated

## Testing plan

n/a

## Risks

n/a

## Metrics

n/a

## References

https://fossa.atlassian.net/browse/ANE-2177

## Checklist

### DIFF
--- a/docs/references/strategies/languages/nodejs/pnpm.md
+++ b/docs/references/strategies/languages/nodejs/pnpm.md
@@ -161,10 +161,17 @@ CLI will infer the package name and version using `/${dependencyName}/${dependen
 You can explicitly specify an analysis target in `.fossa.yml` file. The example below will exclude all analysis targets except for pnpm.
 
 ```yaml
-# .fossa.yml 
+# .fossa.yml
 
 version: 3
 targets:
   only:
     - type: pnpm
 ```
+### Are all versions of `pnpm` supported?
+
+At this time, the latest version of pnpm (v9) and its associated v9 lockfiles are not correctly parsed by FOSSA. Please revert to v8 (v6 lockfile) if your dependencies are not resolved in the FOSSA UI: "FOSSA was unable to analyze this dependency. If it is behind a private registry or auth you may need to configure FOSSA's access, then rebuild this dependency." This is due to the version number being appended to the package name:
+
+<img width="796" alt="image" src="https://github.com/user-attachments/assets/d1461506-d3e7-42da-b9be-2b53a87f79f1" />
+
+We have [requested](https://github.com/pnpm/spec/issues/6#issuecomment-2588100182) more details on the pnpm v9 lockfile spec and hope to be able to prioritize this improvement soon.


docs/references/strategies/languages/nodejs/pnpm.md
```diff
@@ -161,10 +161,17 @@ CLI will infer the package name and version using `/${dependencyName}/${dependen
...
   only:
     - type: pnpm
 ```
+### Are all versions of `pnpm` supported?
+
+At this time, the latest version of pnpm (v9) and its associated v9 lockfiles are not correctly parsed by FOSSA. Please revert to v8 (v6 lockfile) if your dependencies are not resolved in the FOSSA UI: "FOSSA was unable to analyze this dependency. If it is behind a private registry or auth you may need to configure FOSSA's access, then rebuild this dependency." This is due to the version number being appended to the package name:
+
+<img width="796" alt="image" src="https://github.com/user-attachments/assets/d1461506-d3e7-42da-b9be-2b53a87f79f1" />
+
+We have [requested](https://github.com/pnpm/spec/issues/6#issuecomment-2588100182) more details on the pnpm v9 lockfile spec and hope to be able to prioritize this improvement soon.
```
csasarak, line 177: This is unnecessary. I would prefer customers actually ask us to do this in order to help prioritize it rather than letting them think "Well, it says it's coming soon - I'll say nothing." which stifles that signal.

My comments about the spec weren't meant to say that we need more details. It was only to point out _why_ the spec may be a bit sparse: the format of `pnpm-lock.yaml` likely isn't meant for consumption by anything but `pnmp`. In general, we want the CLI to focus on supporting what's out in the wild rather than just the spec anyhow.

csasarak (changes requested):

docs/references/strategies/languages/nodejs/pnpm.md
ryanlink: Makes sense, I can remove that. I'll explicitly call out that they should email us if this is affecting them :)

I thought the spec was actually more sparse than the [previous spec](https://github.com/pnpm/spec/blob/master/lockfile/6.0.md) (v6), that's why I put it that way. I could be wrong!

csasarak: Yeah, I wasn't quite sure how to read their specs. Like maybe they each only describe differences between the spec and the previous spec? In any case, the dev deps thing is what I really want to prioritize fixing - if we can explain the additional with `catalog` as a version then it's unfortunate but still usable. Obviously I'd like to fix both things though.

ryanlink:
Could somebody help me figure out why the integration test is failing? https://github.com/fossas/fossa-cli/actions/runs/12756509884

csasarak:
@ryanlink That should be fixed once this [PR](https://github.com/fossas/fossa-cli/actions/runs/12775713201) merges to master and then you merge `master` into this branch.

PR #1496, opened 2025-01-14
[NOTICKET] Ignore the cowboy erlang spec to unblock other PRs

james-fossa:
# Overview

[This erlang integration test](https://github.com/fossas/fossa-cli/blob/master/integration-test/Analysis/ErlangSpec.hs#L17-L27) has started failing consistently in CI.

Running it locally it seems fine, regardless of being run as a single test or in the suite / with various random seeds that fail in CI. In addition, re-running existing tags in CI replicates that failure.

I'm out of interesting ideas on how to fix it, and as it stands it's an "always negative" result so I don't think it's adding value beyond a todo-item to be fixed later.

This PR just removes its use from the erlang spec for the moment so we can clear the decks and come back to it later.

## Acceptance criteria

The build passes, the team is agreed that this is good enough for the moment.

## Testing plan

_Example:_

1. _Set up test case X._

## Risks

Ignoring a spec potentially means ignoring a bug.

## Metrics

N/A
## References

[ANE-2236](https://fossa.atlassian.net/browse/ANE-2236) to track the work of fixing the spec when time allows.

## Checklist

[ANE-2236]: https://fossa.atlassian.net/browse/ANE-2236?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-test/Analysis/ErlangSpec.hs
+++ b/integration-test/Analysis/ErlangSpec.hs
@@ -1,6 +1,5 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE QuasiQuotes #-}
-{-# OPTIONS_GHC -Wno-unused-top-binds #-}
 
 module Analysis.ErlangSpec (spec) where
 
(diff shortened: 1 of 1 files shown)


integration-test/Analysis/ErlangSpec.hs
```diff
@@ -40,5 +41,4 @@ emqx =
 
 spec :: Spec
 spec = do
-  testSuiteDepResultSummary NonStrict cowboy Rebar3ProjectType (DependencyResultsSummary 2 2 0 1 Complete)
```
csasarak, line 43: [nit, probably just my personal preference] Usually I don't like dead code, but if we're planning to come back to this I'd prefer having a comment near the fixture with a note about the failure and a reference to the ticket number and possibly just commenting out this code? 

Without a note in the code, a future developer would have to stumble across the ticket or remember to know that this file isn't quite where we want it to be. Additionally, it makes the fixture with no references kind of mysterious.

csasarak (approved):

integration-test/Analysis/ErlangSpec.hs
csasarak: It may also be worth just commenting out this `testSuite...` line so that we know what we used to expect the results to be, but I'll leave that up to you.

james-fossa: I think it probably makes sense to just reproduce the old fixture and `testSuite...` line in the body of the ticket. That way it'll be clean here (no dead code, no mystery) and the ticket will be self-contained for a future reader without linking out.

PR #1498, opened 2025-01-16
[ANE-2237] make the preflight check filename unique

spatten:
# Overview

In [ANE-2237](https://fossa.atlassian.net/browse/ANE-2237), Lukas from SEW Eurodrive reported that the preflight check was failing around 10% of the time on their CI instance.

This was happening because they were running multiple runs of `fossa` in parallel, so the processes were sometimes deleting the preflight check file created by another instance of `fossa`, and it was raising an error when it tried to delete the nonexistent file.

Two obvious ways to fix this are to just not fail if you can't delete or to make the filename unique. I opted for the unique filename approach, but I have no strong opinion here.

## Testing plan

I was able to repro the initial problem by running this in one terminal:

```
while true; do rm -f $TMPDIR/preflight-check.txt; done
```

and then running `fossa report` a bunch of times in another:

```
repeat 20 { fossa report --format spdx-json --project zd-5344 --revision 2022-12-21T01:21:22Z attribution > /dev/null }
```

That's no longer going to reproduce the problem. So I just put in some logging and validated that the filename being returned by preflightCheckFilename was different between runs.

## Risks

This is low risk

## References

[ANE-2237](https://fossa.atlassian.net/browse/ANE-2237)

## Checklist

[ANE-2237]: https://fossa.atlassian.net/browse/ANE-2237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[ANE-2237]: https://fossa.atlassian.net/browse/ANE-2237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,7 +1,10 @@
 # FOSSA CLI Changelog
 
+## 3.9.44
+- Preflight: Fix a bug where the preflight check could fail if you ran fossa multiple times simultaneously ([#1498](https://github.com/fossas/fossa-cli/pull/1498))
+
 ## 3.9.43
-- Discovery: Fix a bug where directories in paths.exclude may still be accessed during discovery which causes an error when users don't have permission to read those directories.
+- Discovery: Fix a bug where directories in paths.exclude may still be accessed during discovery which causes an error when users don't have permission to read those directories ([#1493](https://github.com/fossas/fossa-cli/pull/1493))
 
 ## 3.9.42
 - Licensing: Adds support for the Text-Tabs+Wrap License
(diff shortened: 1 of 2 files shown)


spatten:
> This is a nice find and fix!

I can't really take credit for the find. Lukas from SEW Eurodrive was the one who pointed out that it was because they are running multiple scans on the same machine. 

I'm just glad he did. This could have taken a while to figure out otherwise!

PR #1505, opened 2025-02-13
Support labels in `fossa-deps` files

jssblck:
# Overview

Support user-provided labels for dependencies in `fossa-deps` files.

## Acceptance criteria

- Users are able to provide labels to dependencies inside `fossa-deps`.
- Labels are collected at the root of the uploaded build under a new `Labels` key, which is a map of locator strings to an array of labels for that locator.

## Testing plan

Automated tests, mainly- but also manually tested with the below:
```yaml
referenced-dependencies:
- name: lodash
  type: npm
  labels:
  - label: npm-rev
    scope: revision
  - label: npm-proj
    scope: project
  - label: npm-org
    scope: org

```

```shell
cabal run fossa -- analyze ~/scratch/labels -e https://staging.int.fossa.io
```

image

## Risks

No real risk

## Metrics

None

## References

[RFC](https://go/user-dep-labels-fossa-cli)

## Checklist

### DIFF
--- a/docs/references/files/fossa-deps.md
+++ b/docs/references/files/fossa-deps.md
@@ -95,8 +95,6 @@ For more details, please refer to the [feature](../../features/vendored-dependen
 
 Each kind of dependency referenced above can have a `labels` field, which is a list of labels to be added to the dependency.
 These labels are **user-defined**; you may choose any labels. What they mean is up to you and/or your organization.
-For more information on labels, refer to the main [FOSSA documentation](https://docs.fossa.com).
-<!-- Note: We should really link to a more specific page here, but these docs don't yet exist. -->
 
 Labels have a `scope` field, which is the scope of the label. The possible scopes are:
 - `org`: The label is scoped to the organization.
--- a/docs/references/files/fossa-deps.md
+++ b/docs/references/files/fossa-deps.md
@@ -15,7 +15,7 @@ Denotes listing of dependencies, which are to be analyzed in conjunction with th
 - `type`: Type of dependency. (Required)
 - `name`: Name of the dependency. It should be the same name as listed in dependencies registry. (Required)
 - `version`: Revision of the dependency. If left unspecified, the latest version discovered from the registry will be used.
-
+- `labels`: An optional list of labels to be added to the dependency.
 ```yaml
 referenced-dependencies:
 - type: gem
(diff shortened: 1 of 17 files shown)


src/App/Fossa/ManualDeps.hs
```diff
@@ -187,12 +203,12 @@ toSourceUnit root depsFile manualDeps@ManualDependencies{..} maybeApiOpts vendor
   -- validated without endpoint interactions.
   rdeps <- case maybeApiOpts of
```
zlav, line 204: I can't get this to let me give a suggestion for the following 5 lines, but I think this is cleaner
```
  rdeps <- case org of
    Just orgValid ->  
      traverse (`validateRemoteDep` orgValid) remoteDependencies
    Nothing -> pure remoteDependencies
```
Or you could do `maybe

jssblck: 🤔 yeah that's fair!

src/App/Fossa/ManualDeps.hs
```diff
@@ -384,9 +454,20 @@ data RemoteDependency = RemoteDependency
   , remoteVersion :: Text
   , remoteUrl :: Text
   , remoteMetadata :: Maybe DependencyMetadata
+  , remoteLabels :: Maybe [ProvidedPackageLabel]
```
zlav, line 457: I think we can remove the `Maybe []` here (and in vendoredDeps) in favor of `[]` to clean things up. I could be missing something, but I don't see a convincing reason to keep this.

We should be able to remove`liftMaybe` and change the return values `Maybe (Text, [ProvidedPackageLabel])` around line 260 to `Maybe`. I think this ends up looking more accurate.

jssblck: I generally agree that `Maybe [a]` is better written as just `[a]`, and `[]` is functionally the empty case. However, the reason I used `liftMaybe` and friends here is because:
1. I think pattern matching with `Maybe` is better than `if mempty` checks (it's more natural to the language)
2. Normally the answer would be "don't even check `if mempty`" but I don't want to create spurious empty arrays since the map could get relatively large and I don't want to duplicate locators for no reason

But, in thinking about this and typing this out, I realize that I've left open the possibility that users actually provide an empty array here, which would lead to exactly the sort of duplication I want to avoid. So if I want to avoid it, I need to check for an empty array anyway; if I am doing that I may as well collapse `Maybe` into the array itself as you suggest.

So I started out disagreeing for this particular case, but now agree. Will do, thanks for the nudge 😆

zlav: Thats a good point about user empty arrays, I hadn't even thought about empty user arrays.

src/Fossa/API/Types.hs
```diff
@@ -478,6 +478,9 @@ instance Pretty Issues where
 newtype OrgId = OrgId Int
   deriving (Eq, Ord, FromJSON, ToJSON)
 
+instance ToText OrgId where
+  toText (OrgId orgId) = toText orgId
+
```
csasarak, line 483: [nit] Use [DerivingVia](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/deriving_via.html)


```suggestion
deriving ToText via Int
```

test/App/Fossa/ManualDepsSpec.hs
```diff
@@ -77,6 +154,28 @@ spec = do
...
+          let orgId = OrgId 1234
+          let extracted = collectInteriorLabels (Just orgId) yamlDeps
+          extracted `shouldBe` theWorksLabels (Just orgId)
+
+    it "should extract nested labels without orgId" $
+      case Yaml.decodeEither' theWorksLabeledBS of
+        Left err -> expectationFailure $ displayException err
+        Right yamlDeps -> do
+          let extracted = collectInteriorLabels Nothing yamlDeps
+          extracted `shouldBe` theWorksLabels Nothing
```
csasarak, line 177: [Q] Would it be worth round-tripping the `FromJSON/ToJSON` instances?

csasarak (approved):

test/App/Fossa/ManualDepsSpec.hs
jssblck: We can't round trip them: -- | Note: The ToJSON and FromJSON implementations for this type differ
-- because ToJSON is used to upload to Core and FromJSON is used to parse
-- from a fossa-deps file.

jssblck: If we really want to I could of course split them into two types for this, which would then let us round trip, but I didn't think it was worth the effort in this case especially given our future plans around moving towards an observation model. 

If you disagree happy to reconsider!

docs/references/files/fossa-deps.md
```diff
@@ -91,6 +91,59 @@ vendored-dependencies:
 
 For more details, please refer to the [feature](../../features/vendored-dependencies.md) walk through.
 
+## Labels
+
+Each kind of dependency referenced above can have a `labels` field, which is a list of labels to be added to the dependency.
+These labels are **user-defined**; you may choose any labels. What they mean is up to you and/or your organization.
+For more information on labels, refer to the main [FOSSA documentation](https://docs.fossa.com).
```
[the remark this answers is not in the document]
jssblck: Yeah, good point. I'll remove it.
Pinged the Core team about revisiting this when they have docs.

test/App/Fossa/ManualDepsSpec.hs
csasarak: Ah no that sounds good. Sorry I missed that!

PR #1509, opened 2025-02-20
Chunk large license scan uploads to avoid timeouts

jssblck:
# Overview

This PR addresses performance issues with large license scan uploads by chunking them into smaller batches. This helps avoid timeout issues that occur when trying to process hundreds of license scans at once.

## Acceptance criteria

When users upload large numbers of license scans (hundreds), the CLI should now:
- Split the uploads into smaller batches of 100 items
- Process each batch sequentially
- Complete successfully without timeout errors

## Testing plan

1. Set up a project with >200 license scans to upload
2. Run the CLI with license scanning enabled
3. Verify that:
   - The uploads are processed in batches
   - No timeout errors occur
   - All scans complete successfully

I wrote a quick script that just created 250 directories with a random file inside and a `fossa-deps` file and validated this:
image

## Risks

The main risk is that we're introducing more API calls by splitting up the uploads. However, this is preferable to having the entire operation fail due to timeouts.

## References

- [ANE-2272](https://fossa.atlassian.net/browse/ANE-2272): License scan finalize endpoint times out with large number of scans

## Checklist

- [x] No tests added as this is a performance optimization of existing functionality
- [x] No user-visible change in behavior, only in performance
- [x] No documentation updates needed as this is an internal performance improvement
- [x] No changelog update needed as this fixes an internal performance issue without changing user-visible behavior

[ANE-2272]: https://fossa.atlassian.net/browse/ANE-2272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,9 @@
 # FOSSA CLI Changelog
 
 
+## 3.9.48
+- General: Fix a bug where directory traversal could fail if the user does not have permission to read a directory.
+
 ## 3.9.47
 - Licensing: Adds support for Zeebe Community License v1.1 and Camunda License v1.0
 - NuGet: Consolidate `project.assets.json` and `PackageReference` strategies ([#1461](https://github.com/fossas/fossa-cli/pull/1461))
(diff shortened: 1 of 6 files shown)


spatten (approved):
This looks like a good and safe change to me

PR #1510, opened 2025-02-20
Update set FOSSA_API_KEY powershell command

marius-bughiu:
# Overview
It's a minor thing, but might save some time for people getting started. Without quotes, powershell will throw the following error

```
The term 'XXXX' is not recognized as the name of a
cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify
that the path is correct and try again.
At line:1 char:20
```


spatten:
Hi, thank you for this contribution. Unfortunately due to the way our CI works (we pull in a few proprietary executables) I have to make a new PR with it. You will still be credited for the work of course.

The new PR is here: https://github.com/fossas/fossa-cli/pull/1511

PR #1516, opened 2025-03-04
[ANE-2302] Use existing Cargo.lock

spatten:
# Overview

Delivers [ANE-2302](https://fossa.atlassian.net/browse/ANE-2302). 

When we analyze a cargo project, we run `cargo generate-lockfile`, even if the lockfile already exists.

This can result in changes to the existing lockfile, which is causing builds to fail for at least one customer.

This PR changes the behaviour so that we only run `cargo generate-lockfile` if the `Cargo.lock` file does not exist.

## Acceptance criteria

- If the `Cargo.lock` file does not exist, we run `cargo generate-lockfile` to create it.
- If the `Cargo.lock` file does exist, we do not run `cargo generate-lockfile`, and the `Cargo.lock` file does not get changed.
- In both cases the analysis succeeds

## Testing plan

```
make install-dev
cd <some rust project>
fossa-dev analyze --output
```

This should not result in any changes to the `Cargo.lock` file.

delete the `Cargo.lock` file and run `fossa-dev analyze --output` again. This time it should create the `Cargo.lock` file and the analysis should succeed.

## Risks

I think that this is low risk

## References

https://fossa.atlassian.net/browse/ANE-2302

## Checklist

[ANE-2302]: https://fossa.atlassian.net/browse/ANE-2302?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Cargo.hs
+++ b/src/Strategy/Cargo.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE OverloadedRecordDot #-}
+{-# LANGUAGE TemplateHaskell #-}
 
 module Strategy.Cargo (
   discover,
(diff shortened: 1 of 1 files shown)


nficca (approved):
Nice. Tested this on my local and it works like a charm.

PR #1523, opened 2025-03-13
Fix PDM Parser when it encounters multiple OR lines

zlav:
# Overview

The PDM Parser gets tripped up when handling multiple OR markers, such as
```
"greenlet!=0.4.17; (platform_machine == \"win32\" or platform_machine == \"WIN32\" or platform_machine == \"AMD64\" or platform_machine == \"amd64\" or platform_machine == \"x86_64\" or platform_machine = \"ppc64le\" or platform_machine == \"aarch64\") and python_version < \"3.14\"",
```

## Acceptance criteria

pdm.lock files containing complex combinations of "or" and "and" marker's can now be scanned.

## Testing plan

1. I started by testing on the supplied pdm.lock file that was failing. I addressed this issue.
2. I then added tests for this exact issue
3. I then iterated on the combination of "and" and "or"s trying to find another failing case.

## Risks

I missed a certain combination of "and" and "or"s that still fails.

## References

[ANE-2303](https://fossa.atlassian.net/browse/ANE-2303):  Solve PDM parsing bug

## Checklist

[ANE-2303]: https://fossa.atlassian.net/browse/ANE-2303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Python/Util.hs
+++ b/src/Strategy/Python/Util.hs
@@ -259,7 +259,7 @@ requirementParser = specification
     marker_var :: Parser Text
     marker_var = label "marker_var" $ whitespace *> (env_var <|> fmap toText python_str)
     marker_expr :: Parser Marker
-    marker_expr = 
+    marker_expr =
       label "marker_expr" $
         MarkerExpr <$> marker_var <*> marker_op <*> marker_var
           <|> whitespace *> char '(' *> marker_or <* char ')'
--- a/src/Strategy/Python/Util.hs
+++ b/src/Strategy/Python/Util.hs
@@ -271,7 +271,7 @@ requirementParser = specification
     marker_or =
       label "marker_or" $ do
         first <- marker_and
-        rest <- many (whitespace *> string "or" *> whitespace *> marker_and)
+        rest <- many (whitespace *> string "or" *> whitespace *> marker_or)
         pure $ case rest of
           [] -> first
           xs -> foldr MarkerOr first xs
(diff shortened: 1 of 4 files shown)


src/Strategy/Python/Util.hs
```diff
@@ -269,9 +269,12 @@ requirementParser = specification
 
     marker_or :: Parser Marker
     marker_or =
-      label "marker_or" $
-        try (MarkerOr <$> marker_and <* whitespace <* string "or" <*> marker_and)
-          <|> marker_and
+      label "marker_or" $ do
```
spatten, line 272: I think this still fails for mulitple `and`s in a line. E.g. this fails:

```
    "greenlet!=0.4.17; (platform_machine == \"x\" or platform_machine == \"a\" or platform_machine == \"b\") and platform_release < \"3.14\" and platform_version < \"3.1.4\"",
```

If I change either of the `and`s to an `or`, it parses:

```
    "greenlet!=0.4.17; (platform_machine == \"x\" or platform_machine == \"a\" or platform_machine == \"b\") or platform_release < \"3.14\" and platform_version < \"3.1.4\"",
```

I'm not sure that this is blocking, though. This currently fixes the customer's problem, so we could ship it. But if it's easy to fix the problem then I'd do that

spatten (approved):
:+1: with a caveat in a comment

spatten (commented):
Also, it looks like this spec is now failing:

https://github.com/fossas/fossa-cli/blob/fix/ANE-2303-pdm-parsing-multiple-ors/test/Python/RequirementsSpec.hs/#L33

zlav:
Yea this still fails for a number of test cases. I'm adding a bunch more to the reqparserspec and elsewhere. I think I found a generic solution that solves any combination but I'm testing that now.

spatten (approved):
This looks good to me too. I tried to break it as well and haven't found any problems

:+1:

PR #1537, opened 2025-05-07
[ANE-2316] Adds general pyproject.toml support

saramaebee:
# Overview

This PR implements a generic PyProject.toml parser that consolidates the handling of Python dependency management systems that use pyproject.toml files. It introduces a unified analyzer for Poetry, PDM, and standard PEP 621 projects, prioritizing detection of these systems and using available lock files when present.

The implementation creates a more consistent detection and analysis experience for Python projects that follow modern packaging standards, avoids duplicate detection of projects, and improves dependency handling across different Python build systems.

## Acceptance criteria

When users run FOSSA CLI on Python projects that use pyproject.toml configuration:
- The system will automatically detect whether the project uses Poetry, PDM, or standard PEP 621 format
- If multiple build systems are configured in a single pyproject.toml, only one will be analyzed based on a priority system (Poetry > PDM > PEP 621)
- Lock files (poetry.lock, pdm.lock) will be used when available to build complete dependency graphs
- Direct dependencies will be reported from pyproject.toml when no lock files are available
- Complex dependency types (git, path, url) will be properly parsed across all build systems
- Environment-specific dependencies (dev vs production) will be correctly distinguished

## Testing plan

1. Set up a collection of test pyproject.toml files representing Poetry, PDM, and standard PEP 621 projects
2. Run the analysis on each project type with and without lock files
3. Verify that the appropriate dependencies are extracted with correct metadata
4. Test projects with complex dependencies (git, path, url) to ensure proper parsing
5. Test projects with mixed build system configurations to verify proper prioritization
6. Compare results against direct analysis with the specific build system tools (Poetry, PDM)

## Risks

- The prioritization system might not match all user expectations when multiple build systems are configured
- The complexity of pyproject.toml variations might lead to edge cases not covered by the current implementation
- There may be subtle differences in how different build systems handle dependency specifications

## Metrics

The generic PyProject parser should improve the detection and analysis of Python projects using modern packaging standards. We could track:
- Percentage of Python projects using pyproject.toml that are successfully analyzed
- Distribution of detected project types (Poetry vs PDM vs PEP 621)
- Frequency of projects with multiple build system configurations

## References

- [PEP 621 – Storing project metadata in pyproject.toml](https://peps.python.org/pep-0621/)
- [Poetry Documentation](https://python-poetry.org/docs/pyproject/)
- [PDM Documentation](https://pdm.fming.dev/latest/usage/pyproject/)

## Checklist

- [x] If this PR added docs, I added links as appropriate to the user manual's ToC in `docs/README.md` and gave consideration to how discoverable or not my documentation is.

### DIFF
--- a/docs/references/strategies/languages/python/pyproject-generic.md
+++ b/docs/references/strategies/languages/python/pyproject-generic.md
@@ -1,0 +1,60 @@
+# PyProject.toml Generic Support
+
+The FOSSA CLI provides robust support for analyzing Python projects that use the [PEP 621](https://peps.python.org/pep-0621/) `pyproject.toml` format, which is becoming the standard for Python project metadata and configuration.
+
+## Project Discovery
+
+The CLI searches for `pyproject.toml` files and automatically detects which build system is being used by examining the file contents.
+
+## Project Type Detection
+
+When analyzing a `pyproject.toml` file, the CLI will attempt to determine which build system is in use by checking for specific sections:
+
+1. **Poetry**: Detected by the presence of `[tool.poetry]` section
+2. **PDM**: Detected by the presence of `[tool.pdm]` section
+3. **PEP 621**: Detected by the presence of `[project]` section (standard PEP 621 project)
+4. **Unknown**: If none of the above sections are found
+
+### Priority System
+
+While a `pyproject.toml` file can technically contain configuration for multiple build systems, in practice, developers typically choose one primary build system per project. 
+
+When multiple build system configurations are detected in a single `pyproject.toml` file, the CLI uses the following priority order:
+
+1. Poetry (highest priority)
+2. PDM
+3. PEP 621
+4. Unknown (lowest priority)
+
+This prioritization ensures we select the most likely active build system when multiple configurations are present, avoiding confusion and duplicate analysis results. 
+
+## Lock File Handling
+
+For each detected project type, the CLI will look for the appropriate lock file:
+
+- Poetry projects: `poetry.lock`
+- PDM projects: `pdm.lock`
+
+If a matching lock file is found, it will be used to provide a complete dependency graph with transitive dependencies. Otherwise, only direct dependencies from `pyproject.toml` will be reported.
+
+## Analysis Strategy
+
+The analysis strategy varies depending on the detected project type:
+
+| Project Type | With Lock File | Without Lock File |
+|--------------|----------------|-------------------|
+| Poetry       | Complete graph using `poetry.lock` | Direct dependencies only |
+| PDM          | Complete graph using `pdm.lock`    | Direct dependencies only |
+| PEP 621      | Direct dependencies only           | Direct dependencies only |
+
+## Limitations
+
+- When multiple build system configurations are present in a single `pyproject.toml` file, only the highest-priority one will be analyzed.
+- For PEP 621 projects without a specific build system, only direct dependencies are reported.
+- Local path dependencies might not be fully analyzed depending on the configuration.
+
+## References
+
+- [PEP 621 – Storing project metadata in pyproject.toml](https://peps.python.org/pep-0621/)
+- [Poetry Documentation](https://python-poetry.org/docs/pyproject/)
+- [PDM Documentation](https://pdm.fming.dev/latest/usage/pyproject/)
--- a/src/App/Fossa/Analyze/Discover.hs
+++ b/src/App/Fossa/Analyze/Discover.hs
@@ -69,11 +69,10 @@ discoverFuncs =
   , DiscoverFunc Nuspec.discover
   , DiscoverFunc PackagesConfig.discover
   , DiscoverFunc Paket.discover
-  , DiscoverFunc Pdm.discover
+  -- Remove PDM and Poetry discoverers - they're now handled by PyProjectGeneric
   , DiscoverFunc Perl.discover
   , DiscoverFunc Pipenv.discover
-  , DiscoverFunc Poetry.discover
-  , DiscoverFunc PyProjectGeneric.discover  -- Add Generic PyProject handler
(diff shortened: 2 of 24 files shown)


src/App/Fossa/Analyze/Discover.hs
```diff
@@ -68,10 +69,10 @@ discoverFuncs =
   , DiscoverFunc Nuspec.discover
   , DiscoverFunc PackagesConfig.discover
   , DiscoverFunc Paket.discover
-  , DiscoverFunc Pdm.discover
+  -- Remove PDM and Poetry discoverers - they're now handled by PyProjectGeneric
```
csasarak, line 72: This reads like a commit message, not a comment. Please remove it.

src/Strategy/Python/Dependency.hs
```diff
@@ -0,0 +1,553 @@
...
+    OpLtEq -> PythonVersionLtEq ver
+    OpGtEq -> PythonVersionGtEq ver
+    OpLt -> PythonVersionLt ver
+    OpGt -> PythonVersionGt ver
+    OpArbitrary -> PythonVersionArbitrary
+
+-- | Map a package category to a dependency environment
+-- This function provides consistent environment mapping across different package managers
+mapCategoryToEnvironment :: Text -> DepEnvironment
+mapCategoryToEnvironment category = case category of
```
csasarak, line 500: Please link to documentation for these env types if they exist.

src/Strategy/Python/PyProjectGeneric.hs
```diff
@@ -0,0 +1,501 @@
+{-# LANGUAGE TemplateHaskell #-}
```
csasarak, line 1: We already have a `pyproject.toml` parser in `src/Strategy/Python/Poetry/PyProject.hs`. The intent of this ticket wasn't to create a new one: it was to reify that tactic as its own strategy which augmented its responses with `Poetry/PDM` information. 

As a type you could imagine something like this:

```haskell
data PyProject buildSystem = PyProject {
  commonField :: ...,
  commonField2 :: ...,
  projectSpecificInfo :: buildSystem
}
```
^^ Though you don't need to do it this way.

I'm pretty uncomfortable with merging in a totally new pyproject parser unless it's obviously better than what we have. Even in that case we should deprecate the old one. What is your sense here/do you have thoughts about that?

src/Strategy/Python/PyProjectGeneric/Types.hs
```diff
@@ -0,0 +1,333 @@
...
+  deriving (Eq, Ord, Show, Generic)
+
+instance ToJSON PyProjectType where
+  toJSON = Aeson.genericToJSON Aeson.defaultOptions
+
+-- | Types of lock files
+data LockFileType 
+  = PoetryLock 
+  | PDMLock 
+  | UVLock     -- For future support
```
csasarak, line 48: We should design types for what we know now. We can add `UVLock` when we actually need it.

docs/references/strategies/languages/python/pyproject-generic.md
```diff
@@ -0,0 +1,62 @@
...
+The CLI searches for `pyproject.toml` files and automatically detects which build system is being used by examining the file contents.
+
+## Project Type Detection
+
+When analyzing a `pyproject.toml` file, the CLI will attempt to determine which build system is in use by checking for specific sections:
+
+1. **Poetry**: Detected by the presence of `[tool.poetry]` section
+2. **PDM**: Detected by the presence of `[tool.pdm]` section
+3. **PEP 621**: Detected by the presence of `[project]` section (standard PEP 621 project)
+4. **Unknown**: If none of the above sections are found
```
csasarak, line 16: What is the difference for the customer between PEP-621 and Unknown? 

In the "Analysis Strategies" section the "Unknown" project type disappears.

docs/references/strategies/languages/python/pyproject-generic.md
```diff
@@ -0,0 +1,62 @@
...
+While a `pyproject.toml` file can technically contain configuration for multiple build systems, in practice, developers typically choose one primary build system per project. 
+
+When multiple build system configurations are detected in a single `pyproject.toml` file, the CLI uses the following priority order:
+
+1. Poetry (highest priority)
+2. PDM
+3. PEP 621
+4. Unknown (lowest priority)
+
+This prioritization ensures we select the most likely active build system when multiple configurations are present, avoiding confusion and duplicate analysis results.
```
csasarak, line 29: What makes Poetry most likely as opposed to the others? 

The [spec](https://packaging.python.org/en/latest/specifications/pyproject-toml/#declaring-build-system-dependencies-the-build-system-table) only says that these are the lists of dependencies needed to build the project. Pip also [describes](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#build-time-dependencies) it this way. 

I think that using this table value as a way to identify the underlying build system is actually the wrong approach. It'd be better to search for the lock file for the respective build system.

src/Strategy/Python/PyProjectGeneric.hs
```diff
@@ -0,0 +1,501 @@
...
+    UnknownProject -> [] -- Unknown project type, no dependencies
+
+-- | Extract dependencies from a Poetry project using unified conversion functions
+extractPoetryDependencies :: PyProjectGeneric -> [Dependency]
+extractPoetryDependencies pyproject =
+  case poetrySection pyproject of
+    Nothing -> []
+    Just poetry -> 
+      let 
+        -- Convert production dependencies
```
csasarak, line 302: These comments don't really add anything over the names imo. Prefer to use comments to explain "why" and "how". Sometimes it's appropriate to describe "what" but that should be rare - ideally code is clear enough that "what" is easy.

csasarak (changes requested):
I left a few comments but they are mainly questions about approach/what research you did in coming up with this solution. We should discuss approach before you move forward with major changes though.

PR #1551, opened 2025-05-30
[TKT-13372] Fix syntax error in Maven documentation

saramaebee:
# Overview

Fixes a bash syntax error in the Maven documentation where `fossa analyze` was incorrectly on the same line as the `export` statement.

## Acceptance criteria

Users can now copy and execute the Maven MAVEN_ARGS example without syntax errors.

## Testing plan

1. Found the syntax error on line 42 of `docs/references/strategies/languages/maven/maven.md`
2. Split the export and command onto separate lines
3. Verified correct bash syntax

## Risks

None - minor documentation fix only.

## References

- TKT-13372: Customer reported syntax error in Maven documentation

### DIFF
--- a/docs/references/strategies/languages/maven/maven.md
+++ b/docs/references/strategies/languages/maven/maven.md
@@ -39,7 +39,8 @@ Not directly, but `mvn` itself has an [environment variable](https://maven.apach
 For example, to set a custom `settings.xml` file you can use an invocation like this:
 
 ```sh
-export MAVEN_ARGS="--settings /foo/bar/settings.xml" fossa analyze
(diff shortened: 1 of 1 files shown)


spatten (commented):
This works, but I think a better fix is to keep it on one line and remove the `export`. Like this:

```
MAVEN_ARGS="--settings /foo/bar/settings.xml" fossa analyze
```

This is safer as the `MAVEN_ARGS` environment variable is only set for this one line. If I export it, it's set for the rest of the script

spatten (approved):
Looks great. Thank you!

PR #1553, opened 2025-06-02
Ignore the tools and godebug GO directives

zlav:
# Overview

Go introduced a new `tools` directive in version [1.24](https://tip.golang.org/doc/go1.24). [Tools directive documentation](https://tip.golang.org/doc/modules/managing-dependencies#tools)

> Tool dependencies let you manage developer tools that are written in Go and used when working on your module. 

The explicit purpose is to allow users to download developer dependencies that are not shipped with the tool. We currently do not support differentiating between developer dependencies so I believe the correct behavior here is to ignore any dependencies in this directive.

This PR also ignores the `godebug` directive that we previously did not support. This specifies the default `GODEBUG` environment variable settings for the project. Unsure when this was introduced, but this causes a failure.

## Acceptance criteria

GO projects with `tool` and `godebug` directives with single lines declarations properly scan.

Note: In the process of writing this PR I realized that multi-line syntax for `toolchain` and by extension, now `tool` and `godebug` directives is not supported by the CLI. I am going to work on this as a follow up item.

## Testing plan

- I downloaded a project that has this tools directive
- I scanned it with the older version of the CLI and saw it fail
- I scanned it with this new branch and saw it succeed with the correct list of dependencies

## Risks

- Tool support
- The way I'm ignoring tool support accidentally starts ignoring other things. I don't believe this is possible as the parser should encompass only what comes after the `tool` section and nothing else.
- After further testing I've found that `toolchain`, `tool`, and `godebug` support as implemented do not support multi line ignores. I'm investigating this but may put up a PR without this for the time being.

## References

- [PS-655](https://fossa.atlassian.net/browse/PS-665): Implement tool support

## Checklist

[PS-655]: https://fossa.atlassian.net/browse/PS-655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## 3.10.10
 
-- go: support the `tool` directive introduced in go Feb 2025 ([#1546](https://github.com/fossas/fossa-cli/pull/1553))
+- go: support the `tool` directive introduced in go Feb 2025 ([#1553](https://github.com/fossas/fossa-cli/pull/1553))
 
 ## 3.10.9
 


Changelog.md
```diff
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## 3.10.10
+
+- go: support the `tool` directive introduced in go Feb 2025 ([#1546](https://github.com/fossas/fossa-cli/pull/1553))
```
spatten: nit: the `#1546` should be `#1553`

spatten (approved):
Looks good to me! I tested by adding a tool to Themis, watching it fail to scan on master and succeed on this PR

PR #1554, opened 2025-06-05
Make it clearer that our manual dependency support is not specific to…

csasarak:
This just just an update to make it clear that our `git` support is not specific to Github.

## Checklist


james-fossa (approved):
I think this is a clearer line. Good job!

PR #1555, opened 2025-06-12
Fix unzipping JARs that symlink to other layers

zlav:
# Overview

This ticket solves an issue where a symlink between a container images's layers is blocking our JARs in Containers analysis from completing, which results in the overall CLI container analysis failing.

Delivers https://fossa.atlassian.net/browse/ANE-2474

## Acceptance criteria

- Symlinked JARs do not cause a failure of the whole scan.

## Testing plan

- I downloaded the impacted JAR from the ticket
- I ran the old version on it and saw the same failure the user is seeing
- I ran the updated version in this PR and saw that only 1 JAR failed to be unzipped, but that the others succeeded.

- I added a test to the `nested_jars.tar` test that has a symlinked tar. The previous version of the CLI wasn't able to successfully analyze this and failed on the current test suite. 
- You will see a `warn` log line if you run `fossa analyze --debug` when analyzing this tar file with the new CLI.

## Risks

This method warns and skips any JAR it comes across that fails extraction. This could result in unintended consequences with JARs that we truly have issues unarchiving. I believe this trade off is worth it. The customer issue in 2474 is solved by this fix and is the first time we have been pushed to deal with JAR extraction issues. Additionally, the error is not swallowed, it is still warned and we can see what happens.

## References

[ANE-2474](https://fossa.atlassian.net/browse/ANE-2474): Implement a fix for unzipping symlinked jars

## Checklist

[ANE-2474]: https://fossa.atlassian.net/browse/ANE-2474?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## 3.10.11
+
+- container scanning: fix unzipping JARs that symlink to other layers #1555 ([#1555](https://github.com/fossas/fossa-cli/pull/1555))
+  
 ## 3.10.10
 
 - go: support the `tool` directive introduced in go Feb 2025 ([#1553](https://github.com/fossas/fossa-cli/pull/1553))


test/App/Fossa/Container/testdata/nested-jar/README.md
```diff
@@ -1,10 +1,29 @@
-# Making test/Container/testdata/nested-jar.tar
+# Using this test suite
```
spatten, line 1: This README is so much better. Thank you!

PR #1557, opened 2025-06-26
Improve Link Checking

gilfaizon:
# Overview

Jess pointed out that the tests should've caught the broken link in https://github.com/fossas/fossa-cli/pull/1556

In testing, the global enable/disable wasn't playing well with the previously broken link. Using the targeted disables fixed this for me locally.

The dead link has since been removed, but I went ahead and removed the other global disables just in case, and updated the action as the previous action is now deprecated.

## Acceptance criteria

Better link checking! Nothing should fail here, but if the links go down in the future, we should be alerted.

## Testing plan

I just tested the links manually 

## Risks

Ideally little risk, maybe I break our tests.

## References

- Continues the work in https://github.com/fossas/fossa-cli/pull/1556

## Checklist

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Check Markdown links
-        uses: gaurav-nelson/github-action-markdown-link-check@v1
+        uses: tcort/github-action-markdown-link-check@v1.1.0
         with:
           use-quiet-mode: "yes"
           config-file: ".markdown-link-check.json"
--- a/README.md
+++ b/README.md
@@ -2,15 +2,17 @@
 ![FOSSA](https://raw.githubusercontent.com/fossas/fossa-cli/master/docs/assets/header.png)
 
 # FOSSA CLI
-<!-- markdown-link-check-disable -->
 <!-- NOTE: If you change the format of the "FOSSA Downloads" badge, make sure to also update the CI action at `./github/workflows/badges.yml` that updates the download count. -->
+<!-- markdown-link-check-disable-next-line -->
 [![FOSSA Downloads](https://img.shields.io/badge/downloads-5.8M-brightgreen)](https://github.com/fossas/fossa-cli/releases)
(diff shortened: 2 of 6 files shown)


.github/workflows/lint.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Check Markdown links
-        uses: gaurav-nelson/github-action-markdown-link-check@v1
+        uses: tcort/github-action-markdown-link-check@v1.1.0
```
spatten, line 30: I'm consistently seeing this on this PR:

```
=========================> MARKDOWN LINK CHECK <=========================

  ERROR: 3 dead links found in ./docs/differences-from-v1.md !
  [✖] #clojure → Status: 404
  [✖] #haskell → Status: 404
  [✖] #how-do-i-run-only-a-specific-analyzer-for-my-project → Status: 404

  ERROR: 1 dead links found in ./docs/references/subcommands/report.md !
  [✖] #common-fossa-project-flags → Status: 404

=========================================================================
```

I'm not sure why it's failing on only these links and not the rest of the links in the `differences-from-v1.md` document.

I'm guessing that this is due to the change in GH actions here. It makes sense to move to the new, maintained one, but we probably shouldn't if it's going to cause these weird test failures. Did we need to move to the new action, or were you just doing the right thing and moving us to the maintained one?

gilfaizon: I was just doing the right thing! It failed due to trailing spaces, fixed in https://github.com/fossas/fossa-cli/pull/1557/commits/7be311924c652d7cf728f926a1b8f9806c91a91e

gilfaizon: Yeah, I just saw the link check test pass - can confirm it's fixed! I'm kinda glad it caught those links, I was hesitant that the changes I originally made actually wouldn't have any effect or benefit.

spatten: Nice!

spatten (approved):
:+1:

PR #1558, opened 2025-06-27
Add `rendergraph` to releases

jssblck:
# Overview

Cortez and I are working with a customer that would benefit from having our internal `rendergraph` tool being public; this PR adds it to the FOSSA CLI repository and releases.

`rendergraph` itself is sort of a "best effort" tool so it doesn't have tests; I don't think it's worth adding to docs as we would expect users to basically use this when we suggest it for troubleshooting.

## Acceptance criteria

`rendergraph` is available in FOSSA CLI releases.

## Testing plan

Sadly, just manual testing

## Risks

Yet more random stuff plugged into FOSSA CLI releases 🙃 

## Metrics

None

## References

None

## Checklist

### DIFF
--- a/tools/rendergraph/README.md
+++ b/tools/rendergraph/README.md
@@ -1,0 +1,228 @@
+# fossa-rendergraph
+
+Utility to render the output of `fossa analyze -o` as a graph.
+
+This program isn't _guaranteed_ to not be broken by new FOSSA CLI releases,
+but we'll try to keep it from doing so especially since we try to keep output of FOSSA CLI stable.
+
+# Installation from source
+
+0. Install `rust`: https://www.rust-lang.org/tools/install
+1. Clone this repo locally
+2. Run `cargo install --path rendergraph` from the parent directory.
+3. Use as `fossa analyze -o | rendergraph` (you can also `cat output.json | rendergraph`- any stdin works)
+
+# What to expect
+
+Direct deps are roots, and tree branches leading off of them are their transitive deps.
+
+Cycles are denoted by the literal `cycle ->` followed by the name of the node that would have had a cycling edge.
+
+For example, a cycle of `a -> b -> c -> a` reads as:
+```
+a
+└─ b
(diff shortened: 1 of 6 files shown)


tools/rendergraph/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+This program isn't _guaranteed_ to not be broken by new FOSSA CLI releases,
+but we'll try to keep it from doing so especially since we try to keep output of FOSSA CLI stable.
+
+# Installation from source
+
+0. Install `rust`: https://www.rust-lang.org/tools/install
+1. Clone this repo locally
+2. Run `cargo install --path rendergraph` from the parent directory.
+3. Use as `fossa analyze -o | rendergraph` (you can also `cat output.json | rendergraph`- any stdin works)
```
csasarak: [nit] Recommend moving this down to the bottom of the readme. A customer who wants to use this can just download the binary, so the docs about how to use it are way more important than how to build it IMO.

tools/rendergraph/README.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+This tool outputs all dependencies found by FOSSA and doesn't currently account for FOSSA's "unused dependencies" graph pruning.
+
+# Future work
+
+A list of things I've thought of to add. Have an idea? Add it to this list via PR!
+
+- [ ] Account for graph pruning (use source units to do this)
+- [ ] Support graphviz output
```
csasarak: IMO, this really ought to be a part of Ficus. Being able to just instruct the CLI to do this is a better user-experience than "please download this bin from GH." 

This PR is totally fine, I'm just saying we should have that as future work item in this list IMO.

jssblck: Good idea, updated

tools/rendergraph/README.md
jssblck: Good point, moved

PR #1561, opened 2025-07-08
Pnpm 9

csasarak:
# Overview

This PR attempts to add some support for version 9 of PNPM's lockfile. 

## Acceptance criteria

* We can scan both pnpm version 9 lockfiles as well as workspaces.
* PNPM's new `pnpm-workspace.yaml` file is now suppported.
* Dev dependencies are correctly filtered.

## Testing plan

I used both [pnpm](https://github.com/pnpm/pnpm) itself's repo and the [vuejs core](https://github.com/vuejs/core)'s repo to verify. I made sure that dev dependencies were not included in the output and verified that dependencies from multiple sub-modules were included properly.

Additionally, there are integration tests testing both the workspace case and the non-workspace case.

## Risks

There isn't really a very detailed [spec](https://github.com/pnpm/spec/blob/master/lockfile/9.0.md) for the lockfile, so some of this I had to develop empirically rather than to spec. I would expect we'll get some bug reports for the functionality in this PR.

Additionally, the PNPM strategy probably should be refactored into several sub-parsers which each support some lockfile range. As it is now, a lot of the code for different versions is intermingled and just works by conditional branching and/or combining fields from different versions which is a smell IMO. I did not have time to do this refactor myself though so it's left for a future time, or Ficus. Overall, I think this PR is a bit gross but an improvement on the status quo.

## References

[ANE-2235](https://fossa.atlassian.net/browse/ANE-2235)

## Checklist

[ANE-2235]: https://fossa.atlassian.net/browse/ANE-2235?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Changelog.md
```diff
@@ -1,9 +1,13 @@
 # FOSSA CLI Changelog
 
+## 3.10.12
+
+- PNPM: Initial support for lockfile version 9.0 ((#1561)[https://github.com/fossas/fossa-cli/pull/1561])
```
spatten: The link syntax is incorrect here, which is causing the link checker to fail.

```suggestion
- PNPM: Initial support for lockfile version 9.0 ([#1561](https://github.com/fossas/fossa-cli/pull/1561))
```

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -123,21 +130,47 @@ data PnpmLockfile = PnpmLockfile
...
+  , lockFileSnapshots :: PnpmLockFileSnapshots
+  -- ^ Dependency graph in lockfile version > 9
   }
   deriving (Show, Eq, Ord)
 
+type SnapshotDepName = Text
+
+type SnapShotDepRev = Text
+
+-- | Lockfile versions > 9 use snapshots to represent the dependency graph.
```
spatten, line 142: Can you add an example of what this looks like? Just a snippet of the lockfile so that we can see the structure

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -283,29 +318,31 @@ buildGraph lockFile = withoutLocalPackages $
...
-        maybe (pure ()) (edge parentDep) (toResolvedDependency deepName deepVersion)
+        maybe (pure ()) (edge parentDep) (toResolvedDependency False deepName deepVersion)
   where
     getPkgNameVersion :: Text -> Maybe (Text, Text)
     getPkgNameVersion = case lockFileVersion lockFile of
       PnpmLock4Or5 -> getPkgNameVersionV5
       PnpmLock6 -> getPkgNameVersionV6
       PnpmLockLt4 _ -> getPkgNameVersionV5 -- v3 or below are deprecated and are not used in practice, fallback to closest
-      PnpmLockGt6 _ -> getPkgNameVersionV6 -- at the time of writing there is no v7, so default to closest
+      PnpmLockV678 _ -> getPkgNameVersionV6 -- at the time of writing there is no v7, so default to closest
```
spatten, line 351: Do we still need `PnpmLock6` in this enum, or should it be merged into `PnpmLockV678`?

I'm finding it confusing that the `6` is included in `PnpmLockV678`, and it seems like `PnpmLockV678` is the same as `PnpmLock6` in behaviour

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -156,7 +189,7 @@ instance FromJSON PnpmLockfile where
             then Map.insert "." virtualRootWs importers
             else importers
 
-    pure $ PnpmLockfile refinedImporters packages rawLockFileVersion
+    pure $ PnpmLockfile{importers = refinedImporters, packages = packages, lockFileVersion = rawLockFileVersion, lockFileSnapshots = snapshots}
     where
       getVersion (TextLike ver) = case (listToMaybe . toString $ ver) of
```
spatten, line 203: I'm a bit confused about how `TextLike` deals with a version like '9.0', like we see [here](https://github.com/vuejs/core/blob/main/pnpm-lock.yaml#L1). Does that get parsed as an int and then turned into a string of `'9'`?

spatten (approved):
Looks good! I had a few comments and questions, but nothing blocking.

src/Strategy/Node/Pnpm/PnpmLock.hs
csasarak: Good call out, this was tricky and took me a bit of time to figure out. It turns out `listToMaybe` either returns `Nothing` for an empty list or the first element of the list in `Just` for a non-empty one.

src/Strategy/Node/Pnpm/PnpmLock.hs
csasarak: Good call, I've combined them.

PR #1566, opened 2025-07-23
[ane-2575] scan all layers for os info

jagonalez:
# Overview

- change to scan all layers (the "squashed" layers first, then base layer) for os information. This change could technically cause os information to be different between previous scans, we previously only checked the base layer. A layer further down the chain could update the os information. If someone feel strongly we should only check for os info after none is found in the base layer - I'm happy to make the change, this feels more accurate though.

## Acceptance criteria

- os info is retrieved even if it exists in another layer

## Testing plan

- download the tarballs from the ANE on call ticket, specifically `1.52.0`, then an analysis on it and dependencies are reported.
- try it on `cgr.dev/chainguard/glibc-dynamic:latest-dev` and dependencies are also reported

## Risks

- I highlighted it in my overview but this change could cause os-information to change between the previous container scans.

## Metrics

n/a

## References

https://fossa.atlassian.net/browse/ANE-2575

## Checklist


csasarak (approved):
Looks good, thanks!

PR #1568, opened 2025-07-29
[ANE-1322] Apply path filters at discovery stage for maven

ryanlink:
# Overview

This PR fixes Maven POM discovery to respect path filters during the discovery stage rather than only during analysis. Previously, Maven would traverse and load all POMs when building the closure graph, then apply path filters only later during analysis. This caused issues where users experienced discovery failures even when they had excluded problematic paths with discovery filters.

This PR accomplishes path filtering at discovery by modifying the POM resolution logic to check `pathAllowed` before recursively loading POMs, preventing excluded directories from being traversed entirely.

## Acceptance criteria

When users configure path filters to exclude directories (like `.m2/repository`), Maven discovery will no longer attempt to load POMs from those excluded paths, preventing discovery failures that couldn't previously be worked around with filters.

## Testing plan

I validated this PR works by:

1. **Created comprehensive unit test** in `test/Maven/PathFilterSpec.hs` that sets up a realistic Maven project structure with a parent POM located in an excluded `.m2` directory
2. **Verified unit test passes** - when `.m2` is excluded via path filters, only the root POM is loaded (not the parent POM in the excluded directory)
3. **Tested against original reproduction case** from ANE-1322:
   - Downloaded the debug bundle attached to ANE-1322 (fossa.debug.json)
   - Extracted test files using `dtool write-files fossa.debug.json extracted-files`
   - Configured `.fossa.yml` to exclude `.m2/` directory
   - Ran `fossa analyze extracted-files/softin-inecrf -o`
   - **✅ CONFIRMED**: Only root project analyzed (no `.m2/repository/` POMs processed)
   - **✅ BEFORE FIX**: Multiple Maven projects in `.m2/repository/` were being analyzed
   - **✅ AFTER FIX**: Only the root Maven project is analyzed, excluded paths respected

To manually test:
1. Create a Maven project with a parent POM reference pointing to a directory you want to exclude
2. Configure path filters to exclude that directory (e.g., `.m2/`)  
3. Run `fossa analyze` - discovery should succeed without attempting to load POMs from excluded paths

## Risks

The main risk area is ensuring the `AllFilters` context is properly threaded through all the Maven POM resolution functions. I've updated the function signatures for `buildGlobalClosure` and `recursiveLoadPom` to include the `Has (Reader AllFilters) sig m` constraint, and added a compatibility fix for the Scala strategy that also uses Maven POM resolution.

## Metrics

This change should reduce Maven discovery failures for users who have problematic paths that can be excluded with filters. We could potentially track discovery success rates before/after this change, but the impact should be visible in reduced support tickets about Maven discovery issues.

## References

- [ANE-1322](https://fossa.atlassian.net/browse/ANE-1322): Apply path filters at discovery stage for maven

## Checklist

[ANE-1322]: https://fossa.atlassian.net/browse/ANE-1322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -96,7 +96,7 @@ jobs:
         mkdir vendor-bins
         ./vendor_download.sh
       env:
-        GITHUB_TOKEN: ${{ secrets.BASIS_ACCESS_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.ANALYSIS_VENDOR_DOWNLOAD_TOKEN }}
 
     - name: Build
       env:
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -167,7 +167,7 @@ jobs:
         mkdir vendor-bins
         ./vendor_download.sh
       env:
-        GITHUB_TOKEN: ${{ secrets.BASIS_ACCESS_TOKEN }}
+        GITHUB_TOKEN: ${{ secrets.ANALYSIS_VENDOR_DOWNLOAD_TOKEN }}
 
     - name: Build and Unit Test (Linux ARM)
       if: ${{ matrix.os == 'LinuxARM' }}
--- a/.github/workflows/integrations-test.yml
+++ b/.github/workflows/integrations-test.yml
@@ -103,7 +103,7 @@ jobs:
         mkdir vendor-bins
         ./vendor_download.sh
       env:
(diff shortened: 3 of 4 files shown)


ryanlink:
not sure what this CI failure means

```
Marking binaries executable
Compressing index.gob
xz: vendor-bins/index.gob: No such file or directory
Error: Process completed with exit code 1.
```

PR #1572, opened 2025-08-01
Replace the BASIS token with an updated token

zlav:
# Overview

This PR replaces the reference to the BASIS_ACCESS_TOKEN. The token recently expired, and I figured it was a good time to update the name. Basis no longer really exists, and the purpose of the command is to download libraries that have been ripped out from Basis, among other new things like circe. Changing it to ANALYSIS_VENDOR_DOWNLOAD_TOKEN felt appropriate. I'm open to other suggestions.

## Acceptance criteria

This build succeeds.

## Testing plan

I validated that the updated token works by replacing the BASIS_ACCESS_TOKEN with the updated token and retrying another failed build.

## Risks

The only risk is that the permissions could be different for this token than the BASIS_ACCESS_TOKEN was, but it only has read access, so it should be fine.

## References

Failed builds

## Checklist

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -96,7 +96,7 @@ jobs:
         mkdir vendor-bins
         ./vendor_download.sh
       env:
-        GITHUB_TOKEN: ${{ secrets.BASIS_ACCESS_TOKEN }}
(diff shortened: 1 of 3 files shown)


csasarak (approved):
LGTM!

I think we should put this in 1password. Or barring that, make an entry for it in 1password and describe what to do if there are problems in the notes.

PR #1588, opened 2025-09-09
Change token names and update tracing-subscriber

zlav:
# Overview

Scott and I decided that the tokens we use to download vendored archives aren't clear and when they get outdated or broken, its hard to trace the origin. This change at least makes it clear they come from the fossabot account

## Acceptance criteria

- Token name is more clear

## Testing plan

- CI works is the only criteria for this to pass

## References

https://teamfossa.slack.com/archives/C039KE5ERNE/p1757448383332399 - Thread where we found the token had expired and ended up finding out the token had been changed from FOSSABOT

## Checklist

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1388,12 +1388,11 @@ dependencies = [
 
 [[package]]
 name = "nu-ansi-term"
-version = "0.46.0"
+version = "0.50.1"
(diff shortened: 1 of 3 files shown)


spatten (approved):
👍

PR #1597, opened 2025-10-07
Add "exclude manifest strategies" option

tjugdev:
# Overview
- Added a new option to skip any manifest-based strategies for discovering targets, replacing the old workaround of using `--only-target npm --exclude-target npm`. This option can be enabled via the CLI flag `--exclude-manifest-strategies` or in the config file under `targets.excludeManifestStrategies`.
- The logic to resolve the various ways of setting (possibly conflicting) options in the config file and CLI is as follows
  - The options for target filters, path filters, and exclude manifest strategies are applied as a group from a single source, either CLI options or the config file. I.e. we won't take some of those options from the config file and some from CLI arguments. This matches the previous behaviour, with the logic extended to include the new exclude manifest strategies option.
  - If any CLI arguments for options in this group are set, we use the CLI options only and disregard the config file options.
  - Otherwise we will use the config file's options.
  - If exclude manifest strategies is set alongside any other filters, then exclude manifest strategies applies, and a warning is displayed that the other filters are being ignored.

## Acceptance criteria
Users can set `--exclude-manifest-strategies` or `targets.excludeManifestStrategies` to avoid the use of any manifest-based strategies for discovering targets.

## Testing plan
- New unit tests added.
- Test a few combinations of CLI filter options and config file filter options:
  - `--exclude-manifest-filters` and `targets.only` in config. CLI option should apply.
  - `--exclude-manifest-filters` and `--only-target npm`. The `--exclude-manifest-filters` flag should apply.
  - `--only-target npm` and `targets.excludeManifestFilters` in config. Cli option should apply.
  - `targets.excludeManifestFilters` and `targets.only` set to `npm` in config. `targets.excludeManifestFilters` should apply.
  - Any other combinations you can think of.

## References
- [ANE-767](https://fossa.atlassian.net/browse/ANE-767): Add new flag for fossa-deps only scan

## Checklist

### DIFF
--- a/docs/contributing/effects-tutorial.md
+++ b/docs/contributing/effects-tutorial.md
@@ -625,7 +625,7 @@ necessary for understanding how the `Member` typeclass works fundamentally.
 
 [Haskell101]: http://youtu.be/cTN1Qar4HSw
 [Haskell102]: http://youtu.be/Ug9yJnOYR4U
-[MTLTutorial]: http://two-wrongs.com/a-gentle-introduction-to-monad-transformers
+[MTLTutorial]: https://web.cs.dal.ca/~nzeh/Teaching/3137/haskell/monads/transformers/
 [fused-talk]: http://youtu.be/vfDazZfxlNs
 [eff-tutorial]: https://github.com/fused-effects/fused-effects/blob/36bec2d6c0e97f7e01df97acd15012e1735c28bf/docs/defining_effects.md
 [simple-file]: https://github.com/fossas/fossa-cli/blob/aafe05624d102eb746b85c78027983e8c128bf24/src/Control/Carrier/Simple.hs
--- a/docs/references/files/fossa-yml.md
+++ b/docs/references/files/fossa-yml.md
@@ -305,6 +305,9 @@ targets:
       path: prod/docker
 ```
 
+#### `targets.excludeManifestStrategies:`
+If set to true, all manifest-based strategies for discovering targets will be skipped. This has the effect of only searching dependencies explicitly specified in `fossa-deps.yml`. This setting will override any target filters or path filters.
+
 ### `paths:`
 The paths filtering section allows you to specify which paths should be scanned and which should not. The paths should be listed as their location from the root of your project.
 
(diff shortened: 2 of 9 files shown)


docs/contributing/effects-tutorial.md
```diff
@@ -625,7 +625,7 @@ necessary for understanding how the `Member` typeclass works fundamentally.
 
 [Haskell101]: http://youtu.be/cTN1Qar4HSw
 [Haskell102]: http://youtu.be/Ug9yJnOYR4U
-[MTLTutorial]: http://two-wrongs.com/a-gentle-introduction-to-monad-transformers
+[MTLTutorial]: https://web.cs.dal.ca/~nzeh/Teaching/3137/haskell/monads/transformers/
```
tjugdev, line 628: The old link's cert expired and the link-check was failing. Picked a new monad transformers tutorial.

docs/references/files/fossa-yml.md
```diff
@@ -305,6 +305,9 @@ targets:
       path: prod/docker
 ```
 
+#### `targets.excludeManifestStrategies:`
+If set to true, all manifest-based strategies for discovering targets will be skipped. This has the effect of only searching dependencies explicitly specified in `fossa-deps.yml`. This setting will override any target filters or path filters.
```
spatten: We will do more than just fossa-deps.yml. For example, we can still do custom-license scans, first-party license scans and snippet scans if this flag is set and they are configured to run. 

So maybe something like this? It's a bit wordy, but I think it's at least correct:

> If set to true, all manifest-based strategies for discovering targets will be skipped. This will still allow things like finding dependencies explicitly specified in `fossa-deps.yml`, running [first-party license scans](docs/features/first-party-license-scans.md) or running [custom-license scans](docs/features/custom-license-and-keyword-searches.md) if they are configured. This setting will override any target filters or path filters.

src/App/Fossa/Config/Analyze.hs
```diff
@@ -575,16 +582,39 @@ collectFilters ::
...
+        let logMessage =
+              if isMempty filters
+                then
+                  logInfo "Excluding manifest strategies in locating targets"
+                else
+                  logWarn "Ignoring explicit filters because \"exclude manifest strategies\" is set to true"
+         in excludeManifestStrategiesFilters <$ logMessage
+      resolveFilters filters False = pure filters
+
+  case (isMempty cliFilters && not cliExcludeManifestStrategies, isMempty cfgFileFilters && not cfgFileExcludeManifestStrategies) of
```
spatten: [nit] I find the logic here hard to think about because of the double negatives.

It might be worth defining `cliAllowManifestStrategies` and `cfgFileAllowManifestStrategies`, but that's definitely optional

PR #1598, opened 2025-10-07
Remove reachability output from scan summary

tjugdev:
# Overview
Remove reachability log output from the scan summary as the feature is in an alpha state and having the scan summary indicate most projects were skipped is misleading/confusing. The reachability code still runs, it's just not included in the summary. The info level log messages when uploading reachability results have been changed to debug rather than being removed.

## Acceptance criteria
There are no reachability logs in the scan summary.

## Testing plan
Run `fossa analyze` on a project and confirm there are no reachability logs in the scan summary.

## References
- [ANE-2658](https://fossa.atlassian.net/browse/ANE-2658): Remove "reachability" logs from the CLI

## Checklist

[ANE-2658]: https://fossa.atlassian.net/browse/ANE-2658?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Reachability/Upload.hs
+++ b/src/App/Fossa/Reachability/Upload.hs
@@ -39,7 +39,7 @@ import Data.Maybe (mapMaybe)
 import Diag.Result (Result (..))
 import Discovery.Filters (AllFilters)
 import Effect.Exec (Exec)
-import Effect.Logger (Logger, logDebug, logInfo, pretty)
+import Effect.Logger (Logger, logDebug, pretty)
 import Effect.ReadFS (ReadFS)
 import Srclib.Converter (projectToSourceUnit)
 import Srclib.Types (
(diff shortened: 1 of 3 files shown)


src/App/Fossa/Reachability/Upload.hs
```diff
@@ -134,7 +134,7 @@ callGraphOf (Scanned dpi (Success _ projectResult)) = do
     -- it is impossible to perform reachability, as we may not have all symbols
     -- used in the application to perform accurate analysis
     (Partial, _) -> do
-      logInfo . pretty $ "FOSSA CLI does not support reachability analysis, with partial dependencies graph (skipping: " <> displayId <> ")"
+      logDebug . pretty $ "FOSSA CLI does not support reachability analysis, with partial dependencies graph (skipping: " <> displayId <> ")"
```
csasarak, line 137: Is there a reason to keep these at all? I think they were of questionable utility even when it was running. 

Making them debug level does mean a customer won't see them so this isn't blocking.

tjugdev: My thought was that we are still running the reachability analysis, so if somebody is looking at debug logs it makes sense to still include the output there.

PR #1599, opened 2025-10-09
Handle Cargo path deps

tjugdev:
# Overview
Detect Cargo path dependencies and report them correctly as such.

I originally included handling git dependencies as well, as they are also not currently being handled correctly. However, the git locator type doesn't have the ability to encode package name in addition to the repo, which means if multiple packages are found in a single repo, the dependencies are not reported accurately. This is an issue with the git locator type that applies more generally, so I backed this out and we can revisit the handling of git dependencies separately.

## Acceptance criteria
Path dependencies are reported as path dependencies when scanning a Cargo project.

## Testing plan
- New tests
- Scanned a repo with Cargo path deps and confirmed they are now being reported appropriately.  

## References
- [ANE-2527](https://fossa.atlassian.net/browse/ANE-2527): path dependencies for cargo are being reported as normal cargo dependencies

## Checklist

[ANE-2527]: https://fossa.atlassian.net/browse/ANE-2527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Cargo.hs
+++ b/src/Strategy/Cargo.hs
@@ -89,7 +89,7 @@ import Text.Megaparsec (
   takeWhile1P,
   try,
  )
-import Text.Megaparsec.Char (char, digitChar, space)
+import Text.Megaparsec.Char (char, digitChar, space, string)
 import Toml.Schema qualified
 import Types (
   DepEnvironment (EnvDevelopment, EnvProduction),
@@ -369,6 +369,9 @@ instance ToDiagnostic FailedToRetrieveCargoMetadata where
     let header = "Could not retrieve machine readable cargo metadata for: " <> toText path
     Errata (Just header) [] Nothing
 
+type PackageIdSourceKind = Text.Text
+type PackageIdSourceProtocol = Text.Text
+
 toDependency :: PackageId -> Set CargoLabel -> Dependency
 toDependency pkg =
   foldr
@@ -385,8 +388,11 @@ toDependency pkg =
     applyLabel :: CargoLabel -> Dependency -> Dependency
     applyLabel (CargoDepKind env) = insertEnvironment env
 
-    parseDepKindAndProtocol :: Text.Text -> (Text.Text, Text.Text)
-    parseDepKindAndProtocol src = case breakOn "+" $ fst $ breakOn ":" src of
+    -- For example:
+    -- path+file:///some/file/path -> ("path", "file")
+    -- file:///some/file/path      -> ("", "file")
+    parseDepKindAndProtocol :: Text.Text -> (PackageIdSourceKind, PackageIdSourceProtocol)
+    parseDepKindAndProtocol src = case breakOn "+" . Text.takeWhile (/= ':') $ src of
       (protocol, "") -> ("", protocol)
       (kind, protocol) -> (kind, protocol)
 
@@ -399,6 +405,8 @@ toDependency pkg =
       -- (_, "ssh") -> GitType
       _ -> CargoType
 
+    -- For a path dependency, use the path as the package name. For example:
+    -- path+file:///some/file/path -> /some/file/path
     depName =
       let sourceUrl = Text.drop 2 $ snd $ breakOn "//" $ pkgIdSource pkg
        in case depType of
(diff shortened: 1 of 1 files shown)


src/Strategy/Cargo.hs
```diff
@@ -385,6 +385,26 @@ toDependency pkg =
     applyLabel :: CargoLabel -> Dependency -> Dependency
     applyLabel (CargoDepKind env) = insertEnvironment env
 
+    parseDepKindAndProtocol :: Text.Text -> (Text.Text, Text.Text)
+    parseDepKindAndProtocol src = case breakOn "+" $ fst $ breakOn ":" src of
+      (protocol, "") -> ("", protocol)
+      (kind, protocol) -> (kind, protocol)
```
csasarak: Could you add some an example of input text for `parseDepKindAndProtocol`? It's not immediately clear why in one case you seem to flip the order of the pairs and in the other case do not. I think an example of both cases would make it super clear though.

I also strongly recommend making some type aliases for the output `Text` types. Right now, all the type for this function communicates to me is that it outputs a pair of`Text`s. The output of this function obviously doesn't have the semantics of arbitrary text though, even if its _representation_ in the computer is `Text`.

src/Strategy/Cargo.hs
```diff
@@ -385,6 +385,26 @@ toDependency pkg =
     applyLabel :: CargoLabel -> Dependency -> Dependency
     applyLabel (CargoDepKind env) = insertEnvironment env
 
+    parseDepKindAndProtocol :: Text.Text -> (Text.Text, Text.Text)
+    parseDepKindAndProtocol src = case breakOn "+" $ fst $ breakOn ":" src of
```
csasarak: [nit/optional] 

```
breakOn "+" . takeWhile (/= ':') $ src
```
_May_ be a bit clearer than the successive `breakOn`.

src/Strategy/Cargo.hs
```diff
@@ -385,6 +385,26 @@ toDependency pkg =
...
+    depType = case parseDepKindAndProtocol $ pkgIdSource pkg of
+      ("path", _) -> UnresolvedPathType
+      (_, "file") -> UnresolvedPathType
+      -- Using  a git type is probably more correct, but the git type doesn't have the ability
+      -- to track the package name within a repo, and this has poor support for monorepos.
+      -- ("git", _) -> GitType
+      -- (_, "ssh") -> GitType
+      _ -> CargoType
+
+    depName =
```
csasarak, line 410: I'd recommend a comment with some example input text for `depName`.

src/Strategy/Cargo.hs
```diff
@@ -415,17 +435,26 @@ buildGraph meta = stripRoot $
...
-  case Text.splitOn " " t of
-    [a, b, c] -> Right $ PackageId a b c
-    _ -> Left $ "malformed Package ID: " <> t
+oldPkgIdParser :: PkgSpecParser PackageId
+oldPkgIdParser = do
+  name <- takeWhile1P (Just "Package name") (/= ' ')
+  void $ char ' '
+  version <- takeWhile1P (Just "Package version") (/= ' ')
+  void $ char ' ' >> char '('
+  source <- takeWhile1P (Just "Package source") (/= ')')
```
csasarak: [Optional] You can use an `Applicative` style here if you like:

```haskell
  name <- takeWhile1P (Just "Package name") (/= ' ') <* char ' '
  version <- takeWhile1P (Just "Package version") (/= ' ') <* string " ("
  source <- takeWhile1P (Just "Package source") (/= ')')
```

src/Strategy/Cargo.hs
```diff
@@ -503,12 +532,6 @@ parsePkgSpec = eatSpaces (try longSpec <|> simplePkgSpec')
...
+parsePkgId t = either fail pure $ first errorBundlePretty parseResult
   where
-    oldPkgIdParser' = first toString . oldPkgIdParser
-
-    parseNewSpec :: Either String PackageId
-    parseNewSpec =
-      bimap errorBundlePretty (\p -> p{pkgIdSource = "(" <> p.pkgIdSource <> ")"})
-        . parse parsePkgSpec "Cargo Package Spec"
-        $ t
+    parseResult = parse (try oldPkgIdParser <|> try newPkgIdParser) "Cargo package spec" t
```
csasarak: [nit] You may not need the second `try` here: I think we only need the back-tracking behavior if the first alternative fails. 

If the caller wants to backtrack from `parseResult` they should do it themselves IMO.

PR #1601, opened 2025-10-15
Make pipenv strategy differentiate between direct and transitive deps

tjugdev:
# Overview
This PR fixes the pipenv strategy to differentiate between direct and transitive dependencies. We are able to make this distinction via static analysis, but generating edges requires the `pipenv graph` command, and is thus only available as a dynamic strategy. Neither `Pipfile.lock` nor `pipenv graph` adequately identify which dependences are direct, so I've introduced parsing of `Pipfile` in order to accurately make this distinction.

I also included a fix for some broken links in the Swift docs.

## Acceptance criteria
Pipenv correctly differentiates between direct and transitive dependencies.

## Testing plan
- Create a pipenv project and install the `requests` package. Confirm that the dependency graph looks correct
- Install the package `langchain` as well which itself also depends on `requests`. Confirm that `langchain` and `requests` are both reported as direct dependencies.
- Unit tests updated to test direct dependencies, including the second case above where a direct dependency is also a transitive dependency.

## References
- [ANE-1400](https://fossa.atlassian.net/browse/ANE-1400): Pipenv must differentiate between direct and transitive deps

## Checklist

[ANE-1400]: https://fossa.atlassian.net/browse/ANE-1400?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Control/Carrier/ContainerRegistryApi.hs
+++ b/src/Control/Carrier/ContainerRegistryApi.hs
@@ -190,6 +190,7 @@ getImageManifest src = context "Getting Image Manifest" $ do
           (digestOf manifestIndex platformArch)
 
       parseOciManifest
+        =<< fromResponse
         =<< mkRequest manager (registryCred src) (Just supportedManifestKinds)
         =<< (manifestEndpoint $ src{registryContainerRepositoryReference = manifestDigest})
     else do
--- a/src/Control/Carrier/ContainerRegistryApi/Errors.hs
+++ b/src/Control/Carrier/ContainerRegistryApi/Errors.hs
@@ -39,6 +39,7 @@ data ContainerRegistryApiErrorKind
   | Denied
   | Unsupported
   | OtherError
+  | TooManyRequests
   deriving (Eq, Ord)
 
 instance Show ContainerRegistryApiErrorKind where
@@ -49,6 +50,7 @@ instance Show ContainerRegistryApiErrorKind where
   show Unauthorized = "UNAUTHORIZED"
   show Denied = "DENIED"
   show Unsupported = "UNSUPPORTED"
+  show TooManyRequests = "TOOMANYREQUESTS"
   show OtherError = "OTHER_UNKNOWN_ERROR"
 
 instance FromJSON ContainerRegistryApiErrorKind where
@@ -61,6 +63,7 @@ errCodeToErrKind errorKind | show NameInvalid == errorKind = NameInvalid
 errCodeToErrKind errorKind | show NameUnknown == errorKind = NameUnknown
 errCodeToErrKind errorKind | show Unauthorized == errorKind = Unauthorized
 errCodeToErrKind errorKind | show Denied == errorKind = Denied
+errCodeToErrKind errorKind | show TooManyRequests == errorKind = TooManyRequests
 errCodeToErrKind errorKind | show Unsupported == errorKind = Unsupported
 errCodeToErrKind _ = OtherError
 
@@ -103,7 +106,7 @@ instance ToDiagnostic UnknownApiError where
     let header =
           renderIt $
             vsep
-              [ "Caught unexpected error from:" <> pretty ("(" <> show (statusCode status) <> ") " <> show uri)
+              [ "Caught unexpected error from: " <> pretty ("(" <> show (statusCode status) <> ") " <> show uri)
               ]
     Errata (Just header) [] Nothing
 
--- a/src/Strategy/Cargo.hs
+++ b/src/Strategy/Cargo.hs
@@ -76,7 +76,7 @@ import Effect.Grapher (
 import Effect.ReadFS (ReadFS, doesFileExist, readContentsToml)
 import Errata (Errata (..))
 import GHC.Generics (Generic)
-import Graphing (Graphing, stripRoot)
+import Graphing (Graphing, shrinkRoots)
 import Path (Abs, Dir, File, Path, mkRelFile, parent, parseRelFile, toFilePath, (</>))
 import Text.Megaparsec (
   Parsec,
@@ -434,7 +434,10 @@ addEdge node = do
     edge parentId $ nodePkg dep
 
 buildGraph :: CargoMetadata -> Graphing Dependency
-buildGraph meta = stripRoot $
+-- By construction, workspace members are the root nodes in the graph.
+-- Use shrinkRoots to remove them and promote their direct dependencies to the
+-- direct dependencies we report for the project.
+buildGraph meta = shrinkRoots $
   run . withLabeling toDependency $ do
     traverse_ direct $ metadataWorkspaceMembers meta
     traverse_ addEdge $ resolvedNodes $ metadataResolve meta
--- a/test/Cargo/MetadataSpec.hs
+++ b/test/Cargo/MetadataSpec.hs
@@ -10,7 +10,6 @@ import Data.Text (Text)
 import Data.Text qualified as Text
 import DepTypes
 import GraphUtil
-import Graphing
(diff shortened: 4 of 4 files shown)


src/Strategy/Python/Pipenv.hs
```diff
@@ -106,12 +117,12 @@ getDeps project = context "Pipenv" $ do
         . errHelp PipenvCmdFailedHelp
       $ execJson (parent (pipenvLockfile project)) pipenvGraphCmd
 
-  graph <- context "Building dependency graph" $ pure (buildGraph lock maybeDeps)
+  graph <- context "Building dependency graph" $ pure (buildGraph pipfile lock maybeDeps)
```
csasarak: [nit] It's older code, but I think you can make `graph` a `let` binding as it's not in `m` originally:

```haskell
let graph = buildGraph pipfile lock maybeDeps
```

src/Strategy/Python/Pipenv.hs
```diff
@@ -121,8 +132,9 @@ getDepsStatically ::
   PipenvProject ->
   m DependencyResults
 getDepsStatically project = context "Pipenv" $ do
-  lock <- context "Getting direct dependencies" $ readContentsJson (pipenvLockfile project)
-  graph <- context "Building dependency graph" $ pure (buildGraph lock Nothing)
+  lock <- context "Getting dependencies from Pipfile.lock" $ readContentsJson (pipenvLockfile project)
+  pipfile <- context "Getting dependencies from Pipfile" $ readContentsToml (pipenvPipfile project)
+  graph <- context "Building dependency graph" $ pure (buildGraph pipfile lock Nothing)
```
csasarak: Same comment re: using a `let` binding. I suppose the you could make the argument that the `context` is useful output for the customer, but I kind of don't think so. I'll leave both cases up to you.

src/Strategy/Python/Pipenv.hs
```diff
@@ -180,6 +193,9 @@ buildGraph lock maybeDeps = run . withLabeling toDependency $ do
             , dependencyTags = Map.empty
             }
 
+    -- We only have edges if we have graph dependencies. Only prune unreachable in this case
+    prune = if isJust maybeDeps then pruneUnreachable else id
```
csasarak, line 201: [nit] I'm not sure this guard is really worth it. If there are no graph dependencies `pruneUnreachable` can't really do anything can it?

src/Strategy/Python/Pipenv.hs
```diff
@@ -302,3 +319,25 @@ instance FromJSON PipenvGraphDep where
       <*> obj .: "installed_version"
       <*> obj .: "required_version"
       <*> obj .: "dependencies"
+
+---------- Pipfile
+
+data PipfileToml = PipfileToml
+  -- {pipfilePackages :: Map Text (), pipfileDevPackages :: Map Text ()}
```
csasarak: Is this just dead code?

src/Strategy/Python/Pipenv.hs
```diff
@@ -302,3 +319,25 @@ instance FromJSON PipenvGraphDep where
...
+  {pipfilePackages :: Map Text PipfilePackageVersion, pipfileDevPackages :: Map Text PipfilePackageVersion}
```
csasarak: If you haven't, could you run `make fmt-ci`? 

I forget if it's a rule, but we typically put each field definition on its own line.

test/Python/PipenvSpec.hs
```diff
@@ -57,15 +57,15 @@ pipfileLock =
           ]
     }
 
+pipfileToml :: PipfileToml
+pipfileToml =
+  PipfileToml
+    (pipfilePackageList ["pkgOne"])
+    (pipfilePackageList ["pkgTwo"])
```
csasarak, line 64: [Optional] We use records positionally a lot in the CLI. It quickly becomes unwieldy and hard to read when it's a record with many fields. If I wrote a function with seven arguments or something you'd probably, correctly, call it out in a review. Personally, I'd like it if we all moved towards being a bit more explicit.

There's even an action HLS has to do it [for you](https://teamfossa.slack.com/archives/CU3EPRS11/p1751318795964299).

This particular case looks fine to me if you want to keep it though, or disagree with the idea above.

src/Strategy/Python/Pipenv.hs
```diff
@@ -210,10 +226,11 @@ buildNodes PipfileLock{..} = do
...
       PipfileDep ->
       m ()
-    addWithEnv env sourcesMap depName dep = do
-      let pkg = PipPkg depName (Text.drop 2 <$> fileDepVersion dep)
-      -- TODO: reachable instead of direct
-      direct pkg
+    addWithEnv env sourcesMap name dep = do
+      let pkg = PipPkg name (Text.drop 2 <$> fileDepVersion dep)
+      -- Use the Pipfile as the source of truth for which dependencies are direct
+      let graphFn = if Map.member name pipfilePackages || Map.member name pipfileDevPackages then direct else deep
```
csasarak, line 236: [nit] It may be more straightforward to just inline this if expression.

src/Strategy/Python/Pipenv.hs
tjugdev: I do have the formatter set up and it seems to be happy with everything on one line in this case. But I'll change it anyway.

csasarak: Hmm you're right. It's fine then - only make the change if you think it's genuinely better.

src/Strategy/Python/Pipenv.hs
tjugdev: This is necessary: if there are no edges, then `pruneUnreachable` will remove all dependencies that aren't direct dependencies as there's no path to them.

csasarak: Is the way this happens that we build a depgraph where we don't know enough to label a dependency as direct, so it gets labeled as indirect/deep?

tjugdev: We can accurately distinguish between direct and deep dependencies, but without edges, when we try to look for paths from the direct dependencies to the deep dependencies there are none, and so all the deep dependencies get pruned leaving only the direct ones.

csasarak: It's been a while since I've been deep in this code so forgive me - but in what cases would we actually prune anything then? If a dep is not direct, then it is by definition deep, right? 

Or is the distinction here that we need to handle both graphs that have any edges at all (and these should be subject to pruning) but sometimes we'll get a graph with no edges at all in which case we don't want to attempt pruning?

tjugdev: > we need to handle both graphs that have any edges at all (and these should be subject to pruning) but sometimes we'll get a graph with no edges at all in which case we don't want to attempt pruning?

Exactly this. In the case of Pipenv, running with `--static-only-analysis` results in a graph with no edges and so we don't want to attemput pruning in this case.

csasarak: ~This function is called `getDepsStatically`, is that name appropriate when it sounds like it can get graphs from a non-static source?~

Would you mind adding some commentary around how the cases where we want to prune vs. the ones where we don't arise? Because this code is sort of operating at the level of the graph, I had trouble connecting it in my head to the business case it was handling.

tjugdev: The function where this is happening is called`buildGraph`. I will expand on the comment

PR #1604, opened 2025-10-24
Add uv support

tjugdev:
# Overview
Adds a new strategy for scanning uv projects. Everything is done via the `uv.lock` file making this a static strategy.

I opted not to set locations/sources on the dependencies as we don't currently use it. I could be persuaded to add it though.

## Acceptance criteria
Dependencies are correctly identified for uv projects

## Testing plan
- New unit tests and new integration test
- Manual testing with my own uv project
- Tested on a few public repos:
  - [fpgmaas/cookiecutter-uv](https://github.com/fpgmaas/cookiecutter-uv)
  - [commaai/openpilot](https://github.com/commaai/openpilot)

## References
- [ANE-2526](https://fossa.atlassian.net/browse/ANE-2526): Support uv for Python

## Checklist

[ANE-2526]: https://fossa.atlassian.net/browse/ANE-2526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Python/Poetry.hs
+++ b/src/Strategy/Python/Poetry.hs
@@ -115,8 +115,13 @@ findProjects = walkWithFilters' $ \dir _ files -> do
   let poetryLockFile = findFileNamed "poetry.lock" files
   let pyprojectFile = findFileNamed "pyproject.toml" files
 
-  case (poetryLockFile, pyprojectFile) of
-    (poetry, Just pyproject) -> do
+  -- TODO: It's possible to use poetry for the build backend along with uv for package management
+  -- As pyproject.toml is used by many package managers/build back ends, we need to handle this better
+  -- See https://fossa.atlassian.net/browse/ANE-2316
+  let uvlockFile = findFileNamed "uv.lock" files
+
+  case (poetryLockFile, pyprojectFile, uvlockFile) of
+    (poetry, Just pyproject, Nothing) -> do
       poetryProject <- readContentsToml pyproject
       let project = PoetryProject (ProjectDir dir) (PyProjectTomlFile pyproject) (PoetryLockFile <$> poetry)
       let pyprojectBuildBackend = getPoetryBuildBackend poetryProject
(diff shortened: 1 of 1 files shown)


src/Strategy/Python/PDM/Pdm.hs
```diff
@@ -49,9 +49,16 @@ findProjects :: (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters
...
   let pdmlockFile = findFileNamed "pdm.lock" files
-  case pyprojectFile of
-    Just pyprojectToml -> pure ([PdmProject pyprojectToml pdmlockFile dir], WalkSkipSome [".venv"])
-    Nothing -> pure ([], WalkContinue)
+
+  -- TODO: The existence if a pyproject.toml should not be enough to conclude this is a PDM project.
+  -- uv projects also use a pyproject.toml, and we don't want to incorrectly report them as PDM projects
+  -- See https://fossa.atlassian.net/browse/ANE-2316
+  let uvlockFile = findFileNamed "uv.lock" files
+  case uvlockFile of
```
spatten, line 57: Do we need to do the same thing for poetry projects? https://github.com/fossas/fossa-cli/blob/master/src/Strategy/Python/Poetry.hs/#L118

src/Strategy/Python/Uv.hs
```diff
@@ -0,0 +1,263 @@
...
+  DiscoveredProjectType (PipenvProjectType, UvProjectType),
+  GraphBreadth (Complete),
+ )
+
+discover ::
+  (Has ReadFS sig m, Has Diagnostics sig m, Has (Reader AllFilters) sig m) =>
+  Path Abs Dir -> m [DiscoveredProject UvProject]
+discover = simpleDiscover findProjects mkProject PipenvProjectType
+
+findProjects ::
```
spatten, line 72: I think it would be worthwhile to have a very stripped down example of a uv.lock file in here. Maybe one with just a few dependencies and one dev dependency.

src/Strategy/Python/PDM/Pdm.hs
tjugdev: The poetry strategy also checks that the build backend is poetry, but I suppose it's possible to use the poetry build backend with uv for package management meaning we should add a similar check there too. Nice catch.

PR #1610, opened 2025-11-05
Ficus debug logs

spatten:
# Overview

We want to capture Ficus's log output in the debug bundle, but storing it in memory to output to fossa.debug.json would really blow up memory usage on large projects.

So we are changing how debug bundles work. We will write Ficus's log output to two separate files (fossa.ficus-stderr.log and fossa.ficus-stdout.log), and include them plus fossa.debug.json and fossa.telemetry.json in a zip file called fossa.debug.zip.

fossa.debug.zip will be created when the --debug flag is passed to the three commands that currently support the debug bundle: 

- fossa analyze
- fossa container analyze
- fossa sbom analyze

If the `--debug` flag is not passed in, but `FOSSA_TELEMETRY_DEBUG` env variable is set to 1, then we will maintain the current behaviour of writing `fossa.telemetry.json` to the current working directory. If both `--debug` and `FOSSA_TELEMETRY_DEBUG` are set, then we will just write the `fossa.debug.zip` file.

## Acceptance criteria

- We create a `fossa.debug.zip` file when the `--debug` flag is used
- The zip file contains everything we currently output for debugging, plus the new Ficus logs
- We do not blow up memory usage when the `--debug` flag is used along with `--x-snippet-scan`

## Testing plan

We need to test using the three subcommands that have debug bundles.

Checkout this branch and run `make install-dev`.

### fossa analyze

Test with and without `--debug` and with and without `--x-snippet-scan`

You can inspect the contents of `fossa.debug.zip` with

```
unzip -v fossa.debug.zip
```

```
fossa-dev analyze
```

No `fossa.debug.zip` file should be created

```
fossa-dev --debug analyze
```

`fossa.debug.zip` should contain `fossa.debug.json` and `fossa.telemetry.json`

```
fossa-dev --x-snippet-scan --debug analyze
```

### fossa container analyze

```
fossa-dev container analyze alpine:3.14
```

No fossa.debug.zip file should be created.

```
fossa-dev container analyze --debug alpine:3.14

unzip -v fossa.debug.zip
Archive:  fossa.debug.zip
 Length   Method    Size  Cmpr    Date    Time   CRC-32   Name
--------  ------  ------- ---- ---------- ----- --------  ----
   63433  Defl:N    12137  81% 11-06-2025 22:52 38b338c8  fossa.debug.json
    2873  Defl:N     1166  59% 11-06-2025 22:52 6df2beae  fossa.telemetry.json
--------          -------  ---                            -------
   66306            13303  80%                            2 files
```

```
FOSSA_TELEMETRY_DEBUG=1 fossa-dev container analyze alpine:3.14
```

fossa.telemetry.json should be created

```
FOSSA_TELEMETRY_DEBUG=1 fossa-dev container analyze --debug alpine:3.14
```

fossa.debug.zip should be created.

### fossa sbom analyze

download an example SBOM from https://github.com/CortezFrazierJr/my_recipe_book/blob/main/sampleSPDX.json

In an empty directory:

```bash
fossa-dev sbom analyze --debug ~/Downloads/sampleSPDX.json
```

Follow the same variations that you did for `fossa container analyze`
## Risks

## Metrics

## References

_Example:_

## Checklist

### DIFF
--- a/src/App/Fossa/Ficus/Analyze.hs
+++ b/src/App/Fossa/Ficus/Analyze.hs
@@ -169,8 +169,8 @@ runFicus maybeDebugDir ficusConfig = do
     (stdoutFile, stderrFile) <- case maybeDebugDir of
       Just debugDir -> do
         sendIO $ do
-          let stdoutPath = debugDir </> "ficus-stdout.log"
-          let stderrPath = debugDir </> "ficus-stderr.log"
+          let stdoutPath = debugDir </> "fossa.ficus-stdout.log"
+          let stderrPath = debugDir </> "fossa.ficus-stderr.log"
           stdoutH <- openFile stdoutPath WriteMode
           stderrH <- openFile stderrPath WriteMode
           pure (Just stdoutH, Just stderrH)


src/App/Fossa/Ficus/Analyze.hs
```diff
@@ -159,19 +164,46 @@ runFicus ficusConfig = do
...
+          putStrLn $ "Teeing Ficus stderr to: " <> stderrPath
+          pure ((stdoutPath, stdoutH), (stderrPath, stderrH))
+        pure (Just stdoutTuple, Just stderrTuple)
+      Nothing ->
+        -- No debug mode, don't tee to files
+        pure (Nothing, Nothing)
+
     (result, exitCode, stdErrLines) <- sendIO $ withProcessWait processConfig $ \p -> do
-      getCurrentTime >>= \now -> putStrLn $ "[TIMING " ++ formatTime defaultTimeLocale "%H:%M:%S.%3q" now ++ "] Ficus process started, beginning stream processing..."
+      getCurrentTime >>= \now -> hPutStrLn stderr $ "[TIMING " ++ formatTime defaultTimeLocale "%H:%M:%S.%3q" now ++ "] Ficus process started, beginning stream processing..."
```
spatten, line 182: IMO, we should always have been writing these log lines to stdErr

src/App/Fossa/Config/Project/Edit.hs
```diff
@@ -130,11 +130,12 @@ projectIdentiferOptions = optional (projectIdOptions <|> projectLocatorOptions)
 
 mergeOpts ::
   (Has Diagnostics sig m) =>
+  Maybe FilePath ->
   Maybe ConfigFile ->
   EnvVars ->
   EditOpts ->
   m EditConfig
-mergeOpts maybeConfig envVars cliOpts@EditOpts{..} = do
+mergeOpts _ maybeConfig envVars cliOpts@EditOpts{..} = do
```
spatten, line 138: I'm sticking the extra `_` in here and in other subcommands that do not have a debug flag so that we keep the signature of `mergeOpts` the same in all subcommands.

src/App/Fossa/Ficus/Analyze.hs
```diff
@@ -91,9 +93,10 @@ analyzeWithFicus ::
   ProjectRevision ->
   Maybe LicenseScanPathFilters ->
   Maybe Int ->
+  Maybe FilePath -> -- Debug directory (if enabled)
```
csasarak, line 96: [nit] Cases like these are decent for type aliases: `type DebugBundleDir = FilePath`

src/App/Fossa/Ficus/Analyze.hs
```diff
@@ -187,42 +210,50 @@ runFicus ficusConfig = do
...
 
-    streamFicusOutput :: Handle -> IO (Maybe FicusSnippetScanResults)
-    streamFicusOutput handle =
+    streamFicusOutput :: Handle -> Maybe Handle -> IO (Maybe FicusSnippetScanResults)
+    streamFicusOutput handle maybeFile =
       Conduit.runConduit $
         CC.sourceHandle handle
           .| CC.decodeUtf8Lenient
           .| CC.linesUnbounded
+          .| CC.mapM
```
csasarak, line 219: [nit] This is [`iterMC`](https://hackage-content.haskell.org/package/conduit-1.3.6.1/docs/Conduit.html#v:iterMC) I think.

src/App/Fossa/DebugDir.hs
```diff
@@ -0,0 +1,23 @@
...
+  DebugDirRef,
+  newDebugDirRef,
+  readDebugDir,
+  writeDebugDir,
+) where
+
+import Data.IORef (IORef, newIORef, readIORef, writeIORef)
+
+-- | Type alias for the debug directory reference
+type DebugDirRef = IORef (Maybe FilePath)
```
csasarak, line 11: I'm not sure I understand how/why the `IORef` is needed. It seems like you pass the `IORef` around wherever it's needed.

spatten: It's because the telemetry sink needs to know the directory name so that it can write telemetry.json to it, but the telemetry sink is instantiated before runSubCommand creates the directory. I tried a few ways to make it happen without this, but nothing felt great.

I'm definitely open to a different solution, though

src/App/Fossa/Subcommand.hs
```diff
@@ -55,19 +75,47 @@ updateCommandName :: SubCommand cli cfg -> String -> SubCommand cli cfg
...
+            tmpDir <- Dir.getTemporaryDirectory
+            suffix <- UUID.toString <$> UUID.nextRandom
+            let uniqueName = "fossa-debug-bundle-" <> suffix
+            let dirName = tmpDir <> "/" <> uniqueName
+            Dir.createDirectoryIfMissing True dirName
+            writeDebugDir debugDirRef (Just dirName)
+            pure (Just dirName)
+          else do
+            writeDebugDir debugDirRef Nothing
+            pure Nothing
```
csasarak, line 99: IMO, for the `IORef` solution this functionality really belongs as a function in `App.Fossa.DebugDir`.

src/App/Fossa/DebugDir.hs
csasarak: Why is it the subcommand that creates the debug directory? 

I won't block anymore but just wanted to dig into it.

The reason I am being a pain in the neck about it is that now anything that wants to use the debug dir has to be concerned with what code was run before and a state where we haven't created a debug dir now is a valid state that anything using the debug directory has to consider. 

Of course, it's always possible to just lie to a function accepting a directory name, but hopefully the fact that a caller must supply one is enough of a guide in the right direction.

PR #1626, opened 2026-01-08
make_fossa_deps_conan.py conan 2.22 graph info compatability

TheIceman2706:
# Overview

This PR makes make_fossa_deps_conan.py work with conan 2.22.

Conan now wraps the graph in  a "graph" object and exports the nodes as named objects. Also some more information is directly available.

The changes in this PR correspond to the new structure and are actively used at Miele & Cie. KG in multiple CI pipelines to check license compliance.

## Acceptance criteria

When users use conan 2.22 for their package management, they can now integrate fossa into their workflow.

## Risks

This may break compatibility to earlier conan versions.

## Tests

Defining automated tests for this script is not feasible in the current test setup.

## References

- Direct Customer Support 

## Checklist

- [X] I added tests for this PR's change (or explained in the PR description why tests don't make sense).
- [X] If this PR introduced a user-visible change, I added documentation into `docs/`.
- [X] If this PR added docs, I added links as appropriate to the user manual's ToC in `docs/README.ms` and gave consideration to how discoverable or not my documentation is.
- [ ] If this change is externally visible, I updated `Changelog.md`. If this PR did not mark a release, I added my changes into an `## Unreleased` section at the top.
- [X] If I made changes to `.fossa.yml` or `fossa-deps.{json.yml}`, I updated `docs/references/files/*.schema.json` AND I have updated example files used by `fossa init` command. You may also need to update these if you have added/removed new dependency type (e.g. `pip`) or analysis target type (e.g. `poetry`).
- [X] If I made changes to a subcommand's options, I updated `docs/references/subcommands/<subcommand>.md`.

<hr/>
For questions please contact friedrich.may@miele.com (my work email). Submitted on behalf of Miele.

PR #1629, opened 2026-01-12
make_fossa_deps_conan.py conan 2.22 graph info compatability

csasarak:
# Overview

Working with the make_fossa_deps_conan.py script and conan 2.22 some errors were encountered. 

Conan now wraps the graph in  a "graph" object and exports the nodes as named objects. Also some more information is directly available.

The changes in this PR correspond to the new structure and are actively used at Miele & Cie. KG in multiple CI pipelines to check license compliance.

For questions please contact friedrich.may@miele.com (my work email). Submitted on behalf of Miele.

Dupe of #1626 to get around CI.

## Acceptance criteria

_If this PR is successful, what impact does it have on the user experience?_

_Example: When users do X, Y should now happen._

## Testing plan

_Example:_

1. _Set up test case X._

## Risks

## Metrics

## References

_Example:_

## Checklist


docs/walkthroughs/make_fossa_deps_conan.py
```diff
@@ -125,9 +125,10 @@ def mk_fossa_deps(graph):
 
     vendored_deps = []
     custom_deps = []
-    for node in graph.get('nodes', []):
+    for nodeName in graph.get('nodes', []):
+        node = graph['nodes'][nodeName]
```
spatten, line 129: This will break on older versions of conan, right? Do we want to support both versions?

I think we could do that by looking at the type of `graph.get('nodes', [])`. Something like this:

```py
  nodes_data = graph.get('nodes', [])
  if isinstance(nodes_data, list):
      nodes_iter = nodes_data  # Conan 1.x
  else:
      nodes_iter = nodes_data.values()  # Conan 2.x
```

It would be worth testing against both versions, IMO, to make sure that there aren't other breaking changes

GauravB159: @spatten Isn't the script only written to support Conan v2?

spatten: I thought there were differences between the JSON we got on different versions of Conan 2.X. But I did some investigation and this isn't true. Every released version of Conan 2.x has followed the shape that this PR updates us to work with. 

I'm not sure what version this was initially implemented against, but I think it must have been a pre-production version.

spatten (approved):

PR #1632, opened 2026-01-19
[ANE-2760] Fix pnpm v9 lockfile bug

tjugdev:
# Overview
We were not correctly reporting transitive dependencies in some cases for v9 pnpm lockfiles. This ended up being for two reasons:
1. In some situations package names have a peer dependency suffix (for example, `vite@5.4.17(@types/node@25.0.7)`) which needs to be removed in order to look it up in the `packages` section where the suffix is absent.
2. Any dependency needs to be consistently given the same environment. We had two ways of assigning the dev environment, and if these two ways gave different results for the same package, we'd end up with two separate nodes in the graph, each with a different environment. Generally one of these nodes would be dangling and would ultimately get pruned, leading to what appears to be missing dependencies. Fix this by always using the same logic to assign a dependency's environment.

## Acceptance criteria
The problematic repo described in the ticket now correctly reports its dependencies.

## Testing plan
- New unit tests added/updated
- The test repo linked in the ticket now reports the same dependencies for both the v6 and v9 lockfiles

## References
- [ANE-2760](https://fossa.atlassian.net/browse/ANE-2760): PNPM Lockfile v9 Issue

## Checklist

[ANE-2760]: https://fossa.atlassian.net/browse/ANE-2760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Node/Pnpm/PnpmLock.hs
+++ b/src/Strategy/Node/Pnpm/PnpmLock.hs
@@ -151,7 +151,7 @@ type SnapShotDepRev = Text
 --       dependencies:
 --       punycode: 2.3.1
 newtype PnpmLockFileSnapshots = PnpmLockFileSnapshots
-  {snapshots :: Map SnapshotDepName [(SnapshotDepName, SnapShotDepRev)]}
+  {snapshots :: HashMap.HashMap SnapshotDepName [(SnapshotDepName, SnapShotDepRev)]}
   deriving (Show, Eq, Ord, Semigroup, Monoid)
 
 instance FromJSON PnpmLockFileSnapshots where
@@ -165,7 +165,7 @@ instance FromJSON PnpmLockFileSnapshots where
 
       -- Remove the peer dependency suffix. It's present in the snapshot entry, but it's not present in packages
       -- section which is where we look the dependency up.
-      let snapshots' = (Map.mapKeys withoutPeerDepSuffix) . Map.fromList . HashMap.toList . toHashMapText $ snapshots
+      let snapshots' = (HashMap.mapKeys withoutPeerDepSuffix) . toHashMapText $ snapshots
       pure $
         PnpmLockFileSnapshots{snapshots = snapshots'}
 
@@ -333,7 +333,7 @@ buildGraph lockFile = withoutLocalPackages $
       let deepDependencies =
             Map.toList (dependencies pkgMeta)
               <> Map.toList (peerDependencies pkgMeta)
-              <> fromMaybe mempty (Map.lookup pkgKey lockFile.lockFileSnapshots.snapshots)
+              <> fromMaybe mempty (HashMap.lookup pkgKey lockFile.lockFileSnapshots.snapshots)
 
       let (depName, depVersion) = case getPkgNameVersion pkgKey of
             Nothing -> (pkgKey, Nothing)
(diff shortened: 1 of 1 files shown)


src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -162,7 +162,10 @@ instance FromJSON PnpmLockFileSnapshots where
               deps <- ds .:? "dependencies" .!= mempty
               pure . HashMap.toList $ deps
       snapshots <- traverse readTransitiveDepPairs o
-      let snapshots' = Map.fromList . HashMap.toList . toHashMapText $ snapshots
+
+      -- Remove the peer dependency suffix. It's present in the snapshot entry, but it's not present in packages
+      -- section which is where we look the dependency up.
+      let snapshots' = (Map.mapKeys withoutPeerDepSuffix) . Map.fromList . HashMap.toList . toHashMapText $ snapshots
```
csasarak: Could you use [HashMap.mapKeys](https://hackage-content.haskell.org/package/unordered-containers-0.2.21/docs/Data-HashMap-Internal.html#v:mapKeys) directly here?

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -417,17 +420,30 @@ buildGraph lockFile = withoutLocalPackages $
...
+    isPnpm9ProjectDep :: Text -> ProjectMap -> Bool
+    isPnpm9ProjectDep name ProjectMap{directDependencies} = Map.member name directDependencies
+
+    -- Returns true if a dependency is either absent or a dev dependency in each workspace. If it is a
+    -- non-dev dependency in any workspace, then this function returns false. Also returns false if the
+    -- lockfile is not PNPM v9
+    isPnpm9Dev :: Text -> Bool
+    isPnpm9Dev depName =
+      (lockFile.lockFileVersion == PnpmLockV9)
+        && (any ((isPnpm9ProjectDev depName) . snd) (toList lockFile.importers))
```
csasarak: `any` is implemented for `Foldable` - so you shouldn't need to turn `lockFile.importers` into a list first or use `snd`. 

I think this should work: `any (isPnpm9ProjectDev depName) lockFile.importers`

src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -481,3 +488,12 @@ buildGraph lockFile = withoutLocalPackages $
...
     withoutLocalPackages = Graphing.shrink (\dep -> dependencyType dep /= UserType)
+
+-- Sometimes package versions include resolved peer dependency version
+-- in parentheses. This is used by pnpm for dependency resolution, we do
+-- not care about them, as they do not represent package version.
+--
+-- >> withoutPeerDepSuffix "1.2.0" = "1.2.0"
+-- >> withoutPeerDepSuffix "1.2.0(babel@1.0.0)" = "1.2.0"
+withoutPeerDepSuffix :: Text -> Text
+withoutPeerDepSuffix version = fst $ Text.breakOn "(" version
```
csasarak: [nit] eta-reduce: `withoutPeerDepSuffix = fst . Text.breakOn "("`

PR #1643, opened 2026-02-17
[ANE-2655] Expose yarn and npm workspace packages as individual build targets

jagonalez:
## Summary

- Yarn and npm workspace packages (including the root) now appear as individual build targets (e.g. `yarn@./:my-monorepo`, `yarn@./:app`, `npm@./:api`)
- Users can see workspace packages via `fossa list-targets` and filter with `.fossa.yml`
- When specific targets are selected, only those packages' dependencies are included in the analysis
- When no filtering is applied, all targets are selected and all dependencies from every workspace package are included
- No dependencies are silently dropped — root deps are available on the root target

## Changes

- `src/Strategy/Node.hs`: `mkProject` populates `projectBuildTargets` for all workspace project types (yarn + npm). `findWorkspaceBuildTargets` includes root package name alongside workspace members. `getDeps` passes `FoundTargets` through to both `analyzeYarn` and `analyzeNpmLock`. `extractDepListsForTargets` scopes `FlatDeps` to selected packages.
- `test/Node/NodeSpec.hs`: Tests for `findWorkspaceBuildTargets` (root + members, single-package) and `extractDepListsForTargets` (all deps, scoped, root-only, all targets selected)
- `docs/references/strategies/languages/nodejs/yarn.md`: Documents workspace build target behavior
- `docs/references/strategies/languages/nodejs/npm-lockfile.md`: Documents workspace build target behavior with lockfile version caveat
- `Changelog.md`: Unreleased entry

## Test plan

- [x] `cabal build` succeeds
- [x] `cabal test unit-tests` — all 8 Node tests pass
- [x] `hlint` and `fourmolu` clean
- [x] Manual test: `fossa list-targets` shows root + workspace member targets for both yarn and npm workspaces
- [x] Manual test: unfiltered `fossa analyze -o` dep counts match released fossa 3.15.7
- [x] Verified no dependencies are dropped when no filtering is applied

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`.
+- Project edit: Fix 500 error when running `fossa project edit --policy` on existing projects ([#1688](https://github.com/fossas/fossa-cli/pull/1688))
 
 ## 3.17.0
 
--- a/spectrometer.cabal
+++ b/spectrometer.cabal
@@ -661,6 +661,7 @@ test-suite unit-tests
     Extra.TextSpec
     Ficus.FicusSpec
     Fortran.FpmTomlSpec
+    Fossa.API.CoreTypesSpec
     Fossa.API.TypesSpec
     Go.GlideLockSpec
     Go.GoListPackagesSpec
--- a/src/App/Fossa/Project.hs
+++ b/src/App/Fossa/Project.hs
@@ -27,4 +27,4 @@ projectMain ::
   m ()
 projectMain (EditCfg config) = do
   logInfo "Running FOSSA project"
-  context "Add projects to release group" . runStickyLogger SevInfo . ignoreDebug . runFossaApiClient (Edit.apiOpts config) $ editMain config
+  context "Editing project" . runStickyLogger SevInfo . ignoreDebug . runFossaApiClient (Edit.apiOpts config) $ editMain config
--- a/src/Fossa/API/CoreTypes.hs
+++ b/src/Fossa/API/CoreTypes.hs
@@ -27,11 +27,12 @@ import Data.Aeson (
   FromJSON (parseJSON),
   KeyValue ((.=)),
   ToJSON (toJSON),
-  Value (Null),
+  Value (Null, Object),
   object,
   withObject,
   withText,
   (.:),
+  (.?=),
  )
 import Data.Text (Text)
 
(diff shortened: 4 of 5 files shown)


src/Strategy/Node.hs
```diff
@@ -152,39 +157,55 @@ mkProject project = do
     DiscoveredProject
       { projectType = typename
       , projectPath = parent rootManifest
-      , projectBuildTargets = ProjectWithoutTargets
+      , projectBuildTargets = findWorkspaceBuildTargets graph
```
spatten: I think there's a subtle bug here.

`mkProject` calls `findWorkspaceBuildTargets graph` for all project types, but `getDeps` only passes targets through to the analysis for Yarn and NPMLock:

```
  getDeps targets (Yarn yarnLockFile graph) = analyzeYarn targets yarnLockFile graph
  getDeps targets (NPMLock packageLockFile graph) = analyzeNpmLock targets packageLockFile graph
  getDeps _ (Pnpm pnpmLockFile _) = analyzePnpmLock pnpmLockFile   -- targets discarded
  getDeps _ (NPM graph) = analyzeNpm graph                          -- targets discarded
```

This means that pnpm workspace projects would show per-package targets in `fossa list-targets`, and users could exclude specific packages in `.fossa.yml`, but the analysis would silently include dependencies from all packages regardless of the selection.

I haven't tested, but I think that the fix is to scope `findWorkspaceBuildTargets` to only the project types that honor it:

```
  projectBuildTargets = case project of
    Yarn _ _    -> findWorkspaceBuildTargets graph
    NPMLock _ _ -> findWorkspaceBuildTargets graph
    _           -> ProjectWithoutTargets
```

src/Strategy/Node.hs
```diff
@@ -152,39 +157,55 @@ mkProject project = do
...
+-- package name (including the root) becomes a 'BuildTarget'. If there
+-- are no workspace children (single-package project), returns
+-- 'ProjectWithoutTargets'.
+findWorkspaceBuildTargets :: PkgJsonGraph -> FoundTargets
+findWorkspaceBuildTargets graph@PkgJsonGraph{..} =
+  let WorkspacePackageNames childNames = findWorkspaceNames graph
+   in if Set.null childNames
+        then ProjectWithoutTargets
+        else
+          let rootName = findWorkspaceRootManifest graph >>= \m -> maybe (Left "no name") Right (packageName =<< Map.lookup m jsonLookup)
```
spatten, line 184: If the root `package.json` has no name field, the root is silently excluded from the target set:

```
  allNames = case rootName of
    Right n -> Set.insert n childNames
    Left _ -> childNames          -- root excluded
```

  Then in `extractDepListsForTargets` (on line 309), the filter skips packages without a matching target name:

```
  filter (maybe False (`Set.member` targetNames) . packageName)
```

  So the nameless root's deps are dropped. This wouldn't just affect explicit user filtering — when no `.fossa.yml` filters are configured, we pass the full `FoundTargets` set
  (via `filterFoundTargets ResultAll targets = Just targets` in `Discovery.Filters`), so the root's deps would be silently lost even with default settings.

Workspace roots almost always have a name field in practice, but this is a silent data loss when they don't. We could either log a warning when the root name is missing, or fall back
   to `ProjectWithoutTargets` in that case so no deps are dropped.

PR #1648, opened 2026-02-20
add bun support

jagonalez:
# Overview

PR adds support for analyzing bun projects, we'll eventually need support for fossabot and it looks like there's a few requests from customer for support.

## Acceptance criteria

- add support for removing JSONC comments and commas
- add support for bun v1.2 and later projects
- supports workspaces - but not as build targets - I'll come back to this after my PR for npm/yarn workspaces

## Testing plan

- [x] tests pass
- [x] analyze a bun project and deps are discovered

## Risks

- mostly just JSONC file parsing - but that would only affect bun projects

## Metrics

## References

_Example:_

## Checklist

### DIFF
--- a/src/Strategy/Node/Bun/BunLock.hs
+++ b/src/Strategy/Node/Bun/BunLock.hs
@@ -250,7 +250,7 @@ buildGraph lockfile = run . withLabeling vertexToDependency $ do
         <> Map.keys (pkgDepsOptionalDependencies $ pkgDeps pkg)
         <> Map.keys (pkgDepsPeerDependencies $ pkgDeps pkg)
 
-    -- | Convert a package to a vertex (environment-agnostic).
+    -- \| Convert a package to a vertex (environment-agnostic).
     --
     -- Returns Nothing for unsupported resolution types (workspace, file, link,
     -- tarball, root, module). Only npm and git/github packages are included.
@@ -259,7 +259,7 @@ buildGraph lockfile = run . withLabeling vertexToDependency $ do
       let (name, version) = parseResolution (pkgResolution pkg)
        in resolutionToVertex name version
 
-    -- | Convert a parsed resolution to a vertex based on the version prefix.
+    -- \| Convert a parsed resolution to a vertex based on the version prefix.
     -- Only npm (no prefix) and git resolutions produce vertices.
     resolutionToVertex :: Text -> Text -> Maybe BunDepVertex
     resolutionToVertex name version = case stripGitPrefix version of
@@ -268,7 +268,7 @@ buildGraph lockfile = run . withLabeling vertexToDependency $ do
         | isUnsupportedRef version -> Nothing
         | otherwise -> Just $ mkVertex NodeJSType name version
 
-    -- | Strip a known git hosting prefix from a version string.
+    -- \| Strip a known git hosting prefix from a version string.
     stripGitPrefix :: Text -> Maybe Text
     stripGitPrefix v =
       Text.stripPrefix "github:" v
@@ -276,14 +276,14 @@ buildGraph lockfile = run . withLabeling vertexToDependency $ do
         <|> Text.stripPrefix "bitbucket:" v
         <|> Text.stripPrefix "git+" v
 
-    -- | Build a GitType vertex from a git reference like @"user/repo#ref"@
+    -- \| Build a GitType vertex from a git reference like @"user/repo#ref"@
     -- or @"https://github.com/user/repo.git#ref"@.
     mkGitVertex :: Text -> BunDepVertex
     mkGitVertex ref =
       let (repo, refPart) = Text.breakOn "#" ref
        in mkVertex GitType repo (Text.drop 1 refPart)
 
-    -- | Check if a version string refers to a local/unsupported resolution type.
+    -- \| Check if a version string refers to a local/unsupported resolution type.
     isUnsupportedRef :: Text -> Bool
     isUnsupportedRef v =
       "workspace:" `Text.isPrefixOf` v


docs/references/strategies/languages/nodejs/nodejs.md
```diff
@@ -1,10 +1,11 @@
...
 
-The nodejs buildtool ecosystem consists of three major toolchains: the `npm` cli, `pnpm` and `yarn`.
+The nodejs buildtool ecosystem consists of four major toolchains: the `npm` cli, `pnpm`, `yarn`, and `bun`.
 
 | Strategy                      | Direct Deps        | Transitive Deps    | Edges              | Container Scanning |
 | ----------------------------- | ------------------ | ------------------ | ------------------ | ------------------ |
 | [yarnlock](yarn.md)           | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | [npmlock](npm-lockfile.md)    | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
 | [pnpmlock](pnpm.md)           | :white_check_mark: | :white_check_mark: | :white_check_mark: | :white_check_mark: |
+| [bunlock](bun.md)             | :white_check_mark: | :white_check_mark: | :white_check_mark: | :x:                |
```
jagonalez, line 10: technically bun is a also a javascript runtime but the bun cli can be used as a nodejs package manager so that's why I've added it here

src/Strategy/Node/Bun/BunLock.hs
```diff
@@ -0,0 +1,247 @@
...
+parseResolution :: Text -> (Text, Text)
+parseResolution res
+  | "@" `Text.isPrefixOf` res =
+      -- Scoped package: @scope/name@version
+      let withoutAt = Text.drop 1 res
+          (scopeAndName, rest) = Text.breakOn "@" withoutAt
+       in ("@" <> scopeAndName, Text.drop 1 rest)
+  | otherwise =
+      let (name, rest) = Text.breakOn "@" res
+       in (name, Text.drop 1 rest)
```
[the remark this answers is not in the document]
csasarak: They aren't disallowed. They're just discouraged in cases where they get particularly complex which this isn't.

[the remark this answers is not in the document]
jagonalez: Yes update your understanding - guards are only discouraged when they become complex

docs/references/strategies/languages/nodejs/bun.md
```diff
@@ -0,0 +1,60 @@
...
+- **npm:** `["name@version", "registry", {deps}, "integrity"]`
+- **file:** `["name@file:path", {deps}]`
+- **workspace:** `["name@workspace:path"]`
+- **git:** `["name@git+url", {deps}, "hash", "integrity"]`
+
+### Environment Labeling
+
+- Dependencies declared in `devDependencies` of any workspace are labeled as development dependencies.
+- Dependencies declared in `dependencies` or `optionalDependencies` of any workspace are labeled as production dependencies.
+- Workspace packages themselves (those with `workspace:` resolutions) are excluded from the final dependency graph.
```
csasarak: I don't know what a workspace dependency is - what's the justification for excluding them? Or is this something that anyone who knows Bun will immediately understand/think is reasonable?

src/Effect/ReadFS.hs
```diff
@@ -366,6 +368,17 @@ readContentsJson file = context ("Parsing JSON file '" <> toText (toString file)
...
+-- | Read JSONC (JSON with Comments) from a file.
+-- Strips single-line comments, block comments, and trailing commas before parsing as JSON.
+readContentsJsonc :: (FromJSON a, Has ReadFS sig m, Has Diagnostics sig m) => Path Abs File -> m a
+readContentsJsonc file = context ("Parsing JSONC file '" <> toText (toString file) <> "'") $ do
+  contents <- readContentsText file
+  case stripJsonc contents of
+    Left err -> errSupport (fileParseErrorSupportMsg file) $ fatal $ FileParseError (toString file) (toText err)
+    Right stripped -> case eitherDecodeStrict (encodeUtf8 stripped) of
+      Left err -> errSupport (fileParseErrorSupportMsg file) $ fatal $ FileParseError (toString file) (toText err)
+      Right a -> pure a
```
csasarak:
```suggestion
  case eitherDecodeStrict . encodeUtf8 =<< stripJsonc contents of
    Left err -> errSupport (fileParseErrorSupportMsg file) $ fatal $ FileParseError (toString file) (toText err)
    Right a -> pure a
```

src/Strategy/Node/Bun/BunLock.hs
```diff
@@ -0,0 +1,249 @@
...
+import Graphing (Graphing)
+import Graphing qualified
+import Path (Abs, File, Path)
+
+-- | Bun lockfile (bun.lock) in JSONC format.
+--
+-- See @docs/references/strategies/languages/nodejs/bun.md@ for the full
+-- lockfile format documentation.
+data BunLockfile = BunLockfile
+  { lockfileVersion :: Int
```
csasarak, line 55: Is the lockfile version actually used meaningfully here? I can see wanting to include it in something like a debug bundle. 

I'd add a comment saying that this was implemented for version X or X-whatever so in the future if it changes we can get a better sense of what's new/old.

src/Strategy/Node/Bun/BunLock.hs
```diff
@@ -0,0 +1,249 @@
...
+  , workspaces :: Map Text BunWorkspace
+  , packages :: Map Text BunPackage
+  }
+  deriving (Show, Eq)
+
+data BunWorkspace = BunWorkspace
+  { wsName :: Text
+  , wsDependencies :: Map Text Text
```
csasarak: You don't have to use `newtype` (though this is generally preferred to `type`) but I think some type aliases explaining what these `Text` fields are would be helpful. They clearly aren't just opaque text even if that's how we choose to represent them in this structure.

src/Strategy/Node/Bun/BunLock.hs
```diff
@@ -0,0 +1,249 @@
...
+--      and create edges to its transitive dependencies.
+--   4. Filter out workspace packages from the final graph.
+buildGraph :: BunLockfile -> Graphing Dependency
+buildGraph lockfile = withoutWorkspacePackages . run . evalGrapher $ do
+  for_ allWorkspaces $ \workspace -> do
+    markDirectDeps EnvProduction workspace.wsDependencies
+    markDirectDeps EnvDevelopment workspace.wsDevDependencies
+    markDirectDeps EnvProduction workspace.wsOptionalDependencies
+
+  for_ (Map.toList $ packages lockfile) $ \(_, pkg) ->
```
csasarak: [nit] Maps are Foldable so I think this toList is unnecessary:


```suggestion
  for_ (packages lockfile) $ \pkg ->
```

src/Strategy/Node/Bun/BunLock.hs
```diff
@@ -0,0 +1,249 @@
...
+    unless (isWorkspaceRef $ pkgResolution pkg) $ do
```
csasarak: [nit] Personally, I find `unless` to be a bit jarring just because unlike most control flow idioms is not `idiom (true thing) <action>`. The equivalent in Haskell in the other direction is `when`. Yes, you'll have to negate but I think the explicitness is nice.

test/Bun/BunLockSpec.hs
```diff
@@ -0,0 +1,236 @@
...
+-- are stripped correctly and the result parses as a valid bun lockfile.
+jsoncSpec :: Path Abs File -> Spec
+jsoncSpec path =
+  describe "jsonc" $ do
+    it' "parses bun.lock with comments and trailing commas" $ do
+      lockfile <- parseBunLockEffect path
+      lockfileVersion lockfile `shouldBe'` 1
+      wsName (workspaces lockfile Map.! "") `shouldBe'` "jsonc-test"
+      wsDependencies (workspaces lockfile Map.! "") `shouldBe'` Map.fromList [("lodash", "^4.17.21")]
+
```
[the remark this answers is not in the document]
csasarak: Normally I would agree. In tests we care less about partial functions because if it doesn't work, that's a test failure which will stop release/merge.

test/Bun/BunLockSpec.hs
```diff
@@ -0,0 +1,237 @@
...
+        it "creates transitive edges" $
+          expectEdge graph (mkDevDep "@lezer/cpp" "1.1.3") (mkDevDep "@lezer/common" "1.3.0")
+
+        it "excludes workspace packages from graph" $ do
+          let names = map dependencyName (Graphing.vertexList graph)
+          names `shouldNotContain` "@types/bun"
+          names `shouldNotContain` "bun-types"
+
+-- | Parse a bun.lock in the effect stack (for it' tests).
```
csasarak: As best I can tell the only real thing this function does is tell the type checker that `readContentsJsonc` should parse to `BunLockfile`. 

Consider removing the effect language since it doesn't really apply here in the same way that `Diagnostics` or `ReadFS` are effects.

Additionally, you can do this quite easily at call-sites: `readContentsJsonc @BunLockfile ...`

csasarak (approved):

docs/references/strategies/languages/nodejs/bun.md
jagonalez: It's similar to the yarn and pnpm strategy - we're filtering out any dependencies that are a workspace. a workspace package is just a dependency that is prefixed with `workspace:*`, so `"bun/types": "workspace:..."` would not be included as a rep

csasarak: oh gotcha. I was thinking of something different.

PR #1664, opened 2026-03-10
CI Improvements - Parallelize integration tests, cache ghcup, fix tag caching

zlav:
# Overview

Four CI improvements across integration tests, build-all, and benchmarks workflows.

Impact
- Build times when fully cached are reliably in the single digits, as low as 5m.
- Integration tests can be more aggressively added to without adding more absolute test run time.

## 1. Parallelize integration tests

Split the single integration test job into a build job + 3 parallel test matrix jobs:
- **jvm**: Maven, Gradle, Scala, Clojure
- **compiled**: Go, Rust, Erlang, Elixir
- **scripting-and-other**: Python, Ruby, Swift, Cocoapods, Pnpm, Nuget, Carthage, LicenseScanner, Container, Reachability

Build job compiles once and populates caches. Test jobs restore caches and run their subset via `--match`. Adds a concurrency group to cancel stale in-progress runs on the same branch.

A `TEST_GROUPS` env var at the workflow level lists all match patterns. After building, a validation step runs `--dry-run` to count total tests vs matched tests and fails if any test is uncovered by a group. This prevents new tests from not being added to the matrix.

Vendor binaries and Rust release binaries are uploaded as artifacts from the build job and downloaded in test jobs, avoiding redundant `cargo build --release` (~1m) and `vendor_download.sh` (~20s) per test job.

The build job exports the release tag via job outputs so test jobs use the same tag for cache keys (avoids a race if a release happens mid-run).

**Note** This update also stops previous integration tests from running if a new commit is pushed. We have limited runner space, and I've been seeing a lot of tests in the "Waiting to run this spec" state. This should fix that.

## 2. Cache ghcup in build-all.yml

Cache `~/.ghcup/{bin,ghc,cache}` on macOS and Windows. Skips the ~500 MB GHC download on cache hit, reducing the `Setup ghc/cabal` step from ~4m to ~30s. Cache key includes both GHC and cabal versions.

## 3. Fix release tag resolution

`git describe --tags --abbrev=0` requires the tagged commit to be reachable in local history. With `fetch-depth: 2`, the tag is never reachable and always resolves to `none`. Switched to `git tag --sort=-v:refname | head -1` which lists fetched tag refs regardless of shallow clone depth.

## 4. Upgrade actions/cache to v5

Upgrade `actions/cache` from v4 to v5 across all workflows (integration tests, build-all, benchmarks). v5 uses the node22 runtime. v4 uses node20 which GitHub is deprecating (forced migration to node24 starts June 2, 2026). Self-hosted runner is on v2.332.0, above the v2.327.1 minimum.

## Testing plan

1. Pushed to branch, verified all 3 test matrix jobs pass.
2. Verified cache hits on subsequent push (exact key match, no recompilation beyond `App.Version.TH` chain).
3. Verified ghcup cache hit on macOS — `Setup ghc/cabal` dropped from ~4m to ~30s.

## Risks

- Test groups may become unbalanced as tests are added/removed. Rebalance by moving patterns between groups.
- The `--dry-run` grep pattern (`grep -c "^  "`) depends on hspec's output format. If hspec changes its dry-run output, the validation step may need updating.

## Checklist

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -41,16 +41,19 @@ jobs:
           os-name: macOS-intel
           project-file: cabal.project.ci.macos
           ghc: '9.8.4'
+          cabal: '3.10.3.0'
 
         - os: windows-latest
           os-name: Windows
           project-file: cabal.project.ci.windows
           ghc: '9.8.4'
+          cabal: '3.10.3.0'
 
         - os: macos-latest
           os-name: macOS-arm64
           project-file: cabal.project.ci.macos
           ghc: '9.8.4'
+          cabal: '3.10.3.0'
 
     steps:
 
@@ -62,7 +65,7 @@ jobs:
 
     - name: Get latest release tag
       id: latest-tag
-      run: echo "tag=$(git describe --tags --abbrev=0 2>/dev/null || echo none)" >> $GITHUB_OUTPUT
+      run: echo "tag=$(git tag --sort=-v:refname | head -1 2>/dev/null || echo none)" >> $GITHUB_OUTPUT
 
     - name: Install MacOS binary dependencies
       if: ${{ contains(matrix.os, 'macos') }}
@@ -70,13 +73,24 @@ jobs:
         brew install jq
 
     # Set up Haskell.
+    # Cache ghcup installations so the setup action skips downloading GHC (~500 MB).
+    - uses: actions/cache@v5
+      if: ${{ !contains(matrix.os-name, 'Linux') }}
+      name: Cache ghcup
+      with:
+        path: |
+          ~/.ghcup/bin
+          ~/.ghcup/ghc/${{ matrix.ghc }}
+          ~/.ghcup/cache
+        key: ${{ matrix.os-name }}-ghcup-${{ matrix.ghc }}-${{ matrix.cabal }}
+
     - uses: haskell-actions/setup@v2
       id: setup-haskell
       name: Setup ghc/cabal (non-alpine)
       if: ${{ !contains(matrix.os-name, 'Linux') }}
       with:
         ghc-version: ${{ matrix.ghc }}
-        cabal-version: '3.10.3.0'
+        cabal-version: ${{ matrix.cabal }}
 
     # Set up Rust.
     # This action installs the 'minimal' profile.
@@ -129,7 +143,7 @@ jobs:
     # The home directory inside a github container run during a step is different than one run outside of it.
     # This is why there is special logic for 'LinuxARM'.
     # Its builds run inside a container but are cached by an action outside of it.
-    - uses: actions/cache@v4
+    - uses: actions/cache@v5
       name: Cache cabal store
       with:
         path: ${{ steps.setup-haskell.outputs.cabal-store || ( matrix.os == 'LinuxARM' && format('{0}/_github_home/.local/state/cabal', runner.temp) || '~/.local/state/cabal') }}
@@ -137,7 +151,7 @@ jobs:
         restore-keys: |
           ${{ matrix.os-name }}-${{ matrix.ghc }}-cabal-cache-
 
-    - uses: actions/cache@v4
+    - uses: actions/cache@v5
       name: Cache dist-newstyle
       with:
         path: ${{ github.workspace }}/dist-newstyle
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -78,9 +78,9 @@ jobs:
       name: Cache ghcup
       with:
         path: |
-          ~/.ghcup/bin
-          ~/.ghcup/ghc/${{ matrix.ghc }}
-          ~/.ghcup/cache
+          ${{ runner.os == 'Windows' && 'C:\ghcup\bin' || '~/.ghcup/bin' }}
+          ${{ runner.os == 'Windows' && format('C:\ghcup\ghc\{0}', matrix.ghc) || format('~/.ghcup/ghc/{0}', matrix.ghc) }}
+          ${{ runner.os == 'Windows' && 'C:\ghcup\cache' || '~/.ghcup/cache' }}
         key: ${{ matrix.os-name }}-ghcup-${{ matrix.ghc }}-3.10.3.0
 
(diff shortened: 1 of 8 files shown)


.github/workflows/build-all.yml
```diff
@@ -62,14 +62,25 @@ jobs:
...
+    - uses: actions/cache@v5
+      if: ${{ !contains(matrix.os-name, 'Linux') }}
+      name: Cache ghcup
+      with:
+        path: |
+          ~/.ghcup/bin
+          ~/.ghcup/ghc/${{ matrix.ghc }}
+          ~/.ghcup/cache
+        key: ${{ matrix.os-name }}-ghcup-${{ matrix.ghc }}
+
```
zlav, line 85: Done

.github/workflows/integrations-test.yml
```diff
@@ -98,8 +100,103 @@ jobs:
...
-        cabal test --project-file=cabal.project.ci.linux --test-show-details=direct --test-option=--times integration-tests --test-options="+RTS -N -RTS"
+        cabal update
+        MATCH_ARGS=""
+        for pattern in ${{ matrix.matches }}; do
+          MATCH_ARGS="$MATCH_ARGS --test-option=--match --test-option=$pattern"
+        done
+        cabal test --project-file=cabal.project.ci.linux \
+          --test-show-details=direct \
+          --test-option=--times \
+          $MATCH_ARGS \
```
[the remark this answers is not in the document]
zlav: Rejected. cabal update is required — the Hackage index isn't in the cached paths and cabal test fails
   without it.

.github/workflows/integrations-test.yml
```diff
@@ -98,8 +96,103 @@ jobs:
...
+
+    - uses: actions/checkout@v4
+      with:
+        lfs: true
+        fetch-depth: 2
+        fetch-tags: true
+
+    - name: Get latest release tag
+      id: latest-tag
+      run: echo "tag=$(git tag --sort=-v:refname | head -1 2>/dev/null || echo none)" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
zlav: Edge case, but fixed

.github/workflows/build-all.yml
```diff
@@ -54,22 +54,35 @@ jobs:
...
       with:
         lfs: true
         fetch-depth: 2
         fetch-tags: true
 
     - name: Get latest release tag
       id: latest-tag
-      run: echo "tag=$(git describe --tags --abbrev=0 2>/dev/null || echo none)" >> $GITHUB_OUTPUT
+      run: |
+        tag="$(git tag --sort=-v:refname 2>/dev/null | head -1 || true)"
```
csasarak, line 66: Why not `git describe`? It seems tailor made for this: https://git-scm.com/docs/git-describe

.github/workflows/build-all.yml
```diff
@@ -54,22 +54,35 @@ jobs:
...
+    # Cache ghcup installations so the setup action skips downloading GHC (~500 MB).
+    - uses: actions/cache@v5
+      if: ${{ !contains(matrix.os-name, 'Linux') }}
+      name: Cache ghcup
+      with:
+        path: |
+          ${{ runner.os == 'Windows' && 'C:\ghcup\bin' || '~/.ghcup/bin' }}
+          ${{ runner.os == 'Windows' && format('C:\ghcup\ghc\{0}', matrix.ghc) || format('~/.ghcup/ghc/{0}', matrix.ghc) }}
+          ${{ runner.os == 'Windows' && 'C:\ghcup\cache' || '~/.ghcup/cache' }}
+        key: ${{ matrix.os-name }}-ghcup-${{ matrix.ghc }}-3.10.3.0
```
csasarak, line 84: What is this `3.10.3.0` at the end? The version of cabal used? I think it should be in a variable which names it.

.github/workflows/integrations-test.yml
```diff
@@ -1,41 +1,58 @@
...
+  cancel-in-progress: true
+
+# Single source of truth for integration test groups. Used by both the
+# matrix strategy and the build job's coverage validation step.
+env:
+  TEST_MATRIX: >-
+    [
+      {"group": "jvm", "matches": "Analysis.Maven Analysis.Gradle Analysis.Scala Analysis.Clojure"},
+      {"group": "compiled", "matches": "Analysis.Go Analysis.Rust Analysis.Erlang Analysis.Elixir"},
+      {"group": "scripting-and-other", "matches": "Analysis.Python Analysis.Ruby Analysis.Swift Analysis.Cocoapods Analysis.Pnpm Analysis.Nuget Analysis.Carthage Analysis.LicenseScanner Container.Analysis Reachability.Upload"}
```
csasarak, line 21: Why these groups?

.github/workflows/integrations-test.yml
```diff
@@ -1,41 +1,58 @@
...
+    ]
+
 jobs:
-  integration-test:
-    name: integration-test
+  # Build once, then run test groups in parallel.
+  # The build job populates the cabal and dist-newstyle caches.
+  # Each test job restores those caches so it doesn't need to rebuild.
```
csasarak, line 27: This comment just describes what the code does.

.github/workflows/integrations-test.yml
```diff
@@ -98,8 +111,127 @@ jobs:
...
+        path: vendor-bins/
+        if-no-files-found: error
+
+    - uses: actions/upload-artifact@v5
+      with:
+        name: rust-release-bins
+        path: |
+          target/release/berkeleydb
+          target/release/millhone
+        if-no-files-found: error
```
csasarak, line 157: I'm not super familiar with this action, but this is how we upload artifacts to display in our release lists. I don't think we want these exposed in our list of release artifacts. Can you verify for me that they won't be? 

This is blocking for me.

csasarak (changes requested):
My only blocking comment is the one asking about `upload-artifact` behavior.

zlav:
Closing this because it was
- Sequenced with the caching PR that broke MacOS without brew
- Outdated, but can easily be revived once I add some testing for the MacOS issue

PR #1688, opened 2026-04-10
[ANE-2900] Omit unset fields from project edit request body

nficca:
# Overview

Calling `fossa project edit` without `--title` returns a 500. The CLI's `UpdateProjectRequest` `ToJSON` instance emits every field unconditionally, so unset ones become JSON `null`. Core's `PUT /api/projects/:locator` doesn't like this because it assumes we want to _set_ the title (and other null fields) to null.

This PR makes the instance omit `Nothing` fields instead. It also renames a misleading copy-paste leftover wrapper in `App.Fossa.Project.projectMain` from `context "Add projects to release group"` to `context "Editing project"`; that string was showing up at the bottom of every `fossa project edit` error traceback.

## Acceptance criteria

* `fossa project edit --policy-id N` succeeds against an existing project without requiring `--title`.
* Error tracebacks from `fossa project edit` no longer mention "Add projects to release group".

## Testing plan

1. Pick any existing project in your org and a license policy id you can swap to.
2. Build the CLI from this branch and run:
   ```sh
   cabal run fossa -- project edit \
     --project-locator "<your-locator>" \
     --policy-id <your-policy-id>
   ```
3. The command should succeed and the project's policy in the FOSSA UI should reflect the new value. Before this PR it returned a 500.

## Risks

None that I can think of.

## Metrics

None.

## References

- [ANE-2900](https://fossa.atlassian.net/browse/ANE-2900): \`fossa project edit --policy\` returns 500 (release-group error) on existing projects

## Checklist

[ANE-2900]: https://fossa.atlassian.net/browse/ANE-2900?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Fossa/API/CoreTypes.hs
+++ b/src/Fossa/API/CoreTypes.hs
@@ -33,6 +33,7 @@ import Data.Aeson (
   withText,
   (.:),
  )
+import Data.Maybe (catMaybes)
 import Data.Text (Text)
 
 --- | Data types of Core's main endpoints
@@ -246,13 +247,13 @@ data UpdateProjectRequest = UpdateProjectRequest
 
 instance ToJSON UpdateProjectRequest where
   toJSON UpdateProjectRequest{..} =
-    object
-      [ "title" .= maybe Null toJSON updateProjectTitle
-      , "url" .= maybe Null toJSON updateProjectUrl
-      , "issueTrackerProjectIds" .= maybe Null toJSON updateProjectIssueTrackerIds
-      , "labels" .= maybe Null toJSON updateProjectLabelIds
-      , "policyId" .= maybe Null toJSON updateProjectPolicyId
-      , "default_branch" .= maybe Null toJSON updateProjectDefaultBranch
+    object . catMaybes $
+      [ ("title" .=) <$> updateProjectTitle
+      , ("url" .=) <$> updateProjectUrl
+      , ("issueTrackerProjectIds" .=) <$> updateProjectIssueTrackerIds
+      , ("labels" .=) <$> updateProjectLabelIds
+      , ("policyId" .=) <$> updateProjectPolicyId
+      , ("default_branch" .=) <$> updateProjectDefaultBranch
       ]
 
 -- Revision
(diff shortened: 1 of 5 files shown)


csasarak:
If it's an error for FOSSA to receive null values here, then is it forwards compatible to change the endpoint to simply accept null? I'm not totally against this PR, but it only solves the problem for future versions of our CLI. If there's a way to make FOSSA do the right thing in both cases IMO that'd be better. But perhaps I'm not understanding the issue

nficca:
> If it's an error for FOSSA to receive null values here, then is it forwards compatible to change the endpoint to simply accept null?

It's not that it's an error for FOSSA to receive null values, it's that FOSSA interprets them as "set this field to null." 

>  it only solves the problem for future versions of our CLI

Good point. I suppose we could change the endpoint to instead treat nulls as "don't update this field," but I'm not sure if that's broadly acceptable; there may be flows in the app that actually intend to null-out some field on the project when they use the endpoint.

csasarak:
ah OK, in that case this seems like the best option.

src/Fossa/API/CoreTypes.hs
```diff
@@ -246,13 +247,13 @@ data UpdateProjectRequest = UpdateProjectRequest
...
-      , "labels" .= maybe Null toJSON updateProjectLabelIds
-      , "policyId" .= maybe Null toJSON updateProjectPolicyId
-      , "default_branch" .= maybe Null toJSON updateProjectDefaultBranch
+    object . catMaybes $
+      [ ("title" .=) <$> updateProjectTitle
+      , ("url" .=) <$> updateProjectUrl
+      , ("issueTrackerProjectIds" .=) <$> updateProjectIssueTrackerIds
+      , ("labels" .=) <$> updateProjectLabelIds
+      , ("policyId" .=) <$> updateProjectPolicyId
+      , ("default_branch" .=) <$> updateProjectDefaultBranch
```
csasarak: It's been a bit, but I think [.?=](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson.html#v:.-63--61-) is made for this purpose.

csasarak (approved):

PR #1692, opened 2026-04-13
Cargo: Fix transitive dev-dependency classification

spatten:
# Overview

The cargo strategy was labeling transitive dependencies by the kind of the single edge directly pointing at them. This caused deps reachable only via a dev-dep subtree to be reported as `Production` whenever the intermediate edges happened to be normal-kind — for example `workspace → approx (dev) → num-traits (normal)` leaked `num-traits` into Production.

The fix in `src/Strategy/Cargo.hs` replaces per-edge labeling with two edge-kind-aware reachability passes over the resolved graph:

- **Production:** reachable from workspace members through paths of null-kind edges only. A `"build"` or `"dev"` edge breaks the release chain, so its target (and everything transitively below) is not Production via that path.
- **Development:** reachable (via any edge) from the target of any non-null-kind edge. A build or dev edge marks the root of a subtree that is never linked into the release binary, and every descendant of that subtree inherits Development.

A package reachable by both kinds of paths carries both labels — `LabeledGrapher` already supports multi-labels, so no structural changes were needed.

### Why we don't need a separate rule for transitive dev-deps

Cargo only resolves `[dev-dependencies]` for workspace members. Dev-deps declared inside a non-workspace crate are not part of `cargo metadata`'s `resolve` graph, so the only non-null kind we ever see on a non-workspace edge is `"build"`. The reachability rule above handles it uniformly.

## Acceptance criteria

- Transitive deps reached only via dev-dep or build-dep roots are labeled `Development` (previously leaked to `Production` when intermediate edges were normal-kind).
- Build-dependencies of production deps (e.g. `autocfg`, `cc`, `version_check`) and their transitives are labeled `Development` — they're needed to compile the project but are never linked into the release artifact.
- A dep reachable from both a Production root and a Development root gets both labels (multi-label preserved).
- Projects with non-trivial dev-dep or build-dep trees will see their Production set shrink and their Development set grow. This is a correctness fix but is a visible behavior change, called out in the changelog.

### Minimal reproduction

The `Cargo.toml` includes two dependencies. `itoa` has no transitive deps. It is included as a prod dependency, so it should be the only production dependency in the graph.

`approx` is being included as a dev dependency. It has two transitive dependencies: `approx` requires `num-traits` as a normal dependency, and `num-traits` requires `autocfg` as a build-script dependency.

Before this fix, `fossa analyze` reported both `itoa` and `num-traits` as production dependencies. After the fix, it correctly reports only `itoa` as a production dependency.

1. Create `/tmp/cargo-devdep-test/Cargo.toml`:

   ```toml
   [package]
   name = "cargo-devdep-test"
   version = "0.1.0"
   edition = "2021"

   [dependencies]
   itoa = "1"

   [dev-dependencies]
   approx = "0.5"
   ```

2. Create `/tmp/cargo-devdep-test/src/lib.rs`:

   ```rust
   pub fn add(a: u64, b: u64) -> u64 { a + b }
   ```

3. Generate the lockfile:

   ```bash
   cd /tmp/cargo-devdep-test && cargo generate-lockfile
   ```

4. Run fossa (released) and fossa-dev (this branch):

   ```bash
   cd /tmp/cargo-devdep-test
   fossa analyze --output | jq
   make install-dev
   fossa-dev analyze --output | jq
   ```

Compare the two outputs. The output from `fossa-dev` only includes `cargo+itoa`. The output from `fossa` includes both `cargo+itoa` and `cargo+num-traits`.

Run again with `--include-unused-deps`. All four deps appear in the graph; the difference is that `approx`, `num-traits`, and `autocfg` are now labeled Development instead of leaking into Production.

### Real-world verification: sparkle and ficus

This is where the build-dep correctness is visible — the minimal `num-traits`-style repro above doesn't differentiate the fix from released `fossa` (both classify `autocfg` as Development, just by different paths). The sparkle/ficus scans are the evidence that the fix handles the build-dep case at scale without regressing any Production deps.

I scanned both projects with released `fossa` (3.16.6) and `fossa-dev` from this branch, then diffed `sourceUnits.Build.Dependencies`:

| Project | Released | This PR | Delta |
|---|---|---|---|
| ficus | 370 | 355 | −15 |
| sparkle | 724 | 693 | −31 |

All movement is Production → Development — no deps incorrectly leaked the other direction. The reclassified deps are exactly the categories you'd expect:

- **Test frameworks:** `proptest`, `wiremock`, `tokio-test`, `axum-test`, `simple_test_case`, `test-log`, `tracing-test-macro`, `assert-json-diff`
- **Build-script utilities:** `autocfg`, `cc`, `cfg_aliases`, `pkg-config`, `rustc_version`, `vcpkg`, `version_check`, `jobserver`, `shlex`, `find-msvc-tools`
- **Dev utilities:** `xshell`, `xshell-macros`, `cargo_metadata`, `cargo-platform`, `env_logger`, `env_filter`, `tempfile`
- **Platform-specific deps of dev tooling:** `winapi-i686-pc-windows-gnu`, `winapi-x86_64-pc-windows-gnu`, `hermit-abi`

### Automated coverage

New `MetadataSpec` cases:

- Transitive deps of dev-deps labeled Development (the original bug).
- Dep reachable from both prod and dev roots gets both labels.
- Direct dep declared as both normal and dev gets both labels.
- **Build-deps of prod-deps and their transitives are Development** (catches the regression that a naïve propagation approach would reintroduce).
- **Dep reachable via a normal prod path and a build-edge gets both labels** (multi-label under the edge-kind-aware algorithm).

## Risks

Correctness fix with a visible behavior change: Production dep sets will shrink for any project that has dev-dep or build-dep trees with non-trivial transitive closures. This was the intent — previously those trees were partially leaking into Production. Flagged in the changelog. No deps move in the opposite direction (Dev → Prod), so there's no risk of over-reporting.

## Metrics

Not tracked.

## References

- Plan: `cargo-dev-prod-labeling-fix.md`

## Checklist

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,11 +2,8 @@
 
 ## 3.17.2
 
+- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production ([#1692](https://github.com/fossas/fossa-cli/pull/1692)).
 - Poetry: Support PEP 621 `[project].dependencies` for Poetry 2.x projects. Production dependencies declared in the standard `[project]` section are now correctly detected alongside legacy `[tool.poetry.dependencies]`. ([#1683](https://github.com/fossas/fossa-cli/pull/1683))
-
-## 3.17.1
-
-- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production. Projects with non-trivial dev-dep trees may see their Production dependency set shrink and their Development set grow ([#1692](https://github.com/fossas/fossa-cli/pull/1692)).
 
 ## 3.17.1
 - Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`. ([#1643](https://github.com/fossas/fossa-cli/pull/1643))
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,6 +6,7 @@
 
 ## 3.17.1
 
+- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production. Projects with non-trivial dev-dep trees may see their Production dependency set shrink and their Development set grow.
 - Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`.
 - Project edit: Fix 500 error when running `fossa project edit --policy` on existing projects ([#1688](https://github.com/fossas/fossa-cli/pull/1688))
 - UV: Add `directory` source type to uv.lock parser, fixing parse failures on projects with local directory dependencies ([#1690](https://github.com/fossas/fossa-cli/pull/1690))
--- a/Changelog.md
+++ b/Changelog.md
@@ -7,7 +7,9 @@
 ## 3.17.1
 
 - Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production. Projects with non-trivial dev-dep trees may see their Production dependency set shrink and their Development set grow ([#1692](https://github.com/fossas/fossa-cli/pull/1692)).
-- Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`.
+
+## 3.17.1
+- Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`. ([#1643](https://github.com/fossas/fossa-cli/pull/1643))
 - Project edit: Fix 500 error when running `fossa project edit --policy` on existing projects ([#1688](https://github.com/fossas/fossa-cli/pull/1688))
 - UV: Add `directory` source type to uv.lock parser, fixing parse failures on projects with local directory dependencies ([#1690](https://github.com/fossas/fossa-cli/pull/1690))
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -6,7 +6,7 @@
 
 ## 3.17.1
 
-- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production. Projects with non-trivial dev-dep trees may see their Production dependency set shrink and their Development set grow.
+- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production. Projects with non-trivial dev-dep trees may see their Production dependency set shrink and their Development set grow ([#1692](https://github.com/fossas/fossa-cli/pull/1692)).
 - Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`.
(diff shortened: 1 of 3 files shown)


src/Strategy/Cargo.hs
```diff
@@ -485,36 +485,88 @@ toDependency emitGitBackedLocators sourceMap pkg =
...
+    isProdEdge dep = any (isNothing . nodeDepKind) (nodeDepKinds dep)
+    isDevEdge dep = any (isJust . nodeDepKind) (nodeDepKinds dep)
+
+    -- Adjacency containing only edges whose parent declares the child as a
+    -- normal dependency (at least one kind is null). Production reachability
+    -- must only traverse these — a build or dev edge breaks the release chain.
+    prodAdj =
+      Map.fromList $
+        map
+          (\node -> (resolveNodeId node, map nodePkg (filter isProdEdge (resolveNodeDeps node))))
```
csasarak: [nit] Slightly easier (for me) to read.

```suggestion
          (\node -> (resolveNodeId node, map nodePkg . filter isProdEdge . resolveNodeDeps $ node))
```

src/Strategy/Cargo.hs
```diff
@@ -485,36 +485,88 @@ toDependency emitGitBackedLocators sourceMap pkg =
...
+reachable :: Map.Map PackageId [PackageId] -> Set PackageId -> Set PackageId
+reachable adj = go Set.empty . Set.toList
+  where
+    go visited [] = visited
+    go visited (x : xs) =
+      if Set.member x visited
+        then go visited xs
+        else
+          let children = fromMaybe [] (Map.lookup x adj)
+           in go (Set.insert x visited) (children ++ xs)
```
csasarak, line 569: I could be looking to quickly, but this looks like a left fold to me. `Data.Set` defines folds. If it is indeed, prefer to use that simply because it makes it clearer what's happening.

Changelog.md
```diff
@@ -2,11 +2,11 @@
 
 ## 3.17.2
 
+- Cargo: Fix transitive dev-dependency classification. Dependencies reachable only through dev-dep or build-dep roots are now correctly labeled as Development instead of Production ([#1692](https://github.com/fossas/fossa-cli/pull/1692)).
 - Poetry: Support PEP 621 `[project].dependencies` for Poetry 2.x projects. Production dependencies declared in the standard `[project]` section are now correctly detected alongside legacy `[tool.poetry.dependencies]`. ([#1683](https://github.com/fossas/fossa-cli/pull/1683))
 
 ## 3.17.1
-
-- Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`.
+- Node.js: Yarn and npm workspace packages now appear as individual build targets (e.g. `yarn@./:my-package`, `npm@./:my-package`), enabling per-package dependency scoping via `.fossa.yml`. ([#1643](https://github.com/fossas/fossa-cli/pull/1643))
```
csasarak, line 40: Do these need to be updated?

csasarak (approved):

src/Strategy/Cargo.hs
spatten: Done — applied in c459170c.

Changelog.md
spatten: Good catch — they did. master has moved on to 3.17.8 since this branch was cut, and 3.17.2 already shipped with other entries. Rebased onto origin/master and moved the Cargo fix into a fresh 3.17.9 section at the top.

src/Strategy/Cargo.hs
spatten: It looks fold-like but it is not one — `go` is a worklist traversal that computes transitive reachability, so it pushes a node's children back onto the worklist (`children ++ xs`) and keeps going until the list drains. A `Set.fold` over the seed set would only visit the seeds themselves; it has no way to follow edges transitively, which is the whole point here. The `visited` set is the accumulator, but the thing being consumed (the worklist) grows during the traversal, so it cannot be expressed as a fold over a fixed `Set`. Happy to extract it into a named `transitiveClosure`/BFS helper if that reads more clearly, but I think a `Set.fold` would actually obscure what's happening.

spatten: Oops. Damn, Claude took it upon itself to reply to the comments. Sorry for the slop!

csasarak: That sounds reasonable, I wasn't reading closely enough. Do agree on not letting claude do that - if only so I know who I'm actually talking to.

PR #1695, opened 2026-04-15
[ANE-2908] Swift Package.swift parser improvements

tjugdev:
# Overview
Updates the parser for `Package.swift` files to handle a few more cases:
- Updates the parser to not get confused by `dependencies` arguments within `targets`. If these came before the dependencies section, we'd end up parsing these dependencies instead of the actual package dependencies, and fail because they use different syntax.
- Handle no `dependencies` argument. The previous parser didn't handle this section missing completely. Now it does, and works even when there are dependencies listed within `targets` (which we correctly ignore).

## Acceptance criteria
We correctly handle `Package.swift` files that fall into either of the two cases outlined in the Overview section.

## Testing plan
- New unit test added, and updated the existing test to trip the bug from the first bullet point in the overview.
- Run `cabal run fossa -- . --output` on a directory containing either of the `Package.swift` files in test/Swift/testdata/
- Confirmed the `Package.swift` files in ANE-2910 are analyzed successfully.

## References
[ANE-2908](https://fossa.atlassian.net/browse/ANE-2908)

## Checklist

[ANE-2908]: https://fossa.atlassian.net/browse/ANE-2908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Swift/PackageSwift.hs
+++ b/src/Strategy/Swift/PackageSwift.hs
@@ -16,6 +16,7 @@ import Control.Applicative (Alternative ((<|>)), optional)
 import Control.Effect.Diagnostics (Diagnostics, context, errCtx, errDoc, errHelp, fatalText, recover, warnOnErr)
 import Control.Monad (void)
 import Data.Foldable (asum)
+import Data.Functor (($>))
 import Data.Map.Strict qualified as Map
 import Data.Set (Set, fromList, member)
 import Data.Text (Text)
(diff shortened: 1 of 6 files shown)


src/Strategy/Swift/PackageSwift.hs
```diff
@@ -196,8 +199,24 @@ parsePackageDep = try parsePathDep <|> parseGitDep
...
+            key <- parseKey
+            case key of
+              "dependencies" -> parseDeps
+              _ -> parseNonDepArray <|> (parseQuotedText $> []) <|> parseIdentifier
+        )
+        (symbol ",")
+  where
+    parseKey = try $ lexeme $ takeWhile1P (Just "package key") (/= ':') <* symbol ":"
+    parseDeps = betweenSquareBrackets (sepEndBy (lexeme parsePackageDep) $ symbol ",")
+    parseIdentifier = takeWhile1P (Just "parse identifier") (/= ',') $> []
```
nficca: [nit] Is it possible for a swift package to end with an identifier and not have a trailing comma? If that's the case (and if I'm understanding this correctly) then this would just consume all the remaining characters, including the final `)` that `betweenBrackets` expects.

nficca (approved):
I have one nit/question I'm unsure about, but otherwise this looks good.

PR #1696, opened 2026-04-16
[ANE-2704] Support pnpm catalog version specifiers

zlav:
# Overview

This PR adds support for pnpm [catalogs](https://pnpm.io/catalogs) (introduced in pnpm 9.5). When a project uses `catalog:` or `catalog:<name>` specifiers in `package.json`, the resolved versions from the `catalogs` section of `pnpm-lock.yaml` are now used during analysis instead of the raw `catalog:...` string.

Removes a stray `Debug.Trace.traceShow` in `Workspace.hs` that dumped the parsed `pnpm-workspace.yaml` object to stderr.

## Acceptance criteria

When users analyze a pnpm project that uses catalog specifiers (e.g. `"react": "catalog:react19"`), FOSSA should resolve these to the actual package versions (e.g. `react@19.0.0`) rather than reporting `catalog:react19` as the version string.

## Testing plan

1. Added `test/Pnpm/testdata/pnpm-9-catalogs/pnpm-lock.yaml` — a v9 lockfile with both `catalog:` (default) and `catalog:react19` (named) specifiers.
2. Added `pnpmLockV9CatalogsSpec` in `PnpmLockSpec.hs` that verifies:
   - `catalog:` resolves `uri-js` via the `default` catalog to `4.4.1`
   - `catalog:react19` resolves `react` via the `react19` catalog to `19.0.0`
   - Regular deps (`colorjs`) are unaffected
   - Edges through catalog-resolved deps work correctly (`uri-js → punycode`)
3. `cabal build spectrometer` and `cabal build unit-tests` both pass.

Note: `cabal test unit-tests` crashes at startup due to a pre-existing Git LFS issue with `test/Container/testdata/emptypath.tar` (unrelated to this change).

## Risks

- The `PnpmCatalogs` parser uses `.:? "catalogs" .!= mempty`, so missing or malformed catalogs are silently ignored — no regression for lockfiles without catalogs.
- If a lockfile contains `catalog:` references but no `catalogs` section (stale lockfile), `resolveCatalogVersion` falls through to the original version string, matching existing behavior.

## References

- [ANE-2704](https://fossa.atlassian.net/browse/ANE-2704): PNPM Catalog Feature
- [pnpm Catalogs documentation](https://pnpm.io/catalogs)
- [pnpm Catalogs RFC](https://github.com/pnpm/rfcs/blob/main/text/0001-catalogs.md)

## Checklist

[ANE-2704]: https://fossa.atlassian.net/browse/ANE-2704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/Strategy/Node/Pnpm/PnpmLock.hs
+++ b/src/Strategy/Node/Pnpm/PnpmLock.hs
@@ -139,8 +139,38 @@ data PnpmLockfile = PnpmLockfile
   , lockFileVersion :: PnpmLockFileVersion
   , lockFileSnapshots :: PnpmLockFileSnapshots
   -- ^ Dependency graph in lockfile version > 9
+  , lockFileCatalogs :: PnpmCatalogs
+  -- ^ Catalog definitions in lockfile version >= 9
   }
   deriving (Show, Eq, Ord)
+
+-- | Catalogs parsed from lockfile. Maps catalog name to (package name -> resolved version).
+-- See: https://pnpm.io/catalogs
+newtype PnpmCatalogs = PnpmCatalogs
+  { catalogEntries :: Map Text (Map Text Text)
+  }
+  deriving (Show, Eq, Ord, Semigroup, Monoid)
+
+instance FromJSON PnpmCatalogs where
+  parseJSON = withObject "PnpmCatalogs" $ \o -> do
+    parsed <- traverse parseCatalog (toHashMapText o)
+    pure $ PnpmCatalogs (Map.fromList $ HashMap.toList parsed)
+    where
+      parseCatalog :: Yaml.Value -> Parser (Map Text Text)
+      parseCatalog = withObject "Catalog" $ \entries ->
+        (Map.fromList . HashMap.toList)
+          <$> traverse (withObject "CatalogEntry" (.: "version")) (toHashMapText entries)
+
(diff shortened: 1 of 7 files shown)


src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -139,9 +139,39 @@ data PnpmLockfile = PnpmLockfile
...
   -- ^ Dependency graph in lockfile version > 9
+  , lockFileCatalogs :: PnpmCatalogs
+  -- ^ Catalog definitions in lockfile version >= 9
   }
   deriving (Show, Eq, Ord)
 
+-- | Catalogs parsed from lockfile. Maps catalog name to (package name -> resolved version).
+-- See: https://pnpm.io/catalogs
+newtype PnpmCatalogs = PnpmCatalogs
+  { catalogEntries :: Map Text (Map Text Text)
```
tjugdev, line 150: Might be worth defining a type alias for `Map Text Text` so that it's a bit clearer what's happening, and then you could use it later in the signature of `parseCatalog`.

tjugdev (approved):

PR #1701, opened 2026-04-23
CI: UPX-compress Linux release binaries (150 MB → 49 MB)

zlav:
## Overview

Enables UPX compression on the Linux release binaries. macOS and Windows intentionally skipped (reasons below).

Still **draft** pending a decision on whether Linux-only is worth shipping on its own; the spike answer is "yes UPX works on Linux, no macOS is not worth the risk surface."

### UPX on Linux amd64 (and Linux-arm via the same gate)
- Install `upx` via `apk add --no-cache upx` inside the existing Alpine build container (Linux runs in `fossa/haskell-static-alpine:ghc-9.8.4`).
- Run `upx --best --lzma` against `fossa`, `diagnose`, and `millhone` after strip.
- `rendergraph` skipped — its main reads from stdin, and UPX's self-modifying stub is more brittle for piped tools.
- Post-UPX smoke-launch verifies the compressed binaries boot before they get uploaded as artifacts.

**Real compression (measured on this branch's green CI run):**

| Binary | Pre-UPX | Post-UPX | Reduction |
|---|---|---|---|
| fossa | 150 MB | 49 MB | 67.6% |
| millhone | 12 MB | 2.4 MB | 79.8% |
| diagnose | 3 MB | 870 KB | 70.7% |

UPX compression itself takes ~60 seconds in CI.

### macOS UPX — dropped after first CI run
UPX 5.x (brew's current version) refuses Mach-O with `CantPackException: macOS is currently not supported (try --force-macos)`. Stacking UPX's own experimental macOS path on top of the hardened-runtime notarization question made failures diagnostically ambiguous — if `--force-macos` produced a binary Apple's notary rejected, we couldn't tell if UPX, the runtime, or the notary was the culprit. Separate spike if macOS matters enough.

### macOS split-sections — tried, didn't work
Original plan was to also enable `split-sections: True` in `cabal.project.ci.macos`, matching Linux/Windows. CI showed why this was never set:

```
[GHC-74335] [-Winconsistent-flags, Werror=inconsistent-flags]
    -fsplit-sections is not useful on this platform since it uses subsections-via-symbols. Ignoring.
```

Mach-O's `ld64` already does function-granular dead-code elimination via `subsections-via-symbols` by default. The ~10–20 MB I projected for macOS was imaginary — that saving was already baked in. Replaced with an explanatory comment in the cabal project file so nobody repeats the mistake.

### Windows UPX — excluded on principle
AV false-positive surface at our distribution scale: Defender ML (`Trojan:Win32/Wacatac.H!ml`), corporate EDR (CrowdStrike/SentinelOne), Zscaler-class proxies, and SmartScreen all flag UPX-packed Windows binaries regardless of contents. Across thousands of CI environments the support burden would outweigh the ~100 MB saving.

### LinuxARM UPX — deferred
Linux-arm passed the gate (`matrix.os-name == 'Linux'`) doesn't include it, so the current workflow compresses only Linux amd64. ARM would need the equivalent apk install inside its own container step; boring but real work.

## Acceptance criteria

- Linux amd64 `fossa` artifact is noticeably smaller (observed 49 MB vs 150 MB baseline).
- `./fossa --version` launches successfully against the UPX-packed binary (smoke-launch step).
- Windows and macOS artifacts unchanged.

### CI
- Build job matrix all green (verified).
- Smoke-launch step on Linux after UPX passes.

### Manual (before un-drafting)
- Download the `Linux-binaries` artifact from the latest build on this branch.
- `file fossa` should report a UPX-packed executable.
- Run on a fresh Linux box (Docker: `docker run --rm -v $(pwd):/w -w /w alpine:3.19 ./fossa --version` or equivalent).
- Measure cold-start latency: UPX LZMA adds ~300–700 ms at decompression time; tolerable for a CLI invoked once per CI job.

## Risks

- **UPX + AV on Linux**: Far lower surface than Windows but not zero. ClamAV can flag packed binaries. Mitigation: low prior incidence for our Linux customer base; watch for support tickets after first release.
- **UPX startup latency**: ~300–700 ms cold-start for LZMA decompression. Visible if someone scripts `fossa --version` in a hot loop (unusual), invisible in normal CI use.
- **Binary identification**: tools like `ldd`, `strings`, and antivirus heuristics see the UPX stub, not the inner binary. Debug workflow changes: need `upx -d` to inspect.

## Out of scope

- `--force-macos` UPX spike (separate issue if pursued).
- Windows UPX + code-signing reputation warmup.
- Dep audit and HTTP stack consolidation (separate PRs).

## References

- UPX docs on macOS support status: https://github.com/upx/upx/blob/devel/README.md
- GHC -Winconsistent-flags: https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Winconsistent-flags
- Earlier chat thread: binary size breakdown showing `__text` 81 MB + `__const` 75 MB + `__data` 8.5 MB as the macOS 177 MB dominant contributors.

## Checklist

- [x] CI workflow change — validated via green CI.
- [x] No user-visible behavior change (internal CI + packaging).
- [x] No `.fossa.yml` / `fossa-deps` / subcommand changes.
- [ ] Manual bare-box Linux launch verification before un-drafting.
- [ ] Product/release decision: ship Linux-only size reduction, or hold for macOS parity?

### DIFF
--- a/.github/workflows/build-all.yml
+++ b/.github/workflows/build-all.yml
@@ -62,7 +62,7 @@ jobs:
     - name: Install MacOS binary dependencies
       if: ${{ contains(matrix.os, 'macos') }}
       run: |
-        brew install jq upx
+        brew install jq
 
     # Set up Haskell.
     - uses: haskell-actions/setup@v2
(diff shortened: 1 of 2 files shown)


zlav:
Good thought experiment, closing for now

PR #1703, opened 2026-04-24
ANE-1036: Glob file matching for exclusion filters in .fossa.yml

zlav:
# Overview

Allow `paths.only` / `paths.exclude` in `.fossa.yml` to accept glob patterns. An entry containing `*` is a glob; otherwise it keeps the existing "directory and all children" semantics, so the change is backward-compatible.

Glob matching reuses `Data.Glob` over `System.FilePattern` (already a dependency).

## Acceptance criteria

New user config
```yaml
paths:
  exclude:
    - "**/vendor/**"
    - "**/node_modules/**"
    - "build/generated/*"
```

These directories get pruned during analyze. 

At startup, `fossa analyze` now logs all of the active filters and any pruned subtrees once, so users can correlate a missing project with a configured rule.

## Testing plan

- New unit tests in `test/Discovery/FiltersSpec.hs` cover glob/concrete include + exclude composition, root-anchoring, trailing-slash normalization, and backslash-pattern parsing for Windows.
- New walker integration tests in `test/Discovery/WalkSpec.hs` cover include-path pruning, the strategy `WalkSkipSome` ↔ filter merge, and end-to-end YAML → `collectConfigFileFilters` → walker pruning.
- Verified manually with a `.fossa.yml` containing `paths.exclude: ["**/zip/**"]` against the fossa-cli checkout: pruned subtree shows up once, no per-strategy spam.
- Local test binary cannot run because of an LFS-backed fixture (`test/Container/testdata/emptypath.tar`) that isn't materialized here; CI exercises the suite.

## Risks

- `Path.toString` on `Path Rel Dir` appends `/`, which `System.FilePattern` rejects for patterns like `node_modules/*`. `globMatchesDir` strips it locally rather than touching `Data.Glob.matches` (shared with Node and Walk).
- `FilterCombination` gets a third field for globs; telemetry JSON is only different when globs are configured.
- `logPrunedSubtrees` does one extra walk, gated on at least one path filter being set; it skips into pruned subtrees, so work is bounded by the kept tree.

## Metrics

Not currently tracked.

## References

- [ANE-1036](https://fossa.atlassian.net/browse/ANE-1036): Glob file matching for exclusion filters in .fossa.yml

## Checklist

- [x] I added tests for this PR's change.
- [x] I added user-visible documentation (updated `docs/references/files/fossa-yml.md`).
- [x] I updated `Changelog.md`.
- [x] I updated `docs/references/files/fossa-yml.v3.schema.json`.
- [x] No subcommand option changes.

[ANE-1036]: https://fossa.atlassian.net/browse/ANE-1036?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -121,7 +121,7 @@ import Data.Aeson qualified as Aeson
 import Data.ByteString.Lazy qualified as BL
 import Data.Error (createBody)
 import Data.Flag (Flag, fromFlag)
-import Data.Foldable (traverse_)
+import Data.Foldable (for_, traverse_)
 import Data.Functor (($>))
 import Data.Glob (unGlob)
 import Data.List.NonEmpty qualified as NE
@@ -137,6 +137,7 @@ import Diag.Result (resultToMaybe)
 import Discovery.Archive qualified as Archive
 import Discovery.Filters (AllFilters (..), MavenScopeFilters, applyFilters, combinedPathGlobs, combinedPaths, filterIsVSIOnly, ignoredPaths, isDefaultNonProductionPath)
 import Discovery.Projects (withDiscoveredProjects)
+import Discovery.Walk (enumeratePrunedSubtrees)
 import Effect.Exec (Exec)
 import Effect.Logger (
   Logger,
@@ -145,6 +146,7 @@ import Effect.Logger (
   logInfo,
   logStdout,
   renderIt,
+  viaShow,
  )
 import Effect.ReadFS (ReadFS)
 import Errata (Errata (..))
@@ -317,6 +319,23 @@ logActivePathFilters AllFilters{includeFilters = include, excludeFilters = exclu
       logInfo $
         "Active " <> pretty label <> " filters: " <> pretty (Text.intercalate ", " items)
 
+-- | Walk the tree once at startup and surface every directory the path
+-- filters will prune. Each prune is logged once at info level here, instead
+-- of emitting per-strategy duplicates from inside the walker (~28 strategies
+-- would otherwise each report the same prune).
+logPrunedSubtrees ::
+  ( Has Logger sig m
+  , Has ReadFS sig m
+  , Has Diag.Diagnostics sig m
+  ) =>
+  AllFilters ->
+  Path Abs Dir ->
+  m ()
+logPrunedSubtrees filters basedir = do
+  pruned <- enumeratePrunedSubtrees filters basedir
+  for_ pruned $ \p ->
+    logInfo $ "Skipping path " <> viaShow p <> " (excluded by paths filter)"
+
 analyze ::
   ( Has Debug sig m
   , Has Diag.Diagnostics sig m
@@ -353,6 +372,7 @@ analyze cfg = Diag.context "fossa-analyze" $ do
       enableVendetta = Config.xVendetta cfg
 
   logActivePathFilters filters
+  logPrunedSubtrees filters basedir
 
   manualDepsResult <-
     Diag.errorBoundaryIO . diagToDebug $
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -146,7 +146,6 @@ import Effect.Logger (
   logInfo,
   logStdout,
   renderIt,
-  viaShow,
  )
 import Effect.ReadFS (ReadFS)
 import Errata (Errata (..))
--- a/src/App/Fossa/Analyze.hs
+++ b/src/App/Fossa/Analyze.hs
@@ -378,9 +378,13 @@ analyze cfg = Diag.context "fossa-analyze" $ do
       withoutDefaultFilters = Config.withoutDefaultFilters cfg
       enableSnippetScan = Config.snippetScan cfg
       enableVendetta = Config.xVendetta cfg
+      -- Discovery runs with `mempty` when `--no-discovery-exclusion` is set
+      -- (see definition further down). Log against the same filter set so the
+      -- startup output matches what discovery actually applies.
+      discoveryFilters = if fromFlag NoDiscoveryExclusion noDiscoveryExclusion then mempty else filters
 
-  logActivePathFilters filters
-  logPrunedSubtrees filters basedir
+  logActivePathFilters discoveryFilters
+  logPrunedSubtrees discoveryFilters basedir
 
   manualDepsResult <-
     Diag.errorBoundaryIO . diagToDebug $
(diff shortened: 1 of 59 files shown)


zlav:
@coderabbitai review

src/App/Fossa/Analyze.hs
```diff
@@ -297,6 +301,49 @@ runAnalyzers allowedTactics filters withoutDefaultFilters basedir pathPrefix = d
...
+logActivePathFilters AllFilters{includeFilters = include, excludeFilters = exclude} = do
+  emit "include path" (map (toText . toFilePath) (combinedPaths include))
+  emit "include glob" (map (toText . unGlob) (combinedPathGlobs include))
+  emit "exclude path" (map (toText . toFilePath) (combinedPaths exclude))
+  emit "exclude glob" (map (toText . unGlob) (combinedPathGlobs exclude))
+  where
+    emit :: Has Logger sig m => Text -> [Text] -> m ()
+    emit _ [] = pure ()
+    emit label items =
+      logInfo $
```
tjugdev: This seems more like a debug level log to me, as its essentially just outputting what's in your .fossa.yml.

src/Discovery/Filters.hs
```diff
@@ -86,11 +94,56 @@ instance Monoid AllFilters where
...
+data PathFilter
+  = PathFilterDir (Path Rel Dir)
+  | PathFilterGlob (Glob Rel)
+  deriving (Eq, Ord, Show, Generic)
+
+instance FromJSON PathFilter where
+  parseJSON = withText "PathFilter" $ \txt ->
+    let s = toString txt
+     in if '*' `elem` s
+          then pure . PathFilterGlob . Glob.unsafeGlobRel $ normalizeSlashes s
```
tjugdev, line 126: I think we should try to use `Glob.toGlob` instead. It looks like the unsafe version is currently only used in tests.

src/Discovery/Filters.hs
```diff
@@ -291,6 +385,75 @@ applyTarget (TypeDirTargetTarget t p target) u q = if t == u && p == q then Matc
...
+-- converts backslashes to forward slashes so user-supplied forward-slash
+-- patterns match the backslash-separated paths produced on Windows.
+globMatchesDir :: Glob Rel -> Path Rel Dir -> Bool
+globMatchesDir glob dir = unGlob glob FilePattern.?== normalize (toString dir)
+  where
+    normalize :: String -> String
+    normalize = trimTrailingSlash . normalizeSlashes
+
+    trimTrailingSlash :: String -> String
+    trimTrailingSlash s = case reverse s of
```
tjugdev: I think you can use [`Data.List.unsnoc`](https://hackage-content.haskell.org/package/base-4.22.0.0/docs/Data-List.html#v:unsnoc) here instead to avoid reversing the list twice.

src/Discovery/Filters.hs
```diff
@@ -291,6 +385,75 @@ applyTarget (TypeDirTargetTarget t p target) u q = if t == u && p == q then Matc
...
+  let prefix = takeWhile (notElem '*') (filter (not . null) (splitSlash (normalizeSlashes (unGlob g))))
+   in -- An empty literal prefix means the glob's first segment is wildcarded
+      -- (e.g. @**/service/**@, @*/foo@). The wildcard can stand for any
+      -- ancestor segments, so any path is a candidate parent — let the walker
+      -- descend, the actual match check ('isIncludedByGlob') will fire when
+      -- it reaches a real match. Otherwise require @segs@ to be a prefix of
+      -- the literal directory part.
+      null prefix || segs `isPrefixOf` prefix
+  where
+    splitSlash :: String -> [String]
```
tjugdev, line 445: Sorta a vague comment, but is it possible to make these methods not convert paths to strings? I feel like it could be more robust to keep the paths as `Path`s, but I'm not sure. Feel free to ignore if this suggestion doesn't work.

test/Discovery/FiltersSpec.hs
```diff
@@ -260,6 +266,144 @@ spec = do
...
+        pathAllowed filters $(mkRelDir "lib") `shouldBe` False
+
+      it "treats a leading '**' include glob as accepting any ancestor" $ do
+        -- `**/service/**` can match arbitrarily deep, so the walker has to be
+        -- allowed everywhere on the way down or it'll never reach a match.
+        let filters = includeGlob "**/service/**"
+        pathAllowed filters $(mkRelDir "anywhere") `shouldBe` True
+        pathAllowed filters $(mkRelDir "anywhere/else") `shouldBe` True
+        pathAllowed filters $(mkRelDir "deep/nested/service/foo") `shouldBe` True
+
```
tjugdev, line 326: This behaviour seems wrong to me. Shouldn't `anywhere` and `anywhere/else` be not allowed?

zlav: I think you might be reading this wrong? Shouldn't `**/service/**` only catch things with a service path in them? Or are you insinuating that the `**` prefix should omit all directories

src/App/Fossa/Analyze.hs
zlav: I agree with you here. I'm going to remove this entirely.

For context, I realized during manual testing that we don't omit any information about paths that get omitted due to path filters, I experimented with a few methods, couldn't get `logPrunedSubtrees` to work, implemented this as a stopgap, finally got `logPrunedSubtrees` to work, and then likely forgot to remove this. Apologies, good catch 👍

test/Discovery/FiltersSpec.hs
tjugdev: > Shouldn't `**/service/**` only catch things with a service path in them?

I think that's correct, which means I think the assertions should be:
```
        pathAllowed filters $(mkRelDir "anywhere") `shouldBe` False
        pathAllowed filters $(mkRelDir "anywhere/else") `shouldBe` False
        pathAllowed filters $(mkRelDir "deep/nested/service/foo") `shouldBe` True
```
because the first two don't contain `service` in the path.

src/Discovery/Filters.hs
zlav: Will do 👍

test/Discovery/FiltersSpec.hs
zlav:
```
        -- pathAllowed is what determines if the walker should traverse the
        -- directory. For this test, we're asserting that `**/` should traverse
        -- every directory, so all lines should be true. The previous block of
        -- `apps/*` is asserting on a false example.
```

Added a comment to explain this. Basically, this test isn't asserting on what matches the glob, its asserting on what folders are allowed to be traversed with the regex.

src/Discovery/Filters.hs
zlav: `Glob.toGlob :: Path b t -> Glob b` requires a `Path`, but a `Path` can't contain globs. If they _could_ then we could do Glob.toGlob on `parseRelDir` below, but as it stands, we have to do this directly on the string.

As for `unsafaeGlobRel`, some of the risks are mitigated with `normalizeSlashes` on line 131. Also, there is precedent for creating globs this way in `Strategy/Node.hs:400` where we create globs to find npm workspaces.

src/Discovery/Filters.hs
zlav: I don't think so. They have to be converted to `[String]` to compare with the glob. I think this is the best we can do, considering we are only converting `Path` to `[String]` right before comparison with `pathSegments`.

tjugdev (approved):

PR #1705, opened 2026-04-29
[CORE-6867] Smoke-test ficus binary after vendor download

ahamid-fossa:
See: https://github.com/fossas/ficus/pull/169

The pre-existing version-check loop runs `--version` on every binary but swallows failures via `|| echo "failed to get version information"`, so a non-runnable ficus (wrong glibc, wrong arch, corrupt download) prints a warning and the script continues. The binary then gets embedded into the CLI (and downstream consumers like Hubble) via Template Haskell and the failure only manifests at run time.

Add a strict smoke-execute of vendor-bins/ficus that fails the build if the binary cannot load in the build image's libc/arch.

# Overview

Ensure compatibility with downstream consumers by failing when ficus binary cannot be run.

## Acceptance criteria

Incompatible consumers _cannot_ build against in incompatible ficus. Specifically wrt to libc version.

## Testing plan

I think maybe if we accept this PR (without rebuilt ficus from https://github.com/fossas/ficus/pull/169), and then rebuild Hubble against this branch of ficus-cli, we should expect it to _fail_. If it fails then this change has succeeded in preventing a bad Hubble/ficus build.

## Risks

The previous logic was testing, but NOT failing. Maybe there is a use case where we don't want to fail if ficus is not runnable?

## References

_Example:_

## Checklist

### DIFF
--- a/vendor_download.sh
+++ b/vendor_download.sh
@@ -271,3 +271,12 @@ for binary in vendor-bins/*; do
     "$binary" --version || echo "failed to get version information"
   fi
 done
+
+# Strict smoke-test ficus: it is embedded via Template Haskell into the CLI
+# binary (and downstream consumers like Hubble), so a non-runnable ficus
+# (wrong glibc, wrong arch, corrupt download) ships silently otherwise and
+# only manifests at run time.
+./vendor-bins/ficus --version > /dev/null || {
+  echo "ERROR: vendor-bins/ficus does not run in this image"
+  exit 1
+}


spatten (approved):
This looks like a good change to me.

I tested like this. You'll need a GITHUB_TOKEN that has access to ficus, themis, etc, if you want to try it:

```
git checkout aaron/ficus-smoketest
docker run --rm -it --platform=linux/amd64 -v "$PWD":/src -w /src debian:buster bash

sed -i 's|deb.debian.org|archive.debian.org|g; /-updates/d; /-security/d' /etc/apt/sources.list
apt-get -o Acquire::Check-Valid-Until=false update -qq
apt-get install -y -qq curl jq xz-utils ca-certificates

export GITHUB_TOKEN=...
./vendor_download.sh
echo $?
```

I see this when I do that:

```
vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by vendor-bins/ficus)
vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by vendor-bins/ficus)
vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by vendor-bins/ficus)
failed to get version information
lernie: lernie v0.3.1
themis-cli: 2026/05/06 23:31:24 maxprocs: Leaving GOMAXPROCS=14: CPU quota undefined
Version: 1.0.50
./vendor-bins/ficus: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by ./vendor-bins/ficus)
./vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.30' not found (required by ./vendor-bins/ficus)
./vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.32' not found (required by ./vendor-bins/ficus)
./vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.33' not found (required by ./vendor-bins/ficus)
./vendor-bins/ficus: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.34' not found (required by ./vendor-bins/ficus)
ERROR: vendor-bins/ficus does not run in this image
root@efdac803c8bb:/src# echo $?
1
```

ahamid-fossa:
Thanks for the review and for the test plan @spatten. I got the same result.

PR #1707, opened 2026-05-11
Update make_fossa_deps_conan.py with smarter dependency classification

jtrinidad-fossa:
# Overview

Merges improvements from an updated version of the script:

- Replaces vendored-dependencies output with three-tier classification: referenced-dependencies (git), remote-dependencies (archive tarball), and custom-dependencies (fallback), matching fossa-deps.yml spec more closely
- Adds git URL extraction from conandata.sources before falling back to homepage, supporting GitHub, GitLab, Bitbucket, SourceHut, and Codeberg
- Adds git tag extraction from source URLs to handle packages that use non-standard tag formats (e.g. curl-8_17_0 vs 8.17.0)
- Adds archive URL fallback for packages like GNU libraries that have no Git repo
- Skips private-channel packages without conandata to avoid resolving internal version strings as git tags
- Skips header-only libraries with binary == "Skip"
- Adds Option B usage: accept a pre-generated conan graph info JSON file as an argument instead of always running conan via subprocess
- Removes package_id from version string (was opaque to FOSSA's git resolver)

## Acceptance criteria

Customer's will be able to use this updated script as a starting point for scanning their Conan projects. The current hadn't been updated in a while and needed an update.

## Testing plan

1. Took a `conan graph info -f json` output supplied by a customer and ran it through the script (`python3 make_fossa_deps_conan.py customer-graph.json`) to confirm it produces a valid `fossa-deps.yml` (Option B / pre-generated JSON path).
2. To repeat locally without a Conan project on hand, feed the script a minimal fabricated graph JSON exercising each classification branch, e.g.:
   ```json
   {
     "graph": {
       "nodes": {
         "0": {"label": "conanfile.txt", "recipe": "Consumer"},
         "1": {
           "name": "zlib", "ref": "zlib/1.3.1", "version": "1.3.1",
           "license": ["Zlib", "MIT"],
           "homepage": "https://github.com/madler/zlib",
           "conandata": {"sources": {"1.3.1": {"url": ["https://github.com/madler/zlib/releases/download/v1.3.1/zlib-1.3.1.tar.gz"]}}}
         },
         "2": {"name": "noconan", "ref": "noconan/2.0", "version": "2.0", "license": null}
       }
     }
   }
   ```
   Run `python3 make_fossa_deps_conan.py fake-graph.json` and inspect the generated `fossa-deps.yml`: node `1` should land under `referenced-dependencies` with the extracted GitHub URL and tag, and node `2` (no source URL) should land under `custom-dependencies` with `license: NOASSERTION` (confirms the list/null license fallback works and doesn't emit invalid YAML).
3. `python3 -m py_compile make_fossa_deps_conan.py` for a basic syntax/import check — this script lives under `docs/walkthroughs/` and isn't wired into the Haskell/Rust unit or integration test suites, so manual execution against sample graphs is the only current verification path.
4. Verify `fossa analyze` accepts the generated `fossa-deps.yml` without schema errors (run against `docs/references/files/fossa-deps.schema.json` or via `fossa analyze` on a scratch project).

## Risks

Would like someone more familiar with this script to make sure it is correct.

## Metrics

Not applicable — this is a standalone script distributed via `docs/walkthroughs/`, not compiled into the fossa-cli binary, so there's no telemetry/analytics hook to instrument. Adoption/success would show up indirectly as fewer Zendesk tickets about broken or stale Conan integration guidance.

## References

No tracked Jira/Zendesk ticket — this rewrite was driven directly by customer feedback (a customer-supplied `conan graph info.json` surfaced gaps in the previous version's git-URL/tag resolution and vendored-dependency classification).

## Checklist


zlav (changes requested):
Ok, I took a look through this PR and have a few thoughts. The changes seem directionally correct to provide accurate Conan support, but unfortunately I think these count as breaking changes for existing users of this script, and I'm not familiar with who uses this, so I don't feel confident merging this without digging into who the users are. I can do that when I'm back in office. 
Fortunately, @GauravB159 is working on building first class Conan support, and the goal would be to deprecate this script all together

Breaking changes:
First, the header-only skip drops the header deps from the output, which the users this script was built for may consider to be required:

    if node.get("binary") == "Skip" and node.get("package_type") == "header-library":
        continue

Second, dropping `vendored-dependencies` in exchange for referenced reps changes what gets scanned and how the issues are handled. I think as of last week (Git Vulns changes) this would enable us to show vulns, however it would cause large issue churn for users with vendored deps that now show up as new issues under referenced deps.

TL;DR - Changes look good, but break existing users. Gaurav is working on first class Conan support, I will make sure he consults with you to make sure your users are supported by what he ships, and don't need this script.

PR #1711, opened 2026-05-12
Strategy/Scala: route sbt 1.4+ with explicit DependencyTreePlugin to built-in command (TKT-15490)

zlav:
Route sbt 1.4+ `DependencyTreePlugin` to the uppercase `dependencyBrowseTreeHTML` task, and detect plugins from the real `sbt plugins` output layout — restoring complete deep-dependency graphs on sbt 1.4+.

## What changed

- On sbt 1.4+ with `addDependencyTreePlugin`, fossa ran the lowercase `dependencyBrowseTreeHtml` task (added for sbt <1.4 compat); modern sbt only exposes the uppercase task, so it failed and analysis silently fell back to a poms-only graph. Detection now classifies legacy (`net.virtualvoid…DependencyGraphPlugin`, lowercase) vs modern (`sbt.plugins.DependencyTreePlugin`, uppercase) and runs the matching task.
- Plugin detection now parses sbt's actual `plugins` layout — bare FQCNs under `Enabled plugins in <project>:`, with `disablePlugins(...)` entries excluded via the trailing not-enabled section. An interim `: enabled in` suffix anchor matched **nothing** in real output and dropped the built-in `MiniDependencyTreePlugin` on plain sbt 1.4+ projects.
- `hasDependencyPlugins` returns a `DependencyPluginsDetected` record instead of `(Bool, Bool)`; the one caller (`Strategy.Scala.findProjects`) is updated.

## Validation

- `cabal build`; `fourmolu --mode check`.
- `unit-tests --match "detectDependencyPlugins"` → 8/8 pass, including a verbatim `sbt -batch -no-colors plugins` capture as a regression guard.
- Full Scala integration test needs nix+sbt+network (~15 min) — relying on CI.

## References

- Support ticket TKT-15490; prior routing work ANE-2718. Casing regression originated in #1356 (`215a74f8`).

## Why not to merge this

- **String-parse coupling** — keys off sbt's human-readable `plugins` section headers; a future sbt reformat silently breaks detection again (no machine-readable contract).
- **Legacy branches unverified live** — only the sbt 1.9.8 fixtures are real captures; the `net.virtualvoid` and modern-without-mini fixtures are hand-built to the same layout, not captured from old sbt.
- **Substring, not exact-line match** — `isInfixOf` on the enabled section tolerates suffix variants but would miscount a plugin whose FQCN is a substring of another (none exist today).
- **Symptom vs source** — the detection regression was self-inflicted by earlier commits on this branch; squashing them would avoid shipping the bug+fix pair, though keeping them documents the trap.
- Worth merging anyway: restores complete deep graphs for the common sbt 1.4+ case, proven against the exact fixture the integration test exercises.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
-- Scala/sbt: Fix silently dropped deep dependencies on sbt 1.4+ with explicit `addDependencyTreePlugin`. ([#1711](https://github.com/fossas/fossa-cli/pull/1711))
+- Scala/sbt: Route projects using `addDependencyTreePlugin` (sbt 1.4+) to the correct `dependencyBrowseTreeHTML` task, restoring deep dependencies. ([#1711](https://github.com/fossas/fossa-cli/pull/1711))
 
 ## 3.17.6
 
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+
+- Scala/sbt: Run the uppercase `dependencyBrowseTreeHTML` task when the project explicitly enables `addDependencyTreePlugin` on sbt 1.4+. Previously the lowercase `dependencyBrowseTreeHtml` was used unconditionally for the explicit-plugin path, which sbt 1.4+ rejects, causing deep dependencies to be silently dropped. ([TKT-15490](https://fossa.atlassian.net/browse/ANE-2718))
+
 ## 3.17.5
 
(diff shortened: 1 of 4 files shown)


zlav:
@coderabbitai review

spatten (approved):
Looks good to me!

Claude is calling out that there might be a separate mechanism that would cause missing dependencies. If `hasMini = True` and they have a multi-module project, then they get to line 184 of `src/Strategy/Scala.hs`, and fallback to a pom-only partial graph.

I think that's probably not the case here. I'm guessing we have a debug log to show us that the problem was the casing of the task. But I thought I'd mention it just in case it was useful.

Either way, I think that's a separate PR and this one is good to go.

spatten:
I'm going to fix the vulns while I'm in here. This is the same problem we saw on Sparkle - they're all Rust vulns

PR #1713, opened 2026-05-19
[ANE-2955] Fix vendored archive uploads with absolute paths

nficca:
# Overview

`fossa analyze` with archive uploads crashes when any `vendored-dependencies[].path` is absolute. It fails with `withBinaryFile: permission denied` while trying to write the tarball at the filesystem root. The [meta-fossa](https://github.com/fossas/meta-fossa) Yocto layer always emits absolute paths, so every archive-upload run from it hits this.

The temp filename builder kept a leading `/` for absolute inputs and that escaped the intended output directory. Fix strips the root component so the tarball stays where it should.

## Acceptance criteria

Archive-upload `fossa analyze` succeeds when a vendored dep `path` is absolute, including for the meta-fossa Yocto integration.

## Testing plan

1. Stage a vendored source dir at any absolute path:
   ```sh
   mkdir -p /tmp/vendor/foo
   echo "license: MIT" > /tmp/vendor/foo/LICENSE
   ```

2. Create a project dir with a `fossa-deps.json` pointing at it:
   ```sh
   mkdir /tmp/project && cd /tmp/project
   cat > fossa-deps.json <<'JSON'
   {
     "vendored-dependencies": [
       { "name": "foo", "version": "0.0.1", "path": "/tmp/vendor/foo" }
     ]
   }
   JSON
   ```

3. Run analyze, forcing archive upload so this code path is exercised regardless of org default:
   ```sh
   cabal run fossa -- analyze -p abs-path-repro -r v1 \
     --force-vendored-dependency-scan-method ArchiveUpload
   ```

4. Against `master`, the run dies with `withBinaryFile: permission denied` on a path starting with `/_tmp_vendor_foo...`. With this branch, the run succeeds and the vendored dep shows up in the project on app.fossa.com.

## Risks

Behaviour change is scoped to absolute-path inputs; relative paths produce the same filename as before.

## Metrics

None.

## References

- [ANE-2955](https://fossa.atlassian.net/browse/ANE-2955): Yocto integration yielding no results.

## Checklist

No docs/schema/subcommand changes needed. Changelog entry under a new `## 3.17.7` section since this is intended to ship as the next release.

[ANE-2955]: https://fossa.atlassian.net/browse/ANE-2955

### DIFF
--- a/src/App/Fossa/VendoredDependency.hs
+++ b/src/App/Fossa/VendoredDependency.hs
@@ -7,6 +7,7 @@ module App.Fossa.VendoredDependency (
   vendoredDepToLocator,
   forceVendoredToArchive,
   compressFile,
+  safeSeparators,
   hashFile,
   hashBs,
   dedupVendoredDeps,
@@ -197,8 +198,11 @@ hashFile fileToHash = do
   fileContent <- BS.readFile fileToHash
   pure . toText . show $ md5 fileContent
 
+-- Drop root components so absolute inputs don't yield a result starting with
+-- '/', which `(</>)` would treat as absolute and use to overwrite the caller's
+-- intended output directory.
 safeSeparators :: FilePath -> FilePath
-safeSeparators = intercalate "_" . splitDirectories
+safeSeparators = intercalate "_" . filter (/= "/") . splitDirectories
 
 skippedDepsDebugLog :: NeedScanningDeps -> SkippableDeps -> VendoredDependencyScanMode -> SkippedDepsLogMsg
 skippedDepsDebugLog needScanningDeps skippedDeps scanMode =
--- a/src/App/Fossa/VendoredDependency.hs
+++ b/src/App/Fossa/VendoredDependency.hs
@@ -51,7 +51,8 @@ import Fossa.API.Types (
 import Path (Abs, Dir, Path)
 import Prettyprinter (Pretty (pretty), vsep)
 import Srclib.Types (Locator (..), ProvidedPackageLabel)
-import System.FilePath.Posix (splitDirectories, (</>))
+import System.FilePath (dropDrive, splitDirectories)
+import System.FilePath.Posix ((</>))
 
 data VendoredDependency = VendoredDependency
   { vendoredName :: Text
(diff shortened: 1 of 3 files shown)


test/App/Fossa/VendoredDependencySpec.hs
```diff
@@ -42,6 +43,14 @@ spec = do
         compressedFilePath <- sendIO $ withSystemTempDir "fossa-temp" (flippedCompressFile specDir fileToTar)
         compressedFilePath `shouldContain'` fileToTar
 
+  describe "safeSeparators" $ do
+    it "joins relative path components with underscores" $
+      safeSeparators "foo/bar" `shouldBe` "foo_bar"
+    it "leaves bare filenames untouched" $
+      safeSeparators "foo" `shouldBe` "foo"
+    it "drops the root component for absolute paths" $
+      safeSeparators "/foo/bar/baz" `shouldBe` "foo_bar_baz"
```
csasarak: I _strongly_ recommend making these paths in an abstract way. Something like:

```haskell
do currDir <- getCurrentDirectory -- Should make an absolute path
      safeSeparators `shouldSatisfy` (not . any (== pathSeparator))
```

The reason is that our path finding libraries will use the OS native path format. I think if you constructed the paths more abstractly, this test would fail on our Windows build and probably should.

src/App/Fossa/VendoredDependency.hs
```diff
@@ -197,8 +198,11 @@ hashFile fileToHash = do
   fileContent <- BS.readFile fileToHash
   pure . toText . show $ md5 fileContent
 
+-- Drop root components so absolute inputs don't yield a result starting with
+-- '/', which `(</>)` would treat as absolute and use to overwrite the caller's
+-- intended output directory.
 safeSeparators :: FilePath -> FilePath
-safeSeparators = intercalate "_" . splitDirectories
+safeSeparators = intercalate "_" . filter (/= "/") . splitDirectories
```
csasarak: Would it be sufficient to just remove leading `/`? Not requiring a change, just trying to make sure I understand the problem. 

You may also want to consider if this code will work on Windows. https://hackage-content.haskell.org/package/filepath-1.5.5.0/docs/System-FilePath.html#g:1

csasarak (changes requested):
Can you address my comments re: windows? Aside from that nothing is blocking.

csasarak (approved):
LGTM

hellow554:
Sorry to hijack this, I was the person who reported this.

Currently the script is putting all the data into a directory under `/tmp`.

This is - under linux at least - backed up by my RAM.
My files are much bigger than my RAM can handle, therefore this doesn't succeed.

So the question is:
  - Is there a way to modify the path via the cli or fossa.yaml?
  - If not, can we put it into the current working dir instead of temp?

nficca:
Hi @hellow554. You should be able to modify the path of temp directory we use like so:
```sh
TMPDIR=/some/path fossa analyze ...
```

We use [`Path.IO.withSystemTempDir`](https://hackage.haskell.org/package/path-io-1.8.2/docs/src/Path.IO.html#withSystemTempDir) to create temp directories, which on POSIX [reads](https://hackage.haskell.org/package/directory-1.3.9.0/docs/src/System.Directory.Internal.Posix.html#getTemporaryDirectoryInternal) `TMPDIR` before falling back to `/tmp`.

If this doesn't work for you, I'll ask that you file a new support ticket as this is a separate issue than the one fixed in this PR.

PR #1719, opened 2026-06-10
docs(conan): handle list-valued license in make_fossa_deps_conan

cmboling:
## Problem

The Conan walkthrough script (`docs/walkthroughs/make_fossa_deps_conan.py`) fails when a
dependency's recipe declares `license` as a **list** rather than a string — common in
conan-center-index (e.g. [`license = ["..."]`](https://github.com/search?q=repo%3Aconan-io%2Fconan-center-index+%22license+%3D+%5B%22&type=code)).

`license_of()` returned `node.get("license")` unchanged, so a list was written into the
fossa-deps `license` field, which must be a String. `fossa analyze` then fails:

```
Error in $['custom-dependencies'][N].license:
  parsing Text failed, expected String, but encountered Array
```

This is **not** a Conan-version quirk — it's per-recipe. Conan 2's `conan graph info -f json`
preserves the recipe's shape: `ConanFile.serialize()` does
`result["license"] = list(self.license) if not isinstance(self.license, str) else self.license`,
so string-recipes yield a string and list-recipes yield an array. The script must handle both.

## Fix

`license_of()` now normalizes any shape Conan can emit into a single string:

- `"MIT"` → `"MIT"`
- `["MIT", "Apache-2.0"]` → `"MIT AND Apache-2.0"` (joined into one SPDX expression)
- tuples and lists containing `None`/blank entries are cleaned up
- empty list / `None` / missing → `None`

Multiple licenses are joined with `" AND "` (a one-line `MULTI_LICENSE_JOINER` constant) as the
conservative, SPDX-valid default — Conan's list doesn't disambiguate AND vs OR; switch to
`" OR "` for dual-licensed packages.

## Validation

- Unit-checked `license_of()` against string, list, tuple, list-with-None/blanks, single-element,
  empty list, `None`, and missing — all return the expected single string (or `None`).
- Reproduced the original failure (array → YAML list) and confirmed the fix emits a scalar String.
- End-to-end: generated a `fossa-deps.yml` from a graph containing array-license nodes, parsed it,
  and asserted every `custom-dependencies[*].license` is a String.

## Notes

- Docs-only change to the walkthrough script; no `Changelog.md` entry added (happy to add one).
- Related edge not covered here: a recipe with no license still emits `license: null`, which trips
  the same parser ("expected String, but encountered Null"). Can fold a fix into this PR if desired.

### DIFF
--- a/docs/walkthroughs/make_fossa_deps_conan.py
+++ b/docs/walkthroughs/make_fossa_deps_conan.py
@@ -107,8 +107,24 @@ def name_version_of(label: str) -> Tuple[str, str]:
     name, version = label.split("/", 1)
     return name, version
 
+# Conan recipes may declare `license` as a single string ("MIT") or as a list/tuple of
+# strings (["MIT", "Apache-2.0"]). The fossa-deps `license` field must be a single string,
+# so a list is joined into one SPDX expression. We use " AND " (every license's obligations
+# apply) as the conservative default; change MULTI_LICENSE_JOINER to " OR " if your packages
+# are dual-licensed (consumer's choice).
+MULTI_LICENSE_JOINER = " AND "
+
 def license_of(node: dict) -> Optional[str]:
-    return node.get("license")
+    raw = node.get("license")
+    if raw is None:
+        return None
+    if isinstance(raw, str):
+        return raw or None
+    if isinstance(raw, (list, tuple)):
+        parts = [str(item).strip() for item in raw if item is not None and str(item).strip()]
+        return MULTI_LICENSE_JOINER.join(parts) if parts else None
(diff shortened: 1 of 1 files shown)


cmboling:
One of our users reported this error and this PR fixes it:
```
Error:  An issue occurred

  *** Relevant Errors ***

      Error: parsing file: /__w/<app>/fossa-deps.yml
        Aeson exception:
        Error in $['custom-dependencies'][5].license: parsing Text failed, expected String, but encountered Array
      Support: If you believe this to be a defect, please report a bug to FOSSA support at https://support.fossa.com/, with a copy of:  /__w/<app>/fossa-deps.yml
```

tjugdev (approved):

PR #1730, opened 2026-07-08
Maven: recover duplicate-resolved dependency edges hidden by the depgraph plugin (ANE-3081)

saramaebee:
# Overview

Maven attaches a package shared by several parents to a single winning parent, and the depgraph plugin's `aggregate` goal can only emit those winning edges — it hardcodes `NodeResolution.INCLUDED` and has no `showDuplicates` option, and its text format can't express duplicate edges regardless of flags. The uploaded graph therefore shows a shared transitive dependency under only one parent, and dependency paths under-report who actually uses it.

This PR adds a best-effort second run of the plugin's `graph` goal with `-DshowDuplicates=true` (JSON output; runs per reactor module in one `mvn` invocation) and merges the recovered `OMITTED_FOR_DUPLICATE` edges into the aggregate output before the graph is built. Any failure falls back to current behavior with a warning.

Two details worth knowing: edges join to artifacts by string id because the plugin's `numericId`/`numericFrom`/`numericTo` use unrelated counters; and `OMITTED_FOR_CONFLICT` edges are skipped since they point at losing versions the build doesn't ship.

## Acceptance criteria

Given `poi-ooxml -> log4j-api` plus a direct `log4j-core` dependency that also imports `log4j-api`, `fossa analyze` now reports both parent edges instead of only Maven's winner, and dependency paths in the app show all real paths.

## Testing plan

1. `cabal test unit-tests --test-options='-m "Maven"'` — new specs cover the JSON decoding (including the numeric-id mismatch) and merge semantics (ignores INCLUDED edges and unknown artifacts, no self-edges, idempotent).
2. End-to-end: ran the built CLI's `fossa analyze -o` on a demo project with the scenario above. Before: `log4j-core` and `postgresql` had `imports: []`. After: `log4j-core -> log4j-api` and `postgresql -> checker-qual` are present.

## Risks

One extra `mvn` invocation per Maven analysis (plugin is already installed; failures are non-fatal by construction). The legacy 3.3.0 plugin path gets the same step and degrades gracefully if unsupported.

## Metrics

None added. The `Running plugin to recover duplicate-resolved edges` context appears in debug bundles.

## References

Root-caused during a customer engagement where OEM-vendored transitive dependencies were mis-attributed to a single parent, breaking package-label workflows that rely on accurate dependency paths.


src/Strategy/Maven/Plugin.hs
```diff
@@ -123,35 +132,138 @@ installPlugin dir path plugin = do
...
+    readDerivedCandidates paths = do
+      allPresent <- and <$> for paths doesFileExist
+      if allPresent
+        then traverse readContentsJson paths
+        else walkAndRead dir
+
+    walkAndRead :: (Has ReadFS sig m, Has Diagnostics sig m) => Path Abs Dir -> m [VerboseGraph]
+    walkAndRead base = do
+      outputs <- walk' (\_ _ files -> pure (maybeToList (findFileNamed verboseGraphFileName files), WalkContinue)) base
+      traverse readContentsJson outputs
```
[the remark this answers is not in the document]
csasarak: the only reason this would run is if we fall back to actually walking the file tree - normally the module locations are specified pretty strictly defined. This would happen if maven failed to run (which we already guard against elsewhere) or if the expected file is absent.

it's real, but not worth holding up this PR.

src/Strategy/Maven/PluginStrategy.hs
```diff
@@ -11,15 +11,17 @@ import Control.Effect.Diagnostics (
...
-  (<||>),
  )
 import Control.Effect.Lift (Lift)
 import Control.Effect.Path (withSystemTempDir)
 import Control.Monad (when)
 import Data.Foldable (traverse_)
 import Data.Map.Strict (Map)
 import Data.Map.Strict qualified as Map
+import Data.Maybe (fromMaybe)
+import Data.Set (Set)
```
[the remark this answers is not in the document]
csasarak: .hlint.yaml only applies this for imports that are already qualified. Changing this for the whole project is out of scope for this PR.

docs/references/strategies/languages/maven/mavenplugin.md
```diff
@@ -22,16 +22,25 @@ Find `pom.xml` files, and treat those as maven projects. Skip all subdirectories
...
-their imediate deps marked as direct. This is because we don't want to include the users' projects in graphs, but
-do want to be able to analyze the things that they depend on.
+Submodule discovery is performed by parsing the project's POM files and building
+a reachability graph (POM closure). Submodules are identified from this closure
+rather than by running a separate Maven reactor invocation. We then generate a
+graph that excludes all of those submodules, but with their immediate deps marked
+as direct. This is because we don't want to include the users' projects in graphs,
+but do want to be able to analyze the things that they depend on.
+
+A second `:graph` (per-module) invocation is run with the command `mvn com.github.ferstl:depgraph-maven-plugin:4.0.1:graph -DgraphFormat=json -DmergeScopes -DshowDuplicates=true -DoutputFileName=fossa-depgraph-verbose.json`
```
nficca, line 39: [q] Given that we already vendor this jar, would it be a stretch to suggest that we just patch it such that the normal `:aggregate` mode doesn't hide duplicates?

src/Strategy/Maven/PluginStrategy.hs
```diff
@@ -147,8 +166,15 @@ instance ToDiagnostic MayIncludeSubmodule where
...
-buildGraph reactorOutput PluginOutput{..} =
+--
+-- TODO(#maven-parentless-modules): 'knownSubmodules' is derived from the POM
+-- closure graph, which builds edges only from <parent> elements. A module
+-- listed in <modules> but lacking a <parent> element (legal Maven) will not
+-- appear here, so it won't be promoted to direct and its verbose-graph
+-- duplicate edges may be dropped. This is a rare case; fixing it requires
+-- seeding <modules> edges into the closure graph in Pom/Resolver.hs.
+buildGraph :: Set Text -> PluginOutput -> Graphing MavenDependency
+buildGraph knownSubmodules PluginOutput{..} =
```
nficca: Was this also a limitation when we used the `reactor` mode of the plugin? Because if it wasn't this would introduce a regression right?

docs/references/strategies/languages/maven/mavenplugin.md
csasarak: I'd prefer not to maintain a fork of a third-party package if we can. Unless I'm misunderstanding your suggestion?

nficca: I only suggest it because we haven't updated our vendored copy in over 4 years, and the package itself had its last update over 3 years ago. So my impression is that we'd likely have no maintenance or only minimal maintenance. 

However, I'm also assuming that a patch to the plugin would make the implementation here far simpler, but if that's not the case than ignore me. Either way I don't have a problem with going this route, was just curious.

csasarak: I think that's fair, however this doesn't add an additional run of the plugin so I wouldn't consider it a performance regression. Updating the plugin seems like a good choice, but also opens a can of worms I'd rather not open on this specific PR since we have customers that will benefit from it as it is now. I will create a ticket for upgrading the plugin and we can look at it separately.

Is that an OK compromise for you?

nficca: Yes that's totally reasonable :+1:

src/Strategy/Maven/PluginStrategy.hs
csasarak: I had a closer look at this and it is indeed a regression. The most recent set of commits should fix it as well as add tests for this case.

nficca (approved):
I don't have concerns with this, it looks solid to me.

I notice the windows tests are failing but I'm guessing those are test issues and not a flaw with this fix itself. I have no issues merging once those are fixed.

csasarak:
@nficca Yeah, I'm working on getting those fixed up now

PR #1734, opened 2026-07-23
[ANE-3062] Support workspace target scoping for npm v3 lockfiles

nficca:
# Overview

For an npm-workspaces monorepo with a single root `package-lock.json` (lockfileVersion 3), `--only-target 'npm@./:<workspace>'` had no effect: every workspace target returned the same full graph for the whole repo. v1 and v2 lockfiles already scoped to the selected targets; v3 analysis ignored them.

This threads the selected build targets into the v3 analyzer: each target is resolved to its workspace via `package.json`, and the graph is pruned to the union of the selected targets' closures. Scoping to a workspace does not fold in the root package's own dependencies, matching v1 and yarn; select the root target alongside the workspace for both.

## Acceptance criteria

`fossa analyze --only-target 'npm@./:<workspace>'` on a v3-lockfile monorepo returns only that workspace's dependencies, resolved from the shared lockfile.

## Testing plan

1. Build a two-workspace monorepo with distinct deps and a v3 lockfile (needs npm 9+ so the lockfile is v3):
   ```sh
   mkdir -p /tmp/mono/packages/pkg-a /tmp/mono/packages/pkg-b
   echo '{"name":"mono","workspaces":["packages/*"]}' > /tmp/mono/package.json
   echo '{"name":"pkg-a","version":"1.0.0","dependencies":{"is-odd":"3.0.1"}}' > /tmp/mono/packages/pkg-a/package.json
   echo '{"name":"pkg-b","version":"1.0.0","dependencies":{"is-even":"1.0.0"}}' > /tmp/mono/packages/pkg-b/package.json
   npm install --package-lock-only --prefix /tmp/mono
   ```

2. From the fossa-cli checkout, scope to each workspace and compare direct deps:
   ```sh
   cabal run fossa -- analyze -o --only-target 'npm@./:pkg-a' /tmp/mono 2>/dev/null | jq '[.sourceUnits[].Build.Imports[]]'
   cabal run fossa -- analyze -o --only-target 'npm@./:pkg-b' /tmp/mono 2>/dev/null | jq '[.sourceUnits[].Build.Imports[]]'
   ```

3. Against `master`, both print the same full set. With this branch, `pkg-a` prints `["npm+is-odd$3.0.1"]` and `pkg-b` prints `["npm+is-even$1.0.0"]`.

## Risks

Workspace names are matched by the workspace `package.json` name, not the lockfile's `name` field, which npm omits when a folder basename equals its package name. A dependency reached only through a sibling workspace that the selected workspace links to is not pulled in; this is documented in `npm-lockfile.md`. The no-filter path and non-workspace projects are unchanged.

## Metrics

None.

## References

- [ANE-3062](https://fossa.atlassian.net/browse/ANE-3062): Target-level dependency scoping for NPM lockfile v3. Scopes to the selected targets only; unlike the ticket's expected-(b), a workspace scope does not auto-include root deps (select the root target for those).

## Checklist

Docs change is a note in the existing `npm-lockfile.md` (already in the ToC); no new page, schema, or subcommand-option changes. Changelog entry under a new `## 3.17.15` section since this is intended to ship as the next release.

[ANE-3062]: https://fossa.atlassian.net/browse/ANE-3062

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## 3.17.15
+
+- npm: Scoping a scan to a single workspace build target (`--only-target`) now works for v3 `package-lock.json` files (npm v9+). Previously each workspace returned the full repo-wide dependency set instead of its own. ([#1734](https://github.com/fossas/fossa-cli/pull/1734))
+
 ## 3.17.14
 
 - Licensing: Detect Unicode 3.0 license. ([#1731](https://github.com/fossas/fossa-cli/pull/1731))
--- a/docs/references/strategies/languages/nodejs/npm-lockfile.md
+++ b/docs/references/strategies/languages/nodejs/npm-lockfile.md
@@ -36,10 +36,11 @@ included in the analysis.
 When no filtering is applied, all targets are selected and all dependencies
 from every workspace package are included in the analysis.
 
-> Note: Target-level dependency scoping is only supported for lockfile version 1
-> (npm v5–v6) and version 2 (npm v7–v8). Version 3 lockfiles (npm v9+) will
-> show workspace build targets, but filtering to specific targets does not yet
-> scope the dependency results.
+> Note: Selecting a workspace target scopes the analysis to that workspace's own
+> declared dependencies and their transitive closure. Dependencies contributed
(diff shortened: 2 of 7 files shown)


nficca:
Closing this in favour of #1732, which is more robust.

PR #1740, opened 2026-08-05
Detect Go module dependencies from binaries in container images

saramaebee:
## Summary

Go services shipped as bare binaries in distroless/scratch/chainguard images are currently invisible to `fossa container analyze`: container analysis is static-only, Go strategies are excluded from the container discovery list, and no binary inspection exists. But Go binaries embed their complete module list (the data `go version -m` reads), so the dependency data is sitting in the artifact — syft already reads it.

This PR closes that gap end-to-end, prompted by an enterprise customer for whom this is their top-ranked ask (support has a validated repro + escalation context; ask @saramaebee for details):

- **millhone**: new `go_buildinfo` parser (pure Rust, no new binary-format crates) for the Go >= 1.18 inline buildinfo format; older/UPX-packed binaries skip gracefully with a debug log. Wired into the existing `analyze-container` layer walk behind a cheap magic-byte sniff. Output gains an additive `discovered_go_binaries` field (skew-tolerant in both directions).
- **millhone**: transparently gunzip layer blobs. Modern `docker save` emits OCI layout with gzip-compressed layers, which also silently broke JAR-in-container analysis when millhone runs against such archives directly.
- **CLI**: new `App.Fossa.Container.Sources.GoBinary` converts discovered modules into real `go+path$version` locators in per-layer `srcUnits`, reusing `Strategy.Go.Gomod.parsePackageVersion` so pseudo-versions normalize to commit hashes exactly like go.mod scans. **No backend changes required** — this rides the same locator channel go.mod analysis uses.
- **CLI**: the feature is always-on — no flag. (Earlier revisions gated reporting behind `--experimental-enable-go-binary-discovery`, but the buildinfo scan runs on every container scan regardless since millhone already walks the layers for JAR analysis, so the flag only gated reporting of the results. Per review discussion we dropped it. Note there is no scan-side opt-out: the gobinary units bypass target/path filters — see the PR comment on filtering for details. This matches how JAR-in-container observations already ship.)
- **Docs**: new "Container Go binary analysis" section in the container scanner reference (mechanism, Go >= 1.18 + main-module limitations), a mention in the container subcommand reference, a support-table row, and a changelog entry.

PoC scope, deliberately deferred: committed container-tar test fixtures, `fourmolu` pass, and the pre-1.18 pointer-format encoding.

## Test plan

- [x] Rust: 11 tests pass (`cargo test -p millhone`) — 9 parser unit tests on synthetic buildinfo buffers (replacements, `+incompatible`, pointer-format skip, truncation, sentinel validation) + 2 updated expected-JSON fixture tests
- [x] Haskell: 7 new `GoBinarySpec` examples + full Container suite pass (`cabal test unit-tests`), re-verified after making discovery always-on
- [x] Manual e2e on two repro images containing a Go binary with known modules (uuid v1.6.0, logrus v1.9.3, x/sys pseudo-version):
  - `FROM scratch`: previously a fully empty scan; now reports `go+github.com/google/uuid$v1.6.0`, `go+github.com/sirupsen/logrus$v1.9.3`, `go+golang.org/x/sys$c0bba94af5f8`
  - `FROM cgr.dev/chainguard/static`: Wolfi apk system deps and layer placement unchanged, Go modules reported from the app layer
- [x] Backend sanity check that `go+` locators with commit-hash revisions resolve as expected on a test project (same shape go.mod scans emit today)

### 1. Get the PoC binary

Every push to this PR builds release binaries; download the artifact for your platform from the PR's `Build` workflow run (Artifacts section at the bottom of the run summary), or:

```bash
run_id=$(gh run list --repo fossas/fossa-cli --branch poc/container-go-binary-analysis \
  --workflow build-all.yml --limit 1 --json databaseId --jq '.[0].databaseId')
gh run download "$run_id" --repo fossas/fossa-cli -n macOS-arm64-binaries -D fossa-poc
chmod +x fossa-poc/fossa
```

(Other artifact names: `Linux-binaries`, `Linux-arm-binaries`, `macOS-intel-binaries`, `Windows-binaries`.)

### 2. Build a repro image

Any image containing a Go >= 1.18 module-built binary works. A minimal one:

```dockerfile
# Dockerfile
FROM golang:1.25-alpine AS build
WORKDIR /src
RUN go mod init example.com/repro-app \
 && go get github.com/google/uuid@v1.6.0 \
 && printf 'package main\n\nimport (\n\t"fmt"\n\n\t"github.com/google/uuid"\n)\n\nfunc main() { fmt.Println(uuid.NewString()) }\n' > main.go \
 && CGO_ENABLED=0 go build -o /app .
FROM scratch
COPY --from=build /app /app
ENTRYPOINT ["/app"]
```

```bash
docker build -t go-binary-repro . && docker save go-binary-repro -o repro-image.tar
```

(Swap `FROM scratch` for `cgr.dev/chainguard/static` to also see system deps alongside the Go modules.)

### 3. Run the PR binary

```bash
fossa-poc/fossa container analyze repro-image.tar --output \
  | jq '.image.layers[].srcUnits[] | select(.Type == "gobinary") | {Name, Imports: .Build.Imports}'
```

Expected: a `gobinary` source unit for the app binary whose imports include `go+github.com/google/uuid$v1.6.0`.

### 4. Controls

- **PR binary**: discovery is always-on — no flag needed; passing the former `--experimental-enable-go-binary-discovery` flag now errors as an unknown option (it existed only in earlier revisions of this PR, never in a release).
- **Official release** (`brew install fossa` or the install script): a plain `fossa container analyze repro-image.tar --output` of the scratch image reports no source units at all — the status quo this PR fixes.

https://claude.ai/code/session_01XzWtfpTqJtLyL48WfB35gG


src/App/Fossa/Container/Sources/DockerArchive.hs
```diff
@@ -150,20 +155,33 @@ analyzeFromDockerArchive useGitBackedCargo systemDepsOnly filters withoutDefault
...
-  let baseScanImageLayer = ContainerScanImageLayer baseDigest baseUnits baseObservations
+  let goBinaryDiscoveryEnabled = fromFlag GoBinaryDiscovery goBinaryDiscovery
+      (baseGoBinaries, otherGoBinaries) =
+        maybe ([], []) (partitionByBaseLayer . discoveredGoBinaries) observations
+      (baseGoUnits, otherGoUnits) =
+        if goBinaryDiscoveryEnabled
+          then (goBinariesToSourceUnits baseGoBinaries, goBinariesToSourceUnits otherGoBinaries)
+          else ([], [])
+
+  unless (goBinaryDiscoveryEnabled || (null baseGoBinaries && null otherGoBinaries)) $
```
saramaebee: Important callout: right now with this code, we're sniffing and parsing Go binaries on every container scan, up to 512mb in memory, even with flag disabled.  The flag just determines whether the results are thrown away. 

This enables us to give an alert saying "hey you have a go binary in here, use this CLI flag to capture dependencies from it!"

There _is_ a cheaper check we can do, running `is_candidate_binary()` from `extlib/millhone/src/cmd/go_buildinfo.rs` on the first 64 bytes, or alternatively we can just make it explicitly opt-in with no magic hint

src/Container/Types.hs
```diff
@@ -145,11 +149,47 @@ newtype JarObservation = JarObservation
...
+      <$> o .: "path"
+      <*> o .: "go_version"
+      <*> o .:? "main_module"
+      <*> o .: "modules"
+
 -- | Output parse type for millhone.
-newtype DiscoveredJars = DiscoveredJars
+data DiscoveredJars = DiscoveredJars
   { discoveredJars :: Map.Map LayerPath [JarObservation]
+  , discoveredGoBinaries :: Map.Map LayerPath [DiscoveredGoBinary]
```
saramaebee: this is no longer just `DiscoveredJars`, should I rename to `DiscoveredBinaries`?

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -106,16 +144,51 @@ fn unpack(path: &PathBuf) -> Result<Archive<File>> {
     Ok(tar::Archive::new(file))
 }
 
+/// Layer blobs are plain tars in the legacy `docker save` layout, but gzipped
+/// tars in the OCI layout modern Docker emits. Peek at the magic bytes and
+/// transparently decompress when needed.
```
saramaebee, line 150: Note: this changes millhone behavior, but not shipped CLI behavior. Prior to this PR millhone only understands layers that are plain uncompressed tars. Modern `docker save` (containerd image store — default for fresh Docker Desktop ≥4.34 and Engine ≥29 installs) emits OCI layout with gzip-compressed layer blobs.

In the released CLI this gap is unreachable: circe re-exports every image into a normalized plain-tar archive before millhone runs, and the circe-fallback path dies earlier in the Haskell tarball parser on gzipped layers anyway (whole scan fails loudly). So this is not fixing a live customer-facing regression.

What it does do is make millhone self-sufficient when invoked directly on a raw archive — which this PR's Go-binary flow and its testing rely on, and which matters if millhone ever runs outside circe's normalization. The fix is a magic-byte peek in `scan_layer` with transparent gunzip. (zstd layers are still unhandled; containerd can emit those too. Can add if we want completeness.)

Happy to pull this into a separate PR if preferred by ANE.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -51,31 +60,58 @@ struct OciManifest {
...
-/// Extracts the container (saved via `docker save`) and finds JAR files inside any layer.
-/// For each one found, fingerprints it and reports all those fingerprints along with their
+/// Extracts the container (saved via `docker save`) and finds JAR files and Go
+/// binaries inside any layer. JARs are fingerprinted; Go binaries have their
+/// embedded buildinfo module list parsed. Everything is reported along with its
 /// layer and path.
 #[tracing::instrument]
-fn jars_in_container(image_path: &PathBuf) -> Result<JarAnalysis> {
-    // Visit each layer and fingerprint the JARs within.
+fn jars_in_container(image_path: &PathBuf) -> Result<ContainerAnalysis> {
```
csasarak: This name and its doc string is no longer accurate.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -106,16 +144,51 @@ fn unpack(path: &PathBuf) -> Result<Archive<File>> {
...
+    let mut filled = 0;
+    while filled < magic.len() {
+        match reader.read(&mut magic[filled..]) {
+            Ok(0) => break,
+            Ok(n) => filled += n,
+            Err(e) if e.kind() == std::io::ErrorKind::Interrupted => continue,
+            Err(e) => return Err(e).context("detect layer compression"),
+        }
+    }
+    let is_gzip = filled == 2 && magic == [0x1f, 0x8b];
```
csasarak: It'd be best to put these in a `const`/`static` array with a descriptive name.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -106,16 +144,51 @@ fn unpack(path: &PathBuf) -> Result<Archive<File>> {
...
+/// tars in the OCI layout modern Docker emits. Peek at the magic bytes and
+/// transparently decompress when needed.
 #[tracing::instrument(skip(entry))]
-fn jars_in_layer(entry: Entry<'_, impl Read>) -> Result<Vec<DiscoveredJar>> {
+fn scan_layer(
+    entry: Entry<'_, impl Read>,
+) -> Result<(Vec<DiscoveredJar>, Vec<DiscoveredGoBinary>)> {
+    let mut reader = BufReader::new(entry);
+    // Read exactly two bytes rather than trusting a single fill_buf to return
+    // that many, then stitch them back onto the front of the stream.
```
csasarak: Why wouldn't you trust it? The `BufReader` is likely reading more than two already.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -23,6 +25,13 @@ pub struct Subcommand {
 }
 
 const JAR_OBSERVATION: &str = "v1.discover.binary.jar";
+const GO_BINARY_OBSERVATION: &str = "v1.discover.binary.go";
+
+/// Only sniff regular files at least this large; Go binaries are never tiny.
+const MIN_GO_BINARY_SIZE: u64 = 4096;
+
+/// Skip candidates larger than this rather than buffering them in memory.
+const MAX_GO_BINARY_SIZE: u64 = 512 * 1024 * 1024;
```
csasarak: [nit] Typically we use `usize` for sizes. If `u64` is what some API requires you can leave these as is.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -139,7 +212,41 @@ fn jars_in_layer(entry: Entry<'_, impl Read>) -> Result<Vec<DiscoveredJar>> {
...
+        return None;
+    }
+    let size = entry.header().size().unwrap_or(0);
+    if size < MIN_GO_BINARY_SIZE {
+        return None;
+    }
+    if size > MAX_GO_BINARY_SIZE {
+        warn!(?path, size, "skipped: candidate binary exceeds size limit");
+        return None;
+    }
```
csasarak: [Question] Is there a way to do this in a streaming fashion where we don't need to keep the whole entry in memory at once?

extlib/millhone/src/cmd/go_buildinfo.rs
```diff
@@ -0,0 +1,362 @@
...
+//! Go binaries built with module support embed the module list of everything
+//! linked into the binary (the same data `go version -m` prints). For Go >= 1.18
+//! the data is stored inline as varint-length-prefixed strings following a
+//! 16-byte-aligned magic header; this module supports only that inline format.
+//! Older binaries (Go < 1.18) use a pointer-based encoding and are skipped.
+
+use serde::Serialize;
+
+/// The 14-byte magic marking the start of the buildinfo header.
+const BUILDINFO_MAGIC: &[u8] = b"\xff Go buildinf:";
```
csasarak: [nit] Prefer static for a byte buffer: `const`s get inlined.

extlib/millhone/src/cmd/go_buildinfo.rs
```diff
@@ -0,0 +1,362 @@
+//! Extract embedded Go build information ("buildinfo") from compiled Go binaries.
+//!
+//! Go binaries built with module support embed the module list of everything
+//! linked into the binary (the same data `go version -m` prints). For Go >= 1.18
+//! the data is stored inline as varint-length-prefixed strings following a
+//! 16-byte-aligned magic header; this module supports only that inline format.
+//! Older binaries (Go < 1.18) use a pointer-based encoding and are skipped.
```
csasarak, line 7: If there's a spec somewhere that describes the header format then we should reference it here.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -139,7 +212,41 @@ fn jars_in_layer(entry: Entry<'_, impl Read>) -> Result<Vec<DiscoveredJar>> {
...
+
+    debug!(?path, "candidate binary; scanning for Go buildinfo");
+    let mut contents = header.to_vec();
+    if let Err(e) = entry.read_to_end(&mut contents) {
+        warn!(?path, "failed to read candidate binary: {e:?}");
+        return None;
+    }
+    parse_go_buildinfo(&contents).map(|info| DiscoveredGoBinary::new(path.to_path_buf(), info))
```
csasarak: Does `parse_go_buildinfo` expect to take a file without its header info? Does `Entry::read_exact` actually advance some internal value? If it does we should do one of the following:

- Document 
- Create an actual type that can manage the file handle/byte buffers
- Make `go_buildinfo` a submodule of this one and declare it `pub (super)` so that it can't be accessed casually.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -404,6 +517,15 @@ mod tests {
...
             ]
+          }},
+          "discovered_go_binaries": {{
+            "88a896b18358cbccbf66cc1c3dcd0d2d61504c5bf41284c551e47f230675534f/layer.tar": [],
+            "2cd0dec90e9f3f920397ed6bf0ba740493620a99bb20b79d2c4c8159948439e4/layer.tar": [],
+            "5a99cb0cd20c916ca7444b625ff06e3afe6a1b4349c44c3ba11eb054daf5fda4/layer.tar": [],
+            "9fea496e8349f2c33fe177df27e4369f08cff62ad40168183493d9de3d6832e5/layer.tar": [],
+            "c65b9197c11847b3f36c822b9c57f417af6f721ae6719f8eb3bd334c3516796e/layer.tar": [],
+            "792ccfdf114be140500ea1d6b99ae7ff0cae6a19b247f886328d4b8a01b8869c/layer.tar": [],
+            "d8abb0d1e8708fb1b5b79bcdd098898becfe22019d6b70781974743a90415724/layer.tar": []
```
csasarak, line 525: The name of this function (if my indentation is right) is wrong now.

extlib/millhone/src/cmd/go_buildinfo.rs
```diff
@@ -0,0 +1,362 @@
...
+        main_module: None,
+        modules: Vec::new(),
+    };
+    if info.go_version.is_empty() {
+        return None;
+    }
+
+    if let Some(modinfo) = modinfo {
+        parse_modinfo(&String::from_utf8_lossy(modinfo), &mut info);
+    }
```
csasarak: It's generally better not to use mutability when it's so easy to avoid. Make `parse_modinfo` return its result and then: 

```rust
let info = {
    let version = String::from_utf8_lossy(go_version)
                            .is_some_and(|s| ! s.is_empty() )?;
    let info = parse_modinfo(&String::from_utf8_lossy(modinfo);
    GoBuildInfo { ... }
};
```

The reason that I say that is declaring something `mut` basically means that you have to consider everything that might happen to it after you declare it. That's a lot of mental overhead as opposed to knowing "it's this value and always will be." 

In this case, it's perhaps not so bad since the function is small but it's a good habit to exercise everywhere IMO.

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -139,7 +212,41 @@ fn jars_in_layer(entry: Entry<'_, impl Read>) -> Result<Vec<DiscoveredJar>> {
...
+    }
+
+    let mut header = [0u8; 64];
+    if let Err(e) = entry.read_exact(&mut header) {
+        debug!(?path, "skipped: failed to read header: {e:?}");
+        return None;
+    }
+    if !is_candidate_binary(&header) {
+        return None;
+    }
```
csasarak: What's the difference between this header and `entry.header()`?

Changelog.md
```diff
@@ -1,5 +1,9 @@
 # FOSSA CLI Changelog
 
+## Unreleased
+
+- Container scanning: Add `--experimental-enable-go-binary-discovery` flag to `fossa container analyze`. When enabled, Go module dependencies embedded in Go binaries (built with Go >= 1.18) found in container image layers are reported as regular Go dependencies, supporting images without package manager metadata such as `scratch` and distroless images ([#1740](https://github.com/fossas/fossa-cli/pull/1740))
```
csasarak: Bump this for release.

csasarak (changes requested):

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -106,16 +144,51 @@ fn unpack(path: &PathBuf) -> Result<Archive<File>> {
...
+        }
+    }
+    let is_gzip = filled == 2 && magic == [0x1f, 0x8b];
+    let reader = std::io::Cursor::new(magic)
+        .take(filled as u64)
+        .chain(reader);
+    if is_gzip {
+        scan_layer_tar(flate2::read::GzDecoder::new(reader))
+    } else {
+        scan_layer_tar(reader)
```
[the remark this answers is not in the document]
saramaebee: The Go path no longer buffers whole files: as of 62709d31, `scan_go_buildinfo` streams with a rolling 64KiB window and reads only the two inline strings (each capped) once the magic is found. The 512MiB size cap is gone too — it existed to bound buffering, and streaming made it unnecessary. Deliberately not adding a cumulative decompressed-layer budget: the JAR path has always buffered entries whole, and tar must read through layer bytes to advance regardless.

extlib/millhone/src/cmd/go_buildinfo.rs
```diff
@@ -0,0 +1,362 @@
...
+pub fn is_candidate_binary(header: &[u8]) -> bool {
+    if header.len() < 4 {
+        return false;
+    }
+    match &header[..4] {
+        b"\x7fELF" => true,
+        // Mach-O thin: feedface / feedfacf, both byte orders.
+        [0xfe, 0xed, 0xfa, 0xce | 0xcf] | [0xce | 0xcf, 0xfa, 0xed, 0xfe] => true,
+        // Mach-O fat (universal): cafebabe / bebafeca.
+        [0xca, 0xfe, 0xba, 0xbe] | [0xbe, 0xba, 0xfe, 0xca] => true,
```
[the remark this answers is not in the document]
saramaebee: Added `cafebabf` in both byte orders in 62709d31.

extlib/millhone/src/cmd/go_buildinfo.rs
```diff
@@ -0,0 +1,362 @@
...
+/// Decode an unsigned LEB128 varint. Returns the value and bytes consumed.
+fn read_uvarint(buf: &[u8]) -> Option<(u64, usize)> {
+    let mut value: u64 = 0;
+    for (i, byte) in buf.iter().take(10).enumerate() {
+        value |= u64::from(byte & 0x7f) << (7 * i);
+        if byte & 0x80 == 0 {
+            return Some((value, i + 1));
+        }
+    }
+    None
```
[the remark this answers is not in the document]
saramaebee: The premise here is off — the loop already accepts a terminating 0x00/0x01 tenth byte (`take(10)` covers i=0..9). The real adjacent bug was that an overflowing tenth byte (>1) was silently truncated rather than rejected. Fixed in 62709d31 with the same check Go's encoding/binary uses (`i == 9 && byte > 1`), plus tests.

src/App/Fossa/Container/Sources/DockerArchive.hs
```diff
@@ -17,7 +17,9 @@ import App.Fossa.Analyze.Types (
   DiscoveredProjectScan (..),
  )
 import App.Fossa.Config.Analyze (StrategyConfig (StrategyConfig), UseGitBackedCargoLocators (..), WithoutDefaultFilters (..))
+import App.Fossa.Config.Container.Analyze (GoBinaryDiscovery (GoBinaryDiscovery))
 import App.Fossa.Container.Sources.Discovery (layerAnalyzers, renderLayerTarget)
+import App.Fossa.Container.Sources.GoBinary (goBinariesToSourceUnits)
```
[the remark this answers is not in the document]
saramaebee: Declining this one: the style guide only calls for qualified imports on Prelude-conflicting names (docs/contributing/STYLE-GUIDE.md, Naming), and the surrounding container sources all import these modules unqualified with explicit import lists — this PR matches that convention.

src/App/Fossa/Container/Sources/GoBinary.hs
```diff
@@ -0,0 +1,86 @@
...
+import Data.List (nub)
+import Data.Maybe (mapMaybe, maybeToList)
+import Data.SemVer qualified as SemVer
+import Data.SemVer.Internal (Version (..))
+import Data.Text (Text)
+import Data.Text qualified as Text
+import Srclib.Types (Locator (..), SourceUnit (..), SourceUnitBuild (..), SourceUnitDependency (..), textToOriginPath)
+import Strategy.Go.Gomod (PackageVersion (..), parsePackageVersion)
+import Text.Megaparsec (parseMaybe)
+import Types (GraphBreadth (..))
```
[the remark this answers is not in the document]
saramaebee: Declining per the same reasoning as the sibling comment: STYLE-GUIDE.md only mandates qualifying Prelude-conflicting names, and unqualified explicit import lists are the surrounding convention.

src/App/Fossa/Container/Sources/GoBinary.hs
```diff
@@ -0,0 +1,86 @@
...
+      }
+
+normalizeVersion :: Text -> Maybe Text
+normalizeVersion version
+  | Text.null version || version == "(devel)" = Nothing
+  | otherwise = case parseMaybe (parsePackageVersion id) version of
+      Just (Pseudo commitHash) -> Just commitHash
+      Just (Semantic semver) -> Just ("v" <> SemVer.toText semver{_versionMeta = []})
+      Just (NonCanonical v) -> Just v
+      Nothing -> Just version
```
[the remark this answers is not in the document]
saramaebee: Done in eaa4f07b — rewritten with if/then/else.

src/App/Fossa/Container/Sources/DockerArchive.hs
saramaebee: Resolving: keeping sniff-always + the hint. The sniff is effectively free (tar reads through entry bytes to advance regardless), and the scan now streams (62709d31), so flag-off cost is bounded to a rolling 64KiB window — results are computed and discarded, which is what powers the hint.

src/Container/Types.hs
saramaebee: Renamed to `DiscoveredBinaries` in eaa4f07b.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Keeping it in this PR since it was reviewed on its merits — happy to split if ANE prefers.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Renamed to `binaries_in_container` in 62709d31.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Done in 62709d31 — `GZIP_MAGIC`.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: The comment was misleading — reworded in 62709d31. It's not about BufReader's internal buffer: a single `Read::read` may legally return fewer than 2 bytes even when more remain, so the loop guarantees both magic bytes (or EOF) before deciding.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Keeping u64 per your parenthetical — it's what tar's `Header::size()` returns. (The MAX cap is gone entirely as of 62709d31 now that the scan streams; only `MIN_GO_BINARY_SIZE` remains.)

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Implemented in 62709d31 — single pass, no seeking: `scan_go_buildinfo` keeps a rolling 64KiB window to find the aligned magic, then reads just the two varint-prefixed strings. Whole-entry buffering for Go binaries is gone, and the 512MiB cap went with it since it only existed to bound buffering.

extlib/millhone/src/cmd/go_buildinfo.rs
saramaebee: Done in 62709d31.

extlib/millhone/src/cmd/go_buildinfo.rs
saramaebee: Done in 62709d31 — linked `debug/buildinfo/buildinfo.go`; the format has no standalone spec, that reader is the reference implementation.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Good catch — it was a real trap: `read_exact` does advance the `Entry` (forward-only reader), and correctness depended on the caller stitching the prefix back on. The streaming rewrite in 62709d31 absorbs that: `maybe_go_binary` chains the sniffed prefix back onto the stream in one expression, `scan_go_buildinfo` documents that it needs bytes from file offset 0, and `parse_go_buildinfo` is no longer `pub`.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Renamed to `it_analyzes_nested_jars_image` in 62709d31.

extlib/millhone/src/cmd/go_buildinfo.rs
saramaebee: Done in 62709d31 — `parse_modinfo` returns `(main_module, modules)` and `GoBuildInfo` is built in one expression.

extlib/millhone/src/cmd/analyze_container.rs
saramaebee: Just a badly named local — the file's first 64 bytes, nothing to do with `entry.header()`. Renamed to `prefix` in 62709d31.

Changelog.md
saramaebee: Leaving the bump to you since you're stacking your changes before merge — or say the word and I'll push it.

extlib/millhone/src/cmd/analyze_container.rs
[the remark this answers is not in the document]
saramaebee: Declining the layer budget for this PR: gzip expansion is bounded by DEFLATE's ~1032:1 max ratio, so output is finite and proportional to input — and since decompressed bytes stream through the tar walk without accumulating, the exposure is scan *time*, not memory. The input is also a user-supplied local image scanned on the user's own machine (the same expansion cost `docker load` pays), so a bomb here is self-inflicted. Meanwhile any cap value risks real breakage: ML base images legitimately have layers that decompress past 10GB, and failing those scans is worse than a slow one. The genuine remaining gap is the pre-existing JAR path's `read_to_end` (no size gate), which predates this PR — that's worth a follow-up if ANE wants one. Happy to add a generous per-layer meter in a follow-up if the team disagrees.

Changelog.md
saramaebee: Done in f6c8aadc — bumped to 3.18.0, following the 3.16.0 precedent of minting a minor version for new analysis support.

saramaebee:
**Note on opting out (correcting a claim from the review discussion):**

During review we said that anyone not wanting Go binary deps could filter them out by target. Having checked the code, that's not actually true today:

- The `gobinary` units are appended to `srcUnits` after and outside the `applyFiltersToProject` gate (`DockerArchive.hs` — `baseUnits <> baseGoUnits`), so `AllFilters` never sees them.
- `fossa container analyze` exposes no target/path filter flags (filtering is config-file only), and `gobinary` is not a `DiscoveredProjectType`, so a `.fossa.yml` `targets.exclude` entry has nothing to match. The unit's path is an in-image layer path, outside the domain the path filters operate on.

So there is currently no scan-side opt-out. This has direct precedent — JAR-in-container observations are also always-on and bypass filters the same way (`analyzeContainerJars` takes no filters) — but it's worth stating plainly since the review discussion assumed otherwise.

If an opt-out is ever needed, wiring `gobinary` units through the filter engine (or adding a dedicated exclusion) would be a follow-up change; deliberately not built speculatively here.

docs/references/subcommands/container/scanner.md
```diff
@@ -240,6 +266,7 @@ The following package managers are supported in container scanning:
 | Dart (pub)                           | :warning:          | [Dart](./../../strategies/languages/dart/pub.md)                 |
 | Maven                                | :warning:          | [Maven](./../../strategies/languages/maven/maven.md)             |
 | Java Jar Files                       | :white_check_mark: | [Container Jar Analysis](#container-jar-analysis)               |
+| Go Binaries                          | :white_check_mark: | [Container Go Binary Analysis](#container-go-binary-analysis) |
```
csasarak, line 269:
```suggestion
| Golang (Binaries)                          | :white_check_mark: | [Container Go Binary Analysis](#container-go-binary-analysis) |
```

extlib/millhone/src/cmd/analyze_container.rs
```diff
@@ -404,19 +514,168 @@ mod tests {
...
+    fn finds_buildinfo_deep_in_a_large_binary() {
+        // Buildinfo placed past the streaming scan window, so the rolling
+        // search (not just the sniffed prefix) has to find it.
+        let mut buf = vec![0u8; 256 * 1024];
+        buf[..4].copy_from_slice(b"\x7fELF");
+        buf.extend_from_slice(BUILDINFO_FIXTURE);
+        let discovered = sniff_first(&[("usr/bin/big", &buf)]).expect("should discover");
+        assert_eq!(discovered.go_version, "go1.25.6");
+        assert_eq!(discovered.modules.len(), 161);
+    }
```
csasarak, line 680: This all seems good - if it's easy I'd recommend maybe adding an on-disk fixture so you can test it end to end.

src/App/Fossa/Container/Sources/JarAnalysis.hs
```diff
@@ -37,7 +37,7 @@ millhoneJarAnalyzeCmd cmdPath imageTarFile =
     }
 
 -- | Analyze a container for Jar fingerprints using Millhone.
-analyzeContainerJars :: (Has Logger sig m, Has Exec sig m, Has (Lift IO) sig m, Has Diagnostics sig m) => Path Abs File -> m DiscoveredJars
+analyzeContainerJars :: (Has Logger sig m, Has Exec sig m, Has (Lift IO) sig m, Has Diagnostics sig m) => Path Abs File -> m DiscoveredBinaries
```
csasarak, line 40: [nit] A small comment explaining that this is also where we kick of the Go analysis would be good. Ideally we'd just remove the Jar specific nomenclature altogether. If you want to take a stab at that you can but I can also look at it some other time.

csasarak (approved):

PR #1750, opened 2026-08-20
[ANE-1349] Binary discovery support for whl files

GauravB159:
# Overview

Given a [whl](https://peps.python.org/pep-0427) file, we need to identify the dependency that is associated with it. Whl files are zip files that can be unzipped. They all require a “METADATA” file in their [dist-info directory](https://peps.python.org/pep-0427/#the-dist-info-directory).

This PR adds the capability to scan these .whl files on top of our existing binary analysis logic which only supported JAR files separately. One thing to note: JAR file analysis creates user defined deps rather than `mvn` deps so the logic needed to be extended a bit for `.whl` files to create `pip` deps instead of user defined deps. 

## Acceptance criteria

- [x] Scanning a directory with whl files will result in the dependency that corresponds to the whl files being shown in the UI

## Testing plan

Automated testing has been added. For manual testing, copying the automated testing fixtures, or downloading any .whl file and running a fossa analyze on it should suffice. 

```bash
fossa analyze . --experimental-enable-binary-discovery --debug
```

## Risks

Should be low risk since it's extending a feature and it's behind an experimental flag

## Metrics

Some logging has been added, if that surfaces in Datadog, we can monitor failures

## References

[ANE-1349](https://fossa.atlassian.net/browse/ANE-1349)

## Checklist

[ANE-1349]: https://fossa.atlassian.net/browse/ANE-1349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/App/Fossa/BinaryDeps/Whl.hs
```diff
@@ -0,0 +1,120 @@
...
+findDistInfoFolder :: (Has ReadFS sig m, Has Diagnostics sig m, Has Logger sig m) => Path Abs Dir -> m (Path Abs Dir)
+findDistInfoFolder archive = do
+  (dirs, _) <- listDir archive
+  logDebug $ "Listing folders in archive " <> viaShow (map (dropTrailingPathSeparator . toString . dirname) dirs)
+  logDebug $ "Listing folder matching in archive " <> viaShow (map (\d -> dirHasSuffix d [".dist-info"]) dirs)
+  fromMaybeText "Could not find *.dist-info folder in whl archive" $
+    find (\d -> dirHasSuffix d [".dist-info"]) dirs
+
+parseMetadata :: Text -> Map Text Text
+parseMetadata t = Map.fromList . map strip' . filter' $ map (Text.breakOn ":") (Text.lines t)
```
spatten, line 93: The metadata file contains the readme, which can contain a `:`. So this will break for any readme that contains a `:`

An example is the sympy readme: https://files.pythonhosted.org/packages/a2/09/77d55d46fd61b4a135c444fc97158ef34a095e5681d0a6c10b75bf356191/sympy-1.14.0-py3-none-any.whl

If you unzip that, you see lots of lines in sympy-1.14.0.dist-info/METADATA that contain `:`, and those will become keys

I think the fix is to just send the header to `parseMetadata`, which is everything before the first blank line

The value can also be multi-line, like this:

```
License: Copyright (c) 2010 Example
        Permission is hereby granted, free of charge, to any person...
        THE SOFTWARE IS PROVIDED "AS IS": WITHOUT WARRANTY OF ANY KIND
```

We should probably also parse that properly

The spec for this is apparently RFC 822, which is kind of dense. https://datatracker.ietf.org/doc/html/rfc822#section-3.2

spatten (approved):
I think the code is good, but I want to raise one thing.

The dependencies found by this PR don't show their path, and I think this is going to be confusing to a customer. For example, in this screenshot you can see where the jar file is from but not the pip dep:

Image

The only solution that I can think of is to make each .whl file its own source unit and set the OriginPaths field in it. If you do that, then you get the path:

Image

(I did this by just creating some SourceUnits and POSTing them directly to Core. I didn't write any code)

I'm not sure if this is a deal-breaker, but it's definitely a nice to have for this feature

src/App/Fossa/BinaryDeps/Whl.hs
GauravB159: @spatten We really only care about the Name and Version. Even the License is parsed but it's not really used. So for now, I think it's fine to have this simple parsing. What do you think?

GauravB159: Also for the main comment, adding an individual source unit for each dep is a bit more involved, I'd rather hold off on it for now and see if anyone even uses this feature. We can add it in the future if it's needed.

spatten: Both of those decisions make sense to me. Let's ship this :)

PR #1768, opened 2026-09-03
Maven: report declared dependencies as direct in static analysis

spatten:
# Overview

Static Maven analysis reported the project itself as the only direct dependency, and everything the project declared as transitive. In a multi-module project the root pom and every submodule were reported as dependencies too. Analysis using `mvn` was already correct, so this only showed up without `mvn` on `PATH`, under `--static-only-analysis`, and in container scanning.

The dynamic tactics already handle this: `Plugin.buildGraph` marks the project's own packages direct and `shrinkRoots` removes them. Static analysis never got the equivalent step — the TODO marking it as unfinished was deleted in #1331, a week before #1339 added `shrinkRoots` to the dynamic path. This takes the same two steps in `getStaticAnalysis`.

Both steps are needed. `shrinkRoots` alone removes only the root pom and would promote the submodules to direct dependencies rather than dropping them. A `shrink` alone leaves a submodule-filtered graph with no direct dependencies at all, because submodule filtering deletes the root pom's node without rewiring the edges through it — which is a second bug this fixes.

## Acceptance criteria

- A Maven project analyzed without `mvn` reports the dependencies it declares as direct, and does not report itself.
- A multi-module project reports no first-party module as a dependency.
- `--only-target 'maven@./:group:module'` under static analysis reports that module's dependencies as direct, instead of reporting none.
- Analysis with `mvn` present is unchanged.

## Testing plan

Given a `pom.xml` declaring `junit:junit:4.13.2` and `org.apache.commons:commons-lang3:3.12.0`:

1. `fossa analyze -o --static-only-analysis` and read `sourceUnits[0].Build.Imports`.
2. Before: `["mvn+com.example:demo$1.0"]` — the project itself. After: the two declared dependencies.
3. `fossa analyze -o` with `mvn` on `PATH` gives the same two, before and after.

For the multi-module case, add two submodules each declaring one dependency. Before, `Imports` is the root pom alone and the graph carries all three first-party coordinates; after, `Imports` is the three declared dependencies and no first-party coordinate appears.

`cabal test unit-tests --test-options='--match "static analysis of a"'` covers all of the above against real pom fixtures. All five fail on `master` with the symptoms described.

## Risks

**This changes what FOSSA reports for existing projects.** Dependencies that were transitive become direct, and the user's own modules stop being reported as dependencies. That is the correction, but it will show as a diff on any project analyzed statically.

**The same bug exists in `Strategy/Scala.hs:200`**, where `analyzeWithPoms` calls `Pom.analyze'` with no shrink. Scala's intent matches Maven's — its dynamic tactics do call `shrinkRoots` — so its pom fallback is the odd one out there too. Left for a separate PR: it needs an sbt fixture to verify, and the sbt closure's notion of the project's own packages is not `closureSubmodules`.

**Dynamic submodule filtering looks separately broken** on `master`: on a two-module fixture, `--only-target` for one module returns every dependency and for the other returns none. Untouched here.

## Metrics

No.

## References

Reported internally; no ticket.

## Checklist

https://claude.ai/code/session_01QrDtxeRNysKQz4NWXixRzU

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Maven: Static analysis of `pom.xml` files now reports the dependencies a project declares as direct. Previously the project itself was reported as the only direct dependency and everything it declared was reported as transitive, and in a multi-module project the root pom and every submodule were also reported as dependencies. This affects analysis without `mvn` on `PATH`, `--static-only-analysis`, and container scanning; analysis using `mvn` was already correct. Selecting a submodule with `--only-target` now reports that submodule's dependencies as direct instead of reporting none at all.
 - Diagnostics: When an error or warning group contains multiple errors, each error's `Traceback:` header is now printed on its own line instead of being glued onto the last line of the preceding error message (e.g. `...none passed validationTraceback:`). ([#1758](https://github.com/fossas/fossa-cli/pull/1758))
 
 ## 3.18.2
--- a/Changelog.md
+++ b/Changelog.md
@@ -2,7 +2,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


spatten:
Closing, as this is implemented and merged in https://github.com/fossas/fossa-cli/pull/1770

PR #1770, opened 2026-09-04
Maven: report declared dependencies as direct in static analysis

csasarak:
Our static pom.xml strategies always marked every dependency beside the project itself as a transitive dependency. We never seem to have noticed because for a lot of our product it doesn't really matter that much: you have to comply with the license or address the security issues regardless of where in the project graph the dep appears.

That is not true for some product areas now. This PR changes our static pom.xml strategies to correctly label dependencies in the graph.

## Acceptance criteria

When run with `--static-only-analysis` or on fall-back to static analysis `pom.xml` based strategies now correctly 

## Testing plan

It's pretty easy to see this with most Maven projects. I used one from our example-projects repo.

1. Run an analysis on master with `fossa analyze --static-only-analysis <project>`
2. Run an analysis in the same repo with `fossa analyze --static-only-analysis <project>` from this branch
3. You should see that directs are transitives.

Before:
image

After:

image

*Note* the way we cache builds may affect this if you go to test. After it releases that shouldn't be a problem after a day or so.

## Risks

If customers have workflows that are based on the broken behavior this may change the work they have to do - I could see scenarios where it is more and also scenarios where it's less.

## Checklist


Changelog.md
```diff
@@ -599,20 +668,24 @@ Container scanning: Resolved a large number of issues with scanning containers (
...
 
 ## v3.8.23
+
 - Custom License Scans: Support full-file uploads for custom license scans ([#1333](https://github.com/fossas/fossa-cli/pull/1333)).
 
 ## v3.8.22
+
 - path: adds path dependency scanning functionality. ([#1327](https://github.com/fossas/fossa-cli/pull/1327))
-- `pnpm`: Supports `6.0` version of `pnpm-lockfile.yaml` ([#1320])(https://github.com/fossas/fossa-cli/pull/1320)
+- `pnpm`: Supports `6.0` version of `pnpm-lockfile.yaml` ([#1320])(<https://github.com/fossas/fossa-cli/pull/1320>)
```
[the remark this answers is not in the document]
csasarak: These two malformed links (`[#1320])` and `[#1152]]`) are pre-existing — they appear verbatim at the base of this PR in old release entries (v3.8.22 / v3.6.17) and are not introduced by this change. This PR only touches those lines incidentally because the `origin/master` merge wrapped the URLs in `<...>`. To keep the diff focused on the Maven/Scala fix I'm leaving them as-is.


_(Reply written by an LLM assistant, on behalf of the author.)_

src/Strategy/Maven.hs
```diff
@@ -3,6 +3,7 @@ module Strategy.Maven (
   mkProject,
   MavenProject (..),
   getDeps,
+  getDepsStatically,
```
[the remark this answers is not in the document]
csasarak: Ran the project's own `fourmolu` (same version/config that `make fmt` uses) over all affected files — it makes no changes; the files are already compliant. The `comma-style: leading` setting in `fourmolu.yaml` applies to list/tuple *expressions*, while module export/import lists are governed by `import-export-style: diff-friendly`, which keeps trailing commas. So the current formatting is exactly what the project's formatter produces.


_(Reply written by an LLM assistant, on behalf of the author.)_

src/Strategy/Maven.hs
```diff
@@ -27,7 +28,7 @@ import Effect.ReadFS (ReadFS)
 import GHC.Generics (Generic)
 import Graphing (Graphing, gmap, shrinkRoots)
 import Path (Abs, Dir, Path, parent)
-import Strategy.Maven.Common (MavenDependency (..), filterMavenDependencyByScope, filterMavenSubmodules, mavenDependencyToDependency)
+import Strategy.Maven.Common (MavenDependency (..), filterMavenDependencyByScope, filterMavenSubmodules, mavenDependencyToDependency, promoteFirstPartyToDirect)
```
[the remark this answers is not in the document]
csasarak: Explicit-but-unqualified imports (`import Graphing (Graphing, gmap, shrinkRoots)`) are the dominant convention in this codebase (~85% of imports); only a minority use `qualified`. `.hlint.yaml` only pins alias names for a handful of modules (`Data.Set`, `Data.Text`, `Data.Map`) and doesn't mandate qualification. The new imports follow the surrounding style, so I've left them unqualified.


_(Reply written by an LLM assistant, on behalf of the author.)_

src/Strategy/Scala.hs
```diff
@@ -10,6 +10,7 @@ module Strategy.Scala (
   discover,
   findProjects,
   ScalaProject (..),
+  analyzeWithPoms,
```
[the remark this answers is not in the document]
csasarak: Ran the project's own `fourmolu` (same version/config that `make fmt` uses) over this file — it makes no changes. The export list formatting follows `import-export-style: diff-friendly` (trailing commas), which is the project standard; `comma-style: leading` applies to list/tuple expressions, not module export lists. Leaving as-is.


_(Reply written by an LLM assistant, on behalf of the author.)_

src/Strategy/Scala.hs
```diff
@@ -43,7 +44,7 @@ import Effect.Exec (
 import Effect.Logger (Logger, logDebug, viaShow)
 import Effect.ReadFS (ReadFS, readContentsXML)
 import GHC.Generics (Generic)
-import Graphing (gmap)
+import Graphing (gmap, shrinkRoots)
```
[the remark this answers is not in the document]
csasarak: Explicit-but-unqualified imports are the dominant convention in this codebase (~85% of imports); `.hlint.yaml` only pins alias names for a few modules and doesn't mandate qualification. The added `Graphing` / Maven imports follow the surrounding style in this file, so I've left them unqualified.


_(Reply written by an LLM assistant, on behalf of the author.)_

Changelog.md
```diff
@@ -1,9 +1,10 @@
...
+- Scala: the sbt-generated-pom fallback no longer reports the project's own artifact as the only Direct dependency; it now removes the project artifact and promotes declared dependencies to Direct, matching Scala's dependency-tree tactics.
 - Dart: `pubspec.yaml` files using valid dependency forms the parser previously rejected no longer fail analysis with `Aeson exception: ... empty` or `failed parsing pub package's source!`: a bare dependency with no value (any version), a `version:`-only entry, the `hosted: <url>` shorthand introduced in Dart 2.15, a `hosted:` map without a `version`, and a `git:` map without a `ref`. ([#1760](https://github.com/fossas/fossa-cli/pull/1760))
-- Workflows: `fossa analyze --x-workflow <path>` runs a dependency-usage workflow analyzer through the embedded ficus and records its result in the debug bundle. ([#1761](https://github.com/fossas/fossa-cli/pull/1761))
```
spatten, line 6: Should this be added back in?

spatten: Also, were the whitespace changes intentional? They're reasonable - it's nice to have a space after every header. But it adds a lot of change to the Changelog that will be linked to this PR

src/Strategy/Maven.hs
```diff
@@ -187,8 +182,30 @@ getStaticAnalysis ::
...
+-- sole Direct dependency.
+finalizeMavenGraph ::
+  (Has Diagnostics sig m, Has (Reader MavenScopeFilters) sig m) =>
+  Set Text ->
+  Set Text ->
+  Graphing MavenDependency ->
+  m (Graphing Dependency)
+finalizeMavenGraph targetSet submoduleSet graph = do
+  filteredGraph <- applyMavenFilters targetSet submoduleSet graph
+  pure (shrinkRoots filteredGraph)
```
spatten, line 208: Claude noticed that we double-shrink if you scope dependencies with `scope-only`, resulting in no dependencies being found

Repro is a `.fossa.yml` like this:

```
# .fossa.yml
version: 3
maven:
  scope-only:
    - compile
```

And a pom.xml like this:

```
<project xmlns="http://maven.apache.org/POM/4.0.0">
  <modelVersion>4.0.0</modelVersion>
  <groupId>com.example</groupId>
  <artifactId>app</artifactId>
  <version>1.0.0</version>
  <dependencies>
    <dependency>
      <groupId>junit</groupId>
      <artifactId>junit</artifactId>
      <version>4.13.2</version>
      <scope>compile</scope>
    </dependency>
  </dependencies>
</project>
```

And run with `--static-only-analysis`:

```
fossa-dev analyze --output --static-only-analysis | jq .sourceUnits
```

I see no imports in my source units. I see one source unit if I run on main.

Here's more info from Claude:

Adding shrinkRoots to the static path double-shrinks when experimental.maven.scope-only is set. filterMavenDependencyByScope (Common.hs) uses Graphing.shrink, which rewires through deleted vertices. First-party nodes have empty dependencyScopes — only addDep attaches MavenLabelScope, never buildProjectGraph's direct (coordToPackage ...) — so under MavenScopeIncludeFilters with a non-empty set they hit the (True, False) branch and are excluded. shrink then rewires Root → project → dep into Root → dep, promoting declared deps to direct itself.

shrinkRoots then deletes that promoted layer. Because static pom graphs are only two levels deep (declared deps are leaves), shrinkSingle finds an empty postSet and there is nothing to promote:

```
Root → project → junit    -- scope include filter removes `project`
Root → junit              -- ...then shrinkRoots removes `junit`
Root                      -- empty graph
```

So with scope-only: [compile] and a <scope>compile</scope> dep, --static-only-analysis now reports nothing where it previously reported the dep. Exclude filters are unaffected (empty scopes hit (True, _) → True).

The dynamic path has the same latent flaw — but there it only loses the direct layer while transitives survive, since those graphs have real depth. This PR imports it into static, where it's total data loss.

Fix is probably to exempt first-party nodes from scope filtering (they have no scope to filter on), e.g. keep any node whose name is in submoduleSet in isMavenDependencyIncluded. That would fix the dynamic path at the same time.

spatten (changes requested):
Claude found a reasonably important bug, so I'm requesting changes. I think it's fine once that is fixed, though

Changelog.md
csasarak: Yes — good catch. This was restored in 229dadf0; it had been dropped accidentally during the merge from master. I also reverted the whitespace-only churn across the rest of the Changelog, so the diff against master now contains only the new `## Unreleased` section with the two entries for this change.

---
🤖 **AI assistant reply** — this response was generated by an LLM (Claude), not by a FOSSA employee.

csasarak: The whitespace changes were not intentional — a local markdown formatter ran over the whole file during the merge-conflict resolution. I've reverted all whitespace-only changes, so the Changelog diff against master is now limited to the new `## Unreleased` section (the two entries for this PR). The Workflows line is restored as well.

---
🤖 **AI assistant reply** — this response was generated by an LLM (Claude), not by a FOSSA employee.

src/Strategy/Maven.hs
csasarak: Confirmed and fixed in 229dadf0. Root cause matches the analysis: `filterMavenDependencyByScope` uses `Graphing.shrink`, and first-party nodes (which never carry `MavenLabelScope` labels, since `buildProjectGraph` creates them via `direct`/`edge` without scope labels) hit the `(True, False)` branch under a non-empty `scope-only` set and were excluded. `shrink` then rewired their children onto Root, and the subsequent `shrinkRoots` deleted those newly promoted leaf nodes — total data loss for shallow static graphs.

Fix: `filterMavenDependencyByScope` now takes the set of first-party submodule names and always keeps those nodes, so scope filtering only applies to third-party dependencies. Since both the static and dynamic paths go through the shared `finalizeMavenGraph` / `applyMavenFilters`, this fixes the latent flaw in the dynamic path as well.

Regression tests added:
- `Maven.CommonSpec`: first-party node with empty scopes survives an include filter and keeps its edge to a scoped dep
- `Maven.StaticAnalysisSpec`: end-to-end static analysis of a pom with `<scope>compile</scope>` under `scope-only: [compile]` (new fixture `test/Maven/testdata/static-scope-repro`) still reports the dependency as direct

All 78 Maven and 25 Scala unit tests pass.

---
🤖 **AI assistant reply** — this response was generated by an LLM (Claude), not a FOSSA employee.

csasarak: I think this ought to be addressed now. I made a test that I verify breaks on main.

Changelog.md
csasarak: Yeah, sorry this was tooling stuff. A markdownlint was running. I think that that line should be back in the changelog and the blanks are gone now.

spatten (approved):
Nice! Thanks for fixing that bug

PR #1775, opened 2026-09-11
[ANE-3140] Adapt the Conan strategy in the CLI to emit native Conan source units

GauravB159:
# Overview

The CLI currently has a strategy for Conan but it isn't connected to the CLI discovery logic making it dead code. It uses the `conan install` command instead of `conan graph info`. It then converts those to archive locators which cannot have vulns. 

This PR makes the following changes: 

1. It removes all the enrichment logic that converts to archive deps
2. It connects the Conan strategy to the main discovery logic
3. It adds `conan+` locators and emits Conan source units
4. It switches from using `conan install` to `conan graph info` so that we don't need to install anything on the user machine

## Acceptance criteria

- [x] Running fossa analyze in a conan project finds all the expected Conan dependencies

## Testing plan

1. Put this conanfile.py in an empty folder

```python
from conan import ConanFile


class ToyConan(ConanFile):
    """Deliberately NOT statically parseable.

    Requirements branch on options and settings, and one ref is computed at
    runtime. This is the ~67-78% case measured across ConanCenter recipes and
    real consumer conanfiles: you cannot read the dep set off the source, only
    off a resolved graph.
    """

    name = "toy"
    version = "0.1.0"
    settings = "os", "arch", "compiler", "build_type"
    options = {"with_ssl": [True, False], "with_sqlite": [True, False]}
    default_options = {"with_ssl": True, "with_sqlite": False}

    _zlib_version = "1.3.1"

    def requirements(self):
        # computed arg: an f-string, not a literal
        self.requires(f"zlib/{self._zlib_version}")
        self.requires("fmt/10.2.1")

        # conditional on an option
        if self.options.with_ssl:
            self.requires("openssl/3.2.0")

        if self.options.with_sqlite:
            self.requires("sqlite3/3.45.0")

        # conditional on a setting
        if self.settings.os == "Linux":
            self.requires("libbacktrace/cci.20210118")

    def build_requirements(self):
        self.tool_requires("cmake/3.28.1")
```

2. Run the local CLI in that folder

```bash
fossa-cli/dist-newstyle/build/aarch64-osx/ghc-9.8.4/spectrometer-0.1.0.0/x/fossa/noopt/build/fossa/fossa analyze . --output --debug > output.json
```

3. The output should look something like this

```json
{
    "projects": [
        {
            "graph": {
                "assocs": [
                    [
                        2,
                        [
                            3
                        ]
                    ]
                ],
                "deps": [
                    {
                        "locations": [],
                        "name": "cmake",
                        "tags": {},
                        "type": "ConanType",
                        "version": {
                            "type": "EQUAL",
                            "value": "3.28.1"
                        }
                    },
                    {
                        "locations": [],
                        "name": "fmt",
                        "tags": {},
                        "type": "ConanType",
                        "version": {
                            "type": "EQUAL",
                            "value": "10.2.1"
                        }
                    },
                    {
                        "locations": [],
                        "name": "openssl",
                        "tags": {},
                        "type": "ConanType",
                        "version": {
                            "type": "EQUAL",
                            "value": "3.2.0"
                        }
                    },
                    {
                        "locations": [],
                        "name": "zlib",
                        "tags": {},
                        "type": "ConanType",
                        "version": {
                            "type": "EQUAL",
                            "value": "1.3.1"
                        }
                    }
                ],
                "direct": [
                    0,
                    1,
                    2,
                    3
                ]
            },
            "path": "/Users/gauravbhagchandani/Desktop/Projects/playground/conan-toy/py-project/",
            "type": "conan"
        }
    ],
    "sourceUnits": [
        {
            "AdditionalDependencyData": null,
            "Build": {
                "Artifact": "default",
                "Dependencies": [
                    {
                        "imports": [],
                        "locator": "conan+fmt$10.2.1",
                        "metadata": null
                    },
                    {
                        "imports": [
                            "conan+zlib$1.3.1"
                        ],
                        "locator": "conan+openssl$3.2.0",
                        "metadata": null
                    },
                    {
                        "imports": [],
                        "locator": "conan+zlib$1.3.1",
                        "metadata": null
                    }
                ],
                "Imports": [
                    "conan+fmt$10.2.1",
                    "conan+openssl$3.2.0",
                    "conan+zlib$1.3.1"
                ],
                "Succeeded": true
            },
            "Data": null,
            "Files": null,
            "GraphBreadth": "complete",
            "Info": null,
            "Manifest": "/Users/gauravbhagchandani/Desktop/Projects/playground/conan-toy/py-project/",
            "Name": "/Users/gauravbhagchandani/Desktop/Projects/playground/conan-toy/py-project/",
            "NoticeFiles": [],
            "OriginPaths": [
                "conanfile.py"
            ],
            "Title": null,
            "Type": "conan"
        }
    ]
}
```

## Risks

Low risk since it's completely new

## Metrics

I'll keep an eye on the [CLI dashboard](https://app.datadoghq.com/dashboard/zaj-mx6-kei/fossa-cli-dashboard?fromUser=false&refresh_mode=sliding&from_ts=1789433614660&to_ts=1790038414660&live=true) for any conan related errors

## References

[Conan Notes and Scribbles](https://docs.google.com/document/d/1m-muRwQ6kTgOP6mITAkC6X4thMAw7al7BryuQPEYk98/edit?usp=sharing)

## Checklist


src/Strategy/Conan/ConanGraph.hs
```diff
@@ -190,109 +166,52 @@ instance FromJSON ConanGraphNodeContext where
...
 
-    let transitives = mapMaybe (refToDependency . TextLike) (keys $ requires dep)
-    for_ transitives $ \childDep -> do
-      deep childDep
-      edge resolvedDep childDep
+    let transitives = mapMaybe (refToDependency) (keys $ dependencies dep)
+    let resolvedDirectDeps = Map.map dep_direct (dependencies dep)
+    for_ transitives $ \(nodeId, childDep) -> do
+      let isDirectChild = findWithDefault False nodeId resolvedDirectDeps
+      when isDirectChild $ edge resolvedDep childDep
```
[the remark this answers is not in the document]
GauravB159: This doesn't occur as seen in the manual test output

src/Strategy/Conan/ConanGraph.hs
```diff
@@ -190,109 +166,52 @@ instance FromJSON ConanGraphNodeContext where
...
+    refToDependency nodeId = ((nodeId,)) . toDependency <$> Data.Map.lookup nodeId registry
 
 toDependency :: ConanGraphNode -> Dependency
 toDependency cn =
   Dependency
     { dependencyType = ConanType
-    , dependencyName = name cn
-    , dependencyVersion = Just (CEq $ getVersion cn)
-    , dependencyLocations = getLocations cn
+    , dependencyName = fromMaybe "root" (name cn)
```
zlav, line 203: I don't track this. If the dependency name is empty, we call it `root` and upload that to the backend? Won't this result in an unresolvable `root` dep in the users project?

src/Strategy/Conan/ConanGraph.hs
```diff
@@ -109,63 +100,48 @@ data ConanGraphNode = ConanGraphNode
...
-            Nothing -> fail $ "Expected " <> show key <> " to have, / seperator, but recieved: " <> show value
-            Just (name, version) -> pure (name, version)
+      <*> obj .: "dependencies"
+
+data ConanGraphDependency = ConanGraphDependency
+  { -- Dependency reference
+    dep_ref :: TextLike
+  , dep_test :: Bool
+  , dep_direct :: Bool
+  , dep_build :: Bool
```
zlav: Why not camel case for these?

zlav (approved):
One open question about the root dep, otherwise looks like a solid change to me

src/Strategy/Conan/ConanGraph.hs
GauravB159: ToDependency requires a name and the root dep has a blank name so this just fills a default name in for it. It still gets excluded when the source units are pushed up if you check the output.json

src/Strategy/Conan/ConanGraph.hs
GauravB159: Didn't notice the convention, I'll switch them

src/Strategy/Conan/ConanGraph.hs
```diff
@@ -50,42 +46,37 @@ import Effect.Exec (
...
 import Errata (Errata (..))
 import Graphing (Graphing)
-import Network.HTTP.Types (renderQueryText)
 import Path (Abs, Dir, Path)
 import Strategy.Conan.Version (guardConanVersion2Gt)
 
--- | Represents `conan install . -f json`.
--- Creates the dependency graph from install command.
+-- | Represents `conan graph info -f json`.
+-- Creates the dependency graph for a conan project.
```
csasarak, line 51: Is it that the caveat in the comment is no longer true?

src/Strategy/Conan/ConanGraph.hs
```diff
@@ -50,42 +46,37 @@ import Effect.Exec (
...
--- pass this option in `conan graph info` command. This is not documented,
--- but was communicated in: https://github.com/conan-io/conan/issues/13939
---
--- >> conan install . -f json
+-- >> conan graph info -f json
 -- >
 -- > {
--- >  "nodes": [
--- >    {
+-- > "graph": {
```
csasarak, line 56: Is this format documented anywhere? I suspect not but if it is please reference that document.

src/Strategy/Conan/ConanGraph.hs
```diff
@@ -190,109 +166,52 @@ instance FromJSON ConanGraphNodeContext where
...
-  | -- The package is a python require.
-    PythonRequires
-  | -- The type of the package is unknown.
-    Unknown Text
-  deriving (Show, Eq, Ord)
+getDirectDepsMap :: (Applicative m) => ConanGraph -> m (Map Text Bool)
+getDirectDepsMap graph = do
+  case Map.lookup "0" (nodes graph) of
+    Nothing -> pure Map.empty
+    Just node -> pure (Map.map depDirect (dependencies node))
```
csasarak, line 175:
```haskell
pure $ (Map.map depDirect . dependencies) =<< Map.lookup "0" (nodes graph)
```
Optional

src/Strategy/Conan.hs
```diff
@@ -0,0 +1,97 @@
+{-# LANGUAGE RecordWildCards #-}
```
csasarak, line 1: We use this frequently in our code-base, but personally I'm not a fan: it makes it hard to see where some bindings come from. Even though it's more verbose I like using the names rather than the wildcards

csasarak (approved):

src/Strategy/Conan/ConanGraph.hs
GauravB159: Kinda, and we also don't really need to dep to be installed locally anymore since it's only emitting Conan locators now and not analyzing the local dep for licenses or anything

src/Strategy/Conan/ConanGraph.hs
GauravB159: Good call, added a reference to their docs

src/Strategy/Conan/ConanGraph.hs
GauravB159: Directly replacing is throwing an error, if it's optional I'd rather keep it than redo the manual testing

src/Strategy/Conan.hs
GauravB159: Makes sense, it's only used in one place, removed that

PR #1766, opened 2026-09-02
Read optionalDependencies from pnpm lockfiles

spatten:
> **Related PRs — release together**
>
> - [#1763 — Scope pnpm analysis to individual workspace members](https://github.com/fossas/fossa-cli/pull/1763)
> - [#1764 — Document workspace target filters](https://github.com/fossas/fossa-cli/pull/1764)
> - [#1765 — Resolve pnpm catalog references in package.json fallback analysis](https://github.com/fossas/fossa-cli/pull/1765)
> - [#1766 — Read optional dependencies from pnpm lockfiles](https://github.com/fossas/fossa-cli/pull/1766) (this PR)
> - [#1776 — Preserve pnpm peer-resolution contexts](https://github.com/fossas/fossa-cli/pull/1776)
>
> **All five PRs must be merged into `master`, and the combined changes validated, before cutting a release.**
>
> Arrows point from each PR to its base.
>
> ```mermaid
> flowchart RL
>   p1763["#1763 · Workspace targets"] --> p1776["#1776 · Peer resolutions"]
>   p1776 --> p1766["#1766 · Optional dependencies"]
>   p1766 --> main["master"]
>   p1764["#1764 · Documentation"] --> main
>   p1765["#1765 · Catalog fallback"] --> main
> ```

## Description of the PR stack

Together, these changes let users select individual pnpm workspace members, retain their reachable dependencies (including optional dependencies), document workspace target filters, and avoid reporting unresolved workspace-reference strings as package versions when analysis falls back to package.json.

# Overview

FOSSA's pnpm analysis ignored the lockfile's `optionalDependencies` declarations. It still reported the packages themselves, because it included every entry in the lockfile's `packages` section, but it did not record which project or package depended on them.

For example, this project declares `sharp` as an optional dependency and `chokidar` as an ordinary dependency:

```json
{
  "name": "optional-example",
  "dependencies": {
    "chokidar": "^3.6.0"
  },
  "optionalDependencies": {
    "sharp": "^0.33.0"
  }
}
```

Its v9 lockfile records `fsevents` as an optional dependency of `chokidar`. Relevant snapshot excerpt:

```yaml
snapshots:
  chokidar@3.6.0:
    dependencies:
      readdirp: 3.6.0
    optionalDependencies:
      fsevents: 2.3.3
```

The diagram shows only the relationships relevant to this change. All three packages were already present in the result; the missing relationships made `sharp` appear transitive and left `fsevents` disconnected from `chokidar`.

```mermaid
[...]

PR #1776, opened 2026-09-11
Preserve pnpm peer-resolution contexts

spatten:
> **Related PRs — release together**
>
> - [#1763 — Scope pnpm analysis to individual workspace members](https://github.com/fossas/fossa-cli/pull/1763)
> - [#1764 — Document workspace target filters](https://github.com/fossas/fossa-cli/pull/1764)
> - [#1765 — Resolve pnpm catalog references in package.json fallback analysis](https://github.com/fossas/fossa-cli/pull/1765)
> - [#1766 — Read optional dependencies from pnpm lockfiles](https://github.com/fossas/fossa-cli/pull/1766)
> - [#1776 — Preserve pnpm peer-resolution contexts](https://github.com/fossas/fossa-cli/pull/1776) (this PR)
>
> **All five PRs must be merged into `master`, and the combined changes validated, before cutting a release.**
>
> Arrows point from each PR to its base.
>
> ```mermaid
> flowchart RL
>   p1763["#1763 · Workspace targets"] --> p1776["#1776 · Peer resolutions"]
>   p1776 --> p1766["#1766 · Optional dependencies"]
>   p1766 --> main["master"]
>   p1764["#1764 · Documentation"] --> main
>   p1765["#1765 · Catalog fallback"] --> main
> ```

## Description of the PR stack

Together, these changes let users select individual pnpm workspace members, retain their reachable dependencies (including optional dependencies), document workspace target filters, and avoid reporting unresolved workspace-reference strings as package versions when analysis falls back to package.json.

# Overview

Pnpm can resolve the same package version against different peer versions. Analysis previously stripped the peer context from snapshot keys before building the dependency graph. Those snapshots then shared a key, so one could overwrite the other and dependencies could disappear. Combining nodes too early could also mix production and development labels between the two resolutions.

For example, two workspace members can reach `widget@1.0.0` with different resolved versions of its peer dependency, `peer`. The package name and version match, but the dependency relationships differ:

```text
member a → widget@1.0.0(peer@1.0.0) → peer@1.0.0
member b → widget@1.0.0(peer@2.0.0) → peer@2.0.0
```

The relevant v9 lockfile snapshots look like this:

```yaml
snapshots:
  widget@1.0.0(peer@1.0.0):
    dependencies:
      peer: 1.0.0
  widget@1.0.0(peer@2.0.0):
    dependencies:
      peer: 2.0.0
```

Previously, stripping both parenthesized suffixes turned both snapshot keys into `widget@1.0.0`. One entry replaced the other, so only one set of outgoing edges survived.

| Stage | Before | After |
| --- | --- | --- |
| Build the graph | Both snapshots share one key; one overwrites the other. | Each full snapshot key keeps its own outgoing edge. |
| Propagate environments | Distinct resolutions can share labels too early. | Production/development labels follow each resolution separately. |
| Report `widget@1.0.0` | Only one peer edge survives. | One reported widget has edges to both `peer@1.0.0` and `peer@2.0.0`. |

The final graph can therefore combine the reported widget identity without losing either resolution's dependencies.

Keep the full snapshot keys through graph construction and environment propagation, then combine matching reportable packages. Each resolution keeps its outgoing edges and environments until that final step.

This extracts the general graph correction from #1763. Workspace target selection, pruning, and environment propagation through workspace links remain in #1763, which uses this correction to retain each selected member's peer versions.

## Acceptance criteria

When a pnpm lockfile resolves one package version against different peer versions, whole-workspace analysis retains both peer versions and their dependency edges. Production and development labels follow each resolution separately before matching reportable packages are combined.

## Testing plan

From the repository root, with the repository's Haskell/Rust toolchains and `jq` installed:

```sh
make install-dev
pnpm_review_dir="$(mktemp -d)"
```

Run the remaining commands in the same shell. `--include-unused-deps` retains development-only dependencies, which are part of the expected results below. They scan temporary fixtures with `--output`; no upload or package installation is needed.

Create a minimal workspace around the committed two-peer-context lockfile:

```sh
mkdir -p "$pnpm_review_dir/workspace/a" "$pnpm_review_dir/workspace/b"
printf '%s\n' '{"name":"peer-review","private":true,"version":"1.0.0"}' > "$pnpm_review_dir/workspace/package.json"
printf '%s\n' 'packages:' '  - a' '  - b' > "$pnpm_review_dir/workspace/pnpm-workspace.yaml"
printf '%s\n' '{"name":"a","dependencies":{"parent":"1.0.0"}}' > "$pnpm_review_dir/workspace/a/package.json"
printf '%s\n' '{"name":"b","devDependencies":{"parent":"1.0.0"}}' > "$pnpm_review_dir/workspace/b/package.json"
cp test/Pnpm/testdata/pnpm-9-peer-contexts/pnpm-lock.yaml "$pnpm_review_dir/workspace/pnpm-lock.yaml"
fossa-dev analyze --output --include-unused-deps "$pnpm_review_dir/workspace" | jq .sourceUnits
```

Expected: the output reports both `peer@1.0.0` and `peer@2.0.0`. In `Build.Dependencies`:

- The entry with `locator: "npm+parent$1.0.0"` lists `"npm+widget$1.0.0"` in its `imports`: parent depends on widget.
- The entry with `locator: "npm+widget$1.0.0"` lists both `"npm+peer$1.0.0"` and `"npm+peer$2.0.0"` in its `imports`: the combined workspace result retains widget's dependency on each resolved peer version.

If either peer version is missing from widget's `imports`, one resolution's dependency relationship has been lost.

## Risks

Graph nodes now retain snapshot identity until environment propagation finishes. Review the final collapse carefully: it must preserve all outgoing edges while combining environments only for matching reportable identities. Scoped pruning and its regression coverage are in #1763.

This corrects existing documented dependency reporting; it adds no configuration or command options. The changelog describes the correction, and workspace documentation remains in #1763.

## Metrics

No new runtime metrics. Regression tests check dependency versions, edges, and environments.

## References

[ANE-3125](https://fossa.atlassian.net/browse/ANE-3125)

[ANE-3125]: https://fossa.atlassian.net/browse/ANE-3125?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Checklist


src/Strategy/Node/Pnpm/PnpmLock.hs
```diff
@@ -120,52 +120,48 @@ resolveCatalogVersion (PnpmCatalogs cats) depName ver
...
   | otherwise = ver
 
--- | Apply accumulated labels to transform a graph node.
-applyLabels :: Dependency -> Set.Set PnpmLabel -> Dependency
-applyLabels = foldr applyLabel
-  where
-    applyLabel (PnpmEnv env) = insertEnvironment env
+-- | Keep the lockfile key alongside the reportable dependency so peer
+-- contexts remain distinct during reachability and environment propagation.
+type PnpmNode = (Text, Dependency)
```
tjugdev, line 125: An example of what the key looks like would be helpful here.

tjugdev (approved):

PR #1781, opened 2026-09-16
feat: make --x-workflow a switch instead of taking an analyzer path

robszumski:
# Overview

`--x-workflow` required a `PATH` to an analyzer that is never run.

ficus downloads the workflow analyzer itself from its own hardcoded release repository. The run artifact's `executable.program` is parsed, validated, and then discarded. So the path this flag demanded had stopped meaning anything — and `validateFile` (`Config/Analyze.hs:647`) would fail the whole run if that meaningless file happened not to exist.

This PR makes the flag a plain switch and has the run artifact carry a constant executable section naming the analyzer that actually runs, rather than a path this side no longer has.

## Acceptance criteria

Before:

```sh
fossa analyze --x-workflow /some/path/that/must/exist ~/repo
```

After:

```sh
fossa analyze --x-workflow ~/repo
```

A trailing path is now read as the scan target, which is what a user would expect it to mean.

## Changes

| File | Change |
|---|---|
| `Config/Analyze.hs` | `strOption` → `switch`; `xWorkflow :: Maybe (Path Abs File)` → `Bool`; dropped `validateFile` |
| `Ficus/Types.hs` | `toWorkflowExecutable` (path → executable) replaced by the constant `downloadedWorkflowExecutable` |
| `Ficus/Workflow.hs` | `analyzeWithWorkflow` / `runWorkflowWith` / `failWorkflow` no longer take an analyzer path |
| `Analyze.hs` | gate the run on the `Bool` |

### A latent bug this closes

ficus ignores `executable.program` but still places `executable.args` ahead of the target on the downloaded analyzer's command line:

```rust
command.args(&executable.args).arg(artifact.target()).arg("--output").arg(output)
```

and `toWorkflowExecutable` put the bundle path in `args` for a `.js` analyzer. So `--x-workflow foo.js` would have invoked the downloaded analyzer as `analyzer /path/foo.js <target> --output ...` — a stray leading positional argument. Sending no args removes that by construction.

### What is deliberately removed

The `.js`/`.mjs`/`.cjs` → `node` wrapper inference, and its five tests. With no path to inspect there is nothing to infer, and running a bundle under node is ficus's concern now rather than this side's. Flagging it explicitly since it is a capability removal, not a refactor.

## Testing plan

1. Build: `cabal build` — clean, no warnings.
2. Flag behavior: `cabal test unit-tests --test-options='--match "/--x-workflow/"'` — 4 examples, 0 failures, covering the default, the flag set, a trailing path being read as the scan target, and the `--static-only-analysis` conflict.
3. Full suite: `cabal test unit-tests` — **1522 examples, 2 failures, 25 pending**.

The two failures are `App.Fossa.VSI.DynLinked.Util, setuid bit` and are **pre-existing**: they reproduce identically on a stashed clean tree. Nothing in this change touches VSI or file permissions.

Note for anyone reproducing locally: the suite aborts at startup unless git-lfs objects are present (`read tar at test/Container/testdata/emptypath.tar: TarParserError: TruncatedArchive`). `git lfs pull` fixes it and took failures from 14 to 2 here.

One of the new tests caught a wrong assumption while writing it: I had expected `--x-workflow /definitely/not/here` to simply enable the flag, but the trailing path is now the scan target, so `mergeOpts` fails naming that directory. The test asserts the real behavior, which is also the better thing to pin — an old invocation that still passes an analyzer path must not silently analyze it.

## Risks

Low, but it is a **breaking change to an experimental flag**: any existing `--x-workflow /path/to/analyzer` invocation now treats that path as the scan target instead of the analyzer. That is loud rather than silent when the path is not a directory, and `x-` flags are documented as unsupported and removable, so this seems like the right moment.

No ficus change is needed — ficus already ignores the program, so this works against the current released ficus as well as the build in fossas/ficus#191.

I could not run `make fmt` (fourmolu is not installed here); the new lines match the formatting of the adjacent `--x-vendetta` switch.


src/App/Fossa/Analyze.hs
```diff
@@ -117,7 +117,7 @@ import Control.Effect.Git (Git)
 import Control.Effect.Lift (sendIO)
 import Control.Effect.Stack (Stack, withEmptyStack)
 import Control.Effect.Telemetry (Telemetry, trackResult, trackTimeSpent)
-import Control.Monad (join, unless, void, when)
+import Control.Monad (guard, join, unless, void, when)
```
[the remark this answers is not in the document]
robszumski: Not applying this one — qualifying these imports would make the files inconsistent with the rest of the codebase.

The repo's convention is unqualified imports with an explicit symbol list, and it isn't close:

| Module | unqualified + explicit list | qualified |
|---|---|---|
| `Path` | 221 | 6 |
| `Control.Monad` | 109 | 0 |
| `Control.Effect.Lift` | 120 | 0 |

(counted across `src/` and `test/`)

These import lines are also pre-existing, not introduced here. On `master`:

```haskell
-- src/App/Fossa/Analyze.hs:120
import Control.Monad (join, unless, void, when)
-- test/App/Fossa/Config/AnalyzeSpec.hs:16,23,24
import Control.Effect.Lift (Has, Lift, sendIO)
import Path (Abs, Dir, File, Path, parseAbsFile, parseRelFile, toFilePath, (</>))
import Test.Effect (expectFatal', expectationFailure', it', itWithTempDir', shouldBe', shouldEndWith')
```

This PR only adds or removes *symbols* within those existing lists — `guard` added to `Control.Monad`; dead symbols pruned from the others after `toWorkflowExecutable` was removed. Switching four import sites to qualified style would be an unrelated, repo-wide convention change that belongs in its own PR if the team wants it, not a drive-by on the lines this PR happened to touch.

Happy to reconsider if there's a style guide I've missed that says otherwise.

test/Ficus/WorkflowProcessSpec.hs
```diff
@@ -33,7 +33,7 @@ import Data.String.Conversion (decodeUtf8, toString, toText)
 import Data.Text (Text)
 import Data.Text qualified as Text
 import Effect.Exec (AllowErr (Never), Command (..), ExitCode (ExitFailure, ExitSuccess))
-import Path (Abs, Dir, File, Path, parseAbsDir, parseAbsFile, parseRelFile, toFilePath, (</>))
+import Path (Abs, Dir, Path, parseAbsDir, parseRelFile, toFilePath, (</>))
```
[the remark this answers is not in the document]
robszumski: Same answer as the companion comment on `Analyze.hs` — not applying, for consistency reasons.

`test/Ficus/WorkflowProcessSpec.hs:36` imports `Path` unqualified with an explicit list, which is what 221 of the 227 `Path` imports in `src/` + `test/` do. The line is pre-existing; on `master` it reads:

```haskell
import Path (Abs, Dir, File, Path, parseAbsDir, parseAbsFile, parseRelFile, toFilePath, (</>))
```

This PR's only change to it was pruning `File`, `parseAbsFile` and `(</>)` — symbols left dead when `toWorkflowExecutable` and its `Path Abs File` parameter were removed. CI's `-Werror=unused-imports` required that pruning; converting the import to qualified style is a separate concern.

One thing your review did help with indirectly: the same `toWorkflowExecutable` removal orphaned imports and top-level bindings in three modules, and only one of them failed the first CI run. Those are fixed in `2696aefc`, `188feac7`, and `7481696a`. CI is now green on all 19 checks at `85e459f7`, so a re-review against current HEAD would be useful — this review was against `42fad67e`, three pushes back.

docs/references/subcommands/analyze.md
```diff
@@ -192,13 +193,14 @@ then. Nothing else about this flag changes what `fossa analyze` sends to FOSSA.
 
 #### Enabling dependency usage analysis
 
-| Name                  | Description                                                                                                                                                                    |
-|-----------------------|--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
-| `--x-workflow PATH`   | Experimental. Run the dependency-usage workflow analyzer at `PATH` over the project. A `.js`, `.mjs` or `.cjs` path is run under `node`; any other path is run as the program itself. |
+| Name             | Description                                                                                          |
+|------------------|------------------------------------------------------------------------------------------------------|
+| `--x-workflow`   | Experimental. Run the dependency-usage workflow analyzer, which the embedded ficus downloads, over the project. |
```
zlav: Can you remove the refs to ficus here? I'm not sure we publicly make that obvious, and I think in this case it makes it confusing to users.

src/App/Fossa/Config/Analyze.hs
```diff
@@ -376,14 +376,7 @@ cliParser =
...
     <*> switch (applyFossaStyle <> long "x-vendetta" <> stringToHelpDoc "Experimental flag to enable vendored dependency scanning to identify open source components using file hashing.")
-    <*> optional
-      ( strOption
-          ( applyFossaStyle
-              <> long "x-workflow"
-              <> metavar "PATH"
-              <> stringToHelpDoc "Experimental. Run the dependency-usage workflow analyzer at PATH."
-          )
-      )
+    <*> switch (applyFossaStyle <> long "x-workflow" <> stringToHelpDoc "Experimental flag to enable the dependency-usage workflow analyzer, which ficus downloads.")
```
zlav: Same here, remove ficus ref

zlav (approved):

docs/references/subcommands/analyze.md
robszumski: removed in https://github.com/fossas/fossa-cli/pull/1781/commits/6f5f06952ba0c5c11a3d2e1480f8215e80d168f9. Found a few other refs that I also removed.

src/App/Fossa/Config/Analyze.hs
robszumski: fixed in https://github.com/fossas/fossa-cli/pull/1781/commits/6f5f06952ba0c5c11a3d2e1480f8215e80d168f9

PR #1784, opened 2026-09-18
[ANE-3148] pyproject.toml parse error

tjugdev:
# Overview

`pyproject.toml` files can have entries in `[tool.poetry.dependencies]` with no version, git, path or url (e.g. a source-only `pkg = { source = "private" }` that adds settings to a dependency declared in PEP 621 `[project].dependencies`). The parser was choking on these, and now it doesn't. Version info for these dependencies is populated from the lock file when available.

This also contains an extra bug fix to stop a poetry project from incorrectly being reported twice: once as a poetry project (correct) and again as a pdm project (incorrect). The fix follows the already established pattern being used to stop uv projects from being reported as pdm.

## Acceptance criteria

This PR adds a new test fixture which can be used to test:

1. Run `fossa analyze --output test/Python/Poetry/testdata/pep621-source-only`. Note that it fails and the first line says `Analyzing pdm project at...`
2. Run `cabal run fossa --  analyze --output test/Python/Poetry/testdata/pep621-source-only`. Note it succeeds and only one project of type `poetry` is detected.

## Risks

## Metrics

## References

[ANE-3148](https://fossa.atlassian.net/browse/ANE-3148)

## Checklist

[ANE-3148]: https://fossa.atlassian.net/browse/ANE-3148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zlav (approved):
The logic looks correct to me. 

The logic change in this test appears to show that these deps which FOSSA cannot resolve will now show up in the graph. Should these types of dependencies be reported for projects like this? And are some of these types resolvable?

tjugdev:
@zlav I'm not sure what new deps you're referring to. In most cases we previously failed to analyze entirely due to an error that was hit parsing the poetry.toml. The new behaviour is to now not crash and the deps produced are what I believe to be the full correct set.

